Issue #4, opened 2019-05-14
S3 as storage for App Registry (CRUD + search by groups)


PK85:
**AC**
- [x] request AWS subscription
- [ ] create POC which shows how to query data stored in S3
  - [ ] includes query via minio gateway and directly to S3
  - [ ] show what will be a structure of data in S3 to allow nice query capabilities
  - [ ] show investigation of Apache Parquet output format with AWS S3
- [x] check a status of GPC and Azure storages and Parquet support
- [ ] provide a short guide how setup and how work cross-region replication for AWS S3

pkosiec:
The PoC requirement is not valid anymore as during investigation we found that S3 doesn't fit our needs.

PR #7, opened 2019-05-16
Add S3 storage and Parquet investigation

pkosiec:
**Description**

- Add S3 storage and Parquet investigation

**Related issue(s)**
#4


docs/storage/s3/README.md
```diff
@@ -0,0 +1,123 @@
...
+- List all directories under `{tenant_id}/labels/{key}` (1 request)
+
+Total calls: 1
+
+### Pros
+- low cost
+- region replication
+- backup
+
+### Cons
```
tgorgol: Some other cons I can see are:
1. No update functionality.
If we decide to store metadata in `data.json` file then to update a single value from that file we have to replace whole file. In our case this wouldn't be a big problem as the `data.json` file is expected to be small, nonetheless it will increase complexity of code, as we will have to first download current file, patch it locally and then put the new file in the storage. One way of solving this issue is to use separate files for each metadata property with it's value in name, so we would have a directory structure like that:
   ```
   {tenant_id}/
   ├── applications/
   │   ├── {application_id}/
   │   │  ├── name={application_name}
   │   │  ├── description={application_description}
   │   │  ├── annotations/
   │   │  │  ├── {annotation_key} // value in file?
   │   │  │  └── ...
   │   │  ├── apis/
   │   │  │  └── ...
   │   │  ├── labels/
   │   │  │  └── {label_key}={label_value}
   │   │  └── ...
   ...
   ```
   that way updating single property would require only replacing a single file, also, reading application metadata would be faster as most of it would be available after listing the application directory. The downsides to storing this data in filename are possible length/encoding limitations.
2. No support for transactions. For example, if we want to use data redundancy then we need to consider cases where one file write will succeed while other one will fail. This will further increase code complexity.

pkosiec: Good points. I will add them into the document.

docs/storage/s3/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+- United States (multiple regions in the United States) 
+- European Union (multiple regions in the European Union) 
+- Asia (multiple regions in Asia)
+- eur4 (Finland and Netherlands) 
+- nam4 (Iowa and South Carolina) 
+
+We cannot replicate a single bucket across multiple regions like Asia, USA and Europe.
+
+### AWS S3
```
PK85: Please check azure as well and document

docs/storage/s3/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+Amazon has [Cross-region replication for S3 buckets](https://docs.aws.amazon.com/AmazonS3/latest/dev/crr.html). One bucket would be used to write, and others to read.
+
+## Summary
```
PK85: If not s3, suggest databases that we could investigate and use where below issues will be solved and cross-regional replication requirement will be fulfilled.

docs/storage/s3/README.md
```diff
@@ -0,0 +1,150 @@
...
+- Create `data.json` file in `{tenant_id}/runtimes/{runtime_id}` directory (1 request)
+
+For every label of the runtime (N is the number of the runtimes):
+- Create `{key}={value}` file for every label in `{tenant_id}/runtimes/{runtime_id}/labels` (N requests)
+- Create `data.json` file in `{tenant_id}/labels/{key}/{value}/runtimes/{runtime_id}` directory (N requests)
+- Create `key=value` file in `{tenant_id}/labels/{key}/{value}/runtimes/{runtime_id}/labels` directory (N * N requests)
+
+Total calls: N^2 + 2*N + 1
+
+**Application registration**
```
PK85: Keep it consistent , Application creation

docs/storage/s3/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+### S3-based Solution
+
+To achieve querying by groups, the following bucket structure is suggested:
+
+![Bucket structure](./assets/bucket-structure.jpg)
+
+Data redundancy is needed to reduce requests while doing read operations.
+
+**Runtime creation**
```
PK85: Add header: Adding label to Runtime

docs/storage/s3/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+**Application registration**
+
+Similar to Runtime creation - replace `runtimes` with `applications`
+
+Total calls: N^2 + 2*N + 1
+
+**Adding label to Application**
+
+- Create file `{key}={value}` in `{tenant_id}/applications/{application_id}` directory (1 request)
```
PK85: Why adding label to app has less calls than doing same with runtime?

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,38 @@
...
+
+    For Golang, we have [parquet-go](https://github.com/xitongsys/parquet-go) library, which handles read and writes of the `parquet` file.
+
+    In order to do queries in Apache Parquet, we would need a query engine, such as [Apache Drill](https://drill.apache.org/), which is written in Java.
+
+    If we wanted to avoid this engine, in our use case, we would need to define group names and application names as columns.
+
+1. **Concurrent write into the same file**
+
+    We could enable bucket files versioning (for example, in [GCS](https://cloud.google.com/storage/docs/gsutil/addlhelp/ObjectVersioningandConcurrencyControl)). However, we would need to implement a mechanism to always fetch the latest version of the file and retry saving until success.
```
PK85: It's not an option, that's one of the requirement that we need to keep previous revisions of APIs, docs etc

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,38 @@
+# Apache Parquet
+
+There is idea to use Apache Parquet for storing group relations between Runtime and the Application.
```
PK85: I just missing some data structure in S3 itself, something similar you did for s3 documentation. About Parquet file and where specs are stored.

docs/storage/s3/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+There were an idea to use Apache Parquet for storing the pieces of information regarding labels.
+We want to have an easy access to the pieces of information:
+    - list all labels for application 
+    - query application by label
+    - list all labels
+
+### S3-based Solution
+
+To achieve querying by groups, the following bucket structure is suggested:
```
PK85: Please also add information how swagger, asynApi, docs files will be stored.

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,38 @@
+# Apache Parquet
```
PK85: I don't see evaluation of minio and Parquet usage in the context of the providers. Please add

docs/storage/s3/README.md
pkosiec: I've covered runtime creation, and this is the case of adding label to an existing application. I will add another sections to cover all cases for app and runtime

docs/storage/s3/parquet.md
pkosiec: In parquet file we wouldn't store APIs, docs, etc. - just info about all labels, and labels for all runtimes and apps

docs/storage/s3/parquet.md
pkosiec: Hmm, Minio doesn't have anything to do with that, as it would only download/upload a file. BTW Minio supports S3 select.

docs/storage/s3/parquet.md
pkosiec: Done

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,67 @@
...
+    
+    Fetching specific column values doesn't need reading the entire raw data.
+
+### Cons
+
+1. **Doing queries**
+
+    For Golang, we have [parquet-go](https://github.com/xitongsys/parquet-go) library, which handles read and writes of the `parquet` file.
+
+    In order to do queries in Apache Parquet, we would need a query engine, such as [Apache Drill](https://drill.apache.org/), which is written in Java.
```
aszecowka: https://aws.amazon.com/about-aws/whats-new/2018/12/amazon-s3-announces-parquet-output-format-for-inventory/

docs/storage/s3/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+![Bucket structure](./assets/bucket-structure.jpg)
+
+Data redundancy is needed to reduce requests while doing read operations.
+
+In `data.json` files we could store API, event specs and docs. However, we would need to assure that always the latest file is modified (file versioning). We can also store the files separately, but then the number of requests increases. That's why the cases below assumes that we store the data in single `data.json` file.
```
aszecowka: should `data.json` file be encrypted? IMO yes, and this cause that every modification in this file requires decoding it, applying changes, encoding again and storing on S3. A lot of operation when we compare to simple `insert into` for DB.

docs/storage/s3/README.md
```diff
@@ -0,0 +1,164 @@
...
+- European Union (multiple regions in the European Union) 
+- Asia (multiple regions in Asia)
+- eur4 (Finland and Netherlands) 
+- nam4 (Iowa and South Carolina) 
+
+We cannot replicate a single bucket across multiple regions like Asia, USA and Europe.
+
+### AWS S3
+
+Amazon has [Cross-region replication for S3 buckets](https://docs.aws.amazon.com/AmazonS3/latest/dev/crr.html). One bucket would be used to write, and others to read. It is limited to two regions at the same time (1:1 bucket mapping).
```
aszecowka: We should highlight that data replication for AWS is also very limited.

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. **Efficient data store**
+
+    Very good compression. For example, if CSV file is as big as 1 TB, the data stored in Parquet file has 130 GB.
+
+1. **Good performance**
+    
+    Fetching specific column values doesn't need reading the entire raw data.
+
+### Cons
```
tgorgol: If for some reason our parquet file would be 100 GB large, and we wanted to apply any change to it, I think we would have to:
1. Download 100 GB file
2. Uncompress the file
3. Apply our changes
4. Compress the file
5. Upload 100 GB file to replace the old one

Even if we want to focus on optimizing read operations this seems to me like a pretty big problem, not only because we need to transfer large file two times over the internet, but also because of the implied space requirements. I'm also worried the compression/uncompression operations will be costly.

In case we have to uncompress the whole file, it would take 1 TB of management plane space during this process. If, for example, two or more users decide to modify a label at the same time this number will multiply.

According to answers in [this post](https://stackoverflow.com/questions/28837456/updating-values-in-apache-parquet-file/) it might be possible to uncompress only a part of parquet file but we would either need to partition the file (effectively use multiple parquet files instead of one) or use row groups.

pbochynski:
Reading the evaluation I feel that there is not much gain for us in comparison to the complexity. I think we need to explore other possibilities. I agree that probably NoSQL and document storage would be better suited to the use data model. But please remember that it would be nice to have some persistence also in standalone mode. That was the reason we explored S3 option, because minio gateway mode would help in that case. I am not sure what we can do in that case with Firesotre or Dynamo DB. Argument for Firestore would be that we have everything in one google project. DynamoDB offers global tables that would solve the problem with data replication. But both can't be used in local mode.
I would like to add more options to investigation end evaluate them against requirements. We need to also prioritize requirements, because some of them are nice to have but some of them are essential. My list:
- easy operations/ no operations - super high
- possibility to run locally (standalone, batteries included) - high
- global replication (multi region) - medium/low

docs/storage/s3/parquet.md
pkosiec: On Amazon we can use S3 Select to do queries without need to download the whole file. It is written in the "Time-consuming operations" section.

docs/storage/s3/README.md
pkosiec: As far as I remember, from our set of initial meetings, we don't need to encrypt the Runtime and Application metadata, as there are no secrets, which would be held in vault or as separate, encrypted files.

However, I totally agree with you that it would be time-consuming.

docs/storage/requirements.md
```diff
@@ -0,0 +1,16 @@
+# Requirements
+
+## Storage
+
+- Ability to search Applications and Runtimes by labels (e.g. groups)
+    - List all labels
+    - List labels by key
+    - List applications by label
+- Cross-region replication
```
PK85: - same storage type for Runtime and Management Plane mode

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,66 @@
...
+
+1. **Good performance for reading data**
+    
+    Fetching specific column values doesn't need reading the entire raw data.
+
+### Cons
+
+1. **Doing queries**
+
+    For Golang, we have [parquet-go](https://github.com/xitongsys/parquet-go) library, which handles read and writes of the `parquet` file.
```
mmitoraj:
```suggestion
    For Golang, we have [parquet-go](https://github.com/xitongsys/parquet-go) library, which handles the reading and writing of the `parquet` file.
```

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,66 @@
...
+
+    In order to do queries in Apache Parquet, we would need a query engine, such as [Apache Drill](https://drill.apache.org/), which is written in Java.
+
+    If we wanted to avoid this engine, in our use case, we would need to rethink how to structure the data, to be able to cover our all use cases with just the ability to read a specific column.
```
mmitoraj:
```suggestion
    If we wanted to avoid using this engine. We would need to rethink how to structure the data, to be able to cover all our use cases with just the ability to read a specific column.
```

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,66 @@
...
+
+1. **Concurrent write into the same file**
+
+    We could enable bucket files versioning (for example, in [GCS](https://cloud.google.com/storage/docs/gsutil/addlhelp/ObjectVersioningandConcurrencyControl)). However, we would need to implement a mechanism to always fetch the latest version of the file and retry saving until success.
```
mmitoraj:
```suggestion
    We could enable bucket files versioning (for example, in [GCS](https://cloud.google.com/storage/docs/gsutil/addlhelp/ObjectVersioningandConcurrencyControl)). However, we would need to implement a mechanism to always fetch the latest version of the file and retry saving until it is successful.
```

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,66 @@
...
+ 
+1. **Time-consuming operations**
+
+    For example, every time the user adds an application to a group, we would need to download `parquet` file, modify it and then upload modified version. We can't implement caching as we always need latest version of this file, especially when we will use Cross-region replicas.
```
mmitoraj:
```suggestion
    For example, every time a user adds an application to a group, we would need to download the `parquet` file, modify it and then upload the modified version. We can't implement caching as we always need the latest version of this file, especially when we will use Cross-region replicas.
```

docs/storage/s3/parquet.md
```diff
@@ -0,0 +1,66 @@
...
+
+    However, there is S3 Select available on AWS for [Parquet](https://aws.amazon.com/about-aws/whats-new/2018/09/amazon-s3-announces-new-features-for-s3-select/), which would eliminate the need to download the whole file for read-only operations. Minio supports S3 Select. Still, writing will always mean that we need to download the file.
+
+### Summary
+
+The Apache Parquet is good in different use case, when storing big amount of data with complex structure. In our case, we can store labels using standard S3 capabilities: creating and removing files and directories. We can create a directory structure on S3 to enable querying by groups.
```
mmitoraj:
```suggestion
The Apache Parquet is a good solution for a different use case - when storing a big amount of data with complex structure. In our case, we can store labels using standard S3 capabilities: creating and removing files and directories. We can create a directory structure on S3 to enable querying by groups.
```

Issue #3, opened 2019-05-14
APIs mock implementation


PK85:
**AC**
- [ ] ~~create list of test-cases showing full flow, implement main ones (when test-cases ready ask for review @PK85 )~~
- [ ] provide interfaces for APIs defined in #2  and provide mock implementation in gateway component
- [ ] store data in-memory database
- [ ] interfaces needs to be already nicely defined that will help us in next issues to replace mock with production implementation

PR #9, opened 2019-05-17
Create first GraphQL API specification

pkosiec:
**Description**

- Create first GraphQL API specification
- Add GraphQL server based on GQLGen

**Related issue(s)**
See #3


kfurgol:
Do we need to be compatible with CMS api or app-registry api?

kfurgol:
Remember to change also Blob -> Clob inside labels.go

components/gateway/.gitignore
```diff
@@ -1 +1,2 @@
-/bin
\ No newline at end of file
+vendor
+.idea
```
crabtree: Consider including here also a binary file produced by before-commit.sh script, in case it is not removed.

components/gateway/before-commit.sh
```diff
@@ -22,10 +22,9 @@ if [ "$1" == "$CI_FLAG" ]; then
 	buildEnv="env CGO_ENABLED=0 GOOS=linux GOARCH=amd64"
 fi
 
-${buildEnv} go build -o bin/app ./cmd/
-
+${buildEnv} go build -o gateway ./cmd/main.go
```
crabtree: Currently, this will fail unless you run explicitly `dep ensure -vendor-only`. Is this desired behavior?

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,411 @@
...
+    injectHeaders: HttpHeaders
+    injectQueryParams: QueryParams
+    documentations: [DocumentationInput!]
+    spec: APISpecInput
+
+}
+
+input APISpecInput {
+    type: APISpecType!
+    data: Clob!
```
tgorgol: I think this field shouldn't be required.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,411 @@
...
+# HealthCheck Input
+
+
+enum FilterOperator {
+    ALL, ANY
+}
+
+input LabelFilter {
+    label: String!
+    values: [String]!
```
tgorgol: I think we should use `[String!]!` to be consistent. Also, should `values` field be required? Can there be a case where user wants to find all entites that have a label with specific key, no matter what value it has?

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,411 @@
...
+
+type Query {
+    applications(filer: [LabelFilter!]): [Application!]!
+    application(id: ID!): Application
+
+    runtimes(filter: [LabelFilter!]): [Runtime!]!
+    runtime(id: ID!): Runtime
+
+    healthChecks(types: [HealthCheckType], originID: String): [HealthCheck!]!
+    labels(key: String): [String]!
```
tgorgol: I think we should return `[String!]!` to be consistent.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,411 @@
...
+}
+
+type Query {
+    applications(filer: [LabelFilter!]): [Application!]!
+    application(id: ID!): Application
+
+    runtimes(filter: [LabelFilter!]): [Runtime!]!
+    runtime(id: ID!): Runtime
+
+    healthChecks(types: [HealthCheckType], originID: String): [HealthCheck!]!
```
tgorgol: I think `originID` should be of type `ID` to stay consistent with other IDs in schema.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,411 @@
...
+    deleteApplication(id: ID!): Application
+
+    createApplicationLabel(applicationID: ID!, label: String!, values: [String!]!): [String!]!
+    deleteApplicationLabel(applicationID: ID!, label: String!, values: [String!]!): [String!]!
+
+    createApplicationAnnotation(applicationID: ID!, annotation: String!, value: String!): String
+    deleteApplicationAnnotation(applicationID: ID!, annotation: String!): Boolean
+
+    createApplicationWebhook(applicationID: ID!, in: ApplicationWebhookInput!): ApplicationWebhook
+    updateApplicationWebhook(webhookID: ID!, in: ApplicationWebhookInput): ApplicationWebhook
```
tgorgol: I think `in` field should be required.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,411 @@
...
+    annotations: Annotations
+    webhooks: [ApplicationWebhookInput]
+    healthCheckURL: String
+    apis: [APIInput!]
+    events: [EventInput!]
+}
+
+# Runtime Input
+
+input RuntimeInput {
```
tgorgol: I think optional `description` field should be added.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,411 @@
...
+    # Runtime
+    createRuntime(in: RuntimeInput!): Runtime
+    updateRuntime(id: ID!, in: RuntimeInput!): Runtime
+    deleteRuntime(id: ID!): Runtime
+
+    createRuntimeLabel(runtimeID: ID!, label: String!, values: [String!]!): [String!]!
+    deleteRuntimeLabel(id: ID!, label: String! ,values: [String!]!): [String!]!
+
+    createRuntimeAnnotation(runtimeID: ID!, annotation: String!, value: String!): String
+    deleteRuntimeAnnotation(id: ID!, annotation: String!): Boolean
```
tgorgol: Why are we returning `Boolean` only in case of annotations deletion?

components/gateway/internal/gqlschema/labels.go
```diff
@@ -0,0 +1,100 @@
+package gqlschema
```
tgorgol: If we want to have all custom defined scalars in single file, I think we should rename the file to something like `types.go` or `scalars.go`.

components/gateway/internal/gqlschema/schema.graphql
aszecowka: changed to `[String!]!`.

components/gateway/before-commit.sh
aszecowka: Added dep ensure at the beginning of the file

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,412 @@
...
+# Runtime
+
+type Runtime {
+    id: ID!
+    name: String!
+    description: String
+    tenant: Tenant!
+    labels(key: String): Labels!
+    annotations: Annotations!
+    agentCredential: Credential! # TODO why do we expose it?
```
crabtree: This should be hidden, as the agent will be provisioned with its credentials.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,412 @@
...
+type Mutation {
+    # Application
+    createApplication(in: ApplicationInput!): Application
+    updateApplication(id: ID!, in: ApplicationInput!): Application
+    deleteApplication(id: ID!): Application
+
+    createApplicationLabel(applicationID: ID!, label: String!, values: [String!]!): [String!]!
+    deleteApplicationLabel(applicationID: ID!, label: String!, values: [String!]!): [String!]!
+
+    createApplicationAnnotation(applicationID: ID!, annotation: String!, value: String!): String!
```
tgorgol: Should the returned type be not nullable? Also, are we returning only value (applies also to label mutations)?

PK85:
@aszecowka @pkosiec 
General comment:
1) Update readme.md for gateway, explain how to run application and how to see APIs.

components/gateway/README.md
```diff
@@ -1 +1,14 @@
 # Gateway
+
+Gateway exposes GraphQL API.
```
mmitoraj:
```suggestion
The Gateway exposes GraphQL API.
```

components/gateway/README.md
```diff
@@ -1 +1,14 @@
...
+
+## Development
+
+After introducing changes in GraphQL schema, execute a `gqlgen.sh` script.
```
mmitoraj:
```suggestion
After you introduce changes in the GraphQL schema, run the `gqlgen.sh` script.
```

components/gateway/README.md
```diff
@@ -1 +1,14 @@
...
+To run a Gateway, execute the following command:
```
mmitoraj:
```suggestion
To run the Gateway, use the following command:
```

components/gateway/README.md
```diff
@@ -1 +1,14 @@
...
+
+```
+go run cmd/main.go
+```
+
+GraphQL API playground is available at `localhost:3000` address.
```
mmitoraj:
```suggestion
The GraphQL API playground is available at `localhost:3000`.
```

PR #19, opened 2019-05-29
Allow specify different credentials per runtime

aszecowka:
In this PR, the following issues are addressed:
- allow specifying different credentials for given API that differs between runtimes
- allow specifying authentication headers for fetching specification or documentation via `FetchRequest`:
FetchRequest contains auth field of type Auth (previously it was called Credential). Auth contains credential data, but also additional headers and query params. 
- allow grouping API to find APIs which represents the same API but with different versions:
We have field version in `APIDefinition`, but we cannot express that given API is a successor for another API. To show that 2 API represents the same API but in a different version, I added field `group` on `APIDefinition` level. 

Flow for specifying different credentials: 
Use case: there are three Runtimes already defined
  - create the application - at this point, it is not possible to specify credentials - I would like to have only one flow for providing credentials

```
createApplication(in: ApplicationInput!): Application!
```
1. If the application has configured webhook, it will be called 3 times with the following data:
```
{
  "webhookType": "CREDENTIAL_FOR_RUNTIME",
  "runtimeID": {runtimeID},
  "targetURL": {targetURL},
  "apiID": {apiID} 
}
```

  -  an application set authorization for every runtime by calling 3 times following mutation:
```
setAPIAuth(apiID, runtimeID, auth)
```
apiID and runtimeID were provided by webhook. 

2. If Application does not implement webhooks, an administrator has to periodically query for Application. For every `api` (`APIDefinition`) he gets `auths`. If given runtime has not yet specified `Auth`, `AuthForRuntime` returns `auth = null`. Thanks to that, an administrator know which runtimes are not yet configured and for which he should execute `setAPIAuth`. 

In this PR following issues are not addressed:
- CSRF Token: our API is designed in a way that this token is static. In application-registry it is a URL for fetching it: 
https://github.com/kyma-project/kyma/blob/master/components/application-registry/internal/metadata/model/model.go
https://github.com/kyma-project/kyma/issues/2852

### DIFF
--- a/components/gateway/internal/gqlschema/examples/01_get_apps.graphql
+++ b/components/gateway/internal/gqlschema/examples/01_get_apps.graphql
@@ -19,8 +19,8 @@ query {
         webhooks {
             type
             url
-            credential {
-                data {
+            auth {
+                credential {
                     __typename
                     ... on BasicCredentialData {
                         username
@@ -44,7 +44,7 @@ query {
                 credential {
                     additionalHeaders
                     additionalQueryParams
-                    data {
+                    credential {
                         __typename
                         ... on BasicCredentialData {
                             username
--- a/components/gateway/internal/gqlschema/examples/01_get_apps.graphql
+++ b/components/gateway/internal/gqlschema/examples/01_get_apps.graphql
@@ -1,6 +1,6 @@
 query {
     applications(
-        filer: [
+        filter: [
             { label: "group", values: ["production", "experimental"], operator: ANY },
             {label: "region", values:["us"]}
         ]
@@ -22,7 +22,7 @@ query {
             credential {
                 data {
                     __typename
-                   ... on BasicCredentialData {
+                    ... on BasicCredentialData {
                         username
                         password
                     }
@@ -38,59 +38,66 @@ query {
         apis {
             id
             targetURL
-            credential {
-                data {
-                    __typename
-                    ... on BasicCredentialData {
-                        username
-                        password
-                    }
-                    ... on OAuthCredentialData {
-                        clientId
-                        clientSecret
-                        url
-                    }
-                }
-            }
-            spec {
-                type
-                data
-            }
-            additionalHeaders
-            additionalQueryParams
-            version {
-                value
-                deprecated
-                deprecatedSince
-                forRemoval
-            }
-        }
-        eventAPIs {
-            id
-            spec {
-                type
-                data
-                fetchRequest {
-                    url
-                    credential {
-                        data {
-                            __typename
-                            ... on BasicCredentialData {
-                                username
-                                password
-                            }
-                            ... on OAuthCredentialData {
-                                clientId
-                                clientSecret
-                                url
-                            }
+            credentials {
+                runtimeID
+
+                credential {
+                    additionalHeaders
+                    additionalQueryParams
+                    data {
+                        __typename
+                        ... on BasicCredentialData {
+                            username
+                            password
+                        }
+                        ... on OAuthCredentialData {
+                            clientId
+                            clientSecret
+                            url
                         }
                     }
                 }
             }
-            version {
-                value
-            }
+        }
+        spec {
+            type
+            data
+        }
+        additionalHeaders
+        additionalQueryParams
+        version {
+            value
+            deprecated
+            deprecatedSince
+            forRemoval
         }
     }
+    eventAPIs {
+        id
+        spec {
+            type
+            data
+            fetchRequest {
+                url
+                credential {
+                    data {
+                        __typename
+                        ... on BasicCredentialData {
+                            username
+                            password
+                        }
+                        ... on OAuthCredentialData {
+                            clientId
+                            clientSecret
+                            url
+                        }
+                    }
+                }
+            }
+        }
+        version {
+            value
+        }
+    }
+}
 }
--- a/components/gateway/internal/gqlschema/examples/01_get_apps.graphql
+++ b/components/gateway/internal/gqlschema/examples/01_get_apps.graphql
@@ -38,10 +38,9 @@ query {
         apis {
             id
             targetURL
-            credentials {
+            auths {
                 runtimeID
-
-                credential {
+                auth {
                     additionalHeaders
                     additionalQueryParams
                     credential {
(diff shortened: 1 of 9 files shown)


components/gateway/internal/gqlschema/examples/01_get_apps.graphql
```diff
@@ -1,6 +1,6 @@
 query {
```
aszecowka: dear reviewer: please ignore examples for the first time round, I will update it when we agree on the schema.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -51,8 +51,9 @@ type Application {
     status: ApplicationStatus!
     webhooks: [ApplicationWebhook!]!
     healthCheckURL: String
-    apis: [APIDefinition!]!
-    eventAPIs: [EventAPIDefinition!]!
+    """ groupName allows to find different versions of the same API """
```
pkosiec: The parameter name should be equal to the property in `APIDefinition`, so in this case: `group`. However, I don't like the name, as it sounds like it could contain different APIs, not just the same APIs with different versions.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -89,19 +90,24 @@ type Version {
...
     id: ID!
     spec: APISpec
     targetURL: String!
-    credential: Credential
-    additionalHeaders: HttpHeaders #
-    additionalQueryParams: QueryParams
+    """ group allows you to find the same API but in different version """
+    group: String
+    auth(runtimeID: ID!): Auth
+    auths: [AuthForRuntime!]
```
pkosiec: To keep consistency, please change it to `[AuthForRuntime!]!`

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -89,19 +90,24 @@ type Version {
...
     version: Version
 }
 
+type AuthForRuntime {
```
pkosiec: Maybe `RuntimeAuth` would be a better name?

components/gateway/internal/gqlschema/schema.graphql
tgorgol: Should this comment also be added to `eventAPIs` field?

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -395,8 +416,13 @@ type Mutation {
     deleteAPI(id: ID!): APIDefinition
     refetchAPISpec(apiID: ID!): APISpec
 
-    setAPICredential(apiID: ID!, in: CredentialInput!): Credential!
-    deleteAPICredential(apiID: ID!): Credential
+    """
+    If runtimeID not provided, the samec auth will be set for every runtime already available
```
tgorgol: typo (`samec`)

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -89,19 +91,26 @@ type Version {
...
 type APIDefinition {
     id: ID!
     spec: APISpec
     targetURL: String!
-    credential: Credential
-    additionalHeaders: HttpHeaders #
-    additionalQueryParams: QueryParams
+    """ group allows you to find the same API but in different version """
+    group: String
+    """" If runtime does not exist, error will be returned. If runtime exist but does not have Auth defined, nil is returned """
```
pkosiec:
```suggestion
    """"If runtime does not exist, an error will be returned. If runtime exists but does not have Auth defined, null is returned"""
```

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -89,19 +91,26 @@ type Version {
...
+    auth(runtimeID: ID!): Auth
+    """ Auths returns information for all runtimes, even for given runtime Auth is not yet specified """
```
pkosiec:
```suggestion
    """Returns authentication details for all runtimes, even for a runtime, where Auth is not yet specified"""
```

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -395,8 +419,13 @@ type Mutation {
     deleteAPI(id: ID!): APIDefinition
     refetchAPISpec(apiID: ID!): APISpec
 
-    setAPICredential(apiID: ID!, in: CredentialInput!): Credential!
-    deleteAPICredential(apiID: ID!): Credential
+    """
```
pkosiec: Screen Shot 2019-05-30 at 10 17 58

This is how it looks in Playground. We should use sentences ended with a dot.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -395,8 +419,13 @@ type Mutation {
...
+    If runtimeID not provided, the same auth will be set for every runtime already available
```
pkosiec:
```suggestion
    If runtimeID is not provided, the same authentication details will be set for every runtime already available.
```

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -395,8 +419,13 @@ type Mutation {
...
+    Returns information for which runtime it was configured
```
pkosiec:
```suggestion
    Returns information for which runtime it was configured.
```

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -89,19 +91,26 @@ type Version {
...
     id: ID!
     spec: APISpec
     targetURL: String!
-    credential: Credential
-    additionalHeaders: HttpHeaders #
-    additionalQueryParams: QueryParams
+    """ group allows you to find the same API but in different version """
+    group: String
+    """"If runtime does not exist, an error will be returned. If runtime exists but does not have Auth defined, null is returned"""
+    auth(runtimeID: ID!): Auth
```
tgorgol: I think having an additional default `Auth` for each `APIDefinition` would be preferable. In that case, `runtimeID` would be optional, and when no `runtimeID` is specified or runtime with specified `runtimeID` wasn't configured with specific `Auth` the default `Auth` would be returned.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -274,19 +287,26 @@ input FetchRequestInput {
...
-    credential: CredentialInput
+    auth: AuthInput
 }
 
 # API Input
-
+# you need to perform separate call to set auth
 input APIDefinitionInput {
     targetURL: String!
-    credential: CredentialInput
```
tgorgol: If there would be a default `Auth` like I suggested in my previous comment, we could add back optional `auth` field to allow users to specify the default `Auth`.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -395,8 +419,13 @@ type Mutation {
     deleteAPI(id: ID!): APIDefinition
     refetchAPISpec(apiID: ID!): APISpec
 
-    setAPICredential(apiID: ID!, in: CredentialInput!): Credential!
-    deleteAPICredential(apiID: ID!): Credential
+    """
+    If runtimeID is not provided, the same authentication details will be set for every runtime already available.
+    Returns information for which runtime it was configured.
+    """
+    setAPIAuth(apiID: ID!, runtimeID: ID, in: AuthInput!): [RuntimeAuth!]!
```
tgorgol: If there would be a default `Auth` like I suggested in one of my previous comments, providing no `runtimeID` here would set the mentioned default `Auth`.

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -395,8 +419,13 @@ type Mutation {
...
+    deleteAPIAuth(apiID: ID!, runtimeID: ID): [RuntimeAuth!]!
```
tgorgol: If there would be a default `Auth` like I suggested in one of my previous comments, providing no `runtimeID` here would delete the mentioned default `Auth`.

components/gateway/internal/gqlschema/schema.graphql
aszecowka: I added separateField defaultAuth, let me know what you think!

pkosiec:
I like the idea of `defaultAuth` 👍

components/gateway/internal/gqlschema/schema.graphql
tgorgol: I like this approach as well, very nice :) 👍

Issue #15, opened 2019-05-28
Decide how components communicate each other


PK85:
**AC**
- [ ] proposal and decision about what is a list of components and how they communicate each other 
  - [ ] decision about where we will have graphQL layer, 
    - [ ] in dedicated service (not gateway)
    - [ ] in every service
  - [ ] decision about what gateway responsibility is about (e.g. just token validation)
  - [ ] decision about protocol that services will communicate each other

PR #21, opened 2019-05-30
Prepare PoC for GraphQL Gateway

pkosiec:
**Description**

- Prepare PoC for GraphQL Gateway
- Test gRPC communication on Kyma with sidecars enabled

See the `README-PoC.md` file for instructions (link: https://github.com/pkosiec/compass/blob/gql-poc/README-PoC.md)

**Related issue(s)**
#15


README-PoC.md
```diff
@@ -0,0 +1,72 @@
...
+1. Run the following commands:
+```bash
+GATEWAY_ADDRESS=127.0.0.1:3000 DIRECTOR_ADDRESS=127.0.0.1:4000 go run components/gateway/cmd/main.go
+DIRECTOR_ADDRESS=127.0.0.1:4000 go run components/director/cmd/main.go
+```
+1. Navigate to [http://localhost:3000/](http://localhost:3000/)
+
+## Deploy on Kyma
+
+The PoC can be easily deployed on Kyma. Both Gateway and Director have sidecars injected.
```
tgorgol: Maybe it would be good to specify what sidecars are being injected.
```suggestion
The PoC can be easily deployed on Kyma. Both Gateway and Director have istio sidecars injected.
```

README-PoC.md
```diff
@@ -0,0 +1,71 @@
...
+    labels
+    apis {
+      id
+      targetURL
+    }
+  }
+}
+```
+
+Result: Two resolvers will be triggered: applications (once) + apis (two times) (3 requests in total).
```
tgorgol: Maybe we could point out here that the applications call returns two applications and because of that apis gets called twice (once for each returned application)?

README-PoC.md
aszecowka: istio -> Istio

README-PoC.md
```diff
@@ -0,0 +1,71 @@
+# GraphQL Gateway PoC
```
aszecowka: Add information how to handle subscriptions in that architecture. grpc has a streaming concept so we can mention about that.

README-PoC.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Description
```
aszecowka: Please add information how we synchronize code between 2 components: currently, we have a copy of `director.pb.go` in 2 places thanks to:
```
//go:generate protoc --go_out=plugins=grpc:. director.proto
//go:generate protoc --go_out=plugins=grpc:../../gateway/protobuf/ director.proto
```

I really like this idea but we should describe it's advantages.

README-PoC.md
```diff
@@ -0,0 +1,71 @@
...
+
+This proof of concept contains Gateway GraphQL server, which communicates via gRPC with Director. Gateway delegates all operations to Director, which contains actual business logic.
```
aszecowka: We should think about how this PR will be merged into the master. Currently, we cannot merge code, because we don't know if we accept the described solution. Maybe you should describe steps that need to be performed to test the described solution?

README-PoC.md
```diff
@@ -0,0 +1,71 @@
...
+
+From technical point of view, Gateway utilizes connection pool to be easily scalable. gRPC underneath uses HTTP/2 server-side push, so there is no need to reconnect for every request.
+
+## Run locally
```
aszecowka: even if we reuse connections, with many requests, provided solution seems to be not the most efficient:
I queried for:
- 100 apps, each has 10 APIs: response time was ~0.04s
- 100 apps without querying about APIs: 0.01s
- 1000 apps, each has 10 APIs: response time ~ 0.5s
- 1000 apps, without querying about APIs: 0.04s

This was tested on the localhost.

README-PoC.md
pkosiec: I think it is not necessary, as @tgorgol document is pointing on this PR. I don't want to merge it at all, so the README I wrote is done only to instruct, how to launch this PoC.

README-PoC.md
pkosiec: Although we would go with some pagination, I agree, that's another downsize of this solution.

pkosiec:
I'm closing this, as it shouldn't be merged into `master`. However, this PR is referenced in the @tgorgol investigation summary.

PR #26, opened 2019-05-30
Add high level diagrams

aszecowka:
In this PR document that describes communication between Runtime, Application and Management Plane was added. In addition to that, grouping concept was described.


docs/architecture/high-level-components.md
```diff
@@ -0,0 +1,23 @@
...
+To keep Runtimes in-sync with the configuration, an agent installed on a Runtime is responsible for fetching actual configuration. 
+In case an Application has configured optional Webhooks, Management Plane notify an Application about any events that concern given Application.
+Management Plane does not participate in any business flow. After establishing a trusted connection between Application
+and Runtime, they communicate directly, with omitting Management Plane. 
+
+## Grouping
+In order to connect Applications and Runtimes, assign it to the same Group.
+A Group is a simple label with the key `group`. If Application or Runtime is not explicitly assigned to any Group, it belongs to the `default` Group.
+If you add a Group to the Application or Runtime assigned to the `default` group, it is automatically removed from the `default` group and moved to the new one.
+![](./assets/grouping.svg)
```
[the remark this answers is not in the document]
pkosiec: BTW I don't think we can use rounded rectangles here. BTW do we need `LABEL:` in every rectangle? The diagram describes labelling, so I think we can remove it.

docs/architecture/high-level-components.md
```diff
@@ -0,0 +1,23 @@
...
+An administrator adds Runtimes, Applications and configures them through Management Plane GraphQL API. 
+To keep Runtimes in-sync with the configuration, an agent installed on a Runtime is responsible for fetching actual configuration. 
+In case an Application has configured optional Webhooks, Management Plane notify an Application about any events that concern given Application.
+Management Plane does not participate in any business flow. After establishing a trusted connection between Application
+and Runtime, they communicate directly, with omitting Management Plane. 
+
+## Grouping
+In order to connect Applications and Runtimes, assign it to the same Group.
+A Group is a simple label with the key `group`. If Application or Runtime is not explicitly assigned to any Group, it belongs to the `default` Group.
+If you add a Group to the Application or Runtime assigned to the `default` group, it is automatically removed from the `default` group and moved to the new one.
```
pkosiec: Please add an empty line before and after every diagram

docs/architecture/high-level-components.md
pkosiec: Hmm, Maybe it would be good to include Runtime3, which has label `group=default`. It could be not connected with any app, but the diagram will show the concept with default group.

docs/architecture/high-level-components.md
```diff
@@ -0,0 +1,23 @@
+# Overview
+
+This document describes communication between Management Plane, Application and Runtimes. 
+
+## Components
+Management Plane is a central place which stores configuration about Applications and Runtimes and is responsible for propagating that information to Applications and Runtimes. 
+Management Plane plays crucial role in establishing trusted connection between Applications and Runtimes. 
+![](./assets/components-high-level.svg)
```
pkosiec: Hmm, I think this diagram is a bit too complicated on the left side: Application <-> Management Plane relation. I would simplify it in the following way:

- Remove webhook from this diagram, as IMHO it should be as low level of abstraction as possible
- Remove "add runtime, add application, configure" verbs from Admin and replace it with something like "manage applications and runtimes", or even simpler "manage"
- Does app really configure Management Plane? That's what I read it from the diagram. The arrow should be on the opposite side, right? Because Management Plane passes configuration to Application and Management Plane set-ups trust between App and Runtime. 
- Sync between Agent and Management Plane is two way, right? So the arrows should be on both sizes of the line
-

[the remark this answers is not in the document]
aszecowka: Admin got fat, are you happy now?

docs/architecture/high-level-components.md
```diff
@@ -0,0 +1,27 @@
+# Overview
+
+This document describes communication between Management Plane, Application and Runtimes.
```
tgorgol:
```suggestion
This document describes communication between Management Plane, Applications and Runtimes. 
```

docs/architecture/high-level-components.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Components
+Management Plane is a central place which stores configuration about Applications and Runtimes and is responsible for propagating that information to Applications and Runtimes. 
+Management Plane plays crucial role in establishing trusted connection between Applications and Runtimes. 
+
+![](./assets/components-high-level.svg)
+
+An administrator adds Runtimes, Applications and configures them through Management Plane GraphQL API. 
+To keep Runtimes in-sync with the configuration, an agent installed on a Runtime is responsible for fetching actual configuration. 
+An Agent sends information about Runtime healthchecks to Management Plane.
```
tgorgol:
```suggestion
An Agent sends information about Runtime health checks to Management Plane. 
```

docs/architecture/high-level-components.md
```diff
@@ -0,0 +1,27 @@
...
+## Components
+Management Plane is a central place which stores configuration about Applications and Runtimes and is responsible for propagating that information to Applications and Runtimes. 
+Management Plane plays crucial role in establishing trusted connection between Applications and Runtimes. 
+
+![](./assets/components-high-level.svg)
+
+An administrator adds Runtimes, Applications and configures them through Management Plane GraphQL API. 
+To keep Runtimes in-sync with the configuration, an agent installed on a Runtime is responsible for fetching actual configuration. 
+An Agent sends information about Runtime healthchecks to Management Plane. 
+In case an Application has configured optional Webhooks, Management Plane notify an Application about any events that concern given Application.
```
tgorgol:
```suggestion
In case an Application has configured optional Webhooks, Management Plane notifies an Application about any events that concern given Application.
```

docs/architecture/high-level-components.md
```diff
@@ -0,0 +1,27 @@
...
+Management Plane does not participate in any business flow. After establishing a trusted connection between Application
+and Runtime, they communicate directly, with omitting Management Plane. 
+
+## Grouping
+In order to connect Applications and Runtimes, assign it to the same Group.
+A Group is a simple label with the key `group`. If Application or Runtime is not explicitly assigned to any Group, it belongs to the `default` Group.
+If you add a Group to the Application or Runtime assigned to the `default` group, it is automatically removed from the `default` group and moved to the new one.
+
+![](./assets/grouping.svg)
```
tgorgol: After the changes to this diagram, the red arrow still seems a bit confusing to me, because, for example, `App1` can't communicate with `Runtime2` or `Runtime3` either and yet there's no red arrow connecting them. To understand what's happening reader has to read the whole next paragraph. Maybe that's just me, though.

crabtree: I agree, for me, this red arrow is also a bit confusing.

crabtree:
In the grouping diagram, I would add also arrows pointing to the applications from the runtimes to underline the bi-directional communication.

docs/architecture/high-level-components.md
```diff
@@ -0,0 +1,29 @@
+# High Level Components
```
PK85: Please add an entry (as a first one) in /docs/README.md inside architecture section

PR #29, opened 2019-06-03
API: Add Pagination and fix CSRF

aszecowka:
**Description**

- pagination:
I added pagination to all root level queries for applications, runtimes and healtchecks. I added pagination for `Application.apis`, `Application.eventAPIs` and `Documents` because all of them can have many entries, and possibly can contain a lot of data (`data: CLOB`). IMO it does not make sense to add pagination everywhere (like `APIDefinition.auths`, `Application.webhooks`) because these collections will be probably small or contain a small amount of data.
In `19_pagination.grapghql` you can find how to interact with API
- CSRF token: instead of specifying token, you can provide tokenEndpointURL that can be protected with auth. I removed field `type` to be consistent with `Auth` type where we also removed `type` field.
- `setAPIAuth(apiID: ID!, runtimeID: ID!, in: AuthInput!): RuntimeAuth!`
   `deleteAPIAuth(apiID: ID!, runtimeID: ID!): RuntimeAuth!`
If we have to specify apiID, runtimeID, then we modify only one RuntimeAuth so I modified returned type to single object instead of list. The same for `delete`

### DIFF
--- a/components/gateway/internal/gqlschema/examples/19_pagination.graphql
+++ b/components/gateway/internal/gqlschema/examples/19_pagination.graphql
@@ -1,3 +1,15 @@
+fragment givenPage on ApplicationPage {
+    data
+    {
+        name
+    }
+    totalCount
+    pageInfo {
+        startCursor
+        endCursor
+        hasNextPage
+    }
+}
 
 query {
      firstPage: applications(
@@ -7,16 +19,7 @@ query {
         ],
          first: 10, # return only 10 items. First == SQL limit
     ) {
-        data
-        {
-          name
-        }
-        totalCount
-        pageInfo {
-            startCursor
-            endCursor
-            hasNextPage
-        }
+       ...givenPage
     }
 
      nextPage: applications(
@@ -27,15 +30,6 @@ query {
         first: 10,
         after: "aW5kZXg9MA==" # returned from previous call pageInfo.endCursor
     ) {
-        data
-        {
-            name
-        }
-        totalCount
-        pageInfo {
-            startCursor
-            endCursor
-            hasNextPage
-        }
+       ...givenPage
     }
 }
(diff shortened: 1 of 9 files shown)


components/gateway/internal/gqlschema/examples/19_pagination.graphql
```diff
@@ -0,0 +1,41 @@
+
+query {
```
pkosiec: We could use fragments here

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -14,6 +14,8 @@ scalar QueryParams # -> map[string][]string
 
 scalar CLOB # TBD
 
+scalar PageCursor
```
pkosiec: It should be commented that this is just a string. Because scalar can represent dynamic structure as well

components/gateway/internal/gqlschema/schema.graphql
```diff
@@ -52,10 +54,60 @@ type Application {
...
+    documents(first: Int = 100, after: PageCursor): DocumentPage!
+}
+
+""" Every query that implements pagination returns object that implements Pageable interface.
+To specify page details, query specify two parameters: `first` and `after`.
+`first` specify page size, `after` is a cursor for the next page. When requesting first page, set `after` to empty value.
+For requesting next page, set `after` to `pageInfo.endCursor` returned from previous query. """
+interface Pageable {
+    pageInfo: PageInfo!
+    totalCount: Int!
```
pkosiec: Do we really need `totalCount` property? Relay-style pagination doesn't have one. GraphQL client also doesn't need it - `hasNextPage` is sufficient.

aszecowka: having `totalCount` is safer IMO. I can imagine that there will be requirements in the UI to display a number of pages.

PR #32, opened 2019-06-04
Initial connector component with GQL schema

crabtree:
**Description**

- The initial draft for connector GraphQL schema
- Connector component structure

**Related issue(s)**


components/connector/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,49 @@
+# Certificate
```
PK85: Could you already think and propose a concept of generic connector APIs. I mean for example JWT too.

crabtree: Sure, will do.

components/connector/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,59 @@
...
+type Query {
+    # Client-Certificates
+    
+    # Retuns subject that should be placed in the signing request
+    getCSRInfo(token: String): CSRInfo!
+}
+
+type Mutation {
+    # Applications
+    getApplicationToken(appID: ID!): Token! # internal?
```
pkosiec: hmm, maybe `generate` would be a better name. At first I thought that this is simple query, but of course it generates the token under the hood.

components/connector/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,59 @@
...
+
+# CSR
+type CSR {
+    subject: String! # eg.: "OU=Test,O=Test,L=Blacksburg,ST=Virginia,C=US,CN={ID}"
+    keyAlgorithm: String! # eg.: rsa2048
+}
+
+# CSRInfo
+type CSRInfo {
+    csrURL: String! # eg.: "https://connector-service.test.cluster.kyma.cx/v1/runtimes/certificates?token=1edfc34g"
```
pkosiec: hmm, maybe we should put this URL into `CSR` type.

pkosiec: Also I would rather not use abbreviations for type names. I would go with the `CertificateSigningRequest` name

components/connector/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,59 @@
...
+
+# Certificate
+type Certificate {
+    crt: String!
+    caCrt: String!
+    clientCrt: String!
+}
+
+# MPInfo
+type MPInfo {
```
pkosiec: hmm, what about `ManagementPlaneConfiguration`, `ManagementPlaneDetails`, or something like that?

components/connector/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,59 @@
+# Token
+type Token {
+    url: String! # eg.: "https://connector-service.test.cluster.kyma.cx/v1/applications/info?token=1edfc34g"
+    token: String! # eg.: "1edfc34g"
+}
+
+# Certificate
+type Certificate {
```
pkosiec: I would go with full names:

```graphql
certificate: String!
certificateAuthority: String!
clientCertificate: String!
```

components/connector/internal/gqlschema/schema.graphql
crabtree: After further considerations, this URL is not relevant here, so I remove it.

components/connector/internal/gqlschema/schema.graphql
crabtree: I changed it slightly, please take a look

components/connector/internal/gqlschema/schema.graphql
crabtree: I changed it to ManagementPlaneInfo

components/connector/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,50 @@
...
+type CertificateSigningRequestInfo {
+    managementPlaneInfo: ManagementPlaneInfo!
+    subject: String! # eg.: "OU=Test,O=Test,L=Blacksburg,ST=Virginia,C=US,CN={ID}"
+    keyAlgorithm: String! # eg.: rsa2048
+}
+
+type Query {
+    # Client-Certificates
+    
+    # Retuns subject that should be placed in the signing request
```
pkosiec: Returns

pkosiec: Please write comments in `"""comment"""` form, to include it in generated API documentation

components/connector/internal/gqlschema/schema.graphql
```diff
@@ -0,0 +1,50 @@
+# Token
+type Token {
+    token: String! # eg.: "1edfc34g"
+}
+
+# Certificates
+type Certificates {
```
pkosiec: I would think about better name for the type (I don't like the plural form), but `CertificationDetails` is all I can made up for now 😄 and I don't think it's is better than yours, so...

crabtree: Renamed to `CertificationResult`

PR #41, opened 2019-06-06
Do not filter headers on Gateway while proxying requests

### LINKED ISSUE #35: Prepare simple Gateway implementation
**Description**

- [ ] Prepare simple Reverse Proxy implementation for Gateway.
- [ ] Configure charts for Gateway and Director, to work properly (pass requests to Director)

**Reasons**

pkosiec:
**Description**

- Do not filter headers while proxying requests

**Related issue(s)**
Resolves #35

### DIFF
--- a/components/director/Dockerfile
+++ b/components/director/Dockerfile
@@ -9,13 +9,13 @@ WORKDIR ${BASE_APP_DIR}
 
 COPY ./internal/ ${BASE_APP_DIR}/internal/
 COPY ./vendor/ ${BASE_APP_DIR}/vendor/
-COPY ./cmd/main.go ${BASE_APP_DIR}/
+COPY ./cmd/ ${BASE_APP_DIR}/cmd/
 
 #
 # Build app
 #
 
-RUN go build -v -o main .
+RUN go build -v -o main ./cmd/main.go
 RUN mkdir /app && mv ./main /app/main
 
 FROM alpine:3.9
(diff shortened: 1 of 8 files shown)


components/director/Dockerfile
```diff
@@ -9,7 +9,7 @@ WORKDIR ${BASE_APP_DIR}
 
 COPY ./internal/ ${BASE_APP_DIR}/internal/
 COPY ./vendor/ ${BASE_APP_DIR}/vendor/
-COPY ./cmd/main.go ${BASE_APP_DIR}/
+COPY ./cmd/ ${BASE_APP_DIR}/cmd/
```
crabtree: You should also change the line with `go build ...`

components/gateway/cmd/main.go
```diff
@@ -25,23 +23,11 @@ func main() {
...
-			r.URL.Path = singleJoiningSlash(directorUrl.Path, r.URL.Path)
-			r.Header = http.Header{
-				"Content-Type": []string{"application/json"},
-			}
-			r.Host = directorUrl.Host
-		},
-	}
-
 	router := mux.NewRouter()
 	router.HandleFunc("/", directorProxy.ServeHTTP)        // GraphQL Playground
```
crabtree: Isn't it enough to only handle requests on `/` as the handler function is the same?

pkosiec: Hmm, I was thinking about forwarding requests to Connector and Healthchecker - that's why I specified exact paths to support future cases.

PR #45, opened 2019-06-07
Detailed postgres investigation with examples

aszecowka:
- validating if we can execute performant queries against our SQL schema
- validating PostgresJSON


docs/investigations/postgres/postgres.md
```diff
@@ -0,0 +1,97 @@
+# PostgreSQL Evaluation
```
PK85: Add this document under docs/README.md

Issue #18, opened 2019-05-28
Support Application Connector service REST API in the API-gateway


PK85:
**AC**
- [x] Define API as a graphQL (CERT support) and add to Connector Component (JWT skipped, for now, we need additional research) @crabtree  PR: #32 
- architecture diagram added
  - [x] REST to graphQL @crabtree PR: #42 
  - [ ] for many types of auth, like CERT, JWT
- [ ] Implementation on gateway side for REST and proxying to GraphQL

PR #59, opened 2019-06-11
Expose connector through gateway proxy

crabtree:
**Description**

- Host connector through the gateway proxy

**Related issue(s)**

See also #18


chart/compass/values.yaml
```diff
@@ -2,12 +2,15 @@ global:
   images:
     containerRegistry:
       path: eu.gcr.io/kyma-project/incubator
+    connector:
+      dir: pr/
+      version: "PR-32"
```
pkosiec: It doesn't work - you have to modify Connector (parametrize Playground API endpoint), similarly to Director. Just copy `main.go` from `director`, adjust it if necessary and it'll work.

chart/compass/values.yaml
```diff
@@ -2,12 +2,15 @@ global:
...
     director:
       dir: pr/
       version: "PR-39"
```
pkosiec: Please change the image of Director to `PR-41`. Apparently I forgot to change it 😞

crabtree: Done

chart/compass/values.yaml
crabtree: Fixed

Issue #43, opened 2019-06-06
Prepare Director mocks for Runtime domain


tgorgol:
**Description**

- [ ] Prepare interfaces for Runtime API defined in #2
- [ ] Prepare mocks that use an in-memory database

PR #60, opened 2019-06-11
Prepare mocks for Runtime domain

pkosiec:
**Description**

- Prepare mocks for Runtime domain

You can test it:

```graphql
mutation {
  createRuntime(in:{
    name:"Test",
    description:"Test21",
    labels: {
      test: ["hello", "there"],
    },
    annotations:{
      key:value,
    }
  }) {
    id
    description
    name
  }
}
```
and next:

```graphql
# Write your query or mutation here
{
  runtimes {
    data {
      id
      name
      description
      tenant
      labels
      annotations
    }
    pageInfo {
      startCursor
      endCursor
      hasNextPage
    }
    totalCount
  }
}
```

**Related issue(s)**
See #43


components/director/internal/domain/runtime/converter.go
```diff
@@ -1,4 +1,55 @@
 package runtime
 
+import "github.com/kyma-incubator/compass/components/director/internal/graphql"
+
 type Converter struct {
 }
+
+func (c *Converter) ToGraphQL(in *RuntimeModel) *graphql.Runtime {
+	return &graphql.Runtime{
+		ID:     in.ID,
```
aszecowka: what about other fields, like name, description, tenant etc?

components/director/internal/domain/runtime/converter.go
```diff
@@ -1,4 +1,55 @@
...
+
+func (c *Converter) statusToGraphQL(in *RuntimeModelStatus) *graphql.RuntimeStatus {
+	var condition graphql.RuntimeStatusCondition
+	switch in.Condition {
+	case RuntimeStatusConditionInitial:
+		condition = graphql.RuntimeStatusConditionInitial
+		break
+	case RuntimeStatusConditionFailed:
+		condition = graphql.RuntimeStatusConditionFailed
+		break
```
aszecowka: Go's switch statements break implicitly

components/director/internal/domain/runtime/converter.go
```diff
@@ -1,4 +1,55 @@
...
+	case RuntimeStatusConditionReady:
+		condition = graphql.RuntimeStatusConditionReady
+		break
+	default:
+		condition = graphql.RuntimeStatusConditionFailed
```
aszecowka: if it is not set, it should be rather UNKNOWN, or INITIAL

components/director/internal/domain/runtime/model.go
```diff
@@ -0,0 +1,107 @@
+package runtime
+
+import (
+	"fmt"
+	"time"
+)
+
+type RuntimeModel struct {
+	ID          string              `json:"id"`
```
aszecowka: what is the purpose of JSON tags?

components/director/internal/domain/runtime/model.go
```diff
@@ -0,0 +1,107 @@
...
+	// directive for checking auth
+	AgentAuth interface{} `json:"agentAuth"`
+}
+
+type RuntimeModelStatus struct {
+	Condition runtimeModelStatusCondition `json:"condition"`
+	Timestamp time.Time                   `json:"timestamp"`
+}
+
+func NewRuntimeModel(ID string, name string, description *string, tenant string, labels map[string][]string, annotations map[string]string) *RuntimeModel {
```
aszecowka: usage of this constructor is more difficult than creating RuntimeModel object itself

components/director/internal/domain/runtime/model.go
```diff
@@ -0,0 +1,107 @@
...
+
+const (
+	RuntimeStatusConditionInitial runtimeModelStatusCondition = "INITIAL"
+	RuntimeStatusConditionReady   runtimeModelStatusCondition = "READY"
+	RuntimeStatusConditionFailed  runtimeModelStatusCondition = "FAILED"
+)
+
+func (r *RuntimeModel) AddLabel(key string, values []string) error {
+	if _, exists := r.Labels[key]; exists {
+		return fmt.Errorf("Label %s does already exist", key)
```
aszecowka: Intellij says: Error string should not be capitalized, please fix it in other places also.

components/director/internal/domain/runtime/model.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	r.Labels[key] = values
+	return nil
+}
+
+func (r *RuntimeModel) AppendLabelValues(key string, values []string) error {
+	if _, exists := r.Labels[key]; exists {
+		return fmt.Errorf("Label %s does already exist", key)
```
aszecowka: if I append, label should already exist IMO

components/director/internal/domain/runtime/model.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (r *RuntimeModel) DeleteLabelValues(key string, valuesToDelete []string) error {
+	currentValues, exists := r.Labels[key]
+
+	if !exists {
+		return fmt.Errorf("Label %s doesn't exist", key)
+	}
+
+	var filteredValues []string
```
aszecowka: if you create map on the basis of `valuesToDelete` this method will be simpler and more performant.

components/director/internal/domain/runtime/service.go
```diff
@@ -1,8 +1,165 @@
...
+	return s.repo.Delete(r)
+}
+
+func (s *Service) AddLabel(ctx context.Context, runtimeID string, key string, values []string) error {
+	r, err := s.Get(ctx, runtimeID)
+	if err != nil {
+		return err
+	}
+
+	if _, exists := r.Labels[key]; exists {
```
aszecowka: IMO it will be simpler if on the Runtime object you will have one method for adding labels in both cases.

components/director/internal/domain/runtime/service.go
```diff
@@ -1,8 +1,165 @@
...
+	err = r.DeleteLabel(key)
+	if err != nil {
+		return errors.Wrapf(err, "while deleting label with key %s", key)
+	}
+
+	return nil
+}
+
+func (s *Service) AddAnnotation(ctx context.Context, runtimeID string, key string, value string) error {
+	r, err := s.Get(ctx, runtimeID)
```
aszecowka: short names are cool, but sometimes they are too short. Maybe instead of r, we can use `model` name, or `rtm`?

components/director/internal/domain/runtime/service.go
```diff
@@ -1,8 +1,165 @@
...
+		return errors.Wrapf(err, "while updating Runtime")
+	}
+
+	return nil
+}
+
+func (s *Service) DeleteLabel(ctx context.Context, runtimeID string, key string, values []string) error {
+	r, err := s.Get(ctx, runtimeID)
+	if err != nil {
+		return err
```
aszecowka: why sometimes you wrap errors and sometimes you don't do that?

components/director/internal/domain/runtime/converter.go
```diff
@@ -1,4 +1,55 @@
...
+func (c *Converter) ToGraphQL(in *RuntimeModel) *graphql.Runtime {
+	return &graphql.Runtime{
+		ID:     in.ID,
+		Status: c.statusToGraphQL(in.Status),
+		//Annotations:in.Annotations, //TODO: Wait for scalar unmarshalling
+		//Labels:in.Labels, //TODO: Wait for scalar unmarshalling
+	}
+}
+
+func (c *Converter) MultipleToGraphQL(in []*RuntimeModel) []*graphql.Runtime {
```
aszecowka: Where can I find tests?

pkosiec: The tests have been pushed while you were doing the review 😛

components/director/internal/domain/healthcheck/repository.go
```diff
@@ -0,0 +1,13 @@
+package healthcheck
+
+import (
+	"github.com/kyma-incubator/compass/components/director/internal/model"
+)
+
+type Repository struct {
```
aszecowka: do we have to export this type? Maybe we should call it `inMemoryRepository` because when we introduce real repository, we may want to keep both repos (in memory and sql) at the same time for some time.

components/director/internal/domain/runtime/converter.go
```diff
@@ -1,4 +1,82 @@
...
+		Tenant:      graphql.Tenant(in.Tenant),
+		Annotations: in.Annotations,
+		Labels:      in.Labels,
+	}
+}
+
+func (c *Converter) MultipleToGraphQL(in []*model.Runtime) []*graphql.Runtime {
+	var runtimes []*graphql.Runtime
+	for _, r := range in {
+		if r == nil {
```
aszecowka: this `if` is not needed IMO

components/director/internal/domain/runtime/converter.go
```diff
@@ -1,4 +1,82 @@
...
+	}
+
+	return &graphql.Runtime{
+		ID:          in.ID,
+		Status:      c.statusToGraphQL(in.Status),
+		Name:        in.Name,
+		Description: in.Description,
+		Tenant:      graphql.Tenant(in.Tenant),
+		Annotations: in.Annotations,
+		Labels:      in.Labels,
```
aszecowka: what about agentAuth?

components/director/internal/domain/runtime/converter_test.go
```diff
@@ -1 +1,106 @@
...
+			// then
+			assert.Equal(t, testCase.Expected, res)
+		})
+	}
+}
+
+func TestConverter_MultipleToGraphQL(t *testing.T) {
+	// given
+	input := []*model.Runtime{
+		fixModelRuntime("foo", "Foo", "Lorem ipsum"),
```
aszecowka: test for at least 2 valid objects

components/director/internal/domain/runtime/resolver.go
```diff
@@ -3,49 +3,190 @@ package runtime
...
+	labelFilter := labelfilter.MultipleFromGraphQL(filter)
+
+	var cursor string
+	if after != nil {
+		cursor = string(*after)
+	}
+
+	runtimesPage, err := r.svc.List(ctx, labelFilter, first, &cursor)
+	if err != nil {
+		return nil, err
```
aszecowka: what about logging?

aszecowka: in CBS we have a special layer of errors, that are displayed for end-user. Why we don't follow that pattern?
Also, why we don't wrap errors?

components/director/internal/graphql/tenant.go
```diff
@@ -21,7 +22,7 @@ func (y *Tenant) UnmarshalGQL(v interface{}) error {
 }
 
 func (y Tenant) MarshalGQL(w io.Writer) {
-	_, err := w.Write([]byte(y))
+	_, err := io.WriteString(w, strconv.Quote(string(y)))
```
aszecowka: why you don't print error? Additionally, maybe we should decide which logger to use?

components/director/internal/domain/runtime/converter.go
pkosiec: It is, because in GraphQL schema we defined that we won't return `null` in the output array

components/director/internal/domain/runtime/resolver.go
pkosiec: Here's an issue for that: https://github.com/kyma-incubator/compass/issues/66

It will be done as a follow-up task.

pkosiec: It's a part of a follow-up task: https://github.com/kyma-incubator/compass/issues/66

components/director/internal/graphql/tenant.go
pkosiec: The error is now printed. As discussed, we will use `logrus`, but in new task, to not block our colleges.

components/director/internal/domain/runtime/converter.go
pkosiec: It will be done in a follow-up task, as this is more complicated topic: #67

Issue #51, opened 2019-06-07
Custom scalars implementation


kfurgol:
**Description**

Implement custom scalars with user-defined types

**Reasons**

Part of #3

PR #63, opened 2019-06-11
Implement scalars

kfurgol:
**Description**

- Implemented scalars
- Added scalar tests 

**Related issue(s)**
#51


components/director/internal/graphql/scalars/timestamp.go
```diff
@@ -0,0 +1,49 @@
...
+
+	t, err := time.Parse(time.RFC3339, tmpStr)
+	if err != nil {
+		return err
+	}
+
+	*y = Timestamp(t)
+
+	return nil
+}
```
crabtree: Add a new line here :)

components/director/internal/graphql/scalars/helpers.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func convertToString(in interface{}) (string, error) {
+	if in == nil {
+		return "", errors.New("input should not be nil")
+	}
+
+	value, ok := in.(string)
+	if !ok {
+		return "", errors.Errorf("unexpected input type: %T, should be map[string][]string", in)
```
crabtree: This error seems to be a bit misleading here.

components/director/internal/graphql/scalars/timestamp.go
```diff
@@ -0,0 +1,49 @@
...
+
+type Timestamp time.Time
+
+func (y *Timestamp) UnmarshalGQL(v interface{}) error {
+
+	if v == nil {
+		return nil
+	}
+
+	tmpStr, ok := v.(string)
```
crabtree: Is there any purpose you do not use convertToString here?

components/director/internal/graphql/scalars/annotations_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	//when
+	err := a.UnmarshalGQL(fixAnnotations)
+
+	//then
+	require.NoError(t, err)
+	assert.Equal(t, a, expectedAnnotations)
+}
+
+func TestAnnotations_MarshalGQL(t *testing.T) {
```
crabtree: This smells like two test cases. Consider splitting it.

components/director/internal/graphql/scalars/helpers.go
```diff
@@ -0,0 +1,52 @@
+package scalars
+
+import (
+	"encoding/json"
+	"io"
+
+	"github.com/pkg/errors"
+)
+
+func writeResponse(in interface{}, w io.Writer) error {
```
crabtree: Please rename this method to be more self-describing. This has nothing with writing the response.

components/director/internal/graphql/scalars/timestamp.go
kfurgol: done

components/director/internal/graphql/scalars/helpers.go
kfurgol: done

components/director/internal/graphql/scalars/timestamp.go
kfurgol: no, I just missed it! I've changed it

components/director/internal/graphql/scalars/annotations_test.go
kfurgol: done

components/director/internal/graphql/scalars/helpers.go
kfurgol: I've renamed it to `marshalAndWrite`, if you have a better name please give me a hint

components/director/internal/graphql/scalars/annotations_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAnnotations_UnmarshalGQL(t *testing.T) {
```
crabtree: How about covering here case with multiple values, as we have also in Marshalling func.

components/director/internal/graphql/scalars/clob.go
```diff
@@ -0,0 +1,26 @@
+package scalars
+
+import (
+	"io"
+	"log"
+)
+
+type CLOB string
```
crabtree: I think we should store CLOB as []byte.

components/director/internal/graphql/scalars/helpers.go
```diff
@@ -0,0 +1,52 @@
+package scalars
+
+import (
+	"encoding/json"
+	"io"
+
+	"github.com/pkg/errors"
+)
+
+func marshalAndWrite(in interface{}, w io.Writer) error {
```
crabtree: What about marshallToWriter?

components/director/internal/graphql/scalars/clob.go
pkosiec: Hmm, what is the reason of that?

components/director/internal/graphql/scalars/annotations.go
```diff
@@ -0,0 +1,32 @@
+package scalars
```
pkosiec: Scalars shouldn't be in separate package as they are also GraphQL types. Please move them back to `graphql` package, keeping one file per scalar.

components/director/internal/graphql/scalars/annotations_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	//when
+	err := a.UnmarshalGQL(fixAnnotations)
+
+	//then
+	require.NoError(t, err)
+	assert.Equal(t, a, expectedAnnotations)
+}
+
+func TestAnnotations_MarshalGQL_Int(t *testing.T) {
```
pkosiec: I would have single `TestAnnotations` func with two `t.Run` cases inside.

components/director/internal/graphql/scalars/helpers.go
pkosiec: `marshallToWriter` sounds a little bit confusing.
What about `writeMarshalled` or `writeAsBytes`?

components/director/internal/graphql/scalars/helpers.go
```diff
@@ -0,0 +1,52 @@
+package scalars
```
pkosiec: I don't like the idea of `helpers.go` file. Maybe we could move it to `pkg` as exported functions. However, we need a good name for that package 😄  Maybe `pkg/scalar`?

components/director/internal/graphql/scalars/labels_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLabels_UnmarshalGQL(t *testing.T) {
+	//given
```
pkosiec: Yes, error case could be also good to test for all scalars

components/director/internal/graphql/scalars/annotations_test.go
crabtree: As those two following test functions are similar in the implementation only differs by given arguments you can reuse the implementation by providing the test payload and expectations as a list and iterating over those values.

For example https://tutorialedge.net/golang/improving-your-tests-with-testify-go/#combining-testify-with-table-driven-tests

components/director/internal/graphql/scalars/clob.go
crabtree: For me, it is more natural to store some unstructured, possibly binary data in []byte. You can easily then convert to a string if needed.

components/director/internal/graphql/scalars/helpers.go
crabtree:
```suggestion
func marshallToWriter(in interface{}, w io.Writer) error {
```

pkosiec: Hmm, marshallToWriter is confusing. I had two other ideas: 
`writeMarshalled` or `writeBytes`

components/director/internal/graphql/scalars/annotations_test.go
kfurgol: done

components/director/internal/graphql/scalars/annotations.go
kfurgol: done

components/director/internal/graphql/scalars/annotations_test.go
kfurgol: Did it with use of Table-Driven Tests

kfurgol: done

components/director/internal/graphql/scalars/clob.go
kfurgol: changed to []byte

components/director/internal/graphql/scalars/helpers.go
kfurgol: moved to pkg/scalar/

components/director/internal/graphql/scalars/helpers.go
kfurgol: changed to `writeMarshalled`

components/director/internal/graphql/scalars/labels_test.go
kfurgol: done

components/director/pkg/scalar/helpers.go
```diff
@@ -0,0 +1,69 @@
+package scalar
```
pkosiec: Rename the file to scalar.go

components/director/internal/graphql/tenant.go
```diff
@@ -0,0 +1,28 @@
...
+
+	*y = Tenant(val)
+
+	return nil
+}
+
+func (y Tenant) MarshalGQL(w io.Writer) {
+	_, err := w.Write([]byte(y))
+	if err != nil {
+		log.Printf("error with writing %T", y)
```
pkosiec: As @crabtree noticed, this is quite bad to log an error. But of course for now we can't return error. I reported an issue for gqlgen: https://github.com/99designs/gqlgen/issues/743

Issue #67, opened 2019-06-17
Implement Auth in Director


pkosiec:
**Description**

Figure out how to implement Auth, which contains Auth. Auth has circular reference to itself.

Remember about RuntimeAuth collection for APIs, `agentAuth` for Runtime and `Auth` for FetchRequest, ApplicationWebhook.

Do we want to store Auth in a separate table? If so, `agentAuth` and all `auth` fields should be resolvers and there should be separate services.

Anyway, we will need one converter, so maybe it would be better to create `auth` pseudo-domain.

Issue #54, opened 2019-06-07
Implement Director's mocks for Application


pkosiec:
**Description**

Implement mocks for the Application domain. Prepare implementation for:
- resolvers
- services
- converters
- in-memory repositories
- models

Remember about unit tests for all these parts.

PR #70, opened 2019-06-21
Add mocks for Application and Auth in Director

pkosiec:
**Description**

- Add mocks for Application domain in Director
- Implement Auth types conversion in Director

**Related issue(s)**
#54 
#67


components/director/internal/model/application.go
```diff
@@ -0,0 +1,126 @@
...
+func (a *Application) DeleteAnnotation(key string) error {
+	if _, exists := a.Annotations[key]; !exists {
+		return fmt.Errorf("annotation %s doesn't exist", key)
+	}
+
+	delete(a.Annotations, key)
+	return nil
+}
+
+type ApplicationInput struct {
```
aszecowka: In GraphQL layer we have ApplicationInput, but on model/domain we should have different name IMO. In current shape, it looks that graphQL layer leaks to the service layer. Instead if calling in ApplicationInput, we can call it ApplicationExt (Application with dependencies). I am open to other proposals for name.

components/director/internal/domain/application/resolver.go
```diff
@@ -40,40 +58,150 @@ func NewResolver(svc *service, apiSvc APIService, eventAPISvc EventAPIService, d
...
+	if err != nil {
+		return nil, err
+	}
+
+	return deletedApp, nil
 }
-func (r *Resolver) AddApplicationLabel(ctx context.Context, applicationID string, label string, values []string) ([]string, error) {
-	panic("not implemented")
+func (r *Resolver) AddApplicationLabel(ctx context.Context, applicationID string, key string, values []string) (*graphql.Label, error) {
+	err := r.svc.AddLabel(ctx, applicationID, key, values)
```
aszecowka: As you know, Labels/Annotations can be extended soon. If this happens, we will have to change signatures for our services. Please create proper structs for those 2 types and pass it as an argument for the services, instead of passing `key` and `values`.

pkosiec: Okay, I will create a follow-up task

PR #72, opened 2019-06-21
Test Auth and Application models in Director

### LINKED ISSUE #67: Implement Auth in Director

pkosiec:
**Description**

- Test Auth and Application models in Director

**Related issue(s)**
See #70 
Resolves #67

### DIFF
--- a/components/director/internal/model/auth_test.go
+++ b/components/director/internal/model/auth_test.go
@@ -1,3 +1,321 @@
-package model
+package model_test
 
-//TODO: Test AuthInput toAuth
+import (
+	"fmt"
+	"testing"
+
+	"github.com/kyma-incubator/compass/components/director/internal/model"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAuthInput_ToAuth(t *testing.T) {
+	// given
+	testCases := []struct {
+		Name     string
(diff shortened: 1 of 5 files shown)


components/director/internal/model/auth_test.go
```diff
@@ -1,3 +1,321 @@
...
+	testCases := []struct {
+		Name     string
+		Input    *model.AuthInput
+		Expected *model.Auth
+	}{
+		{
+			Name: "All properties given",
+			Input: &model.AuthInput{
+				Credential: &model.CredentialDataInput{
+					Basic: &model.BasicCredentialDataInput{
```
aszecowka: here you test only basic creds, you should also have a test for oauth, to ensure that `ToAuth` is converting it if needed.

PR #79, opened 2019-06-24
Add mocks for Document and Webhook in Director

### LINKED ISSUE #75: Implement Director's Documents mocks
**Description**

Implement mocks for the Application domain. Prepare implementation for:
- resolvers
- services
- converters
- in-memory repositories
- models

Remember about unit tests for all these parts.

### LINKED ISSUE #76: Implement Director's Webhooks mocks
**Description**

Implement mocks for the Application domain. Prepare implementation for:
- resolvers
- services
- converters
- in-memory repositories
- models

Remember about unit tests for all these parts.

tgorgol:
**Description**

- Add mocks for Document domain in director
- Add mocks for Webhook domain in director
- Add Fetch Request converter
- Add models for Document, Webhook and Fetch Request

**Related issue(s)**
Resolves #75 
Resolves #76


components/director/internal/domain/application/automock/api_repository.go
```diff
@@ -1,4 +1,5 @@
 // Code generated by mockery v1.0.0. DO NOT EDIT.
+
```
pkosiec: Remove it

components/director/pkg/scalar/scalar.go
```diff
@@ -38,6 +38,16 @@ func ConvertToByteArray(in interface{}) ([]byte,error){
 		return nil, errors.New("input should not be nil")
 	}
 
+	//var value []byte
```
pkosiec: Why it is commented? Please remove it

components/director/internal/domain/document/converter_test.go
```diff
@@ -1 +1,144 @@
...
+		},
+		{
+			Name:     "Nil",
+			Input:    nil,
+			Expected: nil,
+		},
+	}
+
+	for i, testCase := range testCases {
+		t.Run(fmt.Sprintf("%d: %s", i, testCase.Name), func(t *testing.T) {
```
aszecowka: what do you add "i" to the test name?
```
FAIL: TestConverter_ToGraphQL/0:_All_properties_given (0.00s)
```

components/director/internal/domain/document/converter_test.go
```diff
@@ -1 +1,144 @@
...
+		fixGQLDocument("foo"),
+		fixGQLDocument("bar"),
+		{},
+	}
+	var nilPointer *model.FetchRequest
+
+	// when
+	frConv := &automock.FetchRequestConverter{}
+	frConv.On("ToGraphQL", input[0].FetchRequest).Return(expected[0].FetchRequest)
+	frConv.On("ToGraphQL", nilPointer).Return(nil)
```
aszecowka: You forget about checking if expectations were met, If I add
```
	frConv.On("WhereIsWally", nilPointer).Return(nil)
```
this test will pass. Please add:
```
defer frConv.AssertExpectations(t)
```

aszecowka: Please check other places

components/director/internal/domain/document/converter_test.go
```diff
@@ -1 +1,144 @@
...
+		fixModelDocument("2", "bar"),
+		{},
+		nil,
+	}
+	expected := []*graphql.Document{
+		fixGQLDocument("foo"),
+		fixGQLDocument("bar"),
+		{},
+	}
+	var nilPointer *model.FetchRequest
```
aszecowka: minor: it can be removed by using such syntax:
```
	frConv.On("ToGraphQL", (*model.FetchRequest)(nil)).Return(nil)

```

components/director/internal/domain/document/converter_test.go
```diff
@@ -1 +1,144 @@
...
+		{},
+	}
+	var nilPointer *graphql.FetchRequestInput
+
+	// when
+	frConv := &automock.FetchRequestConverter{}
+	frConv.On("InputFromGraphQL", input[0].FetchRequest).Return(expected[0].FetchRequest)
+	frConv.On("InputFromGraphQL", nilPointer).Return(nil)
+	converter := document.NewConverter(frConv)
+	res := converter.MultipleInputFromGraphQL(input)
```
aszecowka: when section should contain only this:
```
res := converter.MultipleInputFromGraphQL(input)
```

components/director/internal/domain/document/converter_test.go
tgorgol: I was adding it to stay consistent with how previous tests (runtime and a part of application) were written.

components/director/internal/domain/document/resolver_test.go
```diff
@@ -1 +1,201 @@
...
+				conv := &automock.DocumentConverter{}
+				conv.On("InputFromGraphQL", gqlInput).Return(modelInput).Once()
+				conv.On("ToGraphQL", modelDocument).Return(gqlDocument).Once()
+				return conv
+			},
+			ExpectedDocument: gqlDocument,
+			ExpectedErr:      nil,
+		},
+		{
+			Name: "Create Error",
```
aszecowka: here you can use sentences, so instead of `Create Error`, you can use `Returns error when document creation failed`

components/director/internal/domain/document/resolver.go
```diff
@@ -4,27 +4,67 @@ import (
...
+	return gqlDocument, nil
 }
+
 func (r *Resolver) DeleteDocument(ctx context.Context, id string) (*graphql.Document, error) {
-	panic("not implemented")
+	document, err := r.svc.Get(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	if document == nil {
```
aszecowka: it seems that we are not consistent here, in other resolvers we don't have special if for nil object.
Maybe service should return error if object does not exist? 
We need to have clear documentation:
- if a repository returns an error if an object does not exist?
- if a service returns an error if an object does not exist?

Please check if we are consistent among whole project

tgorgol: This if statement was supposed to be removed. We discussed that case earlier with @pkosiec and we decided that repository will return error when trying to get not existing item, so the scenario where get returns nil wouldn't actually occur and that if statement wasn't needed. I checked consistency in runtime, application, document, webhook and fetchrequest domains.

components/director/internal/domain/document/resolver_test.go
tgorgol: I changed error test names to more descriptive ones.

PR #88, opened 2019-06-28
Add models for application registry REST API

crabtree:
**Description**

- Models for Application Registry REST API

**Related issue(s)**


components/gateway/internal/domain/application-registry-rest/model/model.go
```diff
@@ -0,0 +1,94 @@
+package model
+
+// API is an internal representation of a service's API.
+type API struct {
```
aszecowka: /internal/application-registry-adapter/model

PR #92, opened 2019-06-28
Fill AgentAuth during Runtime creation

### LINKED ISSUE #90: AgentAuth is not populated when creating Runtime
When I execute the following query:
```
mutation {
  result: createRuntime(
    in: { name: "runtime-1", description: "runtime-1-description" }
  ) {
    id
    name
    description

    agentAuth {
      credential {
        ... on BasicCredentialData {
          username
          password
        }
        ... on OAuthCredentialData {
          clientId
          clientSecret
          url
        }
      }
    }
  }
}

```

I got the following error:
```
{
  "errors": [
    {
      "message": "must not be null",
      "path": [
        "result",
        "agentAuth"
      ]
    }
  ],
  "data": null
}
```
Additionally, I do not see any errors in the director logs which is strange because of this is a server side error.

pkosiec:
**Description**

- Fill AgentAuth during Runtime creation

**Related issue(s)**
Resolves #90

### DIFF
--- a/components/director/internal/model/runtime.go
+++ b/components/director/internal/model/runtime.go
@@ -114,11 +114,8 @@ func (i *RuntimeInput) ToRuntime(id string, tenant string) *Runtime {
 		Tenant:      tenant,
 		Labels:      i.Labels,
 		Annotations: i.Annotations,
-		AgentAuth: &Auth{},
-		Status: &RuntimeStatus{
-			Condition: RuntimeStatusConditionInitial,
-			Timestamp: time.Now(),
-		},
+		AgentAuth:   &Auth{},
+		Status:      &RuntimeStatus{},
 	}
 }
 
--- a/components/director/internal/model/runtime.go
+++ b/components/director/internal/model/runtime.go
@@ -103,6 +103,10 @@ type RuntimeInput struct {
 }
 
 func (i *RuntimeInput) ToRuntime(id string, tenant string) *Runtime {
+	if i == nil {
+		return nil
+	}
+
 	return &Runtime{
 		ID:          id,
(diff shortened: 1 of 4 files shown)


components/director/internal/model/runtime.go
```diff
@@ -103,13 +103,22 @@ type RuntimeInput struct {
...
+
 	return &Runtime{
 		ID:          id,
 		Name:        i.Name,
 		Description: i.Description,
 		Tenant:      tenant,
 		Labels:      i.Labels,
 		Annotations: i.Annotations,
+		AgentAuth: &Auth{},
+		Status: &RuntimeStatus{
```
aszecowka: move it to service

components/director/internal/model/runtime.go
```diff
@@ -103,13 +103,22 @@ type RuntimeInput struct {
...
+			Condition: RuntimeStatusConditionInitial,
+			Timestamp: time.Now(),
```
aszecowka: don't use `time.Now()` because it cannot be tested, introduce TImeProvider interface

components/director/internal/model/runtime_test.go
```diff
@@ -274,3 +275,65 @@ func TestRuntime_DeleteAnnotation(t *testing.T) {
...
+		},
+		{
+			Name:     "Nil",
+			Input:    nil,
+			Expected: nil,
+		},
+	}
+
+	for i, testCase := range testCases {
+		t.Run(fmt.Sprintf("%d: %s", i, testCase.Name), func(t *testing.T) {
```
aszecowka: what value gives you this index in the name of the test????????????

PR #97, opened 2019-07-01
Extend graphql model to enable integration tests

aszecowka:


components/director/pkg/graphql/models_gen_ext.go
```diff
@@ -0,0 +1,29 @@
+package graphql
```
pkosiec: I would rename the file to `auth.go`

Issue #47, opened 2019-06-07
Define a contract between Runtime Agent and Compass 


PK85:
**Description**
 

**AC**
- Define communication between Runtime Agent and Compass. Provide documentation as an architecture diagram 
  - take into account that Director needs to know EventGateway URL
  - take into account that Director needs to know for all API spec, that runtime has been created (empty credentials)
  - Establishing a trusted relation based on client certificates (Agent -> Connector)
  - Configuring the runtime with configuration provided from Compass (Agent -> Director)
- Define how to setup and configure Runtime Agent. Document as an architecture diagram

Issue #106, opened 2019-07-04
Generate credentials while creating Runtime and Application


PK85:
**Description**
We need a place to store credentials that will be used by RuntimeAgent and then paired with compass via connector.

**AC**
- extend Application GraphQL type with field `auths` (array of auths from SystemAuths table). Implement resolver to get entries from SystemAuth table
- while generating OneTimeToken for Application, create new empty (ID + ref) entry in SystemAuths table. ID of the entry will be used as `auth_id`.
- Endpoint for generating OneTImeToken for application
  - while generating OneTimeToken for Runtime, create new empty (ID + ref) entry in SystemAuths. ID of the entry will be used as `auth_id`.
- Endpoint for generating OneTImeToken for runtime
  - Call for OneTimeToken endpoint in connector. (Ask frogs if they want one endpoint with duration parameter or stay with seperate endpoints for app and runtime).
- add integration tests for new mutation + auths field
- remove skip in e2e tests for system auth.
**Reasons**

We need to add `auths` field for Application, because we will use auth ID to resolve tenant in tenant mapping service. In future we want to support multiple authorization methods.

pkosiec:
See https://github.com/kyma-incubator/compass/issues/91 - we need to consolidate the tasks into one.

PR #110, opened 2019-07-04
Describe runtime agent contract

crabtree:
**Description**

- Describe runtime agent contract

**Related issue(s)**

See also #47


components/director/pkg/graphql/schema.graphql
```diff
@@ -318,6 +318,13 @@ type HealthCheck {
     timestamp: Timestamp!
 }
 
+# RuntimeConfiguration
+
+type RuntimeConfiguration {
```
PK85: Consider Moving to Runtime domain. We will have less models, queries.

docs/architecture/runtime-agent-contract.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Renewing trusted connection
+
+Depending on the type of trusted connection, during the runtime lifecycle, it may be required to periodically [renew trusted connection](./establishing-trusted-connection.md#client-certificate-flow---certificate-renewal).
+
+## Configuring the runtime
+
+Runtime Agent periodically requests for the configuration of its Runtime from the Management Plane. Changes in the configuration for the Runtime are applied by the Runtime Agent on the Runtime. To fetch the Runtime configuration the Runtime Agent calls `runtimeConfiguration(id: ID!)` query offered by [Director](../terminology.md#mp-director). The response for the query contains list of Applications assigned for the Runtime.
+
+Runtime Agent reports the Runtime eventing configuration to the Director. Runtime entity stores URL of the runtime. Based on the configuration the Eventing URL and other useful URLs, like console UI URL, can be computed.
```
PK85: Krystian mentioned additional requirement, that maybe we coudl handle by labels. 
- Namely, not only URL from events, but we ned to expose runtime console UI and other URLs, maybe even kyma/runtime versions etc.
- We can realize that already by LabelDefintions with schemas. We need to decide that maybe those LD should be global, not removable by tenants.
- All LD values are computed onky by runtime Agent and passed to MP
- We can even consider adding new LD by runtime Agent (if new ones supported by runtime)

crabtree: This should be covered in #106

docs/architecture/runtime-agent-contract.md
```diff
@@ -0,0 +1,27 @@
+# Runtime Agent contract
+
+## Overview
+
+This document describes contract between [Management Plane](../terminology.md#management-plane) and [Runtime Agent](../terminology.md#runtime-agent). Main responsibilities of the Runtime Agent are:
+- [Establishing trusted connection](#establishing-trusted-connection)
+- [Renewing trusted connection](#renewing-trusted-connection)
+- [Configuring applications](#configuring-applications)
```
franpog859: The link should look like this to work correctly:
`- [Configuring the runtime](#configuring-the-runtime)`

crabtree:
> Shouldn't the health checks mentioned [here](https://github.com/kyma-incubator/compass/blob/0c12b0f53f7e113f97e4954000d3e84beadc01a6/docs/terminology.md#runtime-agent) be involved in this contract?

As agreed with @PK85 the health checks will be defined in the next iteration of Runtime Agent Contract. For now, we do not focus on this area.

docs/architecture/runtime-agent-contract.md
```diff
@@ -0,0 +1,27 @@
...
+## Overview
+
+This document describes contract between [Management Plane](../terminology.md#management-plane) and [Runtime Agent](../terminology.md#runtime-agent). Main responsibilities of the Runtime Agent are:
+- [Establishing trusted connection](#establishing-trusted-connection)
+- [Renewing trusted connection](#renewing-trusted-connection)
+- [Configuring the runtime](#configuring-the-runtime)
+
+## Establishing trusted connection
+
+Runtime Agent automates the process of [establishing trusted connection](./establishing-trusted-connection.md#establishing-trusted-connection) between [Runtime](../terminology.md#runtime) and Managemt Plane. To initialy trigger the process the [Runtime Provisioner](../terminology.md#mp-runtime-provisioner) creates the instance of CustomResource that holds the one-time token to enable the pairing. Runtime Agent triggered by the creation of CustomResource performs the pairing process and in the end the trusted connection between the Runtime and the Management Plane is established. At this point further communication between Runtime Agent and Management Plane, and [Application](../terminology.md#application) and Runtime is possible.
```
PK85: That description explains the flow and then implementation details for Kyma as a Runtime. That's good. But we need also genberal contract. That means what information will be included in the token etc. 
For example, we will need information about tenant, scenario etc.

docs/architecture/runtime-agent-contract.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Renewing trusted connection
+
+Depending on the type of trusted connection, during the runtime lifecycle, it may be required to periodically [renew trusted connection](./establishing-trusted-connection.md#client-certificate-flow---certificate-renewal).
+
+## Configuring the runtime
+
+Runtime Agent periodically requests for the configuration of its Runtime from the Management Plane. Changes in the configuration for the Runtime are applied by the Runtime Agent on the Runtime. To fetch the Runtime configuration the Runtime Agent calls `applicationsForRuntime(runtimeId: ID!)` query offered by [Director](../terminology.md#mp-director). The response for the query contains list of Applications assigned for the Runtime.
```
PK85: Add a first sentence that Runtime will get condigurations and update itself in management plane with details about a runtime.

docs/architecture/runtime-agent-contract.md
crabtree: @PK85 Why do we need information like tenant and labeling (eg.scenario) on the runtime agent? I assume that runtime agent need only the Runtime ID. Using runtime ID agent fetches the runtime configuration from Management Plane.

PR #111, opened 2019-07-04
Fix handling nil clobs

aszecowka:
Problem: when I created a new application with `ApiSpec`/`EventApiSpec`, but `Data` was not provided, in return I got Data with `""`, instead of nil.

### DIFF
--- a/components/director/internal/domain/api/converter_test.go
+++ b/components/director/internal/domain/api/converter_test.go
@@ -4,6 +4,9 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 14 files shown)


components/director/internal/domain/api/converter_test.go
```diff
@@ -328,3 +331,33 @@ func TestConverter_MultipleInputFromGraphQL(t *testing.T) {
 		})
 	}
 }
+
+func TestApiSpecDataConversionNilStaysNil(t *testing.T) {
```
pkosiec: I would consolidate this test into table tests for the conversion

Issue #80, opened 2019-06-24
Install PostgreSQL


pkosiec:
**AC:**
- Copy PostreSQL chart and paste as `compass` subchart - #131 
- Add ability to toggle local PostreSQL installation and use GCP PostgreSQL instance

Hint: See `Use embedded etcd` flag in `service-catalog` chart in Kyma, which works similarly to what we want to achieve.

PR #131, opened 2019-07-11
Add postgres chart

kfurgol:
**Description**

- Added Postgres chart 

**Related issue(s)**

### DIFF
--- a/chart/compass/README.md
+++ b/chart/compass/README.md
@@ -1,0 +1,21 @@
+# Compass
+
+## Overview
+
+The Compass consists of the following sub-charts:
+
+- `connector` 
+- `director` 
+- `gateway` 
+- `healthchecker`
+- `postgresql`
+
+## Details
+
+To learn more about the Compass, see the [Overview](https://github.com/kyma-incubator/compass#Overview) document.
+
+## Configuration
+
+| Parameter | Description | Values | Default |
+| --- | --- | --- | --- |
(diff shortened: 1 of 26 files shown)


crabtree:
What about the possibility to disable the installation of this chart? As I understand correctly, the issue #80 also covers the possibility to switch between using in-cluster Postgres or externally provided one.

kfurgol:
> What about the possibility to disable the installation of this chart? As I understand correctly, the issue #80 also covers the possibility to switch between using in-cluster Postgres or externally provided one.

It can be done by setting `storage.useEmbedded` value to `false` in  `values.yaml` file  of Compass chart

crabtree:
> > What about the possibility to disable the installation of this chart? As I understand correctly, the issue #80 also covers the possibility to switch between using in-cluster Postgres or externally provided one.
> 
> It can be done by setting `storage.useEmbedded` value to `false` in `values.yaml` file of Compass chart

Ok, so please put this into the documentation. I would suggest creating a README file under compass chart directory.

kfurgol:
> > > What about the possibility to disable the installation of this chart? As I understand correctly, the issue #80 also covers the possibility to switch between using in-cluster Postgres or externally provided one.
> > 
> > 
> > It can be done by setting `storage.useEmbedded` value to `false` in `values.yaml` file of Compass chart
> 
> Ok, so please put this into the documentation. I would suggest creating a README file under compass chart directory.

Done

PR #139, opened 2019-07-15
Update end to end test in check application existence

### LINKED ISSUE #96: Add Application/Runtime checks to appropriate Director services
**Description**

Some Director mutations (related to the creation of Application subresources, e.g. `addApplicationWebhook`) don't check if an Application with the specified ID actually exists. Right now this happens with Webhooks, Documents, APIs and EventAPIs.

There is also a case where when adding AuthRuntime to API, Runtime should be checked. 

Those services should get Application/Runtime service injected and use it to check if Application/Runtime with received ID exists.

**Expected result**

Attempt to create a subresource of Application (eg. Webhook), when there's no application with specified ID should result in an error.

**Actual result**

Subresource is created even when an Application with passed ID doesn't exist.

**Steps to reproduce**

1. Make sure there is no Application with ID `foo`
2. Create Webhook using `addApplicationWebhook` mutation with `applicationID` parameter set to `foo`

dbadura:
**Description**

- Renamed mutations in end to end tests
- Add test case when application not exist

**Related issue(s)**

Fixes #96

### DIFF
--- a/tests/end-to-end/director/application_api_test.go
+++ b/tests/end-to-end/director/application_api_test.go
@@ -405,7 +405,7 @@ func TestUpdateApplicationParts(t *testing.T) {
 
 	t.Run("manage webhooks", func(t *testing.T) {
 		// add
-		webhookInStr, err := tc.graphqlizer.ApplicationWebhookInputToGQL(&graphql.WebhookInput{
+		webhookInStr, err := tc.graphqlizer.WebhookInputToGQL(&graphql.WebhookInput{
 			URL:  "new-webhook",
 			Type: graphql.ApplicationWebhookTypeConfigurationChanged,
 		})
@@ -430,7 +430,7 @@ func TestUpdateApplicationParts(t *testing.T) {
 		// add to non exist application
 
 		//GIVEN
-		webhookInStr, err = tc.graphqlizer.ApplicationWebhookInputToGQL(&graphql.WebhookInput{
+		webhookInStr, err = tc.graphqlizer.WebhookInputToGQL(&graphql.WebhookInput{
 			URL:  "new-webhook",
 			Type: graphql.ApplicationWebhookTypeConfigurationChanged,
 		})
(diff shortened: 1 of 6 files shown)


tests/end-to-end/director/application_api_test.go
```diff
@@ -405,41 +405,64 @@ func TestUpdateApplicationParts(t *testing.T) {
...
 		require.NoError(t, err)
 		assert.Equal(t, "new-webhook", actualWebhook.URL)
 		assert.Equal(t, graphql.ApplicationWebhookTypeConfigurationChanged, actualWebhook.Type)
 		id := actualWebhook.ID
 		require.NotNil(t, id)
 
+		// add to non exist application
+
+		//GIVEN
+		webhookInStr, err = tc.graphqlizer.WebhookInputToGQL(&graphql.WebhookInput{
```
aszecowka: IMO it will be better to have a separate test when we test all cases when we try to add a dependent object for non-existing application. Thanks to that our tests will be shorter. 
```
func TestAddDependentObjectsWhenAppDoesNotExist(t *testing.T) {
	t.Run("add webhooks", func(t *testing.T) {

	})
}
```

tests/end-to-end/director/application_api_test.go
```diff
@@ -257,6 +257,108 @@ func TestCreateApplicationWithDocuments(t *testing.T) {
...
+	applicationId := "foo"
+
+	t.Run("add Webhook", func(t *testing.T) {
+		//GIVEN
+		ctx := context.Background()
+		webhookInStr, err := tc.graphqlizer.WebhookInputToGQL(&graphql.WebhookInput{
+			URL:  "new-webhook",
+			Type: graphql.ApplicationWebhookTypeConfigurationChanged,
+		})
+		tmpWebhook := graphql.Webhook{}
```
aszecowka: do you really need this variable tmpWebhook? you can use nil instead

tests/end-to-end/director/application_api_test.go
```diff
@@ -257,6 +257,108 @@ func TestCreateApplicationWithDocuments(t *testing.T) {
...
+		addReq := gcli.NewRequest(
+			fmt.Sprintf(`mutation {
+			result: addWebhook(applicationID: "%s", in: %s) {
+					%s
+				}
+			}`, applicationId, webhookInStr, tc.gqlFieldsProvider.ForWebhooks()))
+		err = tc.RunQuery(ctx, addReq, &tmpWebhook)
+
+		//THEN
+		require.Error(t, err)
```
aszecowka: here you can check only if error exist, but you can be more specific and check if error represents missing application

Issue #81, opened 2019-06-24
Define DB schema


pkosiec:
**AC:**
- Create pre-install Job for `compass` chart, which creates and upgrades database tables (for now, only one Runtime table). Prepare separate entry point for Director or even whole new Docker image with upgrade job. 
- Use migration tool from investigation: https://github.com/kyma-incubator/compass/pull/69

PR #143, opened 2019-07-16
Add schema-migrator component and migration hook

tgorgol:
**Description**

- Add schema-migrator component to compass
- Add post-install and post-upgrade hooks to compass chart that will perform database migration

**Related issue(s)**
#81


chart/compass/charts/postgresql/templates/secrets.yaml
```diff
@@ -11,7 +11,12 @@ metadata:
     heritage: {{ .Release.Service | quote }}
 type: Opaque
 data:
+  postgresql-username: {{ include "postgresql.username" . | b64enc | quote }}
```
aszecowka: minor: maybe we can remove prefix postgresql?

chart/compass/charts/postgresql/values.yaml
```diff
@@ -109,7 +109,7 @@ postgresqlUsername: postgres
 ## Create a database
 ## ref: https://github.com/bitnami/bitnami-docker-postgresql/blob/master/README.md#creating-a-database-on-first-run
 ##
-# postgresqlDatabase:
+postgresqlDatabase: postgres # Modified
```
aszecowka: remove Modified comment

chart/compass/templates/migrator-job.yaml
```diff
@@ -0,0 +1,111 @@
+apiVersion: v1
+data:
+    check-db-connection.sh: |
```
aszecowka: why this is not part of the migrator image?

components/schema-migrator/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM alpine:3.9
```
aszecowka: recently @pkosiec added such thing to other Dockerfiles:
```
COPY ./licenses ${BASE_APP_DIR}/licenses
```
please check if we need it also here

components/schema-migrator/.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
aszecowka: do we really need this file?

components/schema-migrator/migrations/1_initialize_schema.up.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TYPE runtime_status_condition AS ENUM ('INITIAL', 'READY', 'FAILED');
+
+CREATE TABLE runtime
```
aszecowka: I have found a recommendation that table names should be pluralized: https://www.periscopedata.com/blog/better-sql-schema

components/schema-migrator/migrations/1_initialize_schema.up.sql
```diff
@@ -0,0 +1,15 @@
...
+    id uuid NOT NULL CONSTRAINT runtime_pk PRIMARY KEY,
+    tenant_id uuid NOT NULL,
+    name varchar(256) NOT NULL,
+    description text,
+    status_condition runtime_status_condition DEFAULT 'INITIAL'::runtime_status_condition NOT NULL,
+    status_timestamp timestamp NOT NULL,
+    auth json
+);
+
+CREATE UNIQUE INDEX runtime_id_uindex ON runtime (id);
```
aszecowka: seems to be not necessary:
https://www.postgresql.org/docs/9.4/indexes-unique.html
```
PostgreSQL automatically creates a unique index when a unique constraint or primary key is defined for a table.
```

components/schema-migrator/migrations/1_initialize_schema.up.sql
```diff
@@ -0,0 +1,15 @@
...
+CREATE UNIQUE INDEX runtime_id_name_uindex ON runtime (tenant_id, name);
```
aszecowka: why do we have the second index `runtime_id_name_uindex`?
In GraphQLAPI we query by runtime ID:
```
runtime(id: ID!): Runtime
```

chart/compass/templates/migrator-job.yaml
```diff
@@ -0,0 +1,111 @@
...
+        "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+    template:
+        metadata:
+            labels:
+                app: {{ .Chart.Name }}
+                release: {{ .Release.Name }}
+        spec:
+            shareProcessNamespace: true
+            restartPolicy: OnFailure
```
aszecowka: usually in Kyma we use `restartPolicy` set to `Never`, but I would like to know how it will behave if `restartPolicy` is set to `OnFailure`. Does chart installation will fail if this job will keep failing?

chart/compass/templates/migrator-job.yaml
```diff
@@ -0,0 +1,111 @@
...
+                      - name: DB_SSL
+                        valueFrom:
+                            secretKeyRef:
+                                name: compass-postgresql
+                                key: postgresql-sslMode
+                  command:
+                    - "/bin/sh"
+                  args:
+                    - "-c"
+                    - "apk add --no-cache bash; \
```
aszecowka: this script should be part of the image (except killing pilot-agent)

components/schema-migrator/README.md
```diff
@@ -0,0 +1,9 @@
+# Schema Migrator
+
+## Overview
+
+The Schema Migrator is responsible for database schema migrations.
+
+## Development
+
+If you want to modify database schema used by compass, add migration files (follow [this](https://github.com/golang-migrate/migrate/blob/master/MIGRATIONS.md) instructions) to `migrations` directory and bump component version value in compass chart.
```
aszecowka: please write explicitly that new image of migrator will be produced that contains all migration files

chart/compass/charts/postgresql/values.yaml
tgorgol: I thought it would be easier that way to keep track of changes made to postgres chart in case we will have to update it.

chart/compass/templates/migrator-job.yaml
tgorgol: Initially, I was thinking about using that script in init container running on alpine, now that it's used in migrator job, it makes sense to move it into that image.

chart/compass/templates/migrator-job.yaml
tgorgol: Eventually, helm will return error:
```
Error: timed out waiting for the condition
```

But I decided to change this value to `Never` because as it's pointed in [this](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/?origin_team=T99LHPS1L#pod-backoff-failure-policy) doc it's easier to debug when failed containers are not terminated.

PR #149, opened 2019-07-18
Describe labeling feature

aszecowka:
This PR describes the proposal for implementing the labeling concept.


components/director/pkg/graphql/schema.graphql
```diff
@@ -456,6 +456,15 @@ input LabelFilter {
     operator: FilterOperator = ALL
 }
 
+type LabelDefinition {
+    key: String!
+    schema: String
```
pkosiec: Shouldn't be `CLOB` in this case?

components/director/pkg/graphql/schema.graphql
```diff
@@ -456,6 +456,15 @@ input LabelFilter {
...
+}
+
+input LabelDefinitionInput {
+    key: String!
+    schema: String
```
pkosiec: `CLOB`

components/director/pkg/graphql/schema.graphql
```diff
@@ -510,4 +522,9 @@ type Mutation {
 
     """If values are not specified, it deletes the label. If values are specified, it deletes values from a specific label."""
     deleteRuntimeLabel(runtimeID: ID!, key: String!, values: [String!]): Label
+
+    # Label Definitions
+    createLabelDefinition(in: LabelDefinitionInput!): LabelDefinition
+    updateLabelDefinition(in: LabelDefinitionInput!): LabelDefinition
+    deleteLabelDefinition(key: String!, force: Boolean=false): LabelDefinition
```
pkosiec: Please describe what `force` means. If it means that all labels related to this LabelDefinition are removed, then consider different name like `deleteRelatedLabels` or something like that

docs/architecture/labeling.md
```diff
@@ -0,0 +1,256 @@
...
+3. LabelDefinition is optional, but it will be created automatically if the user adds a label for which LabelDefinition does not exist.
+ 
+### API Changes
+
+1. Extend GraphQL API with  mutations and queries for a new type: **LabelDefinition** 
+
+```graphql
+type LabelDefinition {
+    key: String!
+    schema: String
```
pkosiec: `CLOB`?

docs/architecture/labeling.md
```diff
@@ -0,0 +1,256 @@
...
+}
+
+input LabelDefinitionInput {
+    key: String!
+    schema: String
```
pkosiec: `CLOB`?

docs/architecture/labeling.md
```diff
@@ -0,0 +1,256 @@
...
+
+type Query {
+    labelDefinitions: [LabelDefinition!]!
+    labelDefinition(key: String!): LabelDefinition
+}
+
+type Mutation {
+    createLabelDefinition(in: LabelDefinitionInput!): LabelDefinition!
+    updateLabelDefinition(in: LabelDefinitionInput!): LabelDefinition!
+    deleteLabelDefinition(key: String!, force: Boolean=false): LabelDefinition!
```
pkosiec: The same comment about `force` as above

docs/architecture/labeling.md
```diff
@@ -0,0 +1,256 @@
...
+See [this section](#database-schema)to learn how it can be implemented.
+Unfortunately, this functionality is planned for PostgreSQL 12, which is going to be released in Q3 2019, see [roadmap](https://www.postgresql.org/developer/roadmap/) and [features highlights](https://www.postgresql.org/about/news/1943/).
+We don't know when this version will be available on GCP or AWS, so, for now, we will be forced to use Postgres running inside the cluster.
+Also, not all relational databases support JSON Path Expressions, other than Postgres is [SQL Server](https://docs.microsoft.com/en-us/sql/relational-databases/json/json-path-expressions-sql-server?view=sql-server-2017) 
+
+#### Special case: Scenario Label
+For scenario label we have additional requirements:
+
+- there is an always `default` scenario
+- every application/runtime has to be assigned to at least one scenario. If not specified explicitly, the `default` scenario is used.
```
PK85: I'm not sure if every and each runtime should land in the default scenario. Adding new runtimes will always then already got default configuration.

docs/architecture/labeling.md
```diff
@@ -0,0 +1,256 @@
...
+
+1. On creation of a new tenant, label `Scenario` is created. Because right now we don't have a mutation for creating tenant, we need to
```
PK85: labelDefiniton is crated, and then per app/run label with "default" value

docs/architecture/labeling.md
```diff
@@ -0,0 +1,256 @@
...
+ https://github.com/kubernetes/kubernetes/blob/7faeee22b109d644f9a0ed736447e8867cae728e/staging/src/k8s.io/client-go/util/jsonpath/jsonpath.go
+
+- SQL/JSON Path Expressions (part of SQL-2016 specification http://www.sai.msu.su/~megera/postgres/talks/jsonpath-pgday.it-2019.pdf, https://news.ycombinator.com/item?id=19949240). 
+
+Options enumerated above are not compatible with each other.
+
+The simplest solution will be to use SQL/JSON Path, then we can propagate that value directly to the PostgreSQL. 
+See [this section](#database-schema)to learn how it can be implemented.
+Unfortunately, this functionality is planned for PostgreSQL 12, which is going to be released in Q3 2019, see [roadmap](https://www.postgresql.org/developer/roadmap/) and [features highlights](https://www.postgresql.org/about/news/1943/).
+We don't know when this version will be available on GCP or AWS, so, for now, we will be forced to use Postgres running inside the cluster.
```
PK85: Could we contact GCP and ask how fast they using new version of postgresql?

components/director/pkg/graphql/schema.graphql
aszecowka: IMO it is better to have a string here, because it is definitely simpler, and this schema will be usually much smaller than `ApiSpec.Data`

docs/architecture/labeling.md
```diff
@@ -0,0 +1,256 @@
+# Labeling
+
+## Requirements
+
+- User can label every top-level entity like Application, Runtime
+- User can search Application/Runtime by the specific label and its value
+- User can find out about all label keys used in the given tenant
+- User can define validation rules for a label with a given key but it is optional
```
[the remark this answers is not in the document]
aszecowka: See section: "Label an application without creation label definition"

docs/architecture/labeling.md
```diff
@@ -0,0 +1,256 @@
...
+
+
+setApplicationLabel(applicationID: "123", key: "supportedLanguages", value:"[Go]") {...}
+
+```
+
+#### Editing Label Definition
+Label definition can be edited. This will be used for example for label **Scenarios**.
+When editing definition, we need to ensure that all labels are compatible with the new definition.
+If this is not a case, such mutation has to be rejected, with a clear message that there are Applications or Runtimes that
```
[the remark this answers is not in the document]
aszecowka: You are right. Regarding your second question, we should return the list of Applications/Runtimes that violates a new definition.

aszecowka:
@pkosiec I created an issue about optimistic locking https://github.com/kyma-incubator/compass/issues/155

docs/architecture/labeling.md
aszecowka: I asked a question on PostgreSQL twitter account when the new version will be finally released. In addition to that, I asked on `googlecloud-community` slack when it will be available on GCP, unfortunately, I haven't got any response.

docs/architecture/labeling.md
```diff
@@ -0,0 +1,250 @@
...
+
+1. In JSON schema user can define if given label accepts one or many values. Because of that, we have to change
+or API and allow to specify only one value for a given label. This value can contain many elements, depending on LabelDefinition's schema.  
+Change from
+```graphql
+addApplicationLabel(applicationID: ID!, key: String!, values: [String!]!): Label!
+```
+to:
+```graphql
+setApplicationLabel(applicationID: ID!, key: String!, value: String!): Label!
```
pkosiec: Shouldn't we use `Any` for `value`?

pkosiec: For whole label probably `Any` scalar would be the best guess, because is converted to `interface{}`. If we would like to manipulate these values in future, it will be easier to not unmarshal object from string.

https://gqlgen.com/reference/scalars/#any

docs/architecture/labeling.md
pkosiec: Why do we use simple `String` value here and not `CLOB`? For all strings with schemas, specs etc. we used CLOB, so it would make sense to put it here as well.
We need to be consistent, so it's a good time to decide whether we go for CLOB, or we just abandon it completely (as String will fit as well).

Or, maybe we could use `Any` scalar - the output from GraphQL will be easier to parse as client won't have to parse JS object from string.

docs/architecture/labeling.md
aszecowka: for setting label value and LabelDefinition schema we use JSON which is equivalent to Any

docs/architecture/labeling.md
```diff
@@ -0,0 +1,252 @@
...
+
+1. In JSON schema user can define if given label accepts one or many values. Because of that, we have to change
+or API and allow to specify only one value for a given label. This value can contain many elements, depending on LabelDefinition's schema.  
+Change from
+```graphql
+addApplicationLabel(applicationID: ID!, key: String!, values: [String!]!): Label!
+```
+to:
+```graphql
+setApplicationLabel(applicationID: ID!, key: String!, value: JSON!): Label!
```
pkosiec: Please change `JSON` to `Any`, because the value could be a string, int or whatever. It's kind of confusing using JSON name.

aszecowka: done

Issue #124, opened 2019-07-09
Implement schema validation service


pkosiec:
**AC**
- Expose convenient API for validating value against JSON schema
- Use library to validate values against JSON Schema

https://json-schema.org/implementations.html#validators

PR #157, opened 2019-07-24
Implement json schema validation service

kfurgol:
**Description**

- Added service with json schema validation

**Related issue(s)**
#124

### DIFF
--- a/components/director/pkg/jsonschema/validator.go
+++ b/components/director/pkg/jsonschema/validator.go
@@ -8,10 +8,21 @@ type validator struct {
 	schema *gojsonschema.Schema
 }
 
-func NewValidator(schema *gojsonschema.Schema) *validator {
+func NewValidator(jsonSchema string) (*validator, error) {
+	var schema *gojsonschema.Schema
+	var err error
+
+	if jsonSchema != "" {
+		sl := gojsonschema.NewStringLoader(jsonSchema)
+		schema, err = gojsonschema.NewSchema(sl)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return &validator{
 		schema: schema,
-	}
+	}, nil
 }
 
 func (v *validator) Validate(json string) (bool, error) {
--- a/components/director/pkg/jsonschema/validator.go
+++ b/components/director/pkg/jsonschema/validator.go
@@ -1,0 +1,30 @@
+package jsonschema
+
+import (
+	"github.com/xeipuuv/gojsonschema"
+)
+
+type validator struct {
+}
+
+func NewValidator() *validator {
+	return &validator{}
+}
+
+func (v *validator) Validate(jsonschema, json string) (bool, error) {
+	schemaLoader := gojsonschema.NewSchemaLoader()
+	stringLoader := gojsonschema.NewStringLoader(jsonschema)
+
+	schema, err := schemaLoader.Compile(stringLoader)
+	if err != nil {
+		return false, err
+	}
+
+	jsonLoader := gojsonschema.NewStringLoader(json)
+	result, err := schema.Validate(jsonLoader)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


components/director/pkg/jsonschema/validator.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (v *validator) Validate(jsonschema string, jsons ...string) (bool, error) {
+
+	stringLoader := gojsonschema.NewStringLoader(jsonschema)
+	schema, err := gojsonschema.NewSchema(stringLoader)
+	if err != nil {
+		return false, err
+	}
+
+	for _, json := range jsons {
```
dbadura: Maybe we should return information which json is not invalid?

components/director/pkg/jsonschema/validator_test.go
```diff
@@ -0,0 +1,129 @@
...
+	  "foo": "test",
+	  "bar": "test",
+	  "baz": 123
+	}`
+	invalidInputJson := `{ "abc": 123 }`
+
+	testCases := []struct {
+		Name            string
+		InputJsonSchema string
+		InputJson       string
```
dbadura: Why not use []InputJsons?
And then call method: Validate(schema, inputJsons...)

components/director/pkg/jsonschema/validator.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+type validator struct {
+}
+
+func NewValidator() *validator {
+	return &validator{}
+}
+
+func (v *validator) Validate(jsonschema string, jsons ...string) (bool, error) {
```
aszecowka: 1. Validate should validate only one json - thanks to that error handling will be easier. 
2. Constructor for validator can accept jsonschema - thanks to that schema will be compiled only once

3. Think about the situation when user does not provide schema - then Validation should always succeed.

components/director/pkg/jsonschema/validator.go
kfurgol: it is no longer a case since we decided to check only one json value here

components/director/pkg/jsonschema/validator_test.go
kfurgol: it is no longer a case (same as above)

components/director/pkg/jsonschema/validator.go
kfurgol: done

components/director/pkg/jsonschema/validator.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/xeipuuv/gojsonschema"
+)
+
+type validator struct {
+	schema *gojsonschema.Schema
+}
+
+func NewValidator(schema *gojsonschema.Schema) *validator {
```
aszecowka: input should be string

components/director/pkg/jsonschema/validator_test.go
```diff
@@ -0,0 +1,97 @@
...
+			Name:            "Json schema and json doesn't match",
+			InputJsonSchema: validInputJsonSchema,
+			InputJson:       invalidInputJson,
+			ExpectedBool:    false,
+		},
+		{
+			Name:            "Empty",
+			InputJsonSchema: "",
+			InputJson:       "",
+			ExpectedBool:    true,
```
aszecowka: ExpectedBool - really good name ;)

components/director/pkg/jsonschema/validator.go
kfurgol: done

components/director/pkg/jsonschema/validator_test.go
kfurgol: changed to ExpectedResult

PR #161, opened 2019-07-24
Introduce LabelDefinition table + validate migration scripts

aszecowka:
- introduce SQL table for storing label definitions 
- added validation scripts that checks if all migrations (UP and DOWN) are correct. In the future, we should also propagate DB between specific migrations and probably use `golang-migrate` SDK instead of CLI, but for the time being CLI and bash script seems to be OK.


components/schema-migrator/Makefile
```diff
@@ -4,12 +4,16 @@ TAG = $(DOCKER_TAG)
...
-ci-master: build-image push-image
-ci-release: build-image push-image
+ci-pr: build-image validate push-image
+ci-master: build-image validate push-image
+ci-release: build-image validate push-image
 
 build-image:
 	docker build -t $(IMG_NAME) .
+
+validate:
```
tgorgol: Maybe build-image could be a dependency of validate?
```suggestion
validate: build-image
```

components/schema-migrator/validate.sh
```diff
@@ -0,0 +1,68 @@
...
+
+trap cleanup EXIT
+
+echo -e "${GREEN}Create network${NC}"
+docker network create --driver bridge ${NETWORK}
+
+echo -e "${GREEN}Start Postgres in detached mode${NC}"
+docker run -d --name ${POSTGRES_CONTAINER} \
+            --network=${NETWORK} \
+            -p 5432:5432 \
```
tgorgol: Does this container have to bind to hosts port?

components/schema-migrator/run.sh
```diff
@@ -10,6 +10,15 @@ if [ "${discoverUnsetVar}" = true ] ; then
     exit 1
 fi
 
+if [[ "${DIRECTION}" == "up" ]]; then
```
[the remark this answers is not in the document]
aszecowka: Added to the loop but here in addition to that, I want to validate its value.

components/schema-migrator/Makefile
aszecowka: done

components/schema-migrator/validate.sh
aszecowka: not needed, removed

components/schema-migrator/migrations/2_label_definition.up.sql
```diff
@@ -0,0 +1,9 @@
+create table label_definitions
```
crabtree: please consider renaming this file to avoid sorting issues.

components/schema-migrator/migrations/2_label_definition.down.sql
```diff
@@ -0,0 +1 @@
+drop table label_definitions;
```
crabtree: same as with up.sql file

components/schema-migrator/validate.sh
```diff
@@ -0,0 +1,67 @@
...
+    docker network rm ${NETWORK}
+}
+
+trap cleanup EXIT
+
+echo -e "${GREEN}Create network${NC}"
+docker network create --driver bridge ${NETWORK}
+
+echo -e "${GREEN}Start Postgres in detached mode${NC}"
+docker run -d --name ${POSTGRES_CONTAINER} \
```
crabtree: consider adding `--rm` parameter

components/schema-migrator/validate.sh
```diff
@@ -0,0 +1,67 @@
...
+            --network=${NETWORK} \
+            -e POSTGRES_USER="usr" \
+            -e POSTGRES_PASSWORD="pwd" \
+            -e POSTGRES_DB="compass" \
+            postgres:${POSTGRES_VERSION}
+
+echo -e "${GREEN}Run UP migrations ${NC}"
+
+docker run --rm --network=${NETWORK} \
+        -e DB_USER="usr" \
```
crabtree: maybe it is worth to move all those common env variables values to some ENV variables in this script?

components/schema-migrator/validate.sh
aszecowka: `--rm                             Automatically remove the container when it exits`- but this actually won't happen IMO, because postgres needs to be killed explicitly, which is done cleanup function:
`    docker rm --force ${POSTGRES_CONTAINER}`

Issue #122, opened 2019-07-09
Update GraphQL schema for Labels


pkosiec:
**AC**
- Modify Label scalar  (key: value)
- Update schema for label filtering. Implementation will be done in a separate task.
- Prepare GraphQL types for LabelDefinition
- CRUD for LabelDefinition (queries and mutations)
- Rename `addApplicationLabel` to `setApplicationLabel` (for runtimes too)

These changes should be done according to proposal from https://github.com/kyma-incubator/compass/pull/149. 

Remember about integration tests.

PR #164, opened 2019-07-24
Adjust Director integration tests for new GraphQL schema

pkosiec:
**Description**

- Adjust Director integration tests for new GraphQL schema
- Fix building binaries for before-commit

**Related issue(s)**
See https://github.com/kyma-incubator/compass/issues/122


components/connector/before-commit.sh
```diff
@@ -55,9 +55,9 @@ if [[ "$1" == "$CI_FLAG" ]]; then
 	buildEnv="env CGO_ENABLED=0 GOOS=linux GOARCH=amd64"
 fi
 
-${buildEnv} go build -o connector ./cmd/main.go
+${buildEnv} go build -o bin/connector ./cmd/main.go
```
aszecowka: don't mix many fixes in one PR, this is extremely important PR and you decided to fix storing binaries in our project...

tests/end-to-end/before-commit.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+bin/#!/usr/bin/env bash
```
aszecowka: probably not necessary

components/director/internal/domain/application/resolver.go
```diff
@@ -214,19 +214,21 @@ func (r *Resolver) SetApplicationLabel(ctx context.Context, applicationID string
 }
 
 func (r *Resolver) DeleteApplicationLabel(ctx context.Context, applicationID string, key string) (*graphql.Label, error) {
-	err := r.appSvc.DeleteLabel(ctx, applicationID, key)
+	app, err := r.appSvc.Get(ctx, applicationID)
```
crabtree: Do you really need to fetch the application? Underneath, there may be a lot of database calls, that are required to fetch all application object and you need here only the value of the label. In general, I'm not sure if we should return the actual value of the object that was deleted. For me, it is rather confusing.

components/director/internal/domain/runtime/resolver.go
```diff
@@ -142,18 +142,20 @@ func (r *Resolver) SetRuntimeLabel(ctx context.Context, runtimeID string, key st
 	}, nil
 }
 func (r *Resolver) DeleteRuntimeLabel(ctx context.Context, runtimeID string, key string) (*graphql.Label, error) {
-	err := r.svc.DeleteLabel(ctx, runtimeID, key)
+	runtime, err := r.svc.Get(ctx, runtimeID)
```
crabtree: Same as with the applications.

components/connector/before-commit.sh
pkosiec: Because I accidentally commited one binary in previous PR, so that's why I fixed it "in the meantime".

components/director/internal/domain/application/resolver.go
pkosiec: Unfortunately I need label values to return for GraphQL mutation result. But yes, I agree, we can think about these improvements later.

Issue #142, opened 2019-07-16
Allow to update only basic application attributes


pkosiec:
**AC**
- Update and merge @tgorgol investigation
- Notify UI team about the changes
- Change update mutations for Application: 
  - Use different input for update and creation
  - Define update input without any references to webhooks, documents, etc.
  - check if runtime create & update mutations work the same way
  - update integration tests
- ~handle a case where user wants to remove all APIs(docs, etc) from application~ out of scope

**Reasons**
To update application name, user should use `updateApplication` mutation and send just Application object, without webhooks, docs, apis and any other references. The same thing is with the Runtimes.

pkosiec:
See the PR for more details: https://github.com/kyma-incubator/compass/pull/173

pkosiec:
As discussed, we will change `update` mutations to update only root objects, for example.:

```graphql
input ApplicationCreateInput {
    name: String!
    description: String
    labels: Labels
    webhooks: [WebhookInput!]
    healthCheckURL: String
    apis: [APIDefinitionInput!]
    eventAPIs: [EventAPIDefinitionInput!]
    documents: [DocumentInput!]
}

input ApplicationUpdateInput {
    name: String!
    description: String
    labels: Labels
    healthCheckURL: String
}
```

aszecowka:
Please ensure that [this](https://github.com/kyma-incubator/compass/pull/280/files) investigation document is merged

PR #173, opened 2019-07-26
Allow to update only basic application attributes

tgorgol:
**Description**

- Add investigation document
- Extend GQL `SpecFormat` enum with `XML` value

**Related issue(s)**
#142

### DIFF
--- a/docs/investigations/graphql-partial-updates/graphql-partial-updates.md
+++ b/docs/investigations/graphql-partial-updates/graphql-partial-updates.md
@@ -1,0 +1,129 @@
+# GraphQL Partial Updates
+
+## Overview
+
+This documents describes and compares different approaches to designing partial updates in graphql schema. Every proposed approach has a link to an example it was based on. 
+
+## Partial update mutation designs
+
+### 1. Simple Patch
+
+#### Schema
+```graphql
+input ApplicationPatchInput {
+    name: String
+    description: String
+    webhooks: [WebhookInput!]
+    healthCheckURL: String
+    apis: [APIDefinitionInput!]
+    eventAPIs: [EventAPIDefinitionInput!]
+    documents: [DocumentInput!]
+}
+
+type Mutation {
+    updateApplication(id: ID!, patch: ApplicationPatchInput!): Application!
+}
+```
+
+#### Description
+In this approach every field has to be optional. If the field is defined, we use the passed value to set the new value. If the field is not defined we don't change the current value.
+
+Because we use the `nil` value to determine whether to change current value or not we lose the ability to set the actual value to `nil`. In the example linked below this limitation doesn't exist because in javascript we can distinguish `null` and `undefined` values.
+
+Example: [link](https://medium.com/workflowgen/graphql-mutations-partial-updates-implementation-bff586bda989)
+
+### 2. Simple Patch (single input)
+
+#### Schema
+```graphql
+input ApplicationUpdateInput {
+    id: ID!,
+    patch: ApplicationPatchInput!
+}
+
+input ApplicationPatchInput {
+    name: String
+    description: String
+    webhooks: [WebhookInput!]
+    healthCheckURL: String
+    apis: [APIDefinitionInput!]
+    eventAPIs: [EventAPIDefinitionInput!]
+    documents: [DocumentInput!]
+}
+
+type Mutation {
+    updateApplication(in: ApplicationUpdateInput!): Application!
+}
+```
+
+#### Description
+This approach is very similar to the previous one, with one difference: there has to be only one, unique input object. In the example linked below it is argued that it make client implementation easier.
+
+Patching logic is the same as in previous approach.
+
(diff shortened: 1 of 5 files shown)


docs/investigations/graphql-partial-updates/graphql-partial-updates.md
```diff
@@ -0,0 +1,156 @@
...
+    eventAPIs: [EventAPIDefinitionInput!]
+    documents: [DocumentInput!]
+}
+
+type Mutation {
+    updateApplication(id: ID!, patch: ApplicationPatchInput!): Application!
+}
+```
+
+#### Description
```
aszecowka: 1. please add examples:
- how to set description to empty value
- how to remove all eventAPIs for Application

2. Let assume that name is required, and the description is optional. Previously such validation was done on the schema label, now we have to implement it on our own. Where this validation will be added?

3. Let assume that user wants to update the only name, can you describe how it will be handled on the resolver/service/repository level? Are we going to create a SQL query builder to update only one field for Application?

docs/investigations/graphql-partial-updates/graphql-partial-updates.md
```diff
@@ -0,0 +1,156 @@
...
+    setName: SetApplicationName
+    setDescription: SetApplicationDescription
+    setWebhooks: SetApplicationWebhooks
+    setHealthCheckURL: SetApplicationHealthCheckURL
+    setAPIs: SetApplicationAPIs
+    setEventAPIs: SetApplicationEventAPIs
+    setDocuments: SetApplicationDocuments
+}
+
+input SetApplicationName {
```
aszecowka: to check: 
can we define only one type for string input, `SetString`, instead of having `SetApplicationName`, `SetApplicationDescription` ?

docs/investigations/graphql-partial-updates/graphql-partial-updates.md
```diff
@@ -0,0 +1,156 @@
...
+    documents: [DocumentInput!]
+}
+
+type Mutation {
+    updateApplication(id: ID!, patch: ApplicationPatchInput!): Application!
+}
+```
+
+#### Description
+In this approach every field has to be optional. If the field is defined, we use the passed value to set the new value. If the field is not defined we don't change the current value.
```
aszecowka: For string, this approach seems to be ok, but I am wondering about fields that store for example integer, is `0` correct value? Let assume that we have `size` for runtime :) representing the number of nodes.
- 0 can mean it was scaled down to 0 nodes
- 0 can mean that we don't know

and I am not convinced that this is ok.

docs/investigations/graphql-partial-updates/graphql-partial-updates.md
```diff
@@ -0,0 +1,156 @@
+# GraphQL Partial Updates
```
aszecowka: to check: If some string field is required, can I provide "" as its value?

docs/investigations/graphql-partial-updates/graphql-partial-updates.md
tgorgol: We would need at least two types, one for optional and one for required string.

docs/investigations/graphql-partial-updates/graphql-partial-updates.md
tgorgol: Yes, `""` is a valid string value for a required field.

docs/investigations/graphql-partial-updates/graphql-partial-updates.md
tgorgol: Ad 2. In this approach we're unable to set field to `nil`, so the fields that were required during creation won't be removed during update. We're still able to set name to `""` but it's the same in current version of our API.

Ad 3. I added appendix to the doc about this.

docs/investigations/graphql-partial-updates/graphql-partial-updates.md
kfurgol: If such a case occurred, we came up with several ideas:
First, we thought about providing a wrapper type which would look like that:
```
input nilableInteger {
	value: Int
	isNil: Boolean
}
```
and the size would be of type `nilableInteger`.
In the case when we don't know the value, we would set `isNil` value to true. Otherwise, we would set `isNil` to false and `value` to 0.

Another idea would be that we stick to strings in our api regardless of what value would suggest. So the `size` value would be a string.

Also, we investigated more the `Commands & Actions` approach, where such a problem wouldn't occur, but we would have to specify setter inputs for pretty much all of inputs.

There is also a case to consider with default values. For example, for now, we have `VersionInput` type which has two boolean values with default values. Considering a case when we want to update just value of the version type, and actual booleans differ from the default values, they would've been overridden. One of our propositions to handle that case is to create another input type for the `Version`.

We have to remember that if we want to use `PATCH` updates, we'll have to specify additional inputs for each existing `PUT` input

PR #177, opened 2019-07-29
Handle Create Label Definition

aszecowka:
TBD soon, stay tuned!


components/director/internal/model/labledef_test.go
```diff
@@ -0,0 +1,65 @@
...
+	t.Run("valid input when schema not provided", func(t *testing.T) {
+		// GIVEN
+		in := model.LabelDefinition{ID: "id", Key: "key", Tenant: "tenant"}
+		// WHEN
+		err := in.Validate()
+		// THEN
+		require.NoError(t, err)
+
+	})
+	t.Run("valid input when correct schema provided", func(t *testing.T) {
```
aszecowka: I am aware of that TODO :)

components/director/internal/domain/labeldef/entity.go
```diff
@@ -0,0 +1,8 @@
+package labeldef
+
+type Entity struct {
+	ID         string `db:"id"`
+	TenantID   string `db:"tenantID"`
+	Key        string `db:"key"`
+	SchemaJSON string `db:"schemaJSON"`
```
tgorgol: if SchemaJSON is a `string` and user does not specify the schema while creating new Label Definition it will be an empty string which won't be a correct value for `jsonb` so db will reject the creation.

components/director/internal/domain/labeldef/repository.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	ld, err := r.conv.FromEntity(dest)
+	if err != nil {
+		return nil, errors.Wrap(err, "while converting Label Definition")
+	}
+	return &ld, nil
+}
+
+func (r *repo) Exists(ctx context.Context, tenant string, key string) (bool, error) {
```
pkosiec: It works, but you pull whole object from DB. Exists method should be implemented like in Runtime domain:

```go
func (r *pgRepository) Exists(ctx context.Context, tenant, id string) (bool, error) {
	persist, err := persistence.FromCtx(ctx)
	if err != nil {
		return false, errors.Wrap(err, "while fetching DB from context")
	}

	stmt := fmt.Sprintf(`SELECT 1 FROM %s WHERE "id" = $1 AND "tenant_id" = $2`,
		runtimeTable)

	var count int
	err = persist.Get(&count, stmt, id, tenant)
	if err != nil {
		if err == sql.ErrNoRows {
			return false, nil
		}
		return false, errors.Wrap(err, "while getting runtime from DB")
	}

	return true, nil
}
```

If you don't want to do this modification right now, create an issue for that.

components/director/internal/domain/labeldef/repository.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, err
+	}
+	dest := Entity{}
+
+	q := fmt.Sprintf("select * from %s where tenant_id=$1 and key=$2 ", tableName)
+
+	err = db.Get(&dest, q, tenant, key)
+	switch {
+	case err == sql.ErrNoRows:
+		return nil, nil
```
pkosiec: For later: We need to discuss common approach for the "Not found" situation. I think the repository should return an error, but the one of custom type (our own "Entity not found" error).

components/director/internal/domain/labeldef/service_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+}
+
+func fixUID() string {
+	return "003a0855-4eb0-486d-8fc6-3ab2f2312ca0"
+}
+
+func fixSchema(t *testing.T) *interface{} {
+	sch := `{
+  "$id": "https://example.com/person.schema.json",
```
pkosiec: add tabs

components/director/internal/model/labeldef.go
```diff
@@ -0,0 +1,34 @@
...
+
+type LabelDefinition struct {
+	ID     string
+	Tenant string
+	Key    string
+	Schema *interface{}
+}
+
+func (def *LabelDefinition) Validate() error {
+	if def.ID == "" {
```
pkosiec: Do we really need to validate these fields? Schema should be validated, but what about other values? The non-empty values are enforced in DB and GraphQL API, so it's rather not necessary.

components/director/pkg/jsonschema/validator.go
```diff
@@ -25,8 +26,26 @@ func NewValidator(jsonSchema string) (*validator, error) {
 	}, nil
 }
 
-func (v *validator) Validate(json string) (bool, error) {
+func NewValidatorFromSchema(jsonSchema interface{}) (*validator, error) {
```
pkosiec: Please rename to `NewValidatorFromRawSchema`. But to be honest we won't use `NewValidatorFromStringSchema`, so probably it's safe to delete and then rename the constructor simply as `NewValidator`.

components/director/internal/domain/labeldef/entity.go
aszecowka: schemaJSON is now sql.NullString, it is saved only if it is provided.

components/director/internal/model/labeldef.go
aszecowka: in GraphQL we mark that some fields are required, but still, you can provide empty string there, which is not proper value. If object is not valid, there is not too much sense to call DB at all.

PR #182, opened 2019-07-30
Fix listing runtimes

tgorgol:
**Description**

- Converting runtime entity to runtime model now correctly copies the description


tests/end-to-end/director/runtime_api_test.go
```diff
@@ -357,9 +357,16 @@ func TestQueryRuntimes(t *testing.T) {
 		}
 	}()
 
+	inputRuntimes := []*graphql.Runtime{
+		{Name: "runtime-query-1", Description: ptrString("test description")},
+		{Name: "runtime-query-2", Description: ptrString("another description")},
+		{Name: "runtime-query-3"},
+	}
+
 	for i := 0; i < 3; i++ {
```
aszecowka: iterate over inputRuntimes, if we change it to only 2 elements we will get panic

tests/end-to-end/director/runtime_api_test.go
```diff
@@ -391,7 +399,7 @@ func TestQueryRuntimes(t *testing.T) {
 	require.NoError(t, err)
 	assert.Len(t, actualPage.Data, 3)
 	assert.Equal(t, 3, actualPage.TotalCount)
-
+	runtimesContain(t, actualPage.Data, inputRuntimes)
```
aszecowka: Not tested, but it looks much simpler:
```
	for _, v := range inputRuntimes {
		found := false
		for _, actual := range actualPage.Data {
			if actual.Name == v.Name {
				found = true
				//assert.Equal desc
				// assert.Equal id
				break
			}
		}
		assert.True(t, found)
	}
```

PR #185, opened 2019-07-31
Query for label definitions

aszecowka:
- labelDefinitions
- labelDefinition

### DIFF
--- a/components/director/internal/domain/labeldef/resolver.go
+++ b/components/director/internal/domain/labeldef/resolver.go
@@ -36,6 +36,8 @@ type Converter interface {
 //go:generate mockery -name=Service -output=automock -outpkg=automock -case=underscore
 type Service interface {
 	Create(ctx context.Context, ld model.LabelDefinition) (model.LabelDefinition, error)
+	Get(ctx context.Context, tenant string, key string) (*model.LabelDefinition, error)
+	List(ctx context.Context, tenant string) ([]model.LabelDefinition, error)
 }
 
 func (r *Resolver) CreateLabelDefinition(ctx context.Context, in graphql.LabelDefinitionInput) (*graphql.LabelDefinition, error) {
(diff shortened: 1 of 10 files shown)


components/director/internal/domain/labeldef/resolver.go
```diff
@@ -64,3 +66,60 @@ func (r *Resolver) CreateLabelDefinition(ctx context.Context, in graphql.LabelDe
...
+	def, err := r.srv.Get(ctx, tnt, key)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "while getting Label Definition")
+	}
+	if err := tx.Commit(); err != nil {
+		return nil, errors.Wrap(err, "while committing transaction")
+	}
+
+	if def == nil {
```
tgorgol: Shouldn' we return error if not found?

components/director/internal/domain/labeldef/resolver.go
```diff
@@ -64,3 +67,60 @@ func (r *Resolver) CreateLabelDefinition(ctx context.Context, in graphql.LabelDe
...
+		return nil, fmt.Errorf("label defintion with key='%s' does not exist", key)
```
pkosiec: typo + remove `=` to be consistent with other errors

components/director/internal/domain/labeldef/resolver.go
pkosiec: I believe we shouldn't, as it is top level query.

components/director/internal/domain/labeldef/resolver_test.go
```diff
@@ -82,11 +78,11 @@ func TestCreateLabelDefinition(t *testing.T) {
...
 	t.Run("got error on creating Label Definition", func(t *testing.T) {
 		// GIVEN
-		mockPersistanceCtx := &pautomock.PersistenceTxOp{}
-		defer mockPersistanceCtx.AssertExpectations(t)
+		mockPersistenceCtx := &pautomock.PersistenceTxOp{}
+		defer mockPersistenceCtx.AssertExpectations(t)
 		mockTransactioner := &pautomock.Transactioner{}
-		mockTransactioner.On("Begin").Return(mockPersistanceCtx, nil)
-		mockTransactioner.On("RollbackUnlessCommited", mockPersistanceCtx).Return(nil)
+		mockTransactioner.On("Begin").Return(mockPersistenceCtx, nil)
```
pkosiec: It would be good to check how many times all mocked methods has been called. For example here, it would be good to add `Once()` expectation.

components/director/internal/domain/labeldef/resolver.go
aszecowka: for now, we return an error to be consistent with other endpoints, but follow up task will be created

PR #186, opened 2019-07-31
Add filtering for runtimes repository, support for Scenarios

crabtree:
**Description**

- Support for filtering via labels filter in runtimes repository
- Handling the Scenarios label with SQL/JSON path filter

**Related issue(s)**

### DIFF
--- a/components/director/internal/domain/label/query.go
+++ b/components/director/internal/domain/label/query.go
@@ -1,0 +1,65 @@
+package label
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/lib/pq"
+
+	"github.com/google/uuid"
+	"github.com/kyma-incubator/compass/components/director/internal/labelfilter"
+)
+
+// QueryFor type defines possible types for quering
+type QueryFor string
+
+const (
+	// QueryForApplication defines the filter query for applications
+	QueryForApplication QueryFor = "app_id"
+	// QueryForRuntime defines the filter query for runtimes
+	QueryForRuntime QueryFor = "runtime_id"
+
+	labelsTable       string = `"public"."labels"`
+	scenariosLabelKey string = "SCENARIOS"
+	stmtPrefixFormat  string = `SELECT "%s" FROM %s WHERE "tenant_id" = '%s'`
+)
+
+// FilterQuery builds select query for given filters
+//
+// It supports quering defined by `QueryFor` parameter. All queries are created
+// in the context of given tenant.
+func FilterQuery(queryFor QueryFor, tenant uuid.UUID, filter []*labelfilter.LabelFilter) string {
+	if filter == nil {
+		return ""
+	}
+
+	stmtPrefix := fmt.Sprintf(stmtPrefixFormat, queryFor, labelsTable, tenant)
+
+	var queryBuilder strings.Builder
+	for idx, lblFilter := range filter {
+		if idx > 0 {
+			queryBuilder.WriteString(` INTERSECT `)
+		}
+
+		queryBuilder.WriteString(fmt.Sprintf(stmtPrefix))
+
+		// TODO: for optimization it can be detected if the given Key was already added to the query
+		// if so, it can be ommited
+		queryBuilder.WriteString(fmt.Sprintf(` AND "key" = %s`, pq.QuoteLiteral(lblFilter.Key)))
+
+		if lblFilter.Query != nil {
+			queryValue := *lblFilter.Query
+			// Handling the Scenarios label case - we assume that Query is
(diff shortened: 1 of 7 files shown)


components/director/internal/domain/label/query_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"testing"
+
+	"github.com/google/uuid"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/kyma-incubator/compass/components/director/internal/labelfilter"
+)
+
+func Test_FilterQuery(t *testing.T) {
```
aszecowka: Such query is not going to work:
```
SELECT "runtime_id" FROM "public"."labels" WHERE "tenant_id" = 'aa5eac41-c2a4-4ead-95b6-5574505e81c5' AND "key" = 'Foo' AND "key" = 'Bar'
```

I would expect that the following query will be created:
```
select l1.app_id from labels l1 join labels l2 on l1.app_id =l2.app_id where l1.key='foo' and l2.key='bar';
```

but I am wondering about the performance of such query...

crabtree: I'm not a fan of this approach as when we will have more labels to filter then the join will grow and grow.

crabtree: The alternative to the join queries will be the intersect.

JOIN:
`select l1.runtime_id from labels l1 join labels l2 on l2.runtime_id = l1.runtime_id and l2.key = 'label1' join labels l3 on l3.runtime_id = l1.runtime_id and l3.key = 'label3' where l1.key = 'label1';`

INTERSECT
`select runtime_id from labels where key = 'label1' intersect select runtime_id from labels where key = 'label2' intersect select runtime_id from labels where key = 'label3';`

components/director/internal/domain/label/query.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/google/uuid"
+	"github.com/kyma-incubator/compass/components/director/internal/labelfilter"
+)
+
+// QueryFor type defines possible types for quering
+type QueryFor string
+
+const (
+	// QueryForApplication defines the filter query for applications
+	QueryForApplication QueryFor = "app_id"
```
dbadura: It would be nice to have ability to query for  whole`Application` or `Runtime` object

crabtree: But this does not query the database, it just provides you the adequate subquery that can be then used in your context. As an example, you can refer to the runtime repository List func. In this func, the query produces runtimes entities filtered by the subquery provided by the label filter.

components/director/internal/domain/label/translator.go
```diff
@@ -0,0 +1,15 @@
+package label
+
+import "regexp"
+
+// ExtractValueFromJSONPath returns the value that is placed in the SQL/JSON path query
+func ExtractValueFromJSONPath(jpq string) *string {
```
aszecowka: IMO in this method, we should be strict and if a query does not match regexp, return an error

components/director/internal/domain/label/translator.go
```diff
@@ -0,0 +1,15 @@
+package label
+
+import "regexp"
+
+// ExtractValueFromJSONPath returns the value that is placed in the SQL/JSON path query
```
aszecowka: add example in the comment what this method is doing.

components/director/internal/domain/label/query.go
```diff
@@ -0,0 +1,57 @@
...
+	"strings"
+
+	"github.com/lib/pq"
+
+	"github.com/google/uuid"
+	"github.com/kyma-incubator/compass/components/director/internal/labelfilter"
+)
+
+// QueryFor type defines possible types for quering
+type QueryFor string
```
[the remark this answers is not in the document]
crabtree: Done.

components/director/internal/domain/label/query_test.go
```diff
@@ -0,0 +1,139 @@
...
+	filterScenariosWithFooValues := labelfilter.LabelFilter{
+		Key:   "Scenarios",
+		Query: &scenariosFooQuery,
+	}
+	filterScenariosWithbarPongValues := labelfilter.LabelFilter{
+		Key:   "Scenarios",
+		Query: &scenariosBarPongQuery,
+	}
+
+	stmtPrefix := `SELECT "runtime_id" FROM "public"."labels" WHERE "runtime_id" IS NOT NULL AND "tenant_id" = '` + tenantID.String() + `'`
```
dbadura: Break it, because is too long

dbadura: Can you break this query after `WHERE`

Issue #125, opened 2019-07-09
Create LabelDefinition Scenario with default value


pkosiec:
**AC**
- While creating Runtime/Application ensure that LabelDefinition for Scenarios exists (along with single DEFAULT value). 
- While creating Application without any custom scenarios ensure that Label `scenarios` with `DEFAULT` value is assigned.
- Prevent removing DEFAULT value for `scenarios` LabelDefinition or the LabelDefinition itself
- Remember about integration tests
- Application has to be in at least one scenario, while Runtime doesn't

PR #190, opened 2019-08-02
Create LabelDefinition Scenario with default value

tgorgol:
**Description**

- implement `deleteLabelDefinition` mutation, which:
  - deletes Label Definition if no existing Labels use that Label Definition
  - returns error if any of existing Labels use that Label Definition
  - returns error if someone tries to delete `scenarios` Label Definition
  - **DOESN'T SUPPORT CASCADING DELETE YET**
- define default `scenarios` Label Definition, with value constraints:
  - has to be an array of string enums
  - has to have at least one element (to remove all `scenarios` from Runtime, the Label itself has to be deleted)
  - has to contain `DEFAULT` element
- define schema which validates that given schema is a valid `scenarios` schema (that only enum changed)
- add scenarios service which ensures that `scenarios` Label Definition exists in tenant
- modify Application creation to:
  - ensure `scenarios` Label Definition exists
  - set `DEFAULT` scenario for newly created Applications (unless they already specify other scenario)
- modify Runtime creation to:
  - ensure `scenarios` Label Definition exists
- modify Label deletion from Application to:
  - return error if someone tries to delete `scenarios` Label
- modify validating Labels value to:
  - include list of validation errors
- modify Label Definition validation to:
  - validate `scenarios` Label Definition schema against defined schema for that schema
- modify validator to:
  - include list of validation errors
  - when validating raw input, marshal it and use string validator

**Related issue(s)**
#125


components/director/internal/domain/runtime/service.go
```diff
@@ -270,3 +282,24 @@ func (s *service) DeleteLabel(ctx context.Context, runtimeID string, key string)
 
 	return nil
 }
+
+func (s *service) ensureScenariosLabelDefinitionExists(ctx context.Context, tenant string) error {
+	ldExists, err := s.ldRepo.Exists(ctx, tenant, model.ScenariosKey)
```
aszecowka: this is the same code as for application, we can create a service `ScenarioLabelService` and move this logic there. I don't see any problems with the situation when service depends on another service (even circular dependencies are solvable).

components/director/internal/model/labeldef.go
```diff
@@ -5,6 +5,57 @@ import (
...
+	ScenariosSchema = map[string]interface{}{
+		"type":        "array",
+		"minItems":    1,
+		"uniqueItems": true,
+		"items": map[string]interface{}{
+			"type": "string",
+			"enum": []string{"DEFAULT"},
+		},
+	}
+	ScenariosSchemaDefinitionSchema = map[string]interface{}{
```
aszecowka: maybe better name will be `SchemaForScenariosSchema`, please add a comment to that variable to explain its purpose

components/director/internal/model/labeldef.go
```diff
@@ -5,6 +5,57 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
```
aszecowka: maybe create a new file `scenarios.go` in the same package

components/director/internal/model/labeldef.go
```diff
@@ -32,3 +83,46 @@ func (def *LabelDefinition) Validate() error {
...
+
+	if def.Key == "" {
+		return errors.New("missing Key field")
+	}
+	if def.Schema != nil {
+		if _, err := jsonschema.NewValidatorFromRawSchema(def.Schema); err != nil {
+			return errors.Wrapf(err, "while validating schema: [%+v]", *def.Schema)
+		}
+	}
+	if def.Key == ScenariosKey {
```
aszecowka: the same validation can be executed for Creation

components/director/internal/domain/labeldef/service.go
```diff
@@ -60,3 +64,49 @@ func (s *service) List(ctx context.Context, tenant string) ([]model.LabelDefinit
...
+		return model.LabelDefinition{}, errors.Wrap(err, "while receiving updated Label Definition")
+	}
+
+	return *updatedLabelDef, nil
+}
+
+// TODO: Add deleting related labels logic
+func (s *service) Delete(ctx context.Context, tenant, key string, deleteRelatedLabels bool) error {
+	if key == model.ScenariosKey {
+		return errors.New(fmt.Sprintf("Label Definition with key %s can not be deleted", model.ScenariosKey))
```
aszecowka: use fmt.Errrof

components/director/internal/model/labeldef.go
aszecowka: this approach is really cool, schema for schema 😎

components/director/pkg/jsonschema/validator.go
```diff
@@ -45,30 +47,75 @@ func NewValidatorFromRawSchema(jsonSchema interface{}) (*validator, error) {
...
+			Valid: true,
+			Error: nil,
+		}, nil
 	}
 
 	jsonLoader := gojsonschema.NewStringLoader(json)
 	result, err := v.schema.Validate(jsonLoader)
 	if err != nil {
-		return false, err
+		return ValidationResult{
```
aszecowka: `return ValidationResult{}, err`

components/director/pkg/jsonschema/validator.go
```diff
@@ -45,30 +47,75 @@ func NewValidatorFromRawSchema(jsonSchema interface{}) (*validator, error) {
...
+			Valid: false,
+			Error: nil,
+		}, err
 	}
 
-	return result.Valid(), nil
+	var validationError error
+	if len(result.Errors()) > 0 {
```
aszecowka: to discuss: using mutlierror, for example `https://github.com/uber-go/multierr/blob/master/error.go`

components/director/pkg/jsonschema/validator.go
```diff
@@ -45,30 +47,75 @@ func NewValidatorFromRawSchema(jsonSchema interface{}) (*validator, error) {
...
+			Error: nil,
+		}, errors.Wrapf(err, "while validating value %+v against schema %+v", value, v.schema)
 	}
 
-	return result.Valid(), nil
+	var validationError error
+	if len(result.Errors()) > 0 {
+		errStr := result.Errors()[0].String()
+		for i := 1; i < len(result.Errors()); i++ {
+			errStr += ", " + result.Errors()[i].String()
```
aszecowka: this is probably not covered by the tests

Issue #151, opened 2019-07-22
Enable Agent to fetch configuration for Runtime


PK85:
**AC**

- fetch the Runtime configuration the Runtime Agent by calling applicationsForRuntime(runtimeId: ID!) query offered by Director (as separate PR)

See flow details:
- https://github.com/kyma-incubator/compass/blob/master/docs/architecture/runtime-agent-contract.md

pkosiec:
For now we don't have labeling and pagination, so please return a fixed page with for example all apps.

PR #192, opened 2019-08-05
Application for runtimes with real labels filtering

dbadura:
**Description**

- Add ability to get application for runtime, filtered on scenarios
- Add test e2e

**Related issue(s)**

See #151


components/director/internal/domain/application/repository.go
```diff
@@ -57,12 +67,71 @@ func (r *inMemoryRepository) List(ctx context.Context, tenant string, filter []*
...
+		return nil, errors.Wrap(err, "while fetching DB from context")
+	}
+
+	//TODO: change tenantID from UUID to String when
+	tenantUUID, err := uuid.Parse(tenantID)
+	if err != nil {
+		return nil, errors.New("tenant_ID is not parseable")
+	}
+
+	stmt := fmt.Sprintf(`SELECT VALUE FROM %s WHERE TENANT_ID=$1 AND RUNTIME_ID=$2 AND KEY='SCENARIOS'`, labelTableName)
```
aszecowka: it is really strange that your application repository is responsible for searching in labels table. Getting scenarios values should be done on the application level (by calling labels repository) and exact values should be passed to this method.

components/director/internal/domain/application/repository.go
```diff
@@ -57,12 +67,71 @@ func (r *inMemoryRepository) List(ctx context.Context, tenant string, filter []*
...
+
+	var scenariosJSON string
+	err = persist.Get(&scenariosJSON, stmt, tenantID, runtimeID)
+
+	if scenariosJSON == "" {
+		return nil, errors.New("Runtime scenarios not found")
+	}
+	scenarios := getScenariosValues(scenariosJSON)
+
+	var scenarioFilers []*labelfilter.LabelFilter
```
aszecowka: typo

examples/query-runtimes.graphql
```diff
@@ -1,54 +1,52 @@
 # Code generated by Compass integration tests, DO NOT EDIT.
 query {
-  result: runtimes {
```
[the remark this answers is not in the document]
dbadura: It was turned off.

components/director/internal/domain/application/repository.go
```diff
@@ -57,12 +67,71 @@ func (r *inMemoryRepository) List(ctx context.Context, tenant string, filter []*
 	}, nil
 }
 
-// TODO: Make filtering and paging
-func (r *inMemoryRepository) ListByRuntimeID(ctx context.Context, tenant, runtimeID string, pageSize *int, cursor *string) (*model.ApplicationPage, error) {
+// TODO: @dbadura add pagination when PR-181 is merged
+func (r *inMemoryRepository) ListByScenariosForRuntime(ctx context.Context, tenantID string, runtimeID string, pageSize *int, cursor *string) (*model.ApplicationPage, error) {
```
crabtree: I would go a different way. This method should return applications by filters. The filters should be built on the service or even upper level. IMO This repository should not have the notion of runtimes. @aszecowka what do you think?

components/director/internal/domain/application/repository.go
```diff
@@ -57,12 +67,71 @@ func (r *inMemoryRepository) List(ctx context.Context, tenant string, filter []*
...
+	persist, err := persistence.FromCtx(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "while fetching DB from context")
+	}
+
+	//TODO: change tenantID from UUID to String when
+	tenantUUID, err := uuid.Parse(tenantID)
+	if err != nil {
+		return nil, errors.New("tenant_ID is not parseable")
```
crabtree: Why you do not decorate the original error?

components/director/internal/domain/application/repository.go
```diff
@@ -57,12 +67,71 @@ func (r *inMemoryRepository) List(ctx context.Context, tenant string, filter []*
...
+	}
+
+	stmt := fmt.Sprintf(`SELECT VALUE FROM %s WHERE TENANT_ID=$1 AND RUNTIME_ID=$2 AND KEY='SCENARIOS'`, labelTableName)
+
+	var scenariosJSON string
+	err = persist.Get(&scenariosJSON, stmt, tenantID, runtimeID)
```
crabtree: I agree with @aszecowka this select should not be done here. Nevertheless, every error should be checked.

components/director/internal/domain/application/repository.go
dbadura: Introduced new method name `ListByScenarios`.

components/director/internal/domain/application/service.go
```diff
@@ -93,13 +97,39 @@ func (s *service) List(ctx context.Context, filter []*labelfilter.LabelFilter, p
...
 		return nil, errors.Wrapf(err, "while loading tenant from context")
 	}
 
-	return s.appRepo.ListByRuntimeID(ctx, tenantID, runtimeID, pageSize, cursor)
+	tenantUUID, err := uuid.Parse(tenantID)
+	if err != nil {
+		return nil, errors.New("tenantID is not UUID")
+	}
+
+	label, err := s.labelRepo.GetByKey(ctx, tenantID, model.RuntimeLabelableObject, runtimeID.String(), model.ScenariosKey)
```
aszecowka: you changed runtimeID from string to UUID, but in repo you have string and again you write: `runtimeID.String()`

components/director/internal/domain/application/service_test.go
```diff
@@ -1311,60 +1311,113 @@ func TestService_DeleteLabel(t *testing.T) {
...
+				appRepository.On("ListByScenarios", ctx, tenantUUID, convertToStringArray(t, scenarios), &first, &cursor).
+					Return(nil, testError).Once()
 				return appRepository
 			},
 			ExpectedError:  testError,
 			ExpectedResult: nil,
-			Input:          runtimeID,
+		},
+		{
+			Name:  " Return empty page when scenarios for runtime is empty array",
```
aszecowka: when runtime is not assigned to any scenario

tests/end-to-end/director/runtime_api_test.go
```diff
@@ -453,6 +450,103 @@ func TestQuerySpecificRuntime(t *testing.T) {
...
+			Tenant:          tenantID,
+			ApplicationName: "noneofscenarios",
+			WithinTenant:    false,
+			Scenarios:       []string{},
+		},
+		{
+			Tenant:          tenantID,
+			ApplicationName: "first",
+			WithinTenant:    true,
+			Scenarios:       scenarios[:1],
```
aszecowka: I have to think when I see such line, simplify it to not force me to think too much

components/director/internal/domain/application/repository.go
```diff
@@ -57,12 +64,55 @@ func (r *inMemoryRepository) List(ctx context.Context, tenant string, filter []*
...
+	if err != nil {
+		return nil, errors.Wrap(err, "while creating filter query")
+	}
+
+	var apps []interface{}
+
+	err = persist.Select(&apps, stmt)
+
+	if err != nil {
+		if err == sql.ErrNoRows {
```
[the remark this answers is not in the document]
dbadura: So in case of `Get`, we should check `ErrNoRows`
In case of `Select` we should only check err.

tgorgol:
If there's no runtime with specified runtimeID resolver returns error: `while getting all Application for Runtime: while getting scenarios for runtime: label 'scenarios' not found` instead of something telling us that runtime doesn't exist.

components/director/internal/domain/application/repository_test.go
```diff
@@ -25,15 +22,15 @@ func TestPgRepository_ListByRuntimeScenarios(t *testing.T) {
...
-	//Create Filters for scenarios, because we cannot mock  filter query generator
-	var scenarioFilter []*labelfilter.LabelFilter
-	for _, scenario := range runtimeScenarios {
-		query := fmt.Sprintf(`$[*] ? (@ == "%s")`, scenario)
-		scenarioFilter = append(scenarioFilter, &labelfilter.LabelFilter{Key: model.ScenariosKey, Query: &query})
-	}
-
-	scenarioQuery, err := label.FilterQuery(model.ApplicationLabelableObject, label.UnionSet, tenantID, scenarioFilter)
-	require.NoError(t, err)
+	scenarioQuery := fmt.Sprintf(`SELECT "app_id" FROM "public"."labels"
```
aszecowka: tenor-103228046

components/director/internal/domain/application/service.go
```diff
@@ -109,6 +123,15 @@ func (s *service) ListByRuntimeID(ctx context.Context, runtimeID uuid.UUID, page
 		return nil, errors.New("tenantID is not UUID")
 	}
 
+	bool, err := s.runtimeRepo.Exists(ctx, tenantID, runtimeID.String())
+	if err != nil {
+		return nil, errors.Wrap(err, "while checking if runtime exits")
+	}
+
+	if !bool {
```
aszecowka: consider renaming this var

PR #208, opened 2019-08-13
Introduce Generic Repository for the following operations: Exist, Get, Delete, pageable select

aszecowka:
- Introduced the following helper classes: ExistQuerier, SingleGetter, repo.Deleter, repo.PageableQuerier
- runtime.Repository is using new helper classes
- Created new package testdb to simplify mocking databases


components/director/internal/repo/exist.go
```diff
@@ -0,0 +1,38 @@
...
+	return &ExistQuerier{tableName: tableName, idColumn: idColumn, tenantColumn: tenantColumn}
+}
+
+func (g *ExistQuerier) Exists(ctx context.Context, tenant, id string) (bool, error) {
+	persist, err := persistence.FromCtx(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	q := fmt.Sprintf("SELECT 1 FROM %s WHERE %s=$1 AND %s=$2", g.tableName, g.tenantColumn, g.idColumn)
```
pkosiec: what if there are different number of conditions than two? I'm concerned that this approach will be too restrictive. I'm still a fan of variadic parameter approach for parameters.

components/director/internal/repo/delete.go
```diff
@@ -0,0 +1,41 @@
...
+	return &Deleter{tableName: tableName, idColumn: idColumn, tenantColumn: tenantColumn}
+}
+
+func (g *Deleter) Delete(ctx context.Context, tenant string, id string) error {
+	persist, err := persistence.FromCtx(ctx)
+	if err != nil {
+		return err
+	}
+
+	q := fmt.Sprintf("DELETE FROM %s WHERE %s=$1 AND %s=$2", g.tableName, g.tenantColumn, g.idColumn)
```
pkosiec: In `labels` domain we have 3 parameters needed for delete. Please use variadic parameter, to allow using generic repository in more cases. Please think about variadic parameters or `map[string]string` of required conditions.

components/director/internal/repo/get.go
```diff
@@ -0,0 +1,46 @@
...
+func (g *SingleGetter) Get(ctx context.Context, tenant, id string, dest interface{}) error {
+	if dest == nil {
+		return errors.New("missing destination")
+	}
+	persist, err := persistence.FromCtx(ctx)
+	if err != nil {
+		return err
+	}
+
+	q := fmt.Sprintf("SELECT %s FROM %s WHERE %s = $1 AND  %s = $2", g.selectedColumns, g.tableName, g.tenantColumn, g.idColumn)
```
pkosiec: Again, in labels we have 3 parameters required. It makes the generic repository useless. I would like to see more flexible solution.

components/director/internal/repo/list.go
```diff
@@ -0,0 +1,90 @@
...
+			filterSubquery += fmt.Sprintf(` AND %s`, cond)
+		}
+	}
+
+	paginationSQL, err := pagination.ConvertOffsetLimitAndOrderedColumnToSQL(pageSize, offset, orderByColumn)
+	if err != nil {
+		return nil, -1, errors.Wrap(err, "while converting offset and limit to cursor")
+	}
+
+	stmtWithoutPagination := fmt.Sprintf("SELECT %s FROM %s WHERE %s=$1 %s", g.selectedColumns, g.tableName, g.tenantColumn, filterSubquery)
```
pkosiec: Allow more than 1 condition for listing. For example: `ListForObject` in labels

components/director/internal/domain/runtime/repository.go
```diff
@@ -2,148 +2,94 @@ package runtime
...
-			StartCursor: cursor,
-			EndCursor:   endCursor,
-			HasNextPage: hasNextPage,
-		},
-	}, nil
+		PageInfo:   page}, nil
+
 }
 
 func (r *pgRepository) Create(ctx context.Context, item *model.Runtime) error {
```
pkosiec: What about creating entities? I think we could also improve this part.

components/director/internal/domain/runtime/repository_test.go
```diff
@@ -32,13 +32,14 @@ func TestPgRepository_GetByID_ShouldReturnRuntimeModelForRuntimeEntity(t *testin
 	timestamp, err := time.Parse(time.RFC3339, "2002-10-02T10:00:00-05:00")
 	require.NoError(t, err)
 
-	sqlxDB, sqlMock := mockDatabase(t)
+	sqlxDB, sqlMock := testdb.MockDatabase(t)
```
pkosiec: How the tests should look like for the repository which uses generic repository?

To be honest, if we still need to test all the cases, the benefits of the generic repo are not that significant.

I would like to see a generic test suite for repo that utilizes generic repo. Because writing test cases actually takes most of the time while writing repository.

Consider something like this:

```go
func TestRepository_Exists(t *testing.T) {
labelDefRepo := labeldef.NewRepository(conv)
 ts := genericrepo.NewTestSuite(t, labelDefRepo)
expectedQuery := "select 1 as exists from public.label_definitions where tenant_id=$1 and key=$2"

ts.TestExists(func(ctx context.Context) bool, error { 
    return labelDefRepo.Exists(ctx, "anything", "anything")
}, expectedQuery, "anything", "anything") // it will do all cases for exists
}
```

which would test all cases for the Exists method.

Similar approach I would see for Delete:

```go
func TestRepository_Delete(t *testing.T) {
key := ...
objID := ...
objType := ...
tnt := ...

labelRepo := label.NewRepository(conv)
 ts := genericrepo.NewTestSuite(t, labelRepo)
expectedQuery := "DELETE FROM "public"."labels" WHERE "key" = $1 AND "runtime_id" = $2 AND "tenant_id" = $3"

ts.TestDelete(func(ctx context.Context) error { 
    return repo.Delete(ctx, tnt, objType, objID, key)
}, expectedQuery, tnt, objType, objID, key) // it will do all cases for delete
}
```

I believe we can do it for Get and List operations as well. We just need to specify proper test suite method prototype. For example, we could pass function which constructs an repo object for test purposes (to be able to specify different mock converters).

components/director/internal/repo/exist.go
```diff
@@ -0,0 +1,51 @@
...
+	tableName    string
+	tenantColumn string
+}
+
+func NewExistQuerier(tableName, tenantColumn string) *ExistQuerier {
+	return &ExistQuerier{tableName: tableName, tenantColumn: tenantColumn}
+}
+
+type Conditions []IDAndValue
+type IDAndValue struct {
```
pkosiec: Please rename to `Condition`

components/director/internal/domain/runtime/repository.go
```diff
@@ -2,148 +2,98 @@ package runtime
...
-	err = persist.Get(&count, stmt, id, tenant)
-	if err != nil {
-		if err == sql.ErrNoRows {
-			return false, nil
-		}
-		return false, errors.Wrap(err, "while getting runtime from DB")
-	}
-
-	return true, nil
+	return r.existQuerier.Exists(ctx, tenant, repo.Conditions{{ID: "id", Val: id}})
```
pkosiec: This approach looks really good
tenor-75024046

components/director/internal/repo/exist_test.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func TestExist(t *testing.T) {
+	givenID := uuidA()
+	givenTenant := uuidB()
+	givenQuery := "SELECT 1 FROM users WHERE tenant_col = $1 AND id_col = $2"
+	escapedQuery := regexp.QuoteMeta(givenQuery)
+	sut := repo.NewExistQuerier("users", "tenant_col")
+
+	t.Run("when exist", func(t *testing.T) {
```
dbadura: What do you think about this test name:`success when exist`. It would be compliant with the rest of our tests.

components/director/internal/domain/labeldef/repository.go
```diff
@@ -34,10 +34,7 @@ func (r *repo) Create(ctx context.Context, def model.LabelDefinition) error {
 		return errors.Wrap(err, "while converting Label Definition to insert")
 	}
 
-	columns := []string{"id", "tenant_id", "key"}
-	if entity.SchemaJSON.Valid {
-		columns = append(columns, "schema")
-	}
+	columns := []string{"id", "tenant_id", "key", "schema"}
```
dbadura: What about moving it to the const `labelDefinitionFields` like is done in runtime `repository`

components/director/internal/repo/delete_test.go
```diff
@@ -0,0 +1,89 @@
...
+		require.NoError(t, err)
+	})
+
+	t.Run("returns error on db operation", func(t *testing.T) {
+		// GIVEN
+		givenErr := errors.New("some err")
+		db, mock := testdb.MockDatabase(t)
+		ctx := persistence.SaveToContext(context.TODO(), db)
+		defer mock.AssertExpectations(t)
+		mock.ExpectExec(expectedQuery).WithArgs(givenTenant, givenID).WillReturnError(givenErr)
```
dbadura: You can inline this error here.

components/director/internal/repo/create_test.go
```diff
@@ -0,0 +1,82 @@
...
+		// GIVEN
+		db, mock := testdb.MockDatabase(t)
+		ctx := persistence.SaveToContext(context.TODO(), db)
+		defer mock.AssertExpectations(t)
+		givenUser := User{}
+
+		mock.ExpectExec(regexp.QuoteMeta("INSERT INTO users ( id_col, tenant_col, first_name, last_name, age ) VALUES ( ?, ?, ?, ?, ? )")).
+			WillReturnError(errors.New("some error"))
+		// WHEN
+		err := sut.Create(ctx, givenUser)
```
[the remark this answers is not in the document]
aszecowka: https://en.wikipedia.org/wiki/System_under_test

components/director/internal/repo/common.go
```diff
@@ -0,0 +1,25 @@
...
+	for _, idAndVal := range conditions {
+		allArgs = append(allArgs, idAndVal.Val)
+	}
+	return allArgs
+}
+
+func appendEnumeratedConditions(query string, startIdx int, conditions Conditions) string {
+	var out string
+	for idx, idAndVal := range conditions {
+		out = fmt.Sprintf("%s AND %s = $%d", query, idAndVal.Field, idx+startIdx)
```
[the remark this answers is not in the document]
aszecowka: good catch!

components/director/internal/repo/delete_test.go
aszecowka: now I use `someError()`

PR #211, opened 2019-08-13
Document that describes DB scheme

aszecowka:


docs/architecture/entity-relations.md
```diff
@@ -0,0 +1,5 @@
+# Entity Relations
+
+Diagram below describes missing entity relations for Director.
```
pkosiec: If we add `Entity relations` document, then it would be good to document all entity relations, not only missing ones.
Fortunately, there are no that much relations.

docs/architecture/entity-relations.md
```diff
@@ -0,0 +1,5 @@
...
+
+![](./assets/entity_diagram.svg)
```
pkosiec: Suggestion: Add field types to the diagram

docs/architecture/db-scheme.md
```diff
@@ -0,0 +1,8 @@
+# Database Scheme
+
+Diagram below represents database scheme. 
+Fields marked with the red color are planned for upgrade:
+- `runtimes.auth` field will be changed to `JSONB`
```
pkosiec: When will we update the schema regarding the runtimes? Please create an issue with this change

Issue #126, opened 2019-07-09
Handle cascading delete for labeling


pkosiec:
**AC**

- Handle cascading delete for LabelDefinition when `force` boolean parameter is specified
- Remember about integration tests and examples

PR #220, opened 2019-08-16
Implement cascade delete for labeling

kfurgol:
**Description**

- Labels which have the same key are now being deleted when Label Definition delete request has `deleteRelatedLabels` value set to `true`

**Related issue(s)**
#126

### DIFF
--- a/components/director/internal/domain/label/repository.go
+++ b/components/director/internal/domain/label/repository.go
@@ -177,18 +177,9 @@ func (r *repository) DeleteByKey(ctx context.Context, tenant string, key string)
 	}
 
 	stmt := fmt.Sprintf(`DELETE FROM %s WHERE "key" = $1 AND "tenant_id" = $2`, tableName)
-	result, err := persist.Exec(stmt, key, tenant)
+	_, err = persist.Exec(stmt, key, tenant)
 	if err != nil {
 		return errors.Wrapf(err, `while deleting all Label entities from database with key "%s"`, key)
-	}
-
-	rows, err := result.RowsAffected()
-	if err != nil {
-		return errors.Wrap(err, "while receiving rows affected by query")
-	}
-
-	if rows == 0 {
-		return errors.New("no rows were affected by query")
 	}
 
 	return nil
--- a/components/director/internal/domain/label/repository.go
+++ b/components/director/internal/domain/label/repository.go
@@ -170,6 +170,30 @@ func (r *repository) DeleteAll(ctx context.Context, tenant string, objectType mo
 	return errors.Wrapf(err, "while deleting all Label entities from database for %s %s", objectType, objectID)
 }
 
+func (r *repository) DeleteByKey(ctx context.Context, tenant string, key string) error {
+	persist, err := persistence.FromCtx(ctx)
+	if err != nil {
+		return errors.Wrap(err, "while fetching persistence from context")
+	}
+
+	stmt := fmt.Sprintf(`DELETE FROM %s WHERE "key" = $1 AND "tenant_id" = $2`, tableName)
+	result, err := persist.Exec(stmt, key, tenant)
+	if err != nil {
+		return errors.Wrapf(err, `while deleting all Label entities from database with key "%s"`, key)
+	}
+
+	rows, err := result.RowsAffected()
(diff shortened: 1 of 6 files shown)


components/director/internal/domain/labeldef/service.go
```diff
@@ -107,16 +107,25 @@ func (s *service) Update(ctx context.Context, def model.LabelDefinition) error {
...
 func (s *service) Delete(ctx context.Context, tenant, key string, deleteRelatedLabels bool) error {
-	if deleteRelatedLabels == true {
-		return errors.New("deleting related labels is not yet supported")
-	}
-
 	if key == model.ScenariosKey {
 		return fmt.Errorf("Label Definition with key %s can not be deleted", model.ScenariosKey)
 	}
 
+	if deleteRelatedLabels == true {
```
pkosiec: `if deleteRelatedLabels {`

components/director/internal/domain/labeldef/service.go
```diff
@@ -107,16 +107,25 @@ func (s *service) Update(ctx context.Context, def model.LabelDefinition) error {
...
+		labels, err := s.labelRepo.ListByKey(ctx, tenant, key)
+		if err != nil {
+			return errors.Wrapf(err, "while receiving existing labels with key \"%s\"", key)
+		}
+
+		for _, label := range labels {
+			err = s.labelRepo.Delete(ctx, tenant, label.ObjectType, label.ObjectID, label.Key)
```
pkosiec: I think it would be good to prepare a new repository method for `DeleteByKey` and use it here. Then you won't need to list all labels by key.

components/director/internal/domain/labeldef/service.go
kfurgol: done

components/director/internal/domain/labeldef/service.go
kfurgol: done

components/director/internal/domain/label/repository_test.go
```diff
@@ -821,6 +821,104 @@ func TestRepository_DeleteAll(t *testing.T) {
 		require.Contains(t, err.Error(), "unable to fetch database from context")
 	})
 }
+func TestRepository_DeleteByKey(t *testing.T) {
+
```
pkosiec: Empty line

components/director/internal/domain/labeldef/service.go
```diff
@@ -107,16 +108,18 @@ func (s *service) Update(ctx context.Context, def model.LabelDefinition) error {
...
 func (s *service) Delete(ctx context.Context, tenant, key string, deleteRelatedLabels bool) error {
-	if deleteRelatedLabels == true {
-		return errors.New("deleting related labels is not yet supported")
-	}
-
 	if key == model.ScenariosKey {
 		return fmt.Errorf("Label Definition with key %s can not be deleted", model.ScenariosKey)
 	}
 
+	if deleteRelatedLabels {
```
pkosiec: This if should be moved to the bottom - we need to check if LabelDefinition exists, right?

components/director/internal/domain/label/repository_test.go
kfurgol: removed

components/director/internal/domain/labeldef/service.go
kfurgol: done

components/director/internal/domain/label/repository.go
```diff
@@ -170,6 +170,30 @@ func (r *repository) DeleteAll(ctx context.Context, tenant string, objectType mo
...
+	if err != nil {
+		return errors.Wrapf(err, `while deleting all Label entities from database with key "%s"`, key)
+	}
+
+	rows, err := result.RowsAffected()
+	if err != nil {
+		return errors.Wrap(err, "while receiving rows affected by query")
+	}
+
+	if rows == 0 {
```
tgorgol: What if someone tries deleting LD with "force flag" and there's no labels with that key? Should that result in error?

kfurgol: changed it, it won't result with an error

PR #230, opened 2019-08-21
Extend Deleter: DeleteOne and DeleteMany

aszecowka:
**Description**
Extend Deleter struct: DeleteOne and DeleteMany

### DIFF
--- a/components/director/internal/repo/delete.go
+++ b/components/director/internal/repo/delete.go
@@ -17,7 +17,15 @@ func NewDeleter(tableName, tenantColumn string) *Deleter {
 	return &Deleter{tableName: tableName, tenantColumn: tenantColumn}
 }
 
-func (g *Deleter) Delete(ctx context.Context, tenant string, conditions Conditions) error {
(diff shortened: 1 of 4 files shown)


components/director/internal/repo/delete.go
```diff
@@ -30,12 +38,14 @@ func (g *Deleter) Delete(ctx context.Context, tenant string, conditions Conditio
 	if err != nil {
 		return errors.Wrap(err, "while deleting from database")
 	}
-	affected, err := res.RowsAffected()
-	if err != nil {
-		return errors.Wrap(err, "while checking affected rows")
-	}
-	if affected != 1 {
-		return fmt.Errorf("delete should remove single row, but removed %d rows", affected)
+	if requireSingleRemoval {
```
pkosiec: Invert the condition
```
if !requireSingleRemoval {
   return nil
}
```

aszecowka: discussed already, the current version seems to be more readable.

PR #231, opened 2019-08-21
Add list helper

tgorgol:
**Description**

- Add repo list helper (without pagination)
- Rename existing `list.go` to `list_pageable.go`
- Add repo upsert helper

**Related issue(s)**

kyma-project/kyma#5149


components/director/internal/repo/upsert.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type Upserter struct {
+	tableName       string
+	columns         []string
+	conflictColumns []string
+	updateColumns   []string
+}
+
+func NewUpserter(tableName string, columns []string, conflictColumns []string, updateColumns []string) *Upserter {
```
aszecowka: think about better names, my proposal is:
- createColumns
- updateColumns
- identifyingColumns or idColumns

components/director/internal/repo/upsert.go
```diff
@@ -0,0 +1,61 @@
...
+	var update []string
+	for _, c := range u.updateColumns {
+		update = append(update, fmt.Sprintf("%[1]s=EXCLUDED.%[1]s", c))
+	}
+
+	stmtWithoutUpsert := fmt.Sprintf("INSERT INTO %s ( %s ) VALUES ( %s )", u.tableName, strings.Join(u.columns, ", "), strings.Join(values, ", "))
+	stmtWithUpsert := fmt.Sprintf("%s ON CONFLICT ( %s ) DO UPDATE SET %s", stmtWithoutUpsert, strings.Join(u.conflictColumns, ", "), strings.Join(update, ", "))
+
+	_, err = persist.NamedExec(stmtWithUpsert, dbEntity)
+	if pqerr, ok := err.(*pq.Error); ok {
```
aszecowka: what if we modified more than one row?

components/director/internal/repo/list.go
```diff
@@ -5,39 +5,29 @@ import (
...
 
-	"github.com/kyma-incubator/compass/components/director/internal/persistence"
-	"github.com/kyma-incubator/compass/components/director/pkg/pagination"
 	"github.com/pkg/errors"
+
+	"github.com/kyma-incubator/compass/components/director/internal/persistence"
 )
 
-type PageableQuerier struct {
+type Lister struct {
```
aszecowka: now it looks that you modified PageableQuerier, but you just renamed a file. Can you check if some git magic tricks exist to correctly show your changes?

tgorgol: I divided this PR into a few commits for easier reviewing

components/director/internal/repo/upsert.go
tgorgol: How about:
- createColumns
- conflictingColumns
- updateColumns

components/director/internal/repo/list.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	filterSubquery := ""
+	for _, cond := range additionalConditions {
+		if strings.TrimSpace(cond) != "" {
+			filterSubquery += fmt.Sprintf(` AND %s`, cond)
+		}
+	}
+
+	stmt := fmt.Sprintf("SELECT %s FROM %s WHERE %s=$1 %s", l.selectedColumns, l.tableName, l.tenantColumn, filterSubquery)
```
dbadura: Can we share the way in which we creating this query, because we do in the same way in `list with pagination`.
For example: Extract to method, like we have in `common` `appendEnumeratedConditions` method

components/director/internal/repo/list_test.go
```diff
@@ -0,0 +1,79 @@
...
+			AddRow(peterRow...).
+			AddRow(homerRow...)
+		mock.ExpectQuery(regexp.QuoteMeta(`SELECT id_col, tenant_col, first_name, last_name, age FROM users WHERE tenant_col=$1`)).WithArgs(givenTenant).WillReturnRows(rows)
+		ctx := persistence.SaveToContext(context.TODO(), db)
+		var dest UserCollection
+
+		err := sut.List(ctx, givenTenant, &dest)
+		require.NoError(t, err)
+		assert.Len(t, dest, 2)
+		assert.Equal(t, peter, dest[0])
```
dbadura: Is the order always the same?

components/director/internal/repo/upsert_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/kyma-incubator/compass/components/director/internal/persistence"
+	"github.com/kyma-incubator/compass/components/director/internal/repo"
+	"github.com/kyma-incubator/compass/components/director/internal/repo/testdb"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUpsert(t *testing.T) {
+	expectedQuery := regexp.QuoteMeta("INSERT INTO users ( id_col, tenant_col, first_name, last_name, age ) VALUES ( ?, ?, ?, ?, ? ) ON CONFLICT ( tenant_col, first_name, last_name ) DO UPDATE SET age=EXCLUDED.age")
```
dbadura: Can you break this sql?

components/director/internal/repo/upsert.go
```diff
@@ -0,0 +1,61 @@
...
+	tableName          string
+	createColumns      []string
+	conflictingColumns []string
+	updateColumns      []string
+}
+
+func NewUpserter(tableName string, createColumns []string, conflictingColumns []string, updateColumns []string) *Upserter {
+	return &Upserter{
+		tableName:          tableName,
+		createColumns:      createColumns,
```
dbadura: Why not just`entityColumns` or `columns`?

tgorgol: Just `columns` wouldn't tell much about them when there are 3 "columns" in arguments in total. Not sure about `entityColumns`, why do you think that name would fit better?

components/director/internal/repo/list_test.go
tgorgol: With mocked DB - yes, but I'll change those asserts to `Contains` to make the example better.

PR #233, opened 2019-08-21
Migrate Webhooks Repository from In-Memory to Postgres

aszecowka:


components/director/internal/domain/application/resolver_test.go
```diff
@@ -10,18 +10,13 @@ import (
...
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 )
 
-var contextParam = mock.MatchedBy(func(ctx context.Context) bool {
-	persistenceOp, err := persistence.FromCtx(ctx)
-	return err == nil && persistenceOp != nil
-})
+var contextParam = persistenceautomock.CtxWithDBMatcher()
```
pkosiec: I would move it out from `automock` package, as it is not common to put something not generated.

What about `persistence/test` package or something like that?

components/director/internal/domain/application/service_test.go
```diff
@@ -81,7 +81,7 @@ func TestService_Create(t *testing.T) {
 			},
 			WebhookRepoFn: func() *automock.WebhookRepository {
 				repo := &automock.WebhookRepository{}
-				repo.On("CreateMany", mock.Anything).Return(nil).Once()
+				repo.On("CreateMany", mock.Anything, mock.Anything).Return(nil).Once()
```
pkosiec: replace first `mock.Anything` with `ctx` at least

components/director/internal/domain/webhook/converter.go
```diff
@@ -70,3 +74,53 @@ func (c *converter) MultipleInputFromGraphQL(in []*graphql.WebhookInput) []*mode
 
 	return inputs
 }
+
+func (c *converter) ToEntity(in model.Webhook) (Entity, error) {
+	var optionalAuth sql.NullString
+	if in.Auth != nil {
```
pkosiec: split it to separate method

components/director/internal/domain/webhook/converter.go
```diff
@@ -70,3 +74,53 @@ func (c *converter) MultipleInputFromGraphQL(in []*graphql.WebhookInput) []*mode
...
+		TenantID: in.Tenant,
+		URL:      in.URL,
+		AppID:    in.ApplicationID,
+		Auth:     optionalAuth,
+	}, nil
+}
+
+func (c *converter) FromEntity(in Entity) (model.Webhook, error) {
+	var auth *model.Auth
+	if in.Auth.Valid {
```
pkosiec: split it to separate method

components/director/internal/persistence/automock/extended.go
```diff
@@ -0,0 +1,32 @@
+package automock
```
pkosiec: Move it out from `automock` package. It won't be obvious to have it here

components/director/run.sh
```diff
@@ -59,4 +74,13 @@ echo -e "${GREEN}Populate DB${NC}"
...
\ No newline at end of file
+if [[  ${SKIP_APP_START} ]]; then
+        echo -e "${GREEN}Skipping starting application${NC}"
+        while true
+        do
+            sleep 1
+        done
+fi
+
+echo -e "${GREEN}Starting aapplication${NC}"
```
pkosiec: aaaaaplication

components/director/internal/model/webhook.go
```diff
@@ -1,13 +1,20 @@
...
 type Webhook struct {
 	ApplicationID string
+	Tenant        string
 	ID            string
 	Type          WebhookType
 	URL           string
 	Auth          *Auth
 }
 
+func (w Webhook) PrettyString() string {
```
pkosiec: Good idea, but I don't like this name. We need to highlight that this is used for our internal errors.

I'm also thinking if this is a good time to introduce something like that, because it would be good to discuss, if we need that at all.

components/director/internal/domain/webhook/repository_test.go
```diff
@@ -0,0 +1,522 @@
...
+
+		given := []*model.Webhook{{ID: "one"}, {ID: "two"}, {ID: "three"}}
+		mockConverter.On("ToEntity", *given[0]).Return(webhook.Entity{ID: "one"}, nil)
+		mockConverter.On("ToEntity", *given[1]).Return(webhook.Entity{ID: "two"}, nil)
+		mockConverter.On("ToEntity", *given[2]).Return(webhook.Entity{ID: "three"}, nil)
+
+		db, dbMock := testdb.MockDatabase(t)
+		defer dbMock.AssertExpectations(t)
+
+		dbMock.ExpectExec(regexp.QuoteMeta("INSERT INTO public.webhooks ( id, tenant_id, app_id, type, url, auth ) VALUES ( ?, ?, ?, ?, ?, ? )")).WithArgs(
```
pkosiec: please move the query to const

components/director/internal/domain/webhook/repository_test.go
```diff
@@ -0,0 +1,522 @@
...
+			"three", "", "", "", "", nil).WillReturnResult(sqlmock.NewResult(-1, 1))
+
+		ctx := persistence.SaveToContext(context.TODO(), db)
+		sut := webhook.NewRepository(mockConverter)
+		// WHEN
+		err := sut.CreateMany(ctx, given)
+		// THEN
+		require.NoError(t, err)
+	})
+
```
pkosiec: add test case for DB error

components/director/internal/domain/application/resolver_test.go
aszecowka: pkg `txtest` created

components/director/internal/model/webhook.go
aszecowka: removed

components/director/internal/domain/webhook/repository_test.go
```diff
@@ -235,7 +236,28 @@ func TestRepositoryCreateMany(t *testing.T) {
...
+		mockConverter := &automock.EntityConverter{}
+		defer mockConverter.AssertExpectations(t)
+
+		given := []*model.Webhook{{ID: "one", URL: "unlucky", Type: model.WebhookTypeConfigurationChanged}, {ID: "two"}, {ID: "three"}}
+		mockConverter.On("ToEntity", *given[0]).Return(webhook.Entity{ID: "one"}, nil)
+
+		db, dbMock := testdb.MockDatabase(t)
+		defer dbMock.AssertExpectations(t)
+
+		dbMock.ExpectExec(regexp.QuoteMeta(expectedInsert)).WillReturnError(givenError())
```
pkosiec: It would be better to check what happens, when first item is added properly, and inserting second fails

components/director/internal/domain/application/service_test.go
pkosiec: It is still not changed

components/director/internal/domain/webhook/converter.go
```diff
@@ -70,3 +74,70 @@ func (c *converter) MultipleInputFromGraphQL(in []*graphql.WebhookInput) []*mode
...
+		TenantID: in.Tenant,
+		URL:      in.URL,
+		AppID:    in.ApplicationID,
+		Auth:     optionalAuth,
+	}, nil
+}
+
+func (c *converter) toAuthEntity(in model.Webhook) (sql.NullString, error) {
+	var optionalAuth sql.NullString
+	if in.Auth != nil {
```
pkosiec: Invert the condition

components/director/internal/domain/webhook/converter.go
```diff
@@ -70,3 +74,70 @@ func (c *converter) MultipleInputFromGraphQL(in []*graphql.WebhookInput) []*mode
...
+		Tenant:        in.TenantID,
+		URL:           in.URL,
+		ApplicationID: in.AppID,
+		Auth:          auth,
+	}, nil
+}
+
+func (c *converter) fromEntityAuth(in Entity) (*model.Auth, error) {
+	var auth *model.Auth
+	if in.Auth.Valid {
```
pkosiec: Invert the condition

components/director/internal/domain/webhook/repository_test.go
aszecowka: discussed, it does not introduce any value

PR #240, opened 2019-08-21
Run compass components as non-root user

mwieczorek:
**Description**

- This change enables running compass components in more restrictive k8s environments (with default PSP which enforce running containers as a user other than root). The whole securityContext can be set directly in values files.

**Related issue(s)**

### DIFF
--- a/chart/compass/charts/director/templates/deployment.yaml
+++ b/chart/compass/charts/director/templates/deployment.yaml
@@ -25,6 +25,10 @@ spec:
             - name: http
               containerPort: {{ .Values.deployment.args.containerPort }}
               protocol: TCP
+          {{- with .Values.deployment.securityContext }}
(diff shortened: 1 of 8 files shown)


chart/compass/charts/director/templates/deployment.yaml
```diff
@@ -66,9 +66,6 @@ spec:
           command: ["/cloud_sql_proxy",
                     "-instances={{ .Values.global.database.managedGCP.instanceConnectionName }}=tcp:5432",
                     "-credential_file=/secrets/cloudsql-instance-credentials/credentials.json"]
-          securityContext:
-            runAsUser: 2  # non-root user
-            allowPrivilegeEscalation: false
```
tgorgol: Do we need this set to true?

Issue #223, opened 2019-08-19
Implement APIDefinition and EventAPI repositories


pkosiec:
**Description**
- Implement APIDefinition repository
- Implement EventAPIDefinition repository

Check if proper integration tests exist.

PR #269, opened 2019-08-29
add repository for event api def

dbadura:
**Description**

- Add event api repository

**Related issue(s)**

see #223


components/director/internal/domain/eventapi/repository.go
```diff
@@ -128,3 +130,147 @@ func (r *inMemoryRepository) DeleteAllByApplicationID(id string) error {
...
+}
+
+type EventAPIDefCollection []Entity
+
+func (r EventAPIDefCollection) Len() int {
+	return len(r)
+}
+
+func (r *pgRepository) ListByApplicationID(ctx context.Context, tenantID string, applicationID string, pageSize int, cursor string) (*model.EventAPIDefinitionPage, error) {
+	appCond := fmt.Sprintf("%s = '%s'", "app_id", applicationID)
```
aszecowka: use `QuoteLiteral`:
https://github.com/kyma-incubator/compass/blob/master/components/director/internal/domain/webhook/repository.go#L61

PR #294, opened 2019-09-06
Switch EventAPI Definition from in memory database to Postgres

kfurgol:
**Description**

- Switched EventAPI definition implementation from in memory db to Postgres
- Adjusted tests

**Related issue(s)**
#223


components/director/internal/domain/application/resolver_test.go
```diff
@@ -1288,25 +1290,28 @@ func TestResolver_EventAPIs(t *testing.T) {
...
 	}
 
+	txGen := txtest.NewTransactionContextGenerator(testErr)
+
 	first := 2
 	gqlAfter := graphql.PageCursor("test")
 	after := "test"
-	testErr := errors.New("Test error")
 
 	testCases := []struct {
```
tgorgol: Missing cases for transaction fails.

tgorgol:
Things that should be done before finalizing moving to sql repo:

- Uncomment foreign keys in migrations [here](https://github.com/kyma-incubator/compass/blob/2d99afde9a0f1e7383faf2380a3c49a14e07abb1/components/schema-migrator/migrations/201908141300_initialize_schema.up.sql#L119) and [here](https://github.com/kyma-incubator/compass/blob/2d99afde9a0f1e7383faf2380a3c49a14e07abb1/components/schema-migrator/migrations/201908141300_initialize_schema.up.sql#L225)
- Modify [application delete service function](https://github.com/kyma-incubator/compass/blob/2d99afde9a0f1e7383faf2380a3c49a14e07abb1/components/director/internal/domain/application/service.go#L314):
  - Remove logic related to removing related resource as we now have cascading deletes on database
  - Maybe: remove Get before delete, do we need it?

Issue #258, opened 2019-08-27
Cannot pass valid JSON schema through GraphQL API


pkosiec:
**Description**

For now we specify JSON Schema in GraphQL as objects, but GraphQL doesn't support special characters like "$", which is valid symbol in JSON Schema keys. 

We should pass JSON Schema as a string.

**Expected result**

JSON Schema can be provided properly using Director API

**Steps to reproduce**

Try to pass the following schema while creating LabelDefinition:

```json
 {
 "definitions": {},
 "$schema": "http://json-schema.org/draft-07/schema#",
 "$id": "http://example.com/root.json",
 "type": "object",
 "title": "The Root Schema",
 "required": [
   "checked",
   "dimensions"
  ],
  "properties": {
    "checked": {
      "$id": "#/properties/checked",
      "type": "boolean",
      "title": "The Checked Schema",
      "default": false,
      "examples": [
        false
      ]
    },
    "dimensions": {
      "$id": "#/properties/dimensions",
      "type": "object",
      "title": "The Dimensions Schema",
      "required": [
        "width"
      ],
      "properties": {
        "width": {
          "$id": "#/properties/dimensions/properties/width",
          "type": "integer",
          "title": "The Width Schema",
          "default": 0,
          "examples": [
            5
          ]
        }
      }
    }
  }
}
```

All fields with `$` prefix are forbidden.

Issue #257, opened 2019-08-27
Use middleware to begin a transaction


pkosiec:
**Description**

For now we create, commit and rollback transactions in Resolvers. We could create a middleware before GraphQL server, open a transaction, put it in context, and have another middleware after GraphQL server to commit/rollback it.

Complexity: medium

**Reasons**

We need to remember to open and commit/rollback a transaction in every resolver. Also, in the highest layer (resolver) we highlight that we use DB. Middlewares would make it very easy to handle and it will make code cleaner.

pkosiec:
After discussions we rejected the idea with HTTP middlewares - we want to have transaction per resolver. Another idea is to write plugin to glgen and wrap all resolvers. https://gqlgen.com/reference/plugins/

PR #296, opened 2019-09-09
Pass JSON as string in Director

dbadura:
**Description**

- Change graphql JSON scalar from map[string]interface{} to string, to be able to accept full json schema with special characters as `$`.
- Changed Graphql converters to be able to Marshall/Unmarshall JSON Schema and return error.

**Related issue(s)**

See #258


components/director/internal/domain/labeldef/resolver.go
```diff
@@ -49,6 +47,12 @@ func (r *Resolver) CreateLabelDefinition(ctx context.Context, in graphql.LabelDe
 		return nil, err
 	}
 
+	// TODO: Use LabelDefinitionInput
```
pkosiec: Hmm, what does this TODO mean? the input is used, right?

components/director/internal/domain/labeldef/resolver.go
```diff
@@ -134,6 +148,11 @@ func (r *Resolver) UpdateLabelDefinition(ctx context.Context, in graphql.LabelDe
 	if err != nil {
 		return nil, err
 	}
+	// TODO: Use LabelDefinitionInput
```
pkosiec: Is this still valid?

components/director/internal/domain/labeldef/resolver.go
pkosiec: And also why did you move conversion to the top? If there will be transaction error, then there is no need to convert input to actual LabelDefinition.

components/director/internal/domain/labeldef/resolver.go
pkosiec: And also why did you move conversion to the top? If there will be transaction error, then there is no need to convert input to actual LabelDefinition.

components/director/internal/domain/labeldef/converter.go
```diff
@@ -15,19 +15,37 @@ func NewConverter() *converter {
...
+		Schema: &schema,
 		Tenant: tenant,
-	}
+	}, nil
 }
 
-func (c *converter) ToGraphQL(in model.LabelDefinition) graphql.LabelDefinition {
+func (c *converter) ToGraphQL(in model.LabelDefinition) (graphql.LabelDefinition, error) {
+	var jsonSchema graphql.JSON
+	if in.Schema != nil {
```
pkosiec: Please invert this condition to avoid deep nesting

components/director/internal/domain/labeldef/converter.go
```diff
@@ -15,19 +15,37 @@ func NewConverter() *converter {
 
 type converter struct{}
 
-func (c *converter) FromGraphQL(input graphql.LabelDefinitionInput, tenant string) model.LabelDefinition {
+func (c *converter) FromGraphQL(input graphql.LabelDefinitionInput, tenant string) (model.LabelDefinition, error) {
+	var schema interface{}
+	if input.Schema != nil {
```
pkosiec: Please invert this condition to avoid deep nesting

components/director/internal/domain/labeldef/converter.go
dbadura: Maybe should I extract to private method, because if I invert the condition I had return object in `if input.Schema==nil` condition and copy paste (change to `Schema: nil`) the return statement.

pkosiec: yup, private method would be a good idea

components/director/internal/domain/labeldef/resolver.go
pkosiec: Hmm, right, but I would keep the transaction at the beginning. For now we need to remember about opening it, so it's good to have it unified across all places. When we implement #257 then the problem will be gone. That's why please move back to opening transaction at the beginning of resolver.

PR #304, opened 2019-09-16
Add verbose output to go test in before-commit scripts

crabtree:
**Description**

- Add verbose output to go test in before-commit scripts

**Related issue(s)**

### DIFF
--- a/components/connector/before-commit.sh
+++ b/components/connector/before-commit.sh
@@ -80,14 +80,17 @@ fi
 ##
 # GO TEST
 ##
+TEST_OUTPUT="gotests.log"
 echo "? go test"
-go test ./...
+go test -v ./... > "$TEST_OUTPUT"
 # Check if tests passed
-if [[ $? != 0 ]];
-    then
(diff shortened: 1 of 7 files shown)


components/connector/before-commit.sh
```diff
@@ -80,14 +80,17 @@ fi
...
+go test -v ./... > "$TEST_OUTPUT"
 # Check if tests passed
-if [[ $? != 0 ]];
-    then
-    	echo -e "${RED}✗ go test\n${NC}"
-    	exit 1;
-	else echo -e "${GREEN}√ go test${NC}"
+if [[ $? != 0 ]]; then
+	echo -e "${RED}✗ go test\n${NC}"
+	cat "$TEST_OUTPUT"; rm "$TEST_OUTPUT"
```
dbadura: Switching the output will be more readable. 
We will get log and then error.

crabtree: Right now it is consistent with the rest of the script

PR #333, opened 2019-09-24
Generate scopes directive automatically via plugin

aszecowka:
In this PR:
- `scopes.yaml` was defined. Every GraphQL operation has assigned one required scope, with the following format: `<type>_<operation>`
- In generating code from GraphQL schema, we enabled a new plugin `scopesdecorator` which adds directive `hasScopes` to every query and mutation and persist that in `schema.graphql`
- TennantMapping handler set all available scopes if the request is secured with certificate

If you add a new query, the directive will be automatically added, but because path with required scopes will be not defined in `scopes.yaml` every request to it will fail with an error.


chart/compass/charts/director/scopes.yaml
```diff
@@ -1,2 +1,39 @@
+query:
+  applications: ["application_read"]
```
pkosiec: Hmm, I quickly checked naming conventions for scopes:
https://api.slack.com/docs/oauth-scopes
https://caseysoftware.com/blog/oauth-2-0-scopes-a-thought-experiment

And looks like colon is used widely. I think it would improve readability of the scopes. And, then we will be able to use underscore in other scopes, like `integration_system:read`

components/director/pkg/graphql/schema.graphql
```diff
@@ -1,552 +1,532 @@
-# Directives
+directive @hasScopes(path: String!) on FIELD_DEFINITION
```
pkosiec: Please add comment that this file is modified automatically with `hasScopes` directives

components/director/pkg/graphql/schema.graphql
```diff
@@ -1,552 +1,532 @@
...
+scalar Any
 
-directive @hasScopes(path:String!) on FIELD_DEFINITION
-# Scalars
+scalar CLOB
 
-scalar Any # -> interface{}
+scalar HttpHeaders
```
pkosiec: What about our comments? Can we keep them somehow?

aszecowka: No sir, only graphql comments are preserved

components/director/hack/scopesdecorator/plugin_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/kyma-incubator/compass/components/director/hack/scopesdecorator"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMutateConfig(t *testing.T) {
+	// GIVEN
+	cfg, err := config.LoadConfig("testdata/config.yaml")
+	require.NoError(t, err)
+	fmt.Println(cfg)
```
pkosiec: remove it pls

PR #337, opened 2019-09-26
Remove scoping overrides to compass component only

kubadz:
**Description**

- remove scoping overrides to compass component only as the same overrides have to be applied for other components installed with Compass (for example Dex)

**Related issue(s)**

### DIFF
--- a/installation/resources/installer-overrides-compass-gateway.yaml
+++ b/installation/resources/installer-overrides-compass-gateway.yaml
@@ -5,7 +5,6 @@ metadata:
   namespace: kyma-installer
   labels:
     installer: overrides
-    component: compass
     kyma-project.io/installation: ""
 data:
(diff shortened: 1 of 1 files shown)


installation/resources/installer-overrides-compass-gateway.yaml
```diff
@@ -5,7 +5,6 @@ metadata:
   namespace: kyma-installer
   labels:
     installer: overrides
-    component: compass
     kyma-project.io/installation: ""
 data:
   gateway.gateway.enabled: "true"
```
pkosiec: Hmm, it would be good to move it to compass-specific override

kubadz: Ok. I created another PR for that: https://github.com/kyma-incubator/compass/pull/338

PR #351, opened 2019-10-03
Add application name validation to application creation and update

### LINKED ISSUE #213: Limit maximum length of Application name
**Description**

Application name should be limited to 36 characters.

**Reasons**

The longer name will cause that Application Release will not be installed successfully due to limit on names length on some K8s resources.

**Attachments**

KarolJaksik:
**Description**

- Add application name length validation

**Related issue(s)**
Resolves #213

### DIFF
--- a/components/director/internal/domain/application/resolver.go
+++ b/components/director/internal/domain/application/resolver.go
@@ -14,8 +14,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-const applicaionNameMaxLength = 36
-
 //go:generate mockery -name=ApplicationService -output=automock -outpkg=automock -case=underscore
 type ApplicationService interface {
 	Create(ctx context.Context, in model.ApplicationInput) (string, error)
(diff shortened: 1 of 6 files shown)


components/director/internal/domain/application/resolver.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const applicaionNameMaxLength = 36
```
pkosiec: typo

components/director/internal/domain/application/resolver.go
```diff
@@ -293,6 +299,12 @@ func (r *Resolver) CreateApplication(ctx context.Context, in graphql.Application
 	return gqlApp, nil
 }
 func (r *Resolver) UpdateApplication(ctx context.Context, id string, in graphql.ApplicationInput) (*graphql.Application, error) {
+	convertedIn := r.appConverter.InputFromGraphQL(in)
+	if len(convertedIn.Name) > applicaionNameMaxLength {
```
pkosiec: The validation shouldn't be done here. As you can see the following code, name validation is already done in Service, while ApplicationInput -> Application conversion. Please modify that `Validate()` method which is used.

components/director/internal/model/application.go
```diff
@@ -72,5 +74,9 @@ func (i *ApplicationInput) Validate() error {
 	if errorMgs := validation.NameIsDNSSubdomain(i.Name, false); errorMgs != nil {
 		return errors.Errorf("%v", errorMgs)
 	}
+	if len(i.Name) > applicationNameMaxLength {
+		err := errors.New("Application name is too long, should be max 36 characters")
+		return err
```
pkosiec:
```go
return errors.New("application name is too long, must be maximum 36 characters long")
```

PR #352, opened 2019-10-03
Prepare database schema for Provisioner

Szymongib:
**Description**

- Prepare initial schema for Provisioner

**Related issue(s)**


components/provisioner/assets/database/provisioner.sql
```diff
@@ -0,0 +1,91 @@
...
+(
+    ID uuid PRIMARY KEY CHECK (id <> '00000000-0000-0000-0000-000000000000'),
+    name varchar(256) NOT NULL,
+    node_count integer NOT NULL,
+    disk_size varchar(256) NOT NULL,
+    machine_type varchar(256) NOT NULL,
+    compute_zone varchar(256) NOT NULL,
+    version varchar(256) NOT NULL,
+    infrastructure_provider infrastructure_provider NOT NULL,
+    runtime_id uuid NOT NULL,
```
akgalwas: I'm not sure if runtime_id is a suitable name here. The field foreign key is built on refers to cluster table so maybe cluster_id is better.

components/provisioner/assets/database/provisioner.sql
```diff
@@ -0,0 +1,91 @@
+-- Cluster
+
+CREATE TABLE cluster
+(
+    ID uuid PRIMARY KEY CHECK (id <> '00000000-0000-0000-0000-000000000000'),
```
crabtree: I would go with lowercase to be consistent with the rest.

```suggestion
    id uuid PRIMARY KEY CHECK (id <> '00000000-0000-0000-0000-000000000000'),
```

components/provisioner/assets/database/provisioner.sql
```diff
@@ -0,0 +1,91 @@
...
+
+-- Cluster Config
+
+CREATE TYPE infrastructure_provider AS ENUM (
+    'GARDENER',
+    'GCP',
+    'AKS'
+    );
+
+CREATE TABLE cluster_config
```
crabtree: Just wondering, maybe splitting the cluster options into separate fields is not the best option. I would suggest a slightly different approach. You can persist the infrastructure providers together with some schema of their respective options. Then, when storing the cluster config selecting the provider you will have the appropriate schema to validate the options. This approach will make it more resistant to further changes.

chart/compass/charts/provisioner/templates/deployment.yaml
```diff
@@ -31,3 +31,54 @@ spec:
               value: "0.0.0.0:{{ .Values.global.provisioner.graphql.port }}"
             - name: APP_API_ENDPOINT
               value: "/graphql"
+            - name: APP_DATABASE_USER
+              valueFrom:
+                secretKeyRef:
+                  name: compass-postgresql
+                  key: postgresql-username
+            - name: APP_DATABSE_PASSWORD
```
akgalwas: Typo: APP_DATABSE_PASSWORD

chart/compass/charts/provisioner/templates/deployment.yaml
```diff
@@ -31,3 +31,54 @@ spec:
...
+              valueFrom:
+                secretKeyRef:
+                  name: compass-postgresql
+                  key: postgresql-password
+            - name: APP_DATABSE_HOST
```
akgalwas: Typo: APP_DATABSE_HOST

chart/compass/charts/provisioner/templates/deployment.yaml
```diff
@@ -31,3 +31,54 @@ spec:
...
+              valueFrom:
+                secretKeyRef:
+                  name: compass-postgresql
+                  key: postgresql-serviceName
+            - name: APP_DATABSE_PORT
```
akgalwas: Typo: APP_DATABSE_PORT

chart/compass/charts/provisioner/templates/deployment.yaml
```diff
@@ -31,3 +31,54 @@ spec:
...
+              valueFrom:
+                secretKeyRef:
+                  name: compass-postgresql
+                  key: postgresql-servicePort
+            - name: APP_DATABSE_NAME
```
akgalwas: Typo: APP_DATABSE_NAME

components/provisioner/pkg/gqlschema/schema.graphql
```diff
@@ -1,4 +1,3 @@
-scalar AdditionalProperties
```
akgalwas: Shouldn't we add forceDeprovisionRuntime (or similar) method now?

components/provisioner/assets/database/provisioner.sql
```diff
@@ -0,0 +1,95 @@
...
+    start_timestamp timestamp without time zone NOT NULL,
+    end_timestamp timestamp without time zone NOT NULL,
+    cluster_id uuid NOT NULL,
+    foreign key (cluster_id) REFERENCES "Cluster" (id) ON DELETE CASCADE
+);
+
+
+-- Kyma Config
+
+CREATE TABLE "KymaConfig"
```
akgalwas: We should add Unique constraint here to enforce 1:1 relation (with the current design 1:n is possible)

components/provisioner/assets/database/provisioner.sql
```diff
@@ -0,0 +1,95 @@
...
+
+CREATE TABLE "Cluster"
+(
+    id uuid PRIMARY KEY CHECK (id <> '00000000-0000-0000-0000-000000000000'),
+    kubeconfig text,
+    terraform_state jsonb,
+    creation_timestamp timestamp without time zone NOT NULL,
+    gardener_config_id uuid,
+    gcp_config_id uuid,
+    foreign key (gardener_config_id) REFERENCES "GardenerConfig" (id) ON DELETE CASCADE,
```
akgalwas: After some consideration I think we should change the direction of the relation.
We have the following relations in the current design:
- Cluster have Gardener Configuration with cardinality 0:1
- Cluster have GCP Configuration with cardinality 0:1
- Kyma Configuration is related to Cluster with cardinality 1:1

We can change the design as follows:
- Gardener Configuration is related to Cluster with cardinality 1:1
- GCP Configuration is related to Cluster with cardinality 1:1
- Kyma Configuration is related to Cluster with cardinality 1:1

With the current design we need to create records in the following order:
- Gardener/GCP Configuration table
- Cluster table
- Kyma Configuration
- Operation

In case we add ClusterID field to KymaConfig table the order will be slightly changed:
- Cluster table
- Gardener/GCP Configuration table
- Kyma Configuration
- Operation

The consequence of this change is that before querying for runtime data we need to check what is the type of configuration. Having fields with configuration ids as in the current design is slightly easier (if we change the design as proposed there is a need to do a query on both GCPConfig and GardenerConfig tables)

Issue #311, opened 2019-09-17
Manage Integration Systems in Compass


pkosiec:
**Description**
We need to provide credentials for integration system by which they can call Compass API

**AC**
- Create DB table for Integration System (ID, name, ~scopes~) @tgorgol 
- For now Integration System can operate within any tenant @tgorgol
- Integration System is global entity (no tenancy) @tgorgol
- integration system has multiple auths (SystemAuth), similarly like in Runtime @tgorgol
- Add create, get, list (paginated), update and delete mutations for Integration System @tgorgol 
- add mutations: addClientCredentials (set scopes to allow integration system managing applications), deleteClientCredentials by ID: @pkosiec 
   - generate client_secret, get SystemAuth ID as client_id and call hydra with scopes
   - fetch scopes from static configmap during client_id/client_secret pair during generation
   - [x] fix scopes in configmap for only managing applications 
   - [x] double check if client can introspect available scopes from public hydra endpoint - if not, create follow up task for that: DONE https://github.com/kyma-incubator/compass/issues/362
  - [x] on SystemAuth delete call hydra and remove registered client
- prepare integration tests @kfurgol 

out of scope: UI

**Reasons**
Allow integration system to register any Application within multiple tenants

PK85:
1) Next iteration we will add a list of tenants that IS could work with.
2) list of tenants is updated when application is created from AppClass from given IS

aszecowka:
how IntegrationSystem provide the tenant in which context it is working (by headers?)? What integrationSystem provide, external tenant or internal tenant? Where tenants mapping is performed?

aszecowka:
> generate client_secret
make an HTTP call to hydra, please double check if this endpoint will not be removed in the future.

PK85:
> how IntegrationSystem provide the tenant in which context it is working (by headers?)? What integrationSystem provide, external tenant or internal tenant? Where tenants mapping is performed?
> 
> Our assumtption: we work on internal tenant

I guess header is a way to go. And right now as we don't have external tenant to  our tenant mapping it shoudl IS should send internal one. That's a temporary solution. We will use only exteral tenant in the future.

pkosiec:
In future we need to clean up Hydra clients for all deleted Applications, Runtimes and Integration System:
https://github.com/kyma-incubator/compass/issues/375

PR #358, opened 2019-10-04
Add Integration System domain and GQL types

tgorgol:
**Description**

- Add Integration System domain to Director
- Add Integration System types to gql schema
- Add `integration_system:read` and `integration_system:write` scopes

**Related issue(s)**

#311


components/director/pkg/graphql/schema.graphql
```diff
@@ -399,6 +403,17 @@ type HealthCheckPage implements Pageable {
 	totalCount: Int!
 }
 
+type IntegrationSystem {
```
aszecowka: IntegrationSystem should also have 
`auths: [SystemAuth!]!`

components/director/internal/domain/integrationsystem/service.go
```diff
@@ -0,0 +1,93 @@
...
+		if repo.IsNotUnique(err) {
+			return "", errors.New("Integration System name is not unique")
+		}
+		return "", err
+	}
+
+	return id, nil
+}
+
+func (s *service) Get(ctx context.Context, id string) (*model.IntegrationSystem, error) {
```
pkosiec: Please also add the following method: `Exist(ctx context.Context, id string) (bool, error)`

components/director/internal/domain/integrationsystem/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer r.transact.RollbackUnlessCommited(tx)
+
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	is, err := r.intSysSvc.Get(ctx, id)
+	if err != nil {
+		return nil, err
```
pkosiec: Check if it is not found - if so, return `nil, nil`

Issue #189, opened 2019-08-02
Do not return error in queries for single item


pkosiec:
**Description**

Currently we return error, 
For queries like `runtime(id: ID!) or `application(id: ID!)` we return error, if the ID is not found. We should return null, as our schema says.

Review all queries and mutations to see if the API is consistent.

PR #361, opened 2019-10-04
Return null on not found single item queries

KarolJaksik:
**Description**

- Create a generic package with custom errors
- Return null on single item queries if the ID is not found
- Review the consistency of graphql queries and mutations

**Related issue(s)**
#189


components/director/internal/domain/application/service.go
```diff
@@ -156,14 +155,13 @@ func (s *service) ListByRuntimeID(ctx context.Context, runtimeID uuid.UUID, page
...
 			return &model.ApplicationPage{
-				Data:       []*model.Application{},
-				PageInfo:   &pagination.Page{},
 				TotalCount: 0,
-			}, nil
+				Data:       []*model.Application{},
+				PageInfo:   &pagination.Page{StartCursor: "", EndCursor: "", HasNextPage: false}}, nil
 		}
-		return nil, errors.Wrap(err, "while getting scenarios for runtime")
+		return nil, errors.Wrap(err, "while getting scenarios for runtime") // TODO: Check for specific error in https://github.com/kyma-incubator/compass/issues/66
```
pkosiec: Please remove this TODO, it is already solved

components/director/internal/domain/application/service.go
```diff
@@ -193,6 +191,9 @@ func (s *service) Get(ctx context.Context, id string) (*model.Application, error
 
 	app, err := s.appRepo.GetByID(ctx, appTenant, id)
 	if err != nil {
+		if apperrors.IsNotFoundError(err) {
```
pkosiec: On services level, please always return error. It is a GraphQL-specific case, when we shouldn't return an error - that's why the check should be done in resolver.

components/director/internal/domain/labeldef/resolver.go
```diff
@@ -134,8 +134,9 @@ func (r *Resolver) LabelDefinition(ctx context.Context, key string) (*graphql.La
 	}
 
 	if def == nil {
-		return nil, fmt.Errorf("label definition with key '%s' does not exist", key)
+		return nil, nil
```
pkosiec: Please notify UI team about our breaking change in API - we will always return null for all queries for single items

components/director/internal/domain/labeldef/service.go
```diff
@@ -66,6 +68,9 @@ func (s *service) Create(ctx context.Context, def model.LabelDefinition) (model.
 func (s *service) Get(ctx context.Context, tenant string, key string) (*model.LabelDefinition, error) {
 	def, err := s.repo.GetByKey(ctx, tenant, key)
 	if err != nil {
+		if apperrors.IsNotFoundError(err) {
```
pkosiec: Nope, please move it to resolver layer

components/director/internal/repo/get.go
```diff
@@ -39,7 +41,7 @@ func (g *SingleGetter) Get(ctx context.Context, tenant string, conditions Condit
 	err = persist.Get(dest, q, allArgs...)
 	switch {
 	case err == sql.ErrNoRows:
-		return &notFoundError{}
+		return apperrors.NewNotFoundError(conditions[0].Val)
```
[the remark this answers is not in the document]
pkosiec: It would be better to have two constructors per error:
-  NewNotFoundError() +  NewNotUniqueError()
- NewNotFoundErrorWithID(identifier) / NewNotUniqueErrorWithID(identifier) <- but it has a better name 😛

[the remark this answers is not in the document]
pkosiec: @aszecowka maybe you have a better idea for constructor name 😄

components/director/internal/domain/application/service.go
```diff
@@ -156,13 +155,6 @@ func (s *service) ListByRuntimeID(ctx context.Context, runtimeID uuid.UUID, page
 
 	label, err := s.labelRepo.GetByKey(ctx, tenantID, model.RuntimeLabelableObject, runtimeID.String(), model.ScenariosKey)
 	if err != nil {
-		if strings.Contains(err.Error(), "not found") { // TODO: Check for specific error in https://github.com/kyma-incubator/compass/issues/66
```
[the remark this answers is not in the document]
pkosiec: Still valid

components/director/pkg/apperrors/errors.go
```diff
@@ -34,10 +42,18 @@ type NotUnique interface {
...
+
+func NewNotUniqueError(identifier string) *notUniqueError {
+	return &notUniqueError{
+		identifier: identifier,
+	}
+}
 
 func (e *notUniqueError) Error() string {
-	return "unique constraint violation"
+	return fmt.Sprintf("Object %s is not unique", e.identifier)
```
pkosiec: If identifier is empty, there will be 2 spaces. I would add simple if to append `%s ` if identifier is defined

components/director/internal/domain/labeldef/resolver.go
```diff
@@ -127,15 +129,15 @@ func (r *Resolver) LabelDefinition(ctx context.Context, key string) (*graphql.La
...
+		if apperrors.IsNotFoundError(err) {
+			return nil, nil
+		}
 		return nil, errors.Wrap(err, "while getting Label Definition")
 	}
 	if err := tx.Commit(); err != nil {
 		return nil, errors.Wrap(err, "while committing transaction")
 	}
 
-	if def == nil {
```
pkosiec: I think it is still safer to check if def is not nil, as we return pointer from the service. I know, this situation shouldn't happen, but still :P return new `apperror.NotFound` error, just in case

tests/end-to-end/gen-examples.sh
```diff
@@ -62,7 +62,10 @@ docker run --rm --network=${NETWORK} \
 
 echo -e "${GREEN}Building Director image...${NC}"
 
-cd "${SCRIPT_DIR}/../../components/director/" && make resolve build-image
+cd "${SCRIPT_DIR}/../../components/director/"
```
pkosiec: Why we won't use make resolve build-image anymore?

components/director/pkg/apperrors/errors.go
```diff
@@ -1,19 +1,27 @@
...
-	return &notFoundError{}
+func NewNotFoundError(identifier string) *notFoundError {
+	return &notFoundError{
+		identifier: identifier,
+	}
 }
 
 func (e *notFoundError) Error() string {
-	return "object not found in DB"
+	return fmt.Sprintf("Object %s not found", e.identifier)
```
pkosiec: Just apply changes from not unique error and we're ready to merge it 😉

components/director/internal/repo/create.go
```diff
@@ -46,7 +48,7 @@ func (c *universalCreator) Create(ctx context.Context, dbEntity interface{}) err
 	_, err = persist.NamedExec(stmt, dbEntity)
 	if pqerr, ok := err.(*pq.Error); ok {
 		if pqerr.Code == persistence.UniqueViolation {
-			return &notUniqueError{}
+			return apperrors.NewNotUniqueError("")
```
tgorgol: Maybe we could pass `pqerr.Column` to this constructor to get more detailed errors?

Issue #288, opened 2019-09-05
Implement tenant mapping service for Compass


pkosiec:
**Description**
Implement Tenant (Mapping) Service according to proposal #278 

**AC**
- client_id to tenant mapping for cert and client_credential flow (we will query DB for client_id in Runtme auth or application Auth and get tenant for corresponding app or runtime)
- in production env tenant will be passed with request, on local development we need a mapping between user and tenant and scopes (configMap)
  - assign 3 tenants from chart for UI to given users
- ~use dedicated user for calls to DB~
- tests only for TenantMapping Service
- Fix scopes in cert flow in code (use configmap)

See and verify things to do from proposal: https://github.com/kyma-incubator/compass/blob/master/docs/architecture/authentication-and-authorization.md

PK85:
Cert flow (from where to get attributes), check with corresponding team that works on that

aszecowka:
- can compass UI switch tenants? @PK85 
- proposed configmap structure (depends on prev question):
```
johnDoe@gmail.com: {
  tenants:[{dispayName, id}]
 scopes: ["scope_a", "scope_b"]
}
```

PK85:
> can compass UI switch tenants? @PK85 
Answer: 
- yes for static local user by dex. Such a user could be assigned too many tenants. And tenant selection is required.
- not really for users that come from outside and are authenticated by extel idernal identity providers, cause they are already logged into given tenant. But of course I assume they can switch tenant as well.

PR #364, opened 2019-10-08
Tenant mapping handler implementation

crabtree:
**Description**

- Tenant mapping handler implementation

**Related issue(s)**

See also #288


chart/compass/values.yaml
```diff
@@ -43,6 +43,9 @@ global:
 
   director:
     port: 3000
+    assignedScopes:
```
pkosiec: Please remove it and define it in config.yaml (old scopes.yaml)

components/director/cmd/main.go
```diff
@@ -158,3 +165,18 @@ func configureLogger() {
 	})
 	log.SetReportCaller(true)
 }
+
+func getTenantMappingHanderFunc(transact persistence.Transactioner, staticUsersSrc string) func(writer http.ResponseWriter, request *http.Request) {
+	if v := os.Getenv("TENANT_MAPPING_MOCK"); v != "" {
```
pkosiec: please define it in `config` or delete - because maybe we don't need it at all

components/director/cmd/main.go
```diff
@@ -158,3 +165,18 @@ func configureLogger() {
...
+		return tenantmapping.NewHandler().ServeHTTP
+	}
+
+	uidSvc := uid.NewService()
+	authConverter := auth.NewConverter()
+	systemAuthConverter := systemauth.NewConverter(authConverter)
+	systemAuthRepo := systemauth.NewRepository(systemAuthConverter)
+	systemAuthSvc := systemauth.NewService(systemAuthRepo, uidSvc)
+	staticUsersRepo, _ := tenantmapping.NewStaticUserRepository(staticUsersSrc)
```
pkosiec: why the error is ignored?

components/director/internal/tenantmapping/staticuser.go
```diff
@@ -0,0 +1,55 @@
+package tenantmapping
```
pkosiec: Where are unit tests?

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,64 @@
...
+		return id.(string), Token, true
+	}
+
+	if ids, ok := d.Header[ClientIDCertKey]; ok {
+		clientIds := ids.([]interface{})
+		return clientIds[0].(string), Certificate, true
+	}
+
+	if username, ok := d.Extra[UsernameKey]; ok && username != "" {
+		return username.(string), Username, true
```
pkosiec: check conversion error

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,64 @@
...
+
+// GetCredentialID looks for credential in the parsed request input represented by the DataV2 struct
+func (d *DataV2) GetCredentialID() (string, CredentialSrc, bool) {
+	if id, ok := d.Extra[ClientIDKey]; ok {
+		return id.(string), Token, true
+	}
+
+	if ids, ok := d.Header[ClientIDCertKey]; ok {
+		clientIds := ids.([]interface{})
+		return clientIds[0].(string), Certificate, true
```
pkosiec: check conversion error

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// GetCredentialID looks for credential in the parsed request input represented by the DataV2 struct
+func (d *DataV2) GetCredentialID() (string, CredentialSrc, bool) {
+	if id, ok := d.Extra[ClientIDKey]; ok {
+		return id.(string), Token, true
+	}
+
+	if ids, ok := d.Header[ClientIDCertKey]; ok {
+		clientIds := ids.([]interface{})
```
pkosiec: check conversion error

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,64 @@
...
+type DataV2 struct {
+	Subject string                 `json:"subject"`
+	Extra   map[string]interface{} `json:"extra"`
+	Header  map[string]interface{} `json:"header"`
+}
+
+// GetCredentialID looks for credential in the parsed request input represented by the DataV2 struct
+func (d *DataV2) GetCredentialID() (string, CredentialSrc, bool) {
+	if id, ok := d.Extra[ClientIDKey]; ok {
+		return id.(string), Token, true
```
pkosiec: check conversion error

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,64 @@
...
+
+// DataV2 represents parsed request input to the handler
+type DataV2 struct {
+	Subject string                 `json:"subject"`
+	Extra   map[string]interface{} `json:"extra"`
+	Header  map[string]interface{} `json:"header"`
+}
+
+// GetCredentialID looks for credential in the parsed request input represented by the DataV2 struct
+func (d *DataV2) GetCredentialID() (string, CredentialSrc, bool) {
```
pkosiec: Where are unit tests?

components/director/internal/tenantmapping/http.go
```diff
@@ -0,0 +1,42 @@
+package tenantmapping
```
pkosiec: Where are unit tests?

components/director/internal/tenantmapping/http.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	return data, nil
+}
+
+func respondWithError(writer http.ResponseWriter, err error, wrapperStr string) {
+	wrappedErr := errors.Wrap(err, wrapperStr)
+	log.Error(wrappedErr)
+
+	http.Error(writer, wrappedErr.Error(), http.StatusInternalServerError)
```
pkosiec: Is this really always InternalServerError? What if request body is empty? then it is invalid request, right?

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,64 @@
...
+		return Runtime
+	} else if sysAuth.IntegrationSystemID != nil {
+		return IntegrationSystem
+	}
+
+	return ""
+}
+
+const (
+	Application       RefObjectType = "Application"
```
pkosiec: maybe rename to `XRefObjectType`, such as `ApplicationRefObjectType` etc. - it will make clear that these values are enum-like ones

components/director/internal/tenantmapping/handler.go
```diff
@@ -105,3 +109,105 @@ func (h *Handler) setScopes(extraMap map[string]interface{}) {
 
 	extraMap["scope"] = strings.Join(scopes, " ")
 }
+
```
pkosiec: Please write missing unit tests for not covered test cases:

Screen Shot 2019-10-11 at 12 39 50

chart/compass/charts/director/templates/configmap.yaml
```diff
@@ -1,10 +1,33 @@
...
+metadata:
+  name: {{ template "fullname" . }}-static-users
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Chart.Name }}
+    release: {{ .Release.Name }}
+data:
+  {{- (.Files.Glob "static-users.json").AsConfig | nindent 2 }}
+---
+apiVersion: v1
```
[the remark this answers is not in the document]
crabtree: OK, but I think this should be aligned upfront next time.

chart/compass/charts/director/static-users.json
```diff
@@ -0,0 +1,9 @@
+[{
```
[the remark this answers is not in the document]
crabtree: In the task, we considered using JSON format.

chart/compass/values.yaml
crabtree: Waiting for the #365 to be resolved then.

components/director/cmd/main.go
crabtree: Don't get why we should treat toggles as configuration. But I decided to remove the previous implementation so this won't be needed anymore.

components/director/internal/tenantmapping/http.go
crabtree: Will extend the helper method to be able to define the error code.

components/director/internal/tenantmapping/model.go
crabtree: Well, I don't get the benefit from such change, could you elaborate a bit more?

chart/compass/charts/director/templates/deployment.yaml
```diff
@@ -74,9 +74,21 @@ spec:
               value: /config/scopes.yaml
             - name: APP_ALLOW_JWT_SIGNING_NONE
               value: {{ .Values.deployment.allowJWTSigningNone | quote }}
+            - name: APPLICATION_SCOPES
+              valueFrom:
+                configMapKeyRef:
+                  name: {{ template "fullname" . }}-assigned-scopes
+                  key: application
+            - name: RUNTIME_SCOPES
```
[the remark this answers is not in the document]
crabtree: not needed anymore as switched to common config map with scopes

components/director/internal/tenantmapping/handler.go
```diff
@@ -1,107 +1,143 @@
...
+		refObjType := sysAuthToCredRefObjectType(*sysAuth)
+
+		if refObjType == IntegrationSystemRefObject {
+			// TODO - handle tenant for integration system
+		}
+
+		return sysAuth.ID, sysAuth.TenantID, refObjType, nil
+	case UsernameCredSrc:
+		staticUser, err := h.staticUsersRepo.Get(credID)
+		if err != nil {
```
pkosiec: Add test case for it

components/director/internal/tenantmapping/handler.go
```diff
@@ -1,107 +1,143 @@
...
-	err = json.NewEncoder(writer).Encode(data)
-	if err != nil {
-		wrappedErr := errors.Wrap(err, "while encoding data")
-		log.Error(wrappedErr)
-		http.Error(writer, wrappedErr.Error(), http.StatusInternalServerError)
-		return
+		refObjType := sysAuthToCredRefObjectType(*sysAuth)
+
+		if refObjType == IntegrationSystemRefObject {
+			// TODO - handle tenant for integration system
```
pkosiec: Why this is not handled at all? Just get `Tenant` header and pass the value as tenant (while we don't have internal <-> external tenant mapping.

components/director/internal/tenantmapping/handler.go
```diff
@@ -1,107 +1,143 @@
...
+	writer.Header().Set("Content-Type", "application/json")
+	err = json.NewEncoder(writer).Encode(reqData)
+	if err != nil {
+		respondWithError(writer, http.StatusInternalServerError, err, "while encoding data")
+		return
 	}
+}
 
-	data.Extra = extraMap
+func (h *Handler) lookupForTenant(ctx context.Context, reqData ReqData) (string, string, CredenialRefObjectType, error) {
```
pkosiec: I don't like the idea of 4 return arguments - it's simply too much, especially if you have equal types (like here: string, string, CredenialRefObjectType, error). You cannot guess what is what.

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const (
+	ApplicationRefObject       CredenialRefObjectType = "Application"
+	RuntimeRefObject           CredenialRefObjectType = "Runtime"
+	IntegrationSystemRefObject CredenialRefObjectType = "Integration System"
+	UserRefObject              CredenialRefObjectType = "User"
+
+	CertificateCredSrc CredentialSrc = "Certificate"
+	TokenCredSrc       CredentialSrc = "Token"
```
pkosiec: In current implementation it describes just the OAuth 2.0 token case, so I would rename it

components/director/internal/tenantmapping/handler.go
pkosiec: I've quickly rewritten two `lookup...` methods into one piece of code:

```go
	credID, credSrc, err := reqData.GetCredentialID()
	if err != nil {
		// handle Error
	}

	var tenant string
	var scope string
	switch credSrc {
	case CertificateCredSrc, TokenCredSrc:
		sysAuth, _ := h.systemAuthSvc.GetGlobal(ctx, credID)
		tenant = sysAuth.TenantID
		refObjType := sysAuthToCredRefObjectType(*sysAuth)

		if refObjType == IntegrationSystemRefObject {
			// TODO - handle tenant for integration system
		}
		
		declaredScopes, _ := h.scopesGetter.GetRequiredScopes(buildPath(refObjType))
		scope = strings.Join(declaredScopes, " ")
	case UsernameCredSrc:
		staticUser, _ := h.staticUsersRepo.Get(credID)
		tenant = staticUser.Tenant
		scope = strings.Join(staticUser.Scopes, " ")
	}
```

You can extract it to a separate method and return tenant, scope and error.
I think now it's much simpler. What do you think?

components/director/cmd/main.go
```diff
@@ -53,6 +56,8 @@ type config struct {
 	JWKSSyncPeriod      time.Duration `envconfig:"default=5m"`
 	AllowJWTSigningNone bool          `envconfig:"default=true"`
 
+	StaticUsersSrc string `envconfig:"default=/data/static-users.yaml"`
```
pkosiec: Please modify `run.sh` script and provide a valid path to static-users.yaml for local development.

pkosiec:
Also, before merge, we need to notify UI team that the tenant selectbox should be gone

components/director/cmd/main.go
```diff
@@ -166,3 +176,17 @@ func configureLogger() {
...
 	log.SetReportCaller(true)
 }
+
+func getTenantMappingHanderFunc(transact persistence.Transactioner, staticUsersSrc string, scopeProvider *scope.Provider) (func(writer http.ResponseWriter, request *http.Request), error) {
+	uidSvc := uid.NewService()
+	authConverter := auth.NewConverter()
+	systemAuthConverter := systemauth.NewConverter(authConverter)
+	systemAuthRepo := systemauth.NewRepository(systemAuthConverter)
+	systemAuthSvc := systemauth.NewService(systemAuthRepo, uidSvc)
+	staticUsersRepo, err := tenantmapping.NewStaticUserRepository(staticUsersSrc)
```
pkosiec: BTW I think our approach to load this configmap is confusing. The other configuration we load every X minutes, and this one we load just once, on Director start.

I'm not saying to do it now, but please create follow-up task to also load static users periodically.

components/director/internal/tenantmapping/handler.go
crabtree: not needed after refactor

components/director/cmd/main.go
crabtree: I will refactor it, generally, I don't like the way our `main.go` looks like. Will try to do something with it.

components/director/internal/tenantmapping/handler.go
```diff
@@ -1,107 +1,143 @@
...
+	var scopes []string
+	switch refObjectType {
+	case UserRefObject:
+		staticUser, err := h.staticUsersRepo.Get(credID)
+		if err != nil {
+			return "", errors.Wrap(err, "while searching static user for scopes")
+		}
+
+		scopes = staticUser.Scopes
+	case RuntimeRefObject, ApplicationRefObject:
```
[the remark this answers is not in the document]
crabtree: Not really, the scopes for Integration System can be defined per instance of the Integration System. As I remember correctly you did the implementation of the client_id/client_secret management between director and hydra. 

This PR covers the case for the Application and the Runtime which basically can have scopes defined statically and mostly focuses on the client certificate and static user flows - as both do not have the notion of the scopes.

components/director/cmd/main.go
crabtree: I think we can continue this topic in the issue already defined by https://github.com/kyma-incubator/compass/issues/386

components/director/internal/tenantmapping/handler.go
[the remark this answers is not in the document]
crabtree: @PK85 @aszecowka @pkosiec in the https://github.com/kyma-incubator/compass/issues/311 I see that those scopes for integration systems were de-scoped but I can't find any reasoning why. Can you elaborate a bit on that? 
Does it mean that all Integration System has common scopes assigned?

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if sysAuth.IntegrationSystemID != nil {
+		return IntegrationSystemRefObject
+	}
+
+	return ""
+}
+
+const (
+	ApplicationRefObject       AuthIDRefObjectType = "Application"
```
aszecowka: in `model/system_auth.go` we define similar consts, can we reuse it?

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,89 @@
...
+
+// ReqData represents parsed request input to the handler
+type ReqData struct {
+	Subject string                 `json:"subject"`
+	Extra   map[string]interface{} `json:"extra"`
+	Header  http.Header            `json:"header"`
+}
+
+// GetAuthID looks for auth ID in the parsed request input represented by the ReqData struct
+func (d *ReqData) GetAuthID() (string, AuthIDSrc, error) {
```
aszecowka: minor: maybe we can introduce a struct that contains authID and it's source

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/pkg/errors"
+)
+
+// AuthIDSrc wraps possible sources of auth ID like token, username and certificate
+type AuthIDSrc string
+
+// AuthIDRefObjectType wraps possible options for the origin object represented by the auth ID
+type AuthIDRefObjectType string
+
+func sysAuthToAuthIDRefObjectType(sysAuth model.SystemAuth) AuthIDRefObjectType {
```
aszecowka: minor: maybe we can extend model.SystemAuth and implement this method there?

components/director/internal/tenantmapping/handler_test.go
```diff
@@ -1,38 +1,413 @@
...
 
 func TestHandler(t *testing.T) {
-	t.Run("for request without scopes", func(t *testing.T) {
-		req := httptest.NewRequest(http.MethodPost, "http://example.com/foo", strings.NewReader("{}"))
+	systemAuthID := "088972c1-a42e-4f4b-b494-916db5342476"
+	tenantID := "fad7abea-202b-40f3-b325-17a695a68016"
+
+	t.Run("success for request with the client_id specified", func(t *testing.T) {
+		reqPayload := `{"Extra": {"client_id": "` + systemAuthID + `"}}`
+		respPayload := `{"subject":"","extra":{"client_id":"` + systemAuthID + `","scope":"application:read","tenant":"` + tenantID + `"},"header":null}`
```
aszecowka: minor: expectedRespPayload

components/director/internal/tenantmapping/model_test.go
```diff
@@ -0,0 +1,51 @@
...
+			IntegrationSystemID: str.Ptr("3f2d9157-a0ba-4ff3-9d50-f5d6f9730eed"),
+		}
+
+		refObjType := sysAuthToAuthIDRefObjectType(sysAuth)
+
+		require.Equal(t, IntegrationSystemRefObject, refObjType)
+	})
+}
+
+// func TestGetTenantID(t *testing.T) {
```
aszecowka: commented code

components/director/internal/tenantmapping/staticuser.go
```diff
@@ -0,0 +1,55 @@
...
+	"io/ioutil"
+
+	"github.com/ghodss/yaml"
+
+	"github.com/pkg/errors"
+)
+
+type StaticUser struct {
+	Username string   `json:"username"`
+	Tenant   string   `json:"tenant"`
```
aszecowka: can we use UUID instead of string for tenant? I am wondering what will happen if someone provide invalid tenant, maybe we should fail fast on loading data to the memory?

components/director/internal/tenantmapping/staticuser_test.go
```diff
@@ -0,0 +1,82 @@
...
+- username: "developer"
+  sopes:
+  - "application:read"
+  tenant: "555d9157-a1bf-4aa2-9a22-f5d6f9730aaf"
+`
+
+func TestStaticUserRepository(t *testing.T) {
+	t.Run("NewStaticUserRepository returns repository instance populated from valid file", func(t *testing.T) {
+		filePath := "static-users.tmp.json"
+		ioutil.WriteFile(filePath, []byte(validFileContent), 0644)
```
aszecowka: maybe try to use ioutil.TempFile

aszecowka: test errors from  `WriteFile` and `os.Remove`

components/director/internal/tenantmapping/model.go
crabtree: Similar, but its context is different. It is missing the User as reference object which as System Auth does not exist

components/director/cmd/main.go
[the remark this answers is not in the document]
crabtree: For now yes, we always use DEX as our user provider.

components/director/internal/tenantmapping/mapper_for_system_auth_test.go
```diff
@@ -0,0 +1,307 @@
...
+	systemauthmock "github.com/kyma-incubator/compass/components/director/internal/domain/systemauth/automock"
+	"github.com/kyma-incubator/compass/components/director/internal/model"
+	tenantmappingmock "github.com/kyma-incubator/compass/components/director/internal/tenantmapping/automock"
+	"github.com/kyma-incubator/compass/components/director/pkg/str"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestMapperForSystemAuthGetTenantAndScopes(t *testing.T) {
+	t.Run("GetTenantAndScopes returns error when unable to get SystemAuth from the service", func(t *testing.T) {
```
aszecowka: minor: for me it is easier to read tests if we starts from happy paths. In addition to that `GetTenantAndScopes` appears twice in the test name.

crabtree: Done

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,133 @@
...
+	if tenantVal := d.Header.Get(TenantKey); tenantVal != "" {
+		return tenantVal, nil
+	}
+
+	return "", apperrors.NewKeyDoesNotExistError(TenantKey)
+}
+
+// GetScopes returns scopes from the parsed request input if defined
+func (d *ReqData) GetScopes() (string, error) {
+	if scopesVal := d.Body.Header.Get(ScopesKey); scopesVal != "" {
```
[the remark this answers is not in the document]
crabtree: removed

components/director/internal/tenantmapping/model.go
```diff
@@ -0,0 +1,133 @@
...
+	if scopesVal, ok := d.Body.Extra[ScopesKey]; ok {
+		scopes, err := str.Cast(scopesVal)
+		if err != nil {
+			return "", errors.Wrapf(err, "while parsing the value for %s", ScopesKey)
+		}
+
+		return scopes, nil
+	}
+
+	if scopesVal := d.Header.Get(ScopesKey); scopesVal != "" {
```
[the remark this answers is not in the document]
crabtree: removed

components/director/internal/tenantmapping/mapper_for_system_auth.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return "", "", errors.Wrap(err, "while retrieving system auth from database")
+	}
+
+	refObjType, err := sysAuth.GetReferenceObjectType()
+	if err != nil {
+		return "", "", errors.Wrap(err, "while getting reference object type")
+	}
+
+	var tenantAndScopesFunc func(sysAuth *model.SystemAuth, refObjType model.SystemAuthReferenceObjectType, reqData ReqData, authFlow AuthFlow) (string, string, error)
```
pkosiec: Suggestion: I understand what you wanted to achieve - not duplicating passing parameters. However, I'd say that using  simple switch/case with passing the parameters would be easier to read.

```go
return m.getTenantAndScopesForIntegrationSystem(sysAuth, refObjType, reqData, authFlow)
```

crabtree: modified

components/director/assets/static-users-local.yaml
```diff
@@ -0,0 +1,22 @@
+- username: "admin"
```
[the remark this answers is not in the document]
crabtree: moved

components/director/run.sh
```diff
@@ -88,6 +88,7 @@ APP_DB_USER=${DB_USER} \
 APP_DB_PASSWORD=${DB_PWD} \
 APP_DB_NAME=${DB_NAME} \
 APP_SCOPES_CONFIGURATION_FILE=${ROOT_PATH}/../../chart/compass/charts/director/config.yaml \
+APP_STATIC_USERS_SRC=${ROOT_PATH}/assets/static-users-local.yaml \
```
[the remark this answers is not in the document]
crabtree: in terms of usage context, it differs a bit, but in terms of consistency, it is valid. I will create a local (hack) copy of this file.

components/director/internal/tenantmapping/staticuser_test.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var validFileContent string = `
+- username: "admin"
+  scopes:
+  - "application:write"
+  tenants: 
+  - "3f2d9157-a0ba-4ff3-9d50-f5d6f9730eed"
+- username: "developer"
+  sopes:
```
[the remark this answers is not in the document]
crabtree: library updated and case covered with appropriate test

Issue #316, opened 2019-09-19
Prepare proposal for Package API


aszecowka:
**Description**

We need a way in compass( and then in Runtime) to register a set of APIs where one credential will access all of them.

**Reasons**

**AC**
- proposal how to model that on compass Application API
- registered PackageAPI must be represented as a one ServiceClass in the Kyma Runtime
- proposal what we need to change in a Runtime as well

**Ideas, hints**
- we could use labelDefinition to add a PackageAPI functionality and do not change current Application API (then PackageAPI list all of the APIs under one ID and name and Runtime Agent knows what do do with that)
- see diagrams with flows in our MS Team

PR #404, opened 2019-10-23
Introduce API Package Proposal

pkosiec:
**Description**

- Introduce API Package Proposal

**Related issue(s)**
Resolves https://github.com/kyma-incubator/compass/issues/316


docs/architecture/input-parameters-for-service-classes.md
```diff
@@ -0,0 +1,23 @@
...
+    - For Event Definition: `instance-create-params.{api-name}.instance-create-params`
+    - For API Package: `instance-create-params.api-packages.{api-package-name}`
+    
+**Value**: JSON schema for the `instanceCreateParameterSchema` for given Service Class
+
+On Kyma Runtime, Runtime Agent queries Applications for given Runtime. It reads all labels with prefix `instance-create-params` and updates `instanceCreateParameterSchema` properties in Service Class list to create.
+
+## Client Usage
+
+To be defined later.
```
pkosiec: As agreed with @PK85, this is something we need to discuss in future.

docs/architecture/api-packages.md
```diff
@@ -0,0 +1,156 @@
...
+## Example
+
+> **NOTE:** For clarity, in this example IDs has been changed from UUID type to string.
+
+1. User creates Application `sample-app`.
+1. User registers three different API definitions with the following IDs: `foo`, `bar` and `baz`.
+1. In order to create API Package called `sample-package`, User sets the following label for `sample-app` Application:
+
+   ```yaml
+   api-packages:
```
crabtree: In the example above and in the JSON schema you operate on the JSON object where keys are the names of the api-packages. Here, on the other hand, you present the array of objects and the name is included in the object.

docs/architecture/api-packages.md
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+}
+```
+
+In a result, the `api-packages` label value is validated. Also, the API Package name uniqueness within Application is guaranteed. Event and API Definition existence is not checked, apart from verifying the UUID type of the ID arrays.
+
+### Assumptions
+
+- API Package name doesn't have to be unique within tenant.
```
crabtree: I miss here the fact that we limit the usage of the API Package to the Application scope. In this concept, you cannot mix under one API Package APIs or Events from different applications.

docs/architecture/api-packages.md
```diff
@@ -0,0 +1,156 @@
...
+  }
+}
+```
+
+The root object key defines API Package name. In this example, `foo` is the API Package name.
+
+In order to allow validation of the label value, Compass creates `api-packages` LabelDefinition automatically. It is done while creating new Application. Once Tenant management feature is ready, creating new tenant will be used as a trigger for the LabelDefinition creation.
+The `api-packages` LabelDefinition has the following JSON schema:
+
+```json
```
crabtree: Something that is worth considering here is the fact that we allow the user to edit this JSON schema. I think we need to introduce some concepts that will allow us defining LabelDefinitions that will not be editable by the users.

docs/architecture/input-parameters-for-service-classes.md
```diff
@@ -0,0 +1,23 @@
...
+
+Previously, in Kyma Runtime, while provisioning Service Class created from API/Event Definition, there were no possibility to pass input parameters. The ability may be useful in some specific use cases. This document describes passing input parameters from Kyma Runtime to Application or Integration System.
+
+## Solution
+
+The solution utilizes labels on Application level. User provides JSON schema for Service Class `instanceCreateParameterSchema` parameter by creating new Label Definition in the following format:
+
+ **Key**: 
+    - For API Definition: `instance-create-params.apis.{api-name}`
+    - For Event Definition: `instance-create-params.{api-name}.instance-create-params`
```
crabtree: Not sure if the name is valid, shouldn't it follow this format:
`instance-create-params.events.{event-id}`

docs/architecture/runtime-agent-contract.md
```diff
@@ -15,9 +15,17 @@ Runtime Agent automates the process of [establishing trusted connection](./estab
 
 Depending on the type of trusted connection, during the runtime lifecycle, it may be required to periodically [renew trusted connection](./establishing-trusted-connection.md#client-certificate-flow---certificate-renewal).
 
-## Configuring the runtime
+## Configuring the Runtime
 
-Runtime Agent periodically requests for the configuration of its Runtime from the Management Plane. Changes in the configuration for the Runtime are applied by the Runtime Agent on the Runtime. To fetch the Runtime configuration the Runtime Agent calls `applicationsForRuntime(runtimeId: ID!, first: Int = 100, after: PageCursor)` query offered by Director. The response for the query contains a page with list of Applications assigned for the Runtime and info about next page. Each Application will contain only credentials that are valid for the runtime that called the query. Each Runtime Agent can fetch the configurations for Runtimes that belong to its tenant, there is no validation if the Runtime Agent is fetching the configuration for the runtime on which it runs.
+Runtime configuration is connecting Applications to Runtime, which are situated in the same scenario(s).
```
crabtree: This sentence is a bit confusing. How does this connection process look like? I think this should be rephrased or removed.

crabtree:
Why does this PR points to the closed PR #389? Shouldn't it point to issue #316?

docs/architecture/api-packages.md
pkosiec: Right, I forgot to mention that. I will update this paragraph.

docs/architecture/api-packages.md
pkosiec: I forgot to update this section. Sorry

docs/architecture/api-packages.md
pkosiec: Yes, this is also the assumption. I will update the paragraph

docs/architecture/api-packages.md
```diff
@@ -0,0 +1,159 @@
...
+
+In a result, the `api-packages` label value is validated. Also, the API Package name uniqueness within Application is guaranteed. Event and API Definition existence is not checked, apart from verifying the UUID type of the ID arrays.
+
+This schema cannot be modified by user. We validate this JSON schema against internal JSON schema, similarly to `scenarios` LabelDefinition.
+
+### Assumptions
+
+- API or Event Definition can be a part of API Package only within the same Application entity.
+- API Package name doesn't have to be unique within tenant.
+- One API or Event Definition can belong to only one API Package. If an API or Event Definition is assigned to multiple API packages, or some items with provided IDs doesn't exist, then Runtime Agent reports a configuration sync error.
```
aszecowka: > Runtime Agent reports a configuration sync error.

How?

docs/architecture/api-packages.md
```diff
@@ -0,0 +1,159 @@
...
+
+The following section describes client usage of the API Packages.
+
+### Runtime Agent
+
+To create Service Classes in Kyma Service Catalog from Application APIs and Event Definitions, Runtime Agent (or another component on Runtime side) has to populate the Service Class list.
+
+In order to do so, Runtime Agent executes the following steps:
+
+1. Query Applications for Runtime
```
aszecowka: If we have many agents, every agent has to implement the same logic. Why Compass cannot prepare "ready-to-use" data and store it in another label?

docs/architecture/api-packages.md
```diff
@@ -0,0 +1,159 @@
...
+    "minLength": 3,
+    "maxLength": 35
+  },
+  "additionalProperties": {
+    "anyOf": [{ "$ref": "#/definitions/apiPackage" }]
+  },
+  "definitions": {
+    "apiPackage": {
+      "type": "object",
+      "required": ["events", "apis"],
```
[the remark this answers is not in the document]
pkosiec: You can, but these keys has to be defined explicitly. You can specify empty array.

docs/architecture/api-packages.md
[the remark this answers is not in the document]
pkosiec: Hmm, to be honest, proposed solution doesn't enforce it. What do you think @PK85? Maybe I will just remove this assumption

docs/architecture/api-packages.md
pkosiec: As discussed, in future we will think about doing this work on Director by extending our API.

docs/architecture/api-packages.md
```diff
@@ -0,0 +1,159 @@
...
+  "propertyNames": {
+    "pattern": "^[a-z0-9_-]*$",
+    "minLength": 3,
+    "maxLength": 35
+  },
+  "additionalProperties": {
+    "anyOf": [{ "$ref": "#/definitions/apiPackage" }]
+  },
+  "definitions": {
+    "apiPackage": {
```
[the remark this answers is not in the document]
pkosiec: I agree - let's discuss it offline.

docs/architecture/api-packages.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Runtime Agent
+
+To create Service Classes in Kyma Service Catalog from Application APIs and Event Definitions, Runtime Agent (or another component on Runtime side) has to populate the Service Class list.
+
+In order to do so, Runtime Agent executes the following steps:
+
+1. Query Applications for Runtime
+1. For every Application:
+   - Read `api-packages` label and populate ServiceClasses for API Packages. Agent generates IDs for these Service Classes.
```
Szymongib: Is the simplification here intended?
I think we should consider things on the Runtime side a little more carefully as I assume it will require some major changes in the Application CR.
I wonder if it shouldn't be consider as a part of this proposal as some issues may appear, that can influence the design of this feature.

pkosiec: Yes, in this section it is intended, as the overall Compass API doesn't enforce any specific design of populating Service Classes on Runtime.

However, I also agree, that a detailed document about Kyma Runtime Agent changes should be created - but I don't know if it should be here or in Kyma repo, where Runtime Agent, Application Operator and other components are placed. What do you think @PK85?

docs/architecture/api-packages.md
```diff
@@ -0,0 +1,164 @@
...
+
+To display all API Packages, UI executes the following steps:
+
+1. Get all Applications
+1. Read `api-packages` label value from every Application (if it does exist)
+1. Display API Packages
+
+As API Package name doesn't have to be unique within tenant, UI displays Application name for every API Package, to distinguish API Packages with equal names.
+
+## Future
```
PK85: Please mention, that in the future we can support validation between api-packages and apis, eventAPIs, in bboth directions.

docs/architecture/input-parameters-on-runtime.md
```diff
@@ -0,0 +1,23 @@
...
+- For Event Definition: `instance-create-params.{event-def-name}.instance-create-params`
+- For API Package: `instance-create-params.api-packages.{api-package-name}`
+    
+**Value**: JSON schema for the `instanceCreateParameterSchema` for given Service Class
+
+On Kyma Runtime, Runtime Agent queries Applications for given Runtime. It reads all labels with prefix `instance-create-params` and updates `instanceCreateParameterSchema` properties in Service Class list to create.
+
+## Client Usage
+
+To be defined later.
```
PK85: Define clearly that we still need to define how we will store, many instances of input parameters in compass/application itself.

docs/architecture/input-parameters-on-runtime.md
```diff
@@ -0,0 +1,23 @@
...
+## Overview
+
+Previously, in Kyma Runtime, while provisioning Service Class created from API/Event Definition, there were no possibility to pass input parameters. The ability may be useful in some specific use cases. This document describes passing input parameters from Kyma Runtime to Application or Integration System.
+
+## Solution
+
+The solution utilizes labels on Application level. User provides JSON schema for Service Class `instanceCreateParameterSchema` parameter by creating new Label Definition in the following format:
+
+ **Key**: 
+- For API Definition: `instance-create-params.apis.{api-def-name}`
```
PK85: take into account proposal from integrationsystemName/input-param. Maybe we should have a convention for that.

docs/architecture/api-packages.md
pkosiec: As discussed with @PK85: please think about solution, which changes are needed in Agent, Application CR etc. and create an issue for implementation for next release. Thanks!

PR #407, opened 2019-10-24
Update README.md

akrausesap:
**Description**

- Added create namespace to helm installation

**Related issue(s)**
N/A


akrausesap:
actually it looks like this is not enough. makes the chart install but config is using non-working defaults

klaudiagrz:
@pkosiec Can you help with this one?

pkosiec:
HI @akrausesap, which Helm version do you use? Helm 2.10 does create a namespace, if you specify one that doesn't exist. AFAIK this behavior is changed in Helm 3.

> actually it looks like this is not enough. makes the chart install but config is using non-working defaults

What do you mean by `non-working` defaults? What doesn't work for you?

DimitarPetrov:
Closed due to inactivity.

Issue #255, opened 2019-08-27
Improve GraphQL input validation


pkosiec:
**Description**

Currently we have very limited validation for inputs.
For example, some fields are required, but still user could provide empty string. There should be consistent approach for every input.

- Investigate what we should validate
- Validate strings that are required and shouldn't be empty
- Validate URLs - We can use URI type extension for PostgreSQL or some kind of validator https://github.com/petere/pguri.
- Validate application names (like length etc.)
- create a ticket with solution for validatipon/conversion specs values. (assign to next release)

Complexity level: high

**Reasons**

It is crucial to block incorrect user input to assure everything (agent, integrations, etc.) works properly.

tgorgol:
Now that we support XML SpecFormat (used by OData) we could validate if provided spec type (OData, OpenAPI, AsyncAPI) and spec format (XML, JSON, YAML) are correct (for example, we could return error if someone tries to create spec of type OpenAPI with XML format).

tgorgol:
## Proposed validation library

https://github.com/go-playground/validator - in this library developer uses struct tags to define validation rules. It is possible to define custom rules or struct validation rules. Validator instance caches struct info, so it has to be passed as dependency (a single instance of validator could be constructed in root resolver and passed to services.

## Proposed validation rules for Compass input types

### APIDefinitionInput

Field | Rules (validation tags) | Comment
-- | -- | --
name: String! | required |  
description: String |   |  
targetURL: String! | required,url |  
group: String |   |  
spec: APISpecInput |   | ODATA works with XML and JSON<br>OPEN_API works with YAML and JSON
version: VersionInput |   |  
defaultAuth: AuthInput |   |  

### APISpecInput

Field | Rules (validation tags) | Comment
-- | -- | --
data: CLOB (string) |   |  
type: APISpecType! | required,oneof=ODATA OPEN_API |  
format: SpecFormat! | required,oneof=YAML JSON XML |  
fetchRequest: FetchRequestInput |   |  

### EventAPIDefinitionInput

Field | Rules (validation tags) | Comment
-- | -- | --
name: String! | required |  
description: String |   |  
spec: EventAPISpecInput! | required | ASYNC_API works with YAML and JSON
group: String |   |  
version: VersionInput |   |  

### EventAPISpecInput

Field | Rules (validation tags) | Comment
-- | -- | --
data: CLOB (string) |   |  
eventSpecType: EventAPISpecType! | required,oneof=ASYNC_API |  
[...]

Issue #410, opened 2019-10-24
API spec validation erroneous 


akrausesap:
**Description**

Although the below attached API specs seem valid they fail with error `Error: cannot recognize spec file.`

[API_OPLACCTGDOCITEMCUBE_SRV.yaml.zip](https://github.com/kyma-incubator/compass/files/3767326/API_OPLACCTGDOCITEMCUBE_SRV.yaml.zip)
[api-docs.json.zip](https://github.com/kyma-incubator/compass/files/3767329/api-docs.json.zip)

Kyma Version is 1.6 (upgraded from 1.5). Install via standard installer (image `eu.gcr.io/kyma-project/compass-ui:541467b2`)

**Expected result**

APIs are uploaded

**Actual result**
^
APIs fail 

**Steps to reproduce**

Add Specification to application

**Troubleshooting**

unknown

PK85:
It's already fixed in the UI.

Issue #409, opened 2019-10-24
Support for edmx file extension


akrausesap:
**Description**

OData metadata files often come with file extension .edmx. However today compass only supports .xml. This requires unnecessary renaming steps by the users.

**Expected result**

Extension .edmx is supported

**Actual result**

Extension .edmx is not supported

**Steps to reproduce**

Create Application and upload API Spec

**Troubleshooting**

rename .edmx file to xml

PR #422, opened 2019-10-30
Add api specification support proposal

tgorgol:
**Description**

- Add api specification support proposal

**Related issue(s)**

#255

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -24,6 +24,7 @@ These documents provide you with more architectural details:
 - [Authentication and authorization](./architecture/authentication-and-authorization.md)
 - [Integration Systems](./architecture/integration-systems.md)
 - [Proposal: Input validation](./architecture/input-validation.md)
+- [Proposal: API specifications support](./architecture/api-specs-support.md)
 - [Automatic Application Management](./architecture/auto-app-management.md)
 - [Proposal: API Packages](./architecture/api-packages.md)
 - [Proposal: Input parameters for API, Event Definitions or API Packages in the Runtime](./architecture/input-parameters-on-runtime.md)
--- a/docs/architecture/api-specs-support.md
+++ b/docs/architecture/api-specs-support.md
@@ -1,0 +1,35 @@
+# API Specifications Support
+
+## Overview
+
+This document describes current and planned API specification support in Kyma components.
+
+## Currently supported API specification versions and formats
+
+The following table shows currently supported API specification versions and formats in different components.
+
+> **NOTE:** The last row describes support for extracting API specifications from archive files.
+
+| | Console UI & Compass UI (rendering) | Runtime Rafter (storing) |  Compass UI (uploading) | Compass Backend (storing) | Runtime Agent (storing)
+| --- | --- | --- | --- | --- | --- |
+| OpenAPI<br>(YAML, JSON) | all formats, v2 and v3 | no validation / conversion | all versions and formats, validation by property | all versions and formats, no validation | all versions, JSON only, no validation and conversion |
+| OData<br>(XML, JSON in v4) | XML only, all versions converted to v4 on client side |  no validation / conversion |  all versions, XML only, validation by parsing XML file and checking property | all versions and formats, no validation | all versions and formats, no validation and conversion |     
+| AsyncAPI<br>YAML, JSON | all formats, v2 only | validation and conversion to JSON and v2 | all versions and formats, validation by property | all versions and formats, no validation | all versions, JSON only, no validation and conversion |
+| Archives | no | yes (TAR and ZIP) | no | no | no |
+
+## Planned support for API specification versions and formats
+
+The following table shows the proposed supported API specification versions and formats in different components.
+Rendering would support only a single version of API specifications.
+Rafter would be responsible for validating and converting API specifications for Console UI.
+Because Compass UI does not use Rafter conversion would have to be done in Compass Backend.
+Preferably Rafter's converters and validators should be exposed as services or libraries that could be used by Compass Backend and Compass UI.
+
+> **NOTE:**  The last row describes support for extracting API specifications from archive files. 
+
+| | Console UI & Compass UI (rendering) | Runtime Rafter (storing) |  Compass UI (uploading) | Compass Backend (storing) | Runtime Agent (storing)
+| --- | --- | --- | --- | --- |  --- |
+| OpenAPI<br>(YAML, JSON) | v3, JSON and YAML | validation and conversion to version supported by renderer | all versions and formats, validation using same logic as rafter | all versions and formats, validation and conversion using same logic as rafter | all versions and formats, no validation and conversion |
+| OData<br>(XML, JSON in v4) | v4, JSON |  validation and conversion to version supported by renderer |  all versions and formats, validation using same logic as rafter | all versions and formats, validation and conversion using same logic as rafter | all versions and formats, no validation and conversion |
+| AsyncAPI<br>YAML, JSON | v2, JSON and YAML | validation and conversion to version supported by renderer | all versions and formats, validation using same logic as rafter | all versions and formats, validation and conversion using same logic as rafter | all versions and formats, no validation and conversion |
+| Archives | no | yes (TAR and ZIP) | no | no | no |


docs/investigations/input-validation/input-validation.md
```diff
@@ -0,0 +1,254 @@
...
+
+```
+
+## Validation rules explanation
+
+- `name` - Kubernetes resources can have names up to 253 characters long. The characters allowed in names are: digits (`0`-`9`), lower case letters (`a`-`z`),`-`, and `.`.  We could use the same pattern to validate our names.
+- `required` - Cannot be nil or empty.
+- `not_empty` - Cannot be empty (can be nil if pointer).
+- `url` - Valid URL.
+- `max` - Maximal allowed length.
```
crabtree: I would go with `max_length` as we should use also `max` and `min` as validation rules for numeric values. Such numeric validation may be applied, for example, in values like port number validation which needs to be in between 0 or 1 and 65535.

```suggestion
- `max_length` - Maximal allowed length.
```

tgorgol: I agree, but these aliases are only used here to make the table more readable, they are not used in code.

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+
+This document describes current and planned api specification support in Kyma components.
+
+## Currently supported api specification versions and formats
+
+Following table shows currently supported api specification versions and formats in different components.
+
+Last row describes support for extracting api specifications from archive files.
+
+| | Console UI & Compass UI (rendering) | Rafter (storing) |  Compass UI (uploading) | Compass Backend (storing) | Runtime Agent (storing)
```
PK85: How Rafter (storing) and Runtime Agent (storing) differs?

maybe:
1. Rafter (storing) -> Runtime Rafter  (storing). In both tables.
1. will be nice to see red/gree rows in every column about format/version/validation/conversion. By that we will see which column is a most limited one. Think how to visualize that.

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+| AsyncAPI<br>YAML, JSON | all formats, v2 only | validation and conversion to JSON and v2 | all versions and formats, validation by property | all versions and formats, no validation | all versions, JSON only, no validation and conversion |
+| Archives | no | yes (TAR and ZIP) | no | no | no |
+
+## Planned support for api specification versions and formats
+
+Following table shows proposed supported api specification versions and formats in different components.
+Rendering would support only a single version of api specifications.
+Rafter would be responsible for validating and converting api specifications for Console UI.
+Because Compass UI doesn't use Rafter conversion would have to be done in Compass Backend.
+Preferably Rafter's converters and validators should be exposed as services or libraries that could be used by Compass Backend and Compass UI.
```
PK85: 1. Prio one: as a libraries if possible, cause we save specs in our db, and I want to avoid for now additional components in compass cluster.

docs/architecture/input-validation.md
```diff
@@ -0,0 +1,202 @@
...
+- `name` - Up to 36 characters long. The characters allowed in names are: digits (`0`-`9`), lower case letters (`a`-`z`),`-`, and `.`. Based on Kubernetes resource name format.
+- `required` - Cannot be nil or empty.
+- `not_empty` - Cannot be empty (can be nil if pointer).
+- `url` - Valid URL.
+- `max` - Maximal allowed length.
+- `oneof` - Value has to be one of specified values.
+- `uuid` - Valid UUID.
+
+
+## Proposed validation rules for Compass input types
```
PK85: If you already implemented those rules, remove a "Proposed". In case they are not implemented please go to me and we will create a  ticket for implementation.

docs/architecture/api-specs-support.md
tgorgol: What can be problematic about libraries is that UI won't be able to use them (unless there would be implementations for different languages).

akrausesap:
Also support .edmx. see https://github.com/kyma-incubator/compass/issues/409

akrausesap:
please also support "swagger" files: https://github.com/kyma-incubator/compass/issues/410

docs/architecture/api-specs-support.md
crabtree: Not sure if we should mention the Rafter here. We use it only as a backing service and in terms of our usecases we user nevert interacts with the Rafter directly.

docs/architecture/input-validation.md
```diff
@@ -0,0 +1,202 @@
+# Input Validation
+
+## Overview
+
+This document contains validation rules for all input types.
+
+## Validation rules explanation
+
+- `printable` - Printable unicode characters.
```
dbadura: Add information that `space` is printable also.

tgorgol:
Think about headers and query params, add new inputs from app templates proposal.

docs/architecture/input-validation.md
```diff
@@ -0,0 +1,202 @@
...
+### ApplicationCreateInput
+
+Field | Rules | Comment
+--- | --- | ---
+name: String! | `required`,`name` | max 36 characters so frogs are able to append suffix
+description: String | `not_empty`,`printableWithWhitespace`,`max=128` |  
+labels: Labels (map[string]interface{}) |  |  
+webhooks: [WebhookInput!] | `[required]` |  
+healthCheckURL: String | `not_empty`,`url`,`printable`,`max=256` | varchar(256) in db  
+apis: [APIDefinitionInput!] | `[required]` |
```
dbadura: What does mean `[required]`?
The array cannot be empty or something else?

docs/architecture/input-validation.md
```diff
@@ -0,0 +1,220 @@
...
+## Overview
+
+This document contains validation rules for all input types.
+
+## Validation rules explanation
+
+- `printable` - Printable unicode characters (space included).
+- `printableWithWhitespace` - Printable unicode characters and whitespace characters.
+- `name` - Up to 36 characters long. The characters allowed in names are: digits (`0`-`9`), lower case letters (`a`-`z`),`-`, and `.`. Based on Kubernetes resource name format.
+- `required` - Cannot be nil or empty.
```
dbadura: I would describe more precisely what `required` on objects mean.

As far as I understand, `required` on object means that all defined validations on object properties will be executed.

tgorgol: Not really to validate nested type you just have to specify field validation for it, even without any rules. E.g. `validation.Field(&i.Name)`.

components/director/pkg/inputvalidation/validators_test.go
```diff
@@ -0,0 +1,296 @@
...
+				require.NoError(t, err)
+			} else {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), testCase.ExpectedError.Error())
+			}
+		})
+	}
+}
+
+type TestStruct struct {
```
dbadura: Can it be moved to test definition like `testCases`?

components/director/pkg/inputvalidation/validator_fixtures.go
```diff
@@ -0,0 +1,20 @@
+package inputvalidation
```
dbadura:
```suggestion
package inputvalidation_test
```

components/director/pkg/inputvalidation/validators_string.go
```diff
@@ -0,0 +1,93 @@
...
+
+	for _, r := range s {
+		if !unicode.IsPrint(r) && !unicode.IsSpace(r) {
+			return errors.New("cannot contain not printable or whitespace characters")
+		}
+	}
+	return nil
+}
+
+func ensureInputIsString(in interface{}) (val string, isNil bool, err error) {
```
crabtree:
```suggestion
func ensureIsString(in interface{}) (val string, isNil bool, err error) {
```

docs/README.md
```diff
@@ -23,6 +23,8 @@ These documents provide you with more architectural details:
 - [Database scheme](./architecture/db-scheme.md)
 - [Authentication and authorization](./architecture/authentication-and-authorization.md)
 - [Integration Systems](./architecture/integration-systems.md)
+- [Proposal: Input validation](./architecture/input-validation.md)
+  - [Proposal: API specifications support](./architecture/api-specs-support.md)
```
crabtree:
```suggestion
 - [Proposal: API specifications support](./architecture/api-specs-support.md)
```

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
+# API Specifications Support
+
+## Overview
+
+This document describes current and planned api specification support in Kyma components.
```
crabtree:
```suggestion
This document describes current and planned API specification support in Kyma components.
```

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Currently supported api specification versions and formats
+
+Following table shows currently supported api specification versions and formats in different components.
```
crabtree:
```suggestion
The following table shows currently supported API specification versions and formats in different components.
```

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+
+Last row describes support for extracting api specifications from archive files.
```
crabtree:
```suggestion
The last row describes support for extracting API specifications from archive files.
```

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+| | Console UI & Compass UI (rendering) | Runtime Rafter (storing) |  Compass UI (uploading) | Compass Backend (storing) | Runtime Agent (storing)
+| --- | --- | --- | --- | --- | --- |
+| OpenAPI<br>(YAML, JSON) | all formats, v2 and v3 | no validation / conversion | all versions and formats, validation by property | all versions and formats, no validation | all versions, JSON only, no validation and conversion |
+| OData<br>(XML, JSON in v4) | XML only, all versions converted to v4 on client side |  no validation / conversion |  all versions, XML only, validation by parsing XML file and checking property | all versions and formats, no validation | all versions and formats, no validation and conversion |     
+| AsyncAPI<br>YAML, JSON | all formats, v2 only | validation and conversion to JSON and v2 | all versions and formats, validation by property | all versions and formats, no validation | all versions, JSON only, no validation and conversion |
+| Archives | no | yes (TAR and ZIP) | no | no | no |
+
+## Planned support for api specification versions and formats
+
+Following table shows proposed supported api specification versions and formats in different components.
```
crabtree:
```suggestion
The following table shows the proposed supported API specification versions and formats in different components.
```

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+Rendering would support only a single version of api specifications.
```
crabtree:
```suggestion
Rendering would support only a single version of API specifications.
```

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+Rafter would be responsible for validating and converting api specifications for Console UI.
```
crabtree:
```suggestion
Rafter would be responsible for validating and converting API specifications for Console UI.
```

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+Because Compass UI doesn't use Rafter conversion would have to be done in Compass Backend.
```
crabtree:
```suggestion
Because Compass UI does not use Rafter conversion would have to be done in Compass Backend.
```

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+Preferably Rafter's converters and validators should be exposed as services or libraries that could be used by Compass Backend and Compass UI.
+
+Last row describes support for extracting api specifications from archive files.
```
crabtree:
```suggestion
> **NOTE:**  The last row describes support for extracting API specifications from archive files. 
```

docs/architecture/api-specs-support.md
crabtree:
```suggestion
> **NOTE:** The last row describes support for extracting API specifications from archive files.
```

docs/architecture/input-validation.md
```diff
@@ -0,0 +1,220 @@
...
+## Proposed validation rules for Compass input types
+
+### APIDefinitionInput
+
+Field | Rules | Comment
+--- | --- | ---
+name: String! | `required`,`name`|  
+description: String | `not_empty`,`max=128`,`printableWithWhitespace` |  
+targetURL: String! | `required`,`url`,`printable`,`max=256` |  varchar(256) in db
+group: String |  `not_empty`,`printable`,`max=36` |  varchar(256) in db
```
crabtree:
```suggestion
group: String | `not_empty`, `printable`, `max=36` |  varchar(256) in db
```

docs/architecture/input-validation.md
```diff
@@ -0,0 +1,220 @@
...
+spec: APISpecInput |   | 
+version: VersionInput |   |  
+defaultAuth: AuthInput |   |  
+
+### APISpecInput
+
+- Struct validator ensuring that `type` and `format` work together (ODATA works with XML and JSON, OPEN_API works with YAML and JSON)
```
crabtree:
```suggestion
- Struct validator ensures that `type` and `format` work together (ODATA works with XML and JSON, OPEN_API works with YAML and JSON)
```

docs/architecture/input-validation.md
```diff
@@ -0,0 +1,220 @@
...
+value: String! | `required`,`printable`,`max=256` | varchar(256) in db
+deprecated: Boolean = false | `required` | required because has default value (?)
+deprecatedSince: String | `not_empty`,`printable`,`max=256` | varchar(256) in db
+forRemoval: Boolean = false | `required` | required because has default value (?)
+
+### ApplicationCreateInput
+
+Field | Rules | Comment
+--- | --- | ---
+name: String! | `required`,`name` | max 36 characters so frogs are able to append suffix
```
crabtree:
```suggestion
name: String! | `required`, `name` | max 36 characters
```

docs/architecture/input-validation.md
```diff
@@ -0,0 +1,220 @@
...
+healthCheckURL: String | `not_empty`,`url`,`printable`,`max=256` | varchar(256) in db  
+apis: [APIDefinitionInput!] | `[required]` |  
+eventAPIs: [EventAPIDefinitionInput!] | `[required]` |  
+documents: [DocumentInput!] | `[required]` |  
+
+### ApplicationUpdateInput
+
+Field | Rules | Comment
+--- | --- | ---
+name: String! | `required`,`name` | max 36 characters so frogs are able to append suffix
```
crabtree:
```suggestion
name: String! | `required`, `name` | max 36 characters
```

components/director/pkg/inputvalidation/validators_map.go
```diff
@@ -0,0 +1,64 @@
...
+
+	validation "github.com/go-ozzo/ozzo-validation"
+	"github.com/pkg/errors"
+)
+
+type mapKeyRule struct {
+	rules []validation.Rule
+}
+
+func EachMapKey(rules ...validation.Rule) *mapKeyRule {
```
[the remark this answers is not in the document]
tgorgol: The name is based on how `Each` rule was implemented in ozzo-validation library. I think it's best to keep it like that.

Example of usage in model validation:
```
return validation.ValidateStruct(&i,
		validation.Field(&i.AdditionalHeaders,
			inputvalidation.EachMapKey(validation.Required, inputvalidation.Printable),                            // key
			validation.Each(validation.Required, validation.Each(validation.Required, inputvalidation.Printable)), // value
		),
)
```

docs/README.md
tgorgol: It was supposed to be a part of the input validation hence the sublist.

components/director/pkg/inputvalidation/validators_map.go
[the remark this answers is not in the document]
crabtree: OK, maybe I'm not clear enough here. As I understand you implement the Validate func to slightly change the behavior of the Validate func that comes from the ozzo-validation library, am I right?

First thing, I don't like the name of the func EachMapKey and I suggest renaming it to `RulesToMap` or `MapRules` or maybe `Each` to correspond more with the `Each` func from ozzo-validation.

The second thing is the overall implementation. I do not see any purpose of having a new structure `mapKeyRule` with the property of type `[]validation.Rule`. You can do the following which will behave the same:

```go
type rules []validation.Rule

func Each(r ...validation.Rule) *rules {
	rm := rules(r)
	return &rm
}

func (v rules) Validate(value interface{}) error {
	errs := validation.Errors{}

	t := reflect.ValueOf(value)
	switch t.Kind() {
	case reflect.Map:
		for _, k := range t.MapKeys() {
			val := getInterface(k)
			if err := validation.Validate(val, v...); err != nil {
				errs[getString(k)] = err
			}
		}
	default:
		return errors.New("must be a map")
	}

	if len(errs) > 0 {
		return errs
	}
	return nil
}
```

And one of the test cases:

```go
		{
			Name: "Success",
			Rules: []*validation.FieldRules{
				validation.Field(&structToValidate.Map, inputvalidation.Each(validation.Required)),
			},
			ExpectedError: nil,
		},
```
```

File: components/director/pkg/inputvalidation/validators_string.go [hunk 25]
```diff
@@ -0,0 +1,93 @@
...
+
+	for _, r := range s {
+		if !unicode.IsPrint(r) {
+			return errors.New("cannot contain not printable characters")
+		}
+	}
+	return nil
+}
+
+type printableWithWhitespaceRule struct{}
```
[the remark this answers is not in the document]
crabtree: This is something minor so up to you, just saying you not need to name structures in such a descriptive way. It's not the way Go suggests.

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+
+The following table shows currently supported API specification versions and formats in different components.
+
+> **NOTE:** The last row describes support for extracting API specifications from archive files.
+
+| | Console UI & Compass UI (rendering) | Runtime Rafter (storing) |  Compass UI (uploading) | Compass Backend (storing) | Runtime Agent (storing)
+| --- | --- | --- | --- | --- | --- |
+| OpenAPI<br>(YAML, JSON) | all formats, v2 and v3 | no validation / conversion | all versions and formats, validation by property | all versions and formats, no validation | all versions, JSON only, no validation and conversion |
+| OData<br>(XML, JSON in v4) | XML only, all versions converted to v4 on client side |  no validation / conversion |  all versions, XML only, validation by parsing XML file and checking property | all versions and formats, no validation | all versions and formats, no validation and conversion |     
+| AsyncAPI<br>YAML, JSON | all formats, v2 only | validation and conversion to JSON and v2 | all versions and formats, validation by property | all versions and formats, no validation | all versions, JSON only, no validation and conversion |
```
crabtree: To make it consistent with previous rows
```suggestion
| AsyncAPI<br>(YAML, JSON) | all formats, v2 only | validation and conversion to JSON and v2 | all versions and formats, validation by property | all versions and formats, no validation | all versions, JSON only, no validation and conversion |
```

docs/architecture/api-specs-support.md
```diff
@@ -0,0 +1,35 @@
...
+
+| | Console UI & Compass UI (rendering) | Runtime Rafter (storing) |  Compass UI (uploading) | Compass Backend (storing) | Runtime Agent (storing)
+| --- | --- | --- | --- | --- |  --- |
+| OpenAPI<br>(YAML, JSON) | v3, JSON and YAML | validation and conversion to version supported by renderer | all versions and formats, validation using same logic as rafter | all versions and formats, validation and conversion using same logic as rafter | all versions and formats, no validation and conversion |
+| OData<br>(XML, JSON in v4) | v4, JSON |  validation and conversion to version supported by renderer |  all versions and formats, validation using same logic as rafter | all versions and formats, validation and conversion using same logic as rafter | all versions and formats, no validation and conversion |
+| AsyncAPI<br>YAML, JSON | v2, JSON and YAML | validation and conversion to version supported by renderer | all versions and formats, validation using same logic as rafter | all versions and formats, validation and conversion using same logic as rafter | all versions and formats, no validation and conversion |
```
crabtree: To make it consistent with previous rows
```suggestion
| AsyncAPI<br>(YAML, JSON) | v2, JSON and YAML | validation and conversion to version supported by renderer | all versions and formats, validation using same logic as rafter | all versions and formats, validation and conversion using same logic as rafter | all versions and formats, no validation and conversion |
```

PR #424, opened 2019-10-30
Add possibility to provision Azure and AWS clusters via Gardener

Maladie:
**Description**

- 

**Related issue(s)**


components/provisioner/pkg/gqlschema/schema.graphql
```diff
@@ -31,12 +31,29 @@ type GardenerConfig {
     targetProvider: String
     targetSecret: String
     diskType: String
-    zone: String
-    cidr: String
+    workercidr: String
```
janmedrek: Let's keep it consistent: `workerCidr` should be here (since `vnetCidr` is already present in Azure - specific config)

components/provisioner/pkg/gqlschema/schema.graphql
```diff
@@ -127,12 +144,33 @@ input GardenerConfigInput {
     targetProvider: String!
     targetSecret: String!
     diskType: String!
-    zone: String!
-    cidr: String!
+    workercidr: String!
```
janmedrek: Same as above, `workerCidr` would work best.

components/provisioner/internal/provisioning/converters.go
```diff
@@ -159,24 +206,56 @@ func clusterConfigToGraphQLConfig(config interface{}) gqlschema.ClusterConfig {
...
+		WorkerCidr:             &config.WorkerCidr,
+		Region:                 &config.Region,
+		AutoScalerMin:          &config.AutoScalerMin,
+		AutoScalerMax:          &config.AutoScalerMax,
+		MaxSurge:               &config.MaxSurge,
+		MaxUnavailable:         &config.MaxUnavailable,
+		ProviderSpecificConfig: providerSpecificConfig,
+	}
+}
+func providerSpecificConfigToGQLConfig(config string) gqlschema.ProviderSpecificConfig {
```
akgalwas: We decode from json to model structures. We could decode to graphql structures directly.

components/provisioner/internal/hydroform/configuration.go
```diff
@@ -47,36 +42,67 @@ func buildConfigForGCP(config model.GCPConfig, credentialsFile string) (*types.C
...
-		},
+		Type:                 types.Gardener,
+		ProjectName:          config.ProjectName,
+		CredentialsFilePath:  credentialsFile,
+		CustomConfigurations: customConfiguration,
 	}
 	return cluster, provider, nil
 }
+
+func addProviderSpecificConfig(customConfiguration map[string]interface{}, providerSpecificConfig string) (map[string]interface{}, error) {
```
akgalwas: If we look at operation flow, the following happens:
- graphQL structures are converted to model structures (provider specific config is encoded to json)
- model structures are converted to hydroform specific format (provider specific config is decoded from json)

It looks a bit strange at the first glance: first we encode to json and then decode. 
So the following questions arise: 
- should hydroform package depend on model? 
- should types defined in  model be moved to persistence package? 
  
What we could do is:
-  moving types defined in model package to persistence package. The reasoning for that is strong relation between types defined in the model package and database design.
- changing hydroform to use gqlschema package. The reasoning for that is that we will avoid additional json decoding and simplify model (GCPProviderConfig, AWSProviderConfig and AzureProviderConfig will no longer be needed)

components/provisioner/internal/provisioning/converters_test.go
```diff
@@ -128,24 +135,25 @@ func TestRuntimeConfigFromGraphQLRuntimeConfig(t *testing.T) {
...
+			Provider:               "GCP",
+			Seed:                   "gcp-eu1",
+			TargetSecret:           "secret",
+			WorkerCidr:             "cidr",
+			AutoScalerMin:          1,
+			AutoScalerMax:          5,
+			MaxSurge:               1,
+			MaxUnavailable:         2,
+			ClusterID:              "runtimeID",
+			ProviderSpecificConfig: "{\"Zone\":\"zone\"}",
```
akgalwas: We should also cover AWS and Azure providers.

Issue #413, opened 2019-10-27
Scenario name validation


PK85:
**Description**

Scenario name validation should have rules that prevent creating scenarios with characters that may be potentially unsafe.

**Reasons**

That "///// . * * " scenario name is allowed and in the future could be a potential issue.

**Attachments**

Modify existing LabelDefinition for `scenarios` with regex for enum values. Update also schema that keeps `scenarios` valid.
```
{
  "items": {
    "enum": [
      "DEFAULT"
    ],
    "type": "string"
  },
  "minItems": 1,
  "type": "array",
  "uniqueItems": true
}
```

crabtree:
Validation for enum values is done by the following regex `^[A-Za-z0-9]([-A-Za-z0-9\\s]*[A-Za-z0-9])$`

PR #427, opened 2019-10-31
Add regex pattern for scenario label enum values

crabtree:
**Description**

- Add regex pattern for scenario label enum values

**Related issue(s)**

See also #413

### DIFF
--- a/components/director/internal/model/scenarios.go
+++ b/components/director/internal/model/scenarios.go
@@ -11,8 +11,9 @@ var (
 		"minItems":    1,
 		"uniqueItems": true,
 		"items": map[string]interface{}{
-			"type": "string",
-			"enum": []string{"DEFAULT"},
+			"type":    "string",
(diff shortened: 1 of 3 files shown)


components/director/internal/model/scenarios.go
```diff
@@ -38,10 +39,14 @@ var (
 					"type": map[string]interface{}{
 						"const": "string",
 					},
+					"pattern": map[string]interface{}{
+						"const": "^[A-Za-z0-9]([-A-Za-z0-9\\s]*[A-Za-z0-9])$",
```
pkosiec: hmm, if `DEFAULT` scenario name is uppercased. Do we want to have any convention, like uppercased snake_case? Maybe we should enforce one convention for naming scenarios, like `camelCase`, `dash-case`, or  uppercased `snake_case`.

crabtree: I would not enforce the convention. I would rather go with the allowed set of characters that can be used.

PR #448, opened 2019-11-07
Define GraphQL types, models and DB Schema for Application Types

pkosiec:
**Description**

- Define GraphQL types and models for Application Types
- Add Integration System ID to Application

**Related issue(s)**


components/director/internal/domain/application/fixtures_test.go
```diff
@@ -14,7 +14,8 @@ import (
 )
 
 var (
-	testURL = "https://foo.bar"
+	testURL  = "https://foo.bar"
+	intSysID = "int-sys"
```
crabtree: maybe it is worth to put something in more uuid-like format here to indicate that this is the uuid.

components/director/internal/domain/application/repository_test.go
```diff
@@ -384,15 +384,15 @@ func TestPgRepository_ListByRuntimeScenarios(t *testing.T) {
 	}{
 		{
 			Name: "Success",
-			ExpectedApplicationRows: sqlmock.NewRows([]string{"id", "tenant_id", "name", "description", "status_condition", "status_timestamp", "healthcheck_url"}).
-				AddRow(app1ID, tenantID, "App ABC", "Description for application ABC", "INITIAL", timestamp, "http://domain.local/app1").
-				AddRow(app2ID, tenantID, "App XYZ", "Description for application XYZ", "INITIAL", timestamp, "http://domain.local/app2"),
+			ExpectedApplicationRows: sqlmock.NewRows([]string{"id", "tenant_id", "name", "description", "status_condition", "status_timestamp", "healthcheck_url", "integration_system_id"}).
+				AddRow(app1ID, tenantID, "App ABC", "Description for application ABC", "INITIAL", timestamp, "http://domain.local/app1", "test").
```
crabtree: also here, it is worth to replace test string with some more uuid-like string

PR #475, opened 2019-11-19
Add context object details to Extra map in tenant mapping handler

crabtree:
**Description**

- Add context object details to Extra map in tenant mapping handler

**Related issue(s)**


components/director/internal/tenantmapping/handler.go
```diff
@@ -23,12 +23,12 @@ type ReqDataParser interface {
 
 //go:generate mockery -name=TenantAndScopesForUserProvider -output=automock -outpkg=automock -case=underscore
 type TenantAndScopesForUserProvider interface {
-	GetTenantAndScopes(reqData ReqData, authID string) (string, string, error)
+	GetTenantAndScopes(reqData ReqData, authID string) (ObjectContext, error)
```
aszecowka: to consider: renaming interface name and its method name

components/director/internal/tenantmapping/mapper_for_system_auth.go
```diff
@@ -24,25 +24,39 @@ type mapperForSystemAuth struct {
...
+		tenant, scopes, err = m.getTenantAndScopesForApplicationOrRuntime(sysAuth, refObjType, reqData, authFlow)
+	default:
+		return ObjectContext{}, errors.Errorf("unsupported reference object type (%s)", refObjType)
+	}
+
+	if err != nil {
+		return ObjectContext{}, errors.Wrap(err, fmt.Sprintf("while fetching the tenant and scopes for object of type %s", refObjType))
+	}
+
+	objID, objType, err := getContextObj(refObjType, sysAuth)
```
aszecowka: this function name is a little bit misleading for me, because I would expect that it returns ContextObject. It looks that main purpose of this method is to return referenceObjectID, maybe we can add such method directly to SystemAuth?

components/director/internal/tenantmapping/mapper_for_system_auth.go
```diff
@@ -24,25 +24,39 @@ type mapperForSystemAuth struct {
...
+		return ObjectContext{}, errors.Wrap(err, "while getting reference object type")
 	}
 
+	var scopes string
+	var tenant string
+
 	switch refObjType {
 	case model.IntegrationSystemReference:
-		return m.getTenantAndScopesForIntegrationSystem(reqData)
+		tenant, scopes, err = m.getTenantAndScopesForIntegrationSystem(reqData)
```
dbadura: I would move this switch to private function and put `GetReferenceObjectType` and `GetReferenceObjectID` closer to each other.

Issue #419, opened 2019-10-29
Prepare proposal for securing resolvers


dbadura:
**Description**
 Prepare proposal for [Epic](https://github.com/kyma-incubator/compass/issues/306)

**Reason**
We need to limit visibility of some resources to owner only.

PK85:
@dbadura add link to https://github.com/kyma-incubator/compass/tree/master/docs README.md

PR #476, opened 2019-11-19
Update secure resolver proposal

dbadura:
**Description**

- update proposal with new names and new directive

**Related issue(s)**

See #419


dbadura:
Probably we can use the directive on field, but we need to update `gqlgen` to version **0.9.1**

crabtree:
> Probably we can use the directive on field, but we need to update `gqlgen` to version **0.9.1**

As far as I know, we are now on 0.9.3

docs/investigations/secure-resolvers/secure-auth-resolvers.md
```diff
@@ -0,0 +1,184 @@
...
+4. Runtime musn't read Applications which are not in the same scenario (`applicationsForRuntime`).
+
+### Restrict access to Application/Runtime for IntegrationSystem is which not managed
+When Integration System request specific Application/Runtime, it should be able to read only object managed by itself.
+
+## Solution
+To achieve those restrictions the following solution is proposed:
+
+### Add information about API consumer in Tenant Mapping service
+Tenant Mapping Service adds 2 items to the JWT token
```
crabtree: This is already done, so maybe you can rephrase this that it will not be added but basically just make use of those two items.

docs/investigations/secure-resolvers/secure-auth-resolvers.md
```diff
@@ -0,0 +1,184 @@
...
+This directive will limit access to application/runtime resolvers managed by integration system
+
+Proposed directive:
+```graphql
+enum managedResourceType {
+    RUNTIME
+    APPLICATION
+}
+
+directive @limitIntegrationSystemAccessTo(managedResourceType: managedResourceType!, idField: String!) on FIELD_DEFINITION
```
crabtree: how about the different name: `limitToManaged` or something similar. I somehow do not like the leaked detail like Integration System

docs/investigations/secure-resolvers/secure-auth-resolvers.md
```diff
@@ -0,0 +1,184 @@
...
+The proposed directive does following things:
+1. get object type and ID from JWT token
+2. check if object is the `integration system`
+3. if yes, then the directive check in database if given application/runtime is managed by integration system which called the query.
+If integration system doesn't manage the application/runtime, the following error is returned `Access Denied`
+
+Currently we cannot use this directive on query param ( https://github.com/99designs/gqlgen/issues/760).
+
+### Database schema change
+Every resource managed by integration system have to have column with reference to integration system table record.
```
crabtree: Why this is necessary? I would say that we need to distinguish the ID of the Integration System via the managed object but I would not say that we need to have the exact column. It can be done for example via the technical table that joins the Application with the Integration System.

dbadura: I would write sth like this:
```
Every resource managed by integration system have to be reachable from integration system point of view.
```

docs/investigations/secure-resolvers/secure-auth-resolvers.md
dbadura: `limitManagerAccessTo`?
It would look better, becasue:
`limitManagerAccessTo(Runtime,...)`

docs/investigations/secure-resolvers/secure-auth-resolvers.md
dbadura:
```suggestion
We can use information which is added in Tenant Mapping Service to JWT [PR](https://github.com/kyma-incubator/compass/pull/475):
```

docs/investigations/secure-resolvers/secure-auth-resolvers.md
crabtree: `point of view` => `perspective`

PR #482, opened 2019-11-20
Cleanup main.go file in director component

crabtree:
**Description**

- move building the connection string into a separate file
- move healthz handler into a separate file

**Related issue(s)**

### DIFF
--- a/components/director/internal/persistence/const.go
+++ b/components/director/internal/persistence/const.go
@@ -5,7 +5,6 @@ import "github.com/lib/pq"
 type persistenceCtxKey string
 
 const (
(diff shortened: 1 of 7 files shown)


components/director/internal/persistence/const.go
```diff
@@ -5,6 +5,7 @@ import "github.com/lib/pq"
 type persistenceCtxKey string
 
 const (
+	connStringf string = "host=%s port=%s user=%s password=%s dbname=%s sslmode=%s"
```
dbadura: I don't like the location of `connStringf`, because it's related to config only. It would be more readable to keep all those things together.

Issue #498, opened 2019-11-22
Fix the Labels scalar in graphql schema


kjaksik:
**Description**

Think of a way how to handle the `Labels` scalar used in `ApplicationCreateInput` type so that it will accpet keys with characters other than alphanumeric.

**Reasons**

The label naming convention that we agreed upon uses dashes as separators. Graphql schema accepts only alphanumeric characters in the `Name` type.

KarolJaksik:
Please remember to update the documentation according to our proposed fix

PK85:
closing, invalid.

Issue #460, opened 2019-11-12
Update existing create Application logic


pkosiec:
**Description**

- modify create Application method in Application service
     - check if integration system already exist while creating the application with `integrationSystemID`
     - create `integrationSystemID` (please suggest properkey name) label for Application. Do not validate it 
(do not create LabelDefinition manually)
     - pick one convention for label key naming 
- modify `updateApplication`
    - check if integration system already exist while updating the application with new `integrationSystemID`
    - update `integrationSystemID` label
 - remember about model validation from #255
- remember about integration tests

For details, please see the proposal: https://github.com/kyma-incubator/compass/pull/359

**Reason**
See #428

KarolJaksik:
We have agreed upon the label key naming convention. The key names should be all lower case with dashes as separators, eg: `integration-system-id`

PR #499, opened 2019-11-22
Update existing app creation logic

KarolJaksik:
**Description**

- Block Application creation if Integration System with given ID doesn't exist
- Block Application update if Integration System with given ID doesn't exist

**Related issue(s)**
#460 
#498


components/director/internal/domain/application/service.go
```diff
@@ -244,11 +256,16 @@ func (s *service) Create(ctx context.Context, in model.ApplicationCreateInput) (
 		in.Labels[model.ScenariosKey] = model.ScenariosDefaultValue
 	}
 
+	if in.IntegrationSystemID != nil {
```
dbadura: What do you think about this:
```
in.Labels["integration-system-id"] = ""
if in.IntegrationSystemID != nil {
   in.Labels["integration-system-id"] = *in.IntegrationSystemID
}
```
Also consider extracting key to const.

Issue #490, opened 2019-11-21
Add status message to Clean Up Runtime Data mutation response


majakurcius:
**Description**

Currently the response to the cleanupRuntimeData mutation on success returns only the ID of the Runtime to wipe. There is no information on whether the Runtime was actually wiped from the database. 
The desired state is to have it return a status message along with the ID, which would inform the User whether the given Runtime was actually deleted or no Runtime of this ID was found. 

**Reasons**

If User makes a typo in the Runtime ID or the ID is incorrect for some other reason and they pass an ID of a non-existent Runtime, they have no way of knowing that as the operation is successful, it just does not have anything to delete. User should get a clear status of whether they actually wiped the given Runtime. 

**Attachments**

Clean Up Runtime Data mutation: 

```graphql
mutation { cleanupRuntimeData(id: "61d1841b-ccb5-44ed-a9ec-45f70cd1b0d3")}
```

Current response:
```graphql
{ "data": {
  "cleanupRuntimeData": "61d1841b-ccb5-44ed-a9ec-45f70cd1b0d3"
}}
```

Example desired response:

- If given Runtime exists in the database:
```graphql
{ "data": {
  "cleanupRuntimeData": {
    "runtimeID" : "61d1841b-ccb5-44ed-a9ec-45f70cd1b0d3"
    "statusMessage" : "Runtime data cleanup successful."
}}}
```

- If no Runtime of given ID found in the database:
```graphql
{ "data": {
  "cleanupRuntimeData": {
    "runtimeID" : "61d1841b-ccb5-44ed-a9ec-45f70cd1b0d3"
    "statusMessage" : "No such Runtime found."
}}}
```

PR #503, opened 2019-11-25
Add status message to Clean Up Runtime Data mutation response

majakurcius:
**Description**

- Add status message to Clean Up Runtime Data mutation response

**Related issue(s)**
See also #490

### DIFF
--- a/components/provisioner/internal/api/resolver.go
+++ b/components/provisioner/internal/api/resolver.go
@@ -105,7 +105,7 @@ func (r *Resolver) RuntimeOperationStatus(ctx context.Context, operationID strin
 	return status, nil
 }
 
-func (r *Resolver) CleanupRuntimeData(ctx context.Context, id string) (*gqlschema.CleanUpRuntimeStatus, error) {
+func (r *Resolver) CleanupRuntimeData(ctx context.Context, id string) (*gqlschema.CleanUpRuntimeDataResult, error) {
 	log.Infof("Requested cleaning up Runtime data for Runtime %s.", id)
 
 	res, err := r.provisioning.CleanupRuntimeData(id)
--- a/components/provisioner/internal/api/resolver.go
+++ b/components/provisioner/internal/api/resolver.go
@@ -95,7 +95,7 @@ func (r *Resolver) RuntimeOperationStatus(ctx context.Context, operationID strin
 	return status, err
 }
 
-func (r *Resolver) CleanupRuntimeData(ctx context.Context, id string) (string, error) {
+func (r *Resolver) CleanupRuntimeData(ctx context.Context, id string) (*gqlschema.CleanUpRuntimeStatus, error) {
 	res, err := r.provisioning.CleanupRuntimeData(id)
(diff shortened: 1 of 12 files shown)


components/provisioner/pkg/gqlschema/schema.graphql
```diff
@@ -117,6 +117,11 @@ enum RuntimeAgentConnectionStatus {
     Disconnected
 }
 
+type CleanUpRuntimeStatus {
+    id: String!
+    message: String
```
Szymongib: I think naming could be a little more accurate. I am not sure if `Status` suffix is appropriate here.

components/provisioner/internal/provisioning/service.go
```diff
@@ -167,8 +167,24 @@ func (r *service) RuntimeOperationStatus(operationID string) (*gqlschema.Operati
...
+
+	if err != nil {
+		if err.Code() == dberrors.CodeNotFound {
+			message := fmt.Sprintf("Runtime with ID %s not found in the database", id)
+			return &gqlschema.CleanUpRuntimeStatus{ID: id, Message: &message}, nil
+		}
+
+		if err.Code() == dberrors.CodeInternal {
+			message := fmt.Sprintf("Could not clean data for Runtime with ID %s: %s", id, err)
+			log.Error(message)
```
Szymongib: We already log the error in the Resolver, no need to do it twice :)

components/provisioner/internal/api/resolver.go
```diff
@@ -95,7 +95,7 @@ func (r *Resolver) RuntimeOperationStatus(ctx context.Context, operationID strin
 	return status, err
 }
 
-func (r *Resolver) CleanupRuntimeData(ctx context.Context, id string) (string, error) {
```
Szymongib: I would suggest adding some logging here so that we can know when the operation was called for which Runtime.

docs/provisioner/01-01-provisioner-overview.md
```diff
@@ -16,7 +16,7 @@ It is powered by [Hydroform](https://github.com/kyma-incubator/hydroform) and it
     
 Note that the operations of provisioning and deprovisioning are asynchronous. They return the operation ID, which you can use to [check the Runtime Operation Status](08-03-runtime-operation-status.md).
 
-The Runtime Provisioner also allows you to [clean up Runtime data](08-06-clean-up-runtime-data.md). This operation removes a given Runtime and all its data from the database and frees up the Runtime ID for reuse. It is useful when your cluster has died or when the operation of deprovisioning has failed.
+The Runtime Provisioner also allows you to [clean up Runtime data](08-06-clean-up-runtime-data.md). This operation removes the data for a given Runtime from the database and frees up the Runtime ID for reuse. It is useful when your cluster has died or when the operation of deprovisioning has failed.
```
Szymongib: I would suggest adding some `Caution` section informing the user that this operation does not trigger Runtime deprovisioning and cluster might still exist after running it.

PR #529, opened 2019-12-03
Proposed interface for credentials management

akgalwas:
**Description**

- Interface for handling credentials

**Related issue(s)**

### DIFF
--- a/components/provisioner/internal/credentials/provider.go
+++ b/components/provisioner/internal/credentials/provider.go
@@ -1,0 +1,42 @@
+package credentials
+
+type TargetProvider string
+type Kubeconfig []byte
+type ServiceAccount []byte
+
+const (
+	GCP   TargetProvider = "GCP"
+	Azure TargetProvider = "Azure"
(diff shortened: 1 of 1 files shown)


components/provisioner/internal/credentials/provider.go
```diff
@@ -0,0 +1,38 @@
...
+	GCP   TargetProvider = "GCP"
+	Azure TargetProvider = "Azure"
+	AWS   TargetProvider = "AWS"
+)
+
+type Kubeconfig []byte
+
+type GardenerCredentials struct {
+	ClusterCredentials Kubeconfig
+	SecretName         string
```
rafalpotempa: Shouldn't we use CredentialsSecretName instead, to be consistent with current naming? Maybe these Credentials are sth different from those required for provisioning?

```suggestion
	CredentialsSecretName         string
```

rafalpotempa: All clear :)

Issue #492, opened 2019-11-21
Make compass playground available


PK85:
**Description**
Update OathKeeper configuration for Director .

**Reasons**
- Tenant mapping handler is an internal endpoint.
- GraphQL Playground is a very convenient tool for using Director, which has been available before introducing OathKeeper.

**AC**
- make compass playground available without credentials, then user needs to use credentials to execute calls
- Do not expose tenant mapping handler for OAuth flow (for other flows it is already not exported)
- Expose GraphQL Playground for Director (GET request on Gateway's /director endpoint)
   -  for OAuth 2.0, certs and JWT hosts
   -  make it unsecured
- add a simple e2e test, that endpoint is not secured

pkosiec:
This is a duplicate of https://github.com/kyma-incubator/compass/issues/416

crabtree:
@PK85 What does it mean?

> * Do not expose tenant mapping handler for OAuth flow (for other flows it is already not exported)

crabtree:
One more thing, regarding the endpoint secured with certs. In this case (mtls) we have the istio gateway configured which will not allow sending calls without client certificate present in the request. So even sending the GET request for the playground page the client certificate must be present.

pkosiec:
@crabtree yes.
See the line https://github.com/kyma-incubator/compass/blob/master/chart/compass/charts/gateway/templates/oathkeeper-rules.yaml#L11 - we're exporting all paths under gateway, so we should create more strict rules

> One more thing, regarding the endpoint secured with certs. In this case (mtls) we have the istio gateway configured which will not allow sending calls without client certificate present in the request. So even sending the GET request for the playground page the client certificate must be present.

I confirmed it with @PK85  - it is okay for us, as long as playground will be exported as well (requiring client certificate though).

So, we should create specific OathKeeper rules for the following endpoints:
- 3 x OathKeeper rule for POST `/director/graphql` - secured endpoint (for JWT, OAuth 2.0, Client certificate)
- 3 x OathKeeper rule for GET `/director` - unsecured 
- remove all `/*`rules as we should have strict ones

crabtree:
@pkosiec ok, this was exactly what I was asking for, thanks for the clarification. The only difference here will be regarding the client certificates. The client certificate needs to be present in the request for accessing the playground (it is required for mtsl handshake). In the OAuth 2 and JWT you will not need to set the token in order do reach the playground (only the actual calls to the GraphQL API will need to have the OAuth 2 or JWT tokens).

PR #530, opened 2019-12-03
Expose GraphQL examples for Director playground

pkosiec:
**Description**

- Expose GraphQL examples for Director Playground

**Related issue(s)**
#492


tests/end-to-end/e2e/playground_test.go
```diff
@@ -9,80 +9,111 @@ import (
...
-		DirectorURLformat   = "https://%s.%s/director"
-		OAuth2Subdomain     = "compass-gateway-auth-oauth"
-		JWTSubdomain        = "compass-gateway"
-		ClientCertSubdomain = "compass-gateway-mtls"
-	)
+type playgroundTestConfig struct {
+	Gateway struct {
+		Domain               string `envconfig:"DOMAIN"`
+		JWTSubdomain         string
+		OAuth20Subdomain     string `envconfig:"GATEWAY_OAUTH2_SUBDOMAIN"`
```
tgorgol: I think it'd be best to change this field to use the same name as env variable or add `envconfig` tag to other config fields. As it is now it can be a bit confusing why only one of these fields has specified tag.

tests/end-to-end/e2e/playground_test.go
```diff
@@ -9,80 +9,111 @@ import (
...
 )
 
-func TestDirectorPlaygroundAccess(t *testing.T) {
-	const (
-		DirectorURLformat   = "https://%s.%s/director"
-		OAuth2Subdomain     = "compass-gateway-auth-oauth"
-		JWTSubdomain        = "compass-gateway"
-		ClientCertSubdomain = "compass-gateway-mtls"
-	)
+type playgroundTestConfig struct {
```
tgorgol: Just a suggestion but I think it would be better to move all the boilerplate to the file with url builder and leave only Tests* here. At first look it's difficult to spot where actual tests start.

tests/end-to-end/e2e/playground_test.go
pkosiec: Unfortunately, we have to keep it for `GATEWAY_OAUTH20_SUBDOMAIN`, because envconfig doesn't detect this variable automatically.

PR #540, opened 2019-12-06
App tmpl integration test

dbadura:
**Description**

- Add missing integration tests for `RegisterApplicationFromTemplate`
- Add validation test for `ApplicationFromTemplateInput`

**Related issue(s)**


tests/end-to-end/director/testctx.go
```diff
@@ -38,7 +38,7 @@ type testContext struct {
 	cli               *gcli.Client
 }
 
-const defaultScopes = "runtime:write application:write label_definition:write integration_system:write"
+const defaultScopes = "application_template:write runtime:write application:write label_definition:write integration_system:write"
```
pkosiec: what about `application_template:read`? Why don't we use it?

dbadura: Because we need to create application template in tests.

Issue #457, opened 2019-11-12
Tenant mapping handler should always return success status code


crabtree:
**Description**

Tenant mapping handler should always return success status code, otherwise, hydrator retries the calls which cause poor user experience during request processing.

**AC**
- Unauthorized call for director must return 403 Forbidden (return it on Director middleware)
- add new e2e test (as app1(client credentials) in tenant foo try to create app2 in tenant bar which results with 403 status code from Director authenticator)

PR #547, opened 2019-12-10
Return success response for Tenant Mapping Handler all the time

pkosiec:
**Description**

- Return success response for Tenant Mapping Handler all times

**Related issue(s)**
#457


components/director/internal/authenticator/claims.go
```diff
@@ -1,20 +1,38 @@
...
-	"errors"
-
 	"github.com/dgrijalva/jwt-go"
+	"github.com/pkg/errors"
 )
 
+type invalidTenantError struct{}
+
+func (err *invalidTenantError) Error() string {
+	return "Invalid tenant"
```
crabtree: shouldn't it be lowercase?

components/director/internal/authenticator/middleware.go
```diff
@@ -49,17 +49,18 @@ func (a *Authenticator) Handler() func(next http.Handler) http.Handler {
...
 			if err != nil {
 				wrappedErr := errors.Wrap(err, "while parsing token")
 				log.Error(wrappedErr)
-				http.Error(w, wrappedErr.Error(), http.StatusUnauthorized)
-				return
-			}
 
-			if !token.Valid {
-				http.Error(w, "Invalid token", http.StatusUnauthorized)
+				vErr, ok := err.(*jwt.ValidationError)
```
crabtree: It just can be handled by the `isInvalidTenantError` func.

components/director/internal/authenticator/middleware.go
```diff
@@ -49,17 +49,18 @@ func (a *Authenticator) Handler() func(next http.Handler) http.Handler {
 			}
 
 			claims := Claims{}
-
-			token, err := jwt.ParseWithClaims(bearerToken, &claims, a.getKeyFunc())
+			_, err = jwt.ParseWithClaims(bearerToken, &claims, a.getKeyFunc())
 			if err != nil {
```
crabtree: This error check can be combined with line 52:

```go
if _, err = jwt.ParseWithClaims(bearerToken, &claims, a.getKeyFunc()); err != nil {
```

components/director/internal/authenticator/middleware_test.go
```diff
@@ -98,7 +98,27 @@ func TestAuthenticator_Handler(t *testing.T) {
...
 		assert.Equal(t, http.StatusOK, rr.Code)
+	})
+
+	t.Run("Error - forbidden when tenant is empty", func(t *testing.T) {
+		//given
+		middleware := createMiddleware(t, true)
+		handler := testHandler(t, tnt, scopes)
+		rr := httptest.NewRecorder()
+
+		req, err := http.NewRequest("GET", "/", nil)
```
crabtree: This code repeats in many test cases, maybe it is worth to extract it to some helper func?

For example:
```go
func testRequest(t *testing.T) *http.Request {
    req, err := http.NewRequest("GET", "/", nil)
    require.NoError(t, err)

    return req
}
```

components/director/internal/authenticator/middleware.go
pkosiec: Yup, but in this case I think it is more readable like it is, as the `ParseWithClaims` function call is quite complicated.

components/director/internal/authenticator/middleware.go
pkosiec: Nope, it can't, as it is custom error type and I want to compare types, not error messages. I'm using `isInvalidTenantError` line below, on inner validation error.

crabtree: Maybe then you can wrap this `vErr, ok := err.(*jwt.ValidationError)` into func isJWTValidationError?

pkosiec: Hmm... It will be hard to do so, as I need inner error of the `Validation Error`, so the function would also return two arguments (like `convertToValidationError`). I think there is no benefit to do so.

tests/end-to-end/e2e/fixtures_queries.go
```diff
@@ -11,13 +11,18 @@ import (
 )
 
 //Application
-func createApplicationFromInputWithinTenant(t *testing.T, ctx context.Context, gqlClient *gcli.Client, tenant string, in graphql.ApplicationCreateInput) graphql.ApplicationExt {
+func createApplicationWithinTenant(t *testing.T, ctx context.Context, gqlClient *gcli.Client, tenant string, in graphql.ApplicationCreateInput) (graphql.ApplicationExt, error) {
```
dbadura: I would move this to your test case to private method.
Because here, we only create resources successfully.
And I think that error is a part of your logic and it should be there, but maybe I am wrong.

pkosiec: Hmm, but it would mean that I need to duplicate code - as this is not `_test` file. I will keep it as it is. I will move it to the bottom of this file.

PR #581, opened 2019-12-18
Update Compass documentation

majakurcius:
**Description**

- Unify Runtime Agent naming in the Compass documentation
- Archive outdated documentation that we want to keep


docs/architecture/provisioning-api.md
```diff
@@ -93,9 +93,9 @@ The mutation returns OperationID allowing to retrieve the operation status.
 
 The mutation returns OperationID allowing to retrieve the operation status.
 
-### Reconnecting Compass Runtime Agent mutation
+### Reconnecting Runtime Agent mutation
 
-***reconnectRuntimeAgent*** mutation reconnects Compass Runtime Agent. Pass the RuntimeID as argument. 
+***reconnectRuntimeAgent*** mutation reconnects the Runtime Agent. Pass the RuntimeID as argument.
```
Szymongib: Here we are mentioning specific Runtime Agent in Kyma as Provisioner supports only Kyma Runtimes, so I wonder if we should not be specific to the Kyma Runtime Agent.

Issue #590, opened 2019-12-20
Simplify Compass API for paired Application, Integration System and Runtime


pkosiec:
**Description**

Let assume the following scenario:
1. The user creates in UI application
2. The user generates a one-time token and provides it to application
3. The application generates certificate used for communication with Director
4. The application wants to add API using the following mutation:
```
addAPI(applicationID: ID!, in: APIDefinitionInput!): APIDefinition!
```

The drawback in this approach is the following: Application has to know it's own `applicationID` which is inconvenient. 
Compass can distinguish Application on the basis of information stored in the certificate.

**AC**
- Implement viewer query according to proposal https://github.com/kyma-incubator/compass/pull/551
- Test all cases in integration tests, to make sure it works properly

PR #593, opened 2019-12-24
Add graphql viewer pattern to schema

dbadura:
**Description**

- add new query to graphql

See #590 

**Related issue(s)**

Tests:
https://github.com/kyma-incubator/compass/pull/607


components/director/pkg/graphql/viewer_converter.go
```diff
@@ -0,0 +1,22 @@
...
+		return &Viewer{ID: cons.ConsumerID, Type: ViewerTypeRuntime}, nil
+	case consumer.Application:
+		return &Viewer{ID: cons.ConsumerID, Type: ViewerTypeApplication}, nil
+	case consumer.IntegrationSystem:
+		return &Viewer{ID: cons.ConsumerID, Type: ViewerTypeIntegrationSystem}, nil
+	case consumer.User:
+		return &Viewer{ID: cons.ConsumerID, Type: ViewerTypeUser}, nil
+	}
+
+	return nil, errors.New("Viewer not exist")
```
kfurgol:
```suggestion
	return nil, errors.New("viewer does not exist")
```

according to https://github.com/golang/go/wiki/CodeReviewComments#error-strings

components/director/pkg/graphql/viewer_converter.go
```diff
@@ -0,0 +1,22 @@
+package graphql
```
kfurgol: I think that converter should be in `viewer` package, not `graphql`.

components/director/internal/domain/viewer/resolver_test.go
```diff
@@ -0,0 +1,67 @@
...
+		ctx = consumer.SaveToContext(ctx, invalidConsumer)
+		resolver := viewer.NewViewerResolver()
+		//WHEN
+		_, err := resolver.Viewer(ctx)
+		//THEN
+		require.Error(t, err)
+		assert.EqualError(t, err, "Viewer not exist")
+	})
+
+	t.Run("No consumer in ctx ", func(t *testing.T) {
```
kfurgol:
```suggestion
	t.Run("No consumer in ctx", func(t *testing.T) {
```

components/director/internal/domain/viewer/resolver_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+		//WHEN
+		vwr, err := resolver.Viewer(ctx)
+		//THEN
+		require.NoError(t, err)
+		require.NotNil(t, vwr)
+		assert.Equal(t, expectedViewer, *vwr)
+	})
+
+	t.Run("error while converting", func(t *testing.T) {
```
kfurgol:
```suggestion
	t.Run("Error while converting", func(t *testing.T) {
```

components/director/internal/domain/viewer/converter.go
```diff
@@ -0,0 +1,23 @@
+package viewer
+
+import (
+	"github.com/kyma-incubator/compass/components/director/internal/consumer"
+	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
+	"github.com/pkg/errors"
+)
+
+func ToViewer(cons consumer.Consumer) (*graphql.Viewer, error) {
```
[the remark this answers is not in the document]
dbadura: The first one , `SystemAuthReferenceObjectType` is used in SystemAuth logic. Additionally we cannot add `USER`, because it will introduce confusion.

`Consumer` type is used in TenantMappingService and it describe who is using our API. I would call it `input` type. 

`Viewer` type is graphql type and it's only used in Viewer. I treat it as `response` type.

From my perspective I would keep it as it's because those types are used in different contexts. Additionally those different types, describe precisely how they should be used and what they represent.
I would look at those types, like `application` entity. We have `ApplicationRegisterInput` and `Application`.
I am aware that this is duplication of code.

Issue #486, opened 2019-11-21
Service Class from APIs defined in the Compass has empty Provider field


aszecowka:
**Problem**: How I can provide information that given API is provided by given company?

So far, runtime agent has following implementation:
```
ProviderDisplayName: "", // not available in the Director's API
```
kyma/components/compass-runtime-agent/internal/kyma/applications/converter.go:107

Information about the provider should be stored as a root attribute of the Application type. 

**AC**
- check how provider is implmeneted in the ApplicationCR
- copy solution to compass Application
- agent fetch that information and sets provider in the ApplicationCR
- UI can display the provider value (populate provider in runtime agent)
- modify integration tests

dbadura:
reestimated from 2 to 5 (bump in kyma, changes in runtimeagent and in console UI)

PR #595, opened 2019-12-30
Rename provider display name field

dbadura:
**Description**

- rename field `providerDisplayName` to `providerName`.
- test are available here: #597 

**Related issue(s)**

See #486

### DIFF
--- a/components/schema-migrator/migrations/201912131700_add_providername_app.up.sql
+++ b/components/schema-migrator/migrations/201912131700_add_providername_app.up.sql
@@ -1,2 +1,8 @@
 ALTER TABLE applications
-    ADD COLUMN provider_display_name varchar(256) NOT NULL;
+    ADD COLUMN provider_name varchar(256);
+
+UPDATE APPLICATIONS AS A
+SET provider_name = name;
+
+ALTER TABLE applications
+    ALTER COLUMN provider_name SET NOT NULL;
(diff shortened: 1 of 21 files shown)


components/schema-migrator/migrations/201912131700_add_providername_app.up.sql
```diff
@@ -1,2 +1,8 @@
 ALTER TABLE applications
-    ADD COLUMN provider_display_name varchar(256) NOT NULL;
```
aszecowka: create separate migration for renaming.
If someone has this migration (201912131700_add_providername_app.up.sql), it has column provider_display_name and provider_name is missing, but migration won't be executed.

components/schema-migrator/migrations/201912131700_add_providername_app.up.sql
```diff
@@ -1,2 +1,8 @@
...
+    ADD COLUMN provider_name varchar(256);
+
+UPDATE APPLICATIONS AS A
```
aszecowka: please check if alias `A` can be omitted

components/schema-migrator/migrations/201912131700_add_providername_app.up.sql
dbadura: I've got question to 2 cases:
 1. somebody has en empty database, so the migration is successful.
 2. somebody has something in database, in this case the old migration will fail.

How we should apply fixing migration?

Changing current migration causes:
1. If somebody has empty database, he has invalid database schema.
2. If somebody has sth in database, he will reapply this migration and the database schema will be in good state.

Adding new migration causes:
1. if somebody has empty database, after migrations the database will have valid schema
2. if somebody has data in db, he won't execute this migration, because the old faulty will fail.

PR #604, opened 2020-01-03
Reorganize Director tests

aszecowka:
**Description**

- rename end-to-end test to director tests 
- director and director-tests use go modules instead of dep
- director-tests still uses vendored director - because of the limitations of generic makefiles. If we would like to have a dependency to director, we need to modify go.mod with replace statement like this:
`replace github.com/kyma-incubator/compass/components/director => ./../../components/director`. 
Unfortunately, build steps are executed in the container that do not have whole compass codebase, but only director-test, so we cannot access  `./../../components`  
- remove dependency to connector - instead copied 4 structs


tests/director-tests/auth-related-tests/README.md
```diff
@@ -1,4 +1,4 @@
-# Compass end to end tests
+#  Director auth tests
```
klaudiagrz:
```suggestion
#  Director authentication tests
```

tests/director-tests/auth-related-tests/README.md
```diff
@@ -1,4 +1,4 @@
...
 
 Compass end to end test scenario consists of the following steps:
```
klaudiagrz:
```suggestion
Director authentication test scenario consists of the following steps:
```

tests/director-tests/auth-related-tests/README.md
```diff
@@ -1,4 +1,4 @@
...
 - Get a Dex token, register the Integration System using this token, and request client credentials for the IntegrationSystem
```
klaudiagrz: No changes in the steps below?

components/director/Makefile
```diff
@@ -8,3 +8,15 @@ include $(SCRIPTS_DIR)/generic_make_go.mk
 VERIFY_IGNORE := /vendor\|/automock\|/testdata
 
 verify:: gqlgen check-gqlgen
+
+errcheck-local:
+	 GO111MODULE="on" errcheck -blank -asserts -ignoregenerated ./...
```
dbadura: Can we set GO111MODULe=on at the top on makefile and then we have to define it once.

components/director/Dockerfile
```diff
@@ -7,17 +7,11 @@ WORKDIR ${BASE_APP_DIR}
...
 
-COPY ./internal/ ${BASE_APP_DIR}/internal/
-COPY ./pkg/ ${BASE_APP_DIR}/pkg/
-COPY ./vendor/ ${BASE_APP_DIR}/vendor/
-COPY ./cmd/ ${BASE_APP_DIR}/cmd/
-COPY ./hack/ ${BASE_APP_DIR}/hack/
-COPY ./licenses/ ${BASE_APP_DIR}/licenses/
-
-#
+COPY . .
```
pkosiec: Then we copy not only needed files, but also scripts, makefile, dockerfile and gitignore. I would keep it as it is, as the list here won't change anywhere soon.

chart/compass/templates/director-auth-test.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: "testing.kyma-project.io/v1alpha1"
 kind: TestDefinition
 metadata:
-  name: {{ .Chart.Name }}-e2e
+  name: {{ .Chart.Name }}-director-auth-related-tests
```
pkosiec: Hmm, we have there tests for:
- different auth methods
- accessing playground
- tenant mapping handler
- viewer query

Basically, these are tests, that are executed through gateway. So, what do you think about the name `-director-gateway-integration` or something like that?

components/director/Dockerfile
```diff
@@ -7,17 +7,11 @@ WORKDIR ${BASE_APP_DIR}
...
 # Build app
 #
 
+RUN go version
```
pkosiec: is this really necessary?

components/director/Makefile
```diff
@@ -8,3 +8,15 @@ include $(SCRIPTS_DIR)/generic_make_go.mk
...
+
+test-local:
+	GO111MODULE=on go test ./...
+
+resolve-local: ;
```
pkosiec: I think it's bad to have it empty, we can easily specify that resolving deps is `go mod download`. You can add a message that this project uses Go modules and this step is not necessary. Eventually the target will be removed, when we move all components to Go modules.

pkosiec: Similarly with `ensure-local`

components/director/Makefile
```diff
@@ -8,3 +8,15 @@ include $(SCRIPTS_DIR)/generic_make_go.mk
...
+
+ensure-local: ;
+
+dep-status-local: ;
```
pkosiec: `go mod graph`? Or just echo a message that this project uses Go modules and this step is skipped. Eventually the target will be removed, when we move all components to Go modules.

chart/compass/templates/director-api-test.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: "testing.kyma-project.io/v1alpha1"
 kind: TestDefinition
 metadata:
-  name: {{ .Chart.Name }}-director
+  name: {{ .Chart.Name }}-director-api-tests
```
pkosiec: Remove suffix `-tests`, as TestDefinition kind with this name is self-explanatory.

I would keep `-director` as it is

tests/director-tests/Makefile
```diff
@@ -1 +1,51 @@
-release: ;
```
pkosiec: Rename `tests/directors-tests` directory to simply `tests/director`. Other components should follow the convention.

What we have right now is against a very old approved proposal: https://github.com/kyma-project/community/blob/master/collaboration/sig-core/proposals/test-folder-consolidation.md

tests/director-tests/Makefile
```diff
@@ -1 +1,51 @@
...
+.PHONY: generate-director-examples clean-up
+
+VERIFY_IGNORE := /vendor\|/automock\|/testdata
+
+verify:: generate-director-examples
+
+$(eval $(call test-local))
+
+# We have to override test-local, because we need to run director with database as docker containers and connected with custom network
+# and the container itsefl has to be connected to the network
```
pkosiec: itsefl -> itself

tests/director-tests/Makefile
```diff
@@ -1 +1,51 @@
...
+test-local: ;
```
pkosiec: Can we run here the `run.sh` script?

components/director/Dockerfile
aszecowka: but we copy to builder image, not to the destination image

tests/director-tests/Makefile
aszecowka: run.sh requires Director to be up and running, `test-local` can work in such case, but `make test` will always fail, because it will run it in the Docker container. IMO it is better to keep it as it is.

components/director/Dockerfile
aszecowka: no, removed

tests/director/gateway-integration-tests/README.md
```diff
@@ -1,6 +1,13 @@
-# Compass end to end tests
+#  Gateway Integration tests
 
-Compass end to end test scenario consists of the following steps:
+Gateway Integration tests are a collection of the following tests:
+- Playground tests check if Playground is available through Gateway
```
klaudiagrz:
```suggestion
- Playground tests check if the playground is available through Gateway
```

tests/director/gateway-integration-tests/README.md
```diff
@@ -1,6 +1,13 @@
...
+- Viewer tests check the correctness of the `viewer` mutation
+- Tenant Mapping Handler checks tenant separation
+- Compass Authentication checks different access methods to Director via Gateway
```
klaudiagrz:
```suggestion
- Compass Authentication checks different methods of access to Director via Gateway
```

tests/director/gateway-integration-tests/README.md
```diff
@@ -24,4 +31,5 @@ To run the test locally, set these environment variables:
 | GATEWAY_JWTSUBDOMAIN | Default gateway for handling requests with a JWT | compass-gateway |
 | GATEWAY_CLIENT_CERTS_SUBDOMAIN | Default gateway for handling requests with a certificate | compass-gateway-mtls |
 | GATEWAY_OAUTH20_SUBDOMAIN | Default gateway for handling requests with an OAuth access token | compass-gateway-auth-oauth|
-Then run `go test ./... -count=1 -v` inside `./tests/end-to-end/e2e` directory.
+
+Then run `go test ./... -count=1 -v` inside `./tests/director/gateway-integration-tests` directory.
```
klaudiagrz:
```suggestion
Then run `go test ./... -count=1 -v` inside the `./tests/director/gateway-integration-tests` directory.
```

components/director/Dockerfile
pkosiec: Fair enough, you can leave it as it is

pkosiec: Yup,  but it will be slightly slower, while copying local binary and other stuff... but anyway, it's not a blocker of this PR

tests/director/api-tests/fixtures_request_test.go
```diff
@@ -1,12 +1,12 @@
-package director
```
pkosiec: Please consider omitting the `-tests` suffix for `api-tests` and `gateway-integration-tests` directories

PR #610, opened 2020-01-05
Implement storage for KEB

polskikiel:
**Description**

- Implemented storage for KEB for the instances including their provisioning parameters.

**Related issue(s)**


components/kyma-environment-broker/internal/storage/dbsession/read.go
```diff
@@ -0,0 +1,30 @@
...
+	session *dbr.Session
+}
+
+func (r readSession) GetInstanceByID(instanceID string) (internal.Instance, dberr.Error) {
+	var instance internal.Instance
+
+	err := r.session.
+		Select("*").
+		From(schema.InstancesTableName).
+		Where(dbr.Eq(schema.InstancesTableName+".instance_id", instanceID)).
```
jasiu001: IMO table name is not neccessary 
see https://godoc.org/github.com/gocraft/dbr#Eq

components/kyma-environment-broker/internal/storage/service/test/instance_test.go
```diff
@@ -0,0 +1,61 @@
+package test
```
jasiu001: It also should be in separate dir with tests

components/kyma-environment-broker/internal/broker/broker.go
```diff
@@ -146,6 +132,19 @@ func (b *KymaEnvBroker) Provision(ctx context.Context, instanceID string, detail
...
 	}
 
+	err = b.Storage.Instances().Insert(internal.Instance{
+		InstanceID:             instanceID,
+		RuntimeID:              "",
+		GlobalAccountID:        ersContext.GlobalaccountID,
+		ServiceID:              details.ServiceID,
+		ServicePlanID:          details.PlanID,
+		DashboardURL:           fixedDummyURL,
+		ProvisioningParameters: string(details.RawParameters),
```
jasiu001: let's discuss `subaccount_id` and SM parameters, should we save it also?

components/kyma-environment-broker/internal/broker/broker.go
```diff
@@ -146,6 +132,19 @@ func (b *KymaEnvBroker) Provision(ctx context.Context, instanceID string, detail
 		SecretName: b.Config.SecretName,
 	}
 
+	err = b.Storage.Instances().Insert(internal.Instance{
+		InstanceID:             instanceID,
+		RuntimeID:              "",
```
jasiu001: In table schema, this value cannot be null

polskikiel: discussed, I will set the `runtimeID` after the response from the provisioner

components/kyma-environment-broker/internal/broker/broker.go
polskikiel: discussed, we don't need them

PR #622, opened 2020-01-10
Set Provider Name as optional

dbadura:
**Description**

- Set provider name as optional in Application domain.

**Related issue(s)**

See #486 
Updated tests: https://github.com/kyma-incubator/compass/pull/624

### DIFF
--- a/components/director/pkg/graphql/application.go
+++ b/components/director/pkg/graphql/application.go
@@ -3,7 +3,7 @@ package graphql
 type Application struct {
 	ID                  string             `json:"id"`
 	Name                string             `json:"name"`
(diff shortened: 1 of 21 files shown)


components/director/pkg/graphql/application.go
```diff
@@ -3,7 +3,7 @@ package graphql
 type Application struct {
 	ID                  string             `json:"id"`
 	Name                string             `json:"name"`
-	ProviderName        string             `json:"ProviderName"`
+	ProviderName        *string            `json:"ProviderName"`
```
pkosiec: why this is from passed as `ProviderName` in JSON? It should be `providerName`

Issue #620, opened 2020-01-08
Create compatibility layer for REST Connectors to work seamlessly with Compass


PK85:
**Description**
Compass must work with REST Connectors (e.g. Wordpress Application). Take into account all findings from our [Investigation](https://github.com/kyma-incubator/compass/pull/612).

**Reasons**

We'd like to give more time for all existing REST connectors to be reimplemented to use new GraphQL APIs.

**AC**
- implement Application Connector and Applications Registry REST APIs in Compass and make sure they work e2e from External Application point of view.
- run existing integration test for REST API and ensure that everything works as expected
- test REST APIs with varkes
- test with existing commerce instance
- mark REST APIs as deprecated

**Related tickets**
- [Support Application Connector service REST API in the API-gateway](https://github.com/kyma-incubator/compass/issues/18)
- [Support Metadata service REST API in the API-gateway](https://github.com/kyma-project/kyma/issues/4240)

aszecowka:
Proposal for tasks split:
- [ ] Scaffold project structure
  - [ ] create Go project in /components/compatibility-layer with Makefile, Dockerfile, consider using go modules
  - [ ] create Go project in /tests/comatibility-layer-tests
  - [ ] helm chart
  - [ ] configure Prow Jobs
- [ ] PoC about propagating credentials. There is a call to REST Adapter with client certs, how to propagate client identity to do director and connector graphql endpoints? @crabtree **IN PROGRESS**
- [ ] Implement Application Registry API] https://kyma-project.io/docs/1.8/components/application-connector/specifications/metadataapi/
  For not supported cases (for example, lack of API Packages, FetchRequests) we should return an error.
- [ ] Implement connector API  https://kyma-project.io/docs/1.8/components/application-connector/specifications/connectorapi/

PR #625, opened 2020-01-10
Add description regarding the REST security flow

crabtree:
**Description**

- Description of the legacy REST API security layer

**Related issue(s)**

See also #620


docs/investigations/rest-adapters/security.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Details
+
+The Compass GraphQL API supports the security mechanism used by the legacy REST API. Having this in mind we should reuse as much of the implementation as possible. Having this in mind, we can route the incoming REST API traffic through the same components as the new GraphQL API goes. Thanks to that the incoming traffic will be handled by the Istio and then routed to the Ory Oathkeeper reusing already implemented hydrators. The hydrators will handle all the details related to the one-time tokens, client certificate, tenant and scopes assignment. Having the security-related topics done we can transform the REST API request to the corresponding GraphQL call(s) and post it to the internal endpoint of the Compass Gateway. The following diagram depicts the described flow.
```
klaudiagrz:
```suggestion
The Compass GraphQL API supports the security mechanism used by the legacy REST API. Having this in mind, we should reuse as much of the implementation as possible. We can route the incoming REST API traffic in the same way as the new GraphQL API. Thanks to that, the incoming traffic will be handled by Istio and then routed to the Ory Oathkeeper, reusing already implemented hydrators. The hydrators will handle all the details related to the one-time tokens, client certificate, tenant and scopes assignment. Having the security-related topics done, we can transform the REST API request to the corresponding GraphQL call(s) and post it to the internal endpoint of the Compass Gateway. This diagram depicts the described flow:
```

PR #653, opened 2020-01-15
Prepare routes for Connectivity Adapter

pkosiec:
**Description**

- Prepare routes for Connectivity Adapter

**Related issue(s)**
https://github.com/kyma-incubator/compass/issues/620

### DIFF
--- a/components/connectivity-adapter/cmd/main.go
+++ b/components/connectivity-adapter/cmd/main.go
@@ -1,16 +1,49 @@
 package main
 
-import "net/http"
+import (
+	"fmt"
+	"log"
+	"net/http"
 
-func main() {
-	http.HandleFunc("/healthz", healthz)
-	err := http.ListenAndServe(":8080", nil)
-	if err != nil {
-		panic(err)
-	}
+	"github.com/gorilla/mux"
+	"github.com/kyma-incubator/compass/components/connectivity-adapter/internal/appregistry"
+	"github.com/kyma-incubator/compass/components/connectivity-adapter/internal/connector"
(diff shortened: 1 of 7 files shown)


components/connectivity-adapter/cmd/main.go
```diff
@@ -1,16 +1,47 @@
...
-		panic(err)
-	}
+	cfg := config{}
+	err := envconfig.InitWithPrefix(&cfg, "APP")
+	exitOnError(err, "while loading app config")
 
+	router := mux.NewRouter()
+	router.HandleFunc("/healthz", health.HandleFunc)
+
+	appRegistryRouter := router.PathPrefix("/app-registry/v1").Subrouter()
```
crabtree: Shouldn't this be the same as we have the URLs in the application registry and connector? Or do you plan to handle the redirections in a different layer?

I would rather set here the prefix `/v1/metadata`.

components/connectivity-adapter/cmd/main.go
```diff
@@ -1,16 +1,47 @@
...
+	appregistry.RegisterHandler(appRegistryRouter, cfg.AppRegistry)
+
+	connectorRouter := router.PathPrefix("/connector/v1").Subrouter()
```
crabtree: Same as with the `appRegistryRouter` router, currently there is no such route. I would go with `/v1/applications` prefix.

pkosiec: I agree 👍 Fixed

components/connectivity-adapter/cmd/main.go
pkosiec: I agree 👍 Fixed

PR #663, opened 2020-01-16
HB sends optional components in json schema

piotrmiskiewicz:
**Description**

- enhance provisioning json schemas by list of components


mszostok:
please add sth like `Components     []string` field to [ProvisioningParametersDTO](https://github.com/kyma-incubator/compass/blob/d5f267202176a5870802e79fda867fed61aa14df/components/kyma-environment-broker/internal/dto.go#L3)

Issue #642, opened 2020-01-14
Implement Application Registry API in Compass compatibility layer


pkosiec:
**Description**

Implement Application Registry API: https://kyma-project.io/docs/1.8/components/application-connector/specifications/metadataapi/
For not supported cases (for example, lack of API Packages, FetchRequests) we should return an error.

- Update App routing (include appName in path), get app details by name and save it into context @dbadura 
- Prepare conversion between types @aszecowka 
- Write HTTP handlers @pkosiec 

**Related issues**
https://github.com/kyma-incubator/compass/issues/620

PR #665, opened 2020-01-16
Delete Services with Connectivity Adapter

pkosiec:
**Description**

- Delete Services with Connectivity Adapter
- Copy App Registry models
- Initialize Converter GraphQL <-> Model

**Related issue(s)**
#642


components/connectivity-adapter/internal/appregistry/service/converter.go
```diff
@@ -0,0 +1,24 @@
+package service
```
crabtree: What is the reasoning for having this sort of not implemented file?

components/connectivity-adapter/internal/appregistry/service/graphql.go
```diff
@@ -0,0 +1,16 @@
+package service
```
crabtree: I would expect this file will grow during the implementation, it would be valuable to create a separate package for it.

components/connectivity-adapter/internal/appregistry/service/handler.go
```diff
@@ -1,37 +1,72 @@
...
+
+	err := gqlCli.Run(context.Background(), gqlRequest, nil)
+	if err != nil {
+		if apperrors.IsNotFoundError(err) {
+			message := fmt.Sprintf("entity with ID %s not found", id)
+			reqerror.WriteErrorMessage(writer, message, apperrors.CodeNotFound)
+			return
+		}
+
+		h.logger.Error(errors.Wrapf(err, "while deleting service with ID %s", id))
```
crabtree: Just a minor general comment, when you create a log message this way and you have some parameter, which has a type of string, has an empty string value then it is hard to catch this in the log output. Maybe it would be more helpful to format the log output, something like `fmt.Sprintf("while deleting service [ID=%s]", id)`

components/connectivity-adapter/internal/appregistry/service/handler_test.go
```diff
@@ -0,0 +1,116 @@
...
+			assert.Equal(t, tc.ExpectedResponseBody, string(bodyBytes))
+
+			if tc.LoggerAssertionsFn != nil {
+				tc.LoggerAssertionsFn(t, hook)
+			}
+		})
+	}
+}
+
+func fixRequest(id string, body io.Reader) *http.Request {
```
crabtree: Rename/refactor this function or move it inside the `TestHandler_Delete`. Right now when you add another test case for different HTTP methods or URL vars, this function cannot be reused.

components/connectivity-adapter/internal/appregistry/service/graphql.go
pkosiec: Let's wait with this decision until this file grows, okay? Because my idea was to have only GraphQL requests here, and for the app registry it shouldn't be much. Let's see what the future brings 🙂

components/connectivity-adapter/internal/appregistry/service/converter.go
pkosiec: I created this empty file to show my idea for the project structure. Now we can split development: one person can write handlers, and one - conversion methods. I will define interface in the main handler file for the converter and keep this file as it is, if you don't mind. What do you think?

components/connectivity-adapter/internal/appregistry/service/handler.go
pkosiec: Hmm, ok, so I will use `WithFields` helper method from logrus - it will be cleaner solution for that

components/connectivity-adapter/internal/appregistry/service/handler_test.go
pkosiec: Hmm, techincally it can be used for every request, because we rely on gorilla/mux, when it comes to routing - that's why the request method and URL can be anything, only the body and variable counts here. Then, this function could be used for all methods for endpoint `/services/{serviceId}`. I will rename it to `fixServiceDetailsRequest`. Is that OK for you?

crabtree: Not really, you hardcoded the `http.MethodDelete` inside.

components/connectivity-adapter/internal/appregistry/service/converter.go
crabtree: OK, with the interface defined it makes more sense to me.

components/connectivity-adapter/internal/appregistry/service/handler_test.go
pkosiec: I changed the method to `POST`. As I wrote before, it doesn't matter, the method can be anything as we don't use it.

pkosiec: As discussed, I provided there `Anything` method.

components/connectivity-adapter/README.md
```diff
@@ -1,5 +1,25 @@
...
+go run cmd//main.go
+```
+
+## Configuration
+
+The Connectivity Adapter binary allows to override some configuration parameters. You can specify following environment variables.
+
+| ENV              | Default        | Description                                       |
+| ---------------- | -------------- | ------------------------------------------------- |
+| APP_ADDRESS      | 127.0.0.1:8080 | The address and port for the service to listen on |
```
klaudiagrz:
```suggestion
| **APP_ADDRESS**      | `127.0.0.1:8080` | Address and port for the service to listen on |
```

components/connectivity-adapter/README.md
```diff
@@ -1,5 +1,25 @@
...
+| APP_DIRECTOR_URL | 127.0.0.1:3000 | The host of running Director                      |
```
klaudiagrz:
```suggestion
| **APP_DIRECTOR_URL** | `127.0.0.1:3000` | Host of the running Director component                      |
```

PR #668, opened 2020-01-19
Remove seed param from azure config

mszostok:
**Description**

- Remove seed param from azure config

Reason: 
The quota for az-eu1 was exceeded. What's more the seed param in the near future will not be supported by the Gardener. To benefit from auto-selected seed we need to remove it from our default config params.


mszostok:
this is blocked because of the workin in provisioner/hydraform

PR #669, opened 2020-01-19
On demand release support in provisioner

mszostok:
**Description**

- On demand release support in provisioner

This PR is necessary for testing the component installation because the required charts are currently missing from the releases on github and kyma installer does not support providing components from external sources, yet.

**Issue:**
- https://github.com/kyma-incubator/compass/issues/691

### DIFF
--- a/components/kyma-environment-broker/Gopkg.lock
+++ b/components/kyma-environment-broker/Gopkg.lock
@@ -245,6 +245,19 @@
   version = "v1.3.2"
 
 [[projects]]
+  digest = "1:2e3c336fc7fde5c984d2841455a658a6d626450b1754a854b3b32e7a8f49a07a"
+  name = "github.com/google/go-cmp"
+  packages = [
+    "cmp",
+    "cmp/internal/diff",
+    "cmp/internal/function",
+    "cmp/internal/value",
+  ]
+  pruneopts = "UT"
+  revision = "3af367b6b30c263d47e8895973edcca9a49cf029"
+  version = "v0.2.0"
+
+[[projects]]
   digest = "1:a6181aca1fd5e27103f9a920876f29ac72854df7345a39f3b01e61c8c94cc8af"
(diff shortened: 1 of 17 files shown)


components/provisioner/internal/installation/release/aggregator.go
```diff
@@ -0,0 +1,59 @@
...
+		if found {
+			return found, nil
+		}
+	}
+
+	return false, nil
+}
+
+// Register adds new release provider
+func (a Aggregator) Register(releaseProvider ReadRepository) {
```
piotrmiskiewicz: you are not using a pointer receiver so the sync does not work. There is a copy of mutex.
See an example: https://play.golang.org/p/GhoGRQGISBg

components/provisioner/internal/installation/release/on_demand.go
```diff
@@ -0,0 +1,137 @@
...
+	return rel, nil
+}
+
+func (o *OnDemand) saveRelease(rel model.Release) dberrors.Error {
+	_, err := o.session.InsertInto("kyma_release").
+		Columns("id", "version", "tiller_yaml", "installer_yaml").
+		Record(rel).
+		Exec()
+	if err != nil {
+		return dberrors.Internal("Failed to save Kyma release artifacts for version %s: %s", rel.Version, err.Error())
```
Szymongib: Same here, I would rather not duplicate the querries logic.

mszostok:
HI

thanks for the review.

> It would be good to add some tests
I didnt add a test coverage because of two reasons,
- First you do not have the test coverage for the current impl, so I thought that u didnt add it because u will refactor that. 
- Second thing is that is only a temp solution until 1.10 will go out (around 1week). 

> This is the exact same logic as in Release repository, I think it would be better to use it and not to duplicate the sql queries code.

This is only a temp code IMO is not worth to unify that when this will be removed after one week.

PR #692, opened 2020-01-21
Create a pre-upgrade job for KEB

polskikiel:
**Description**

- Created a pre-upgrade job for KEB to unblock the compass upgrade pipeline.

**Related issue(s)**

### DIFF
--- a/chart/compass/charts/kyma-environment-broker/templates/pre-upgrade-job.yaml
+++ b/chart/compass/charts/kyma-environment-broker/templates/pre-upgrade-job.yaml
@@ -1,0 +1,139 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "fullname" . }}-pre-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ template "fullname" . }}-pre-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "fullname" . }}-pre-upgrade
+subjects:
+  - kind: ServiceAccount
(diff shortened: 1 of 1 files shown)


chart/compass/charts/kyma-environment-broker/templates/pre-upgrade-job.yaml
```diff
@@ -0,0 +1,139 @@
...
+  name: {{ template "fullname" . }}-pre-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation
+rules:
+  - apiGroups: ["*"]
+    resources: ["secrets"]
+    verbs: ["get", "create", "update", "patch", "list"]
```
jasiu001: Do you need all these actions? `list` for sure is to necessary, IMO only `get` and `update (or) patch` is necessary. `create` can be removed because you check if Secret exist.
Maybe "rbacs" from `chart/compass/charts/kyma-environment-broker/templates/rbac.yaml` can be use, there are also `secret` in resources, maybe you can add one action there

jasiu001:
Do you aware of warning a message during job execution?
```
Secret Exist
Updating secret...
Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply
Successfully updated secret
```

chart/compass/charts/kyma-environment-broker/templates/pre-upgrade-job.yaml
polskikiel: `get` permission is also necessary due to the below command:
```
local result=$(kubectl get secret -n compass-system $APP_SECRET_NAME -o json |
                jq '.data["postgresql-brokerDatabaseName"]={{ .Values.global.database.embedded.brokerDBName | b64enc | quote }}' |
                kubectl apply -f - )
```
I removed `create` and `list` permissions

polskikiel:
@jasiu001 
Yes, I saw that message but I would ignore it as it doesn't affect our job. I can always silence that but that will only increase the complexity here: 
```
local result=$(kubectl get secret -n compass-system $APP_SECRET_NAME -o json |
                jq '.data["postgresql-brokerDatabaseName"]={{ .Values.global.database.embedded.brokerDBName | b64enc | quote }}' |
                kubectl apply -f - )
```
To be honest, I wouldn't even know how to silence that :)

chart/compass/charts/kyma-environment-broker/templates/pre-upgrade-job.yaml
polskikiel: I also prefer to use separate RBACs for this job, because we can define the same helm hooks on it as on the job.

PR #696, opened 2020-01-21
Create service via Connectivity Adapter 

pkosiec:
**Description**

- Create service via Connectivity Adapter 

**Related issue(s)**
#642


components/connectivity-adapter/internal/appregistry/service/converter.go
```diff
@@ -11,14 +12,23 @@ func NewConverter() *converter {
 	return &converter{}
 }
 
-func (c *converter) DetailsToGraphQL(in model.ServiceDetails) graphql.Application {
-	panic("not implemented")
+func (c *converter) DetailsToGraphQLInput(in model.ServiceDetails) (graphql.ApplicationRegisterInput, error) {
+	return graphql.ApplicationRegisterInput{
```
dbadura: Why you don't use `in` param?

dbadura: It would be nice to have test for this method.

pkosiec: Added a comment to clarify what's going on here 🙂

components/connectivity-adapter/internal/appregistry/service/handler.go
```diff
@@ -82,3 +165,10 @@ func (h *Handler) Delete(writer http.ResponseWriter, request *http.Request) {
 
 	writer.WriteHeader(http.StatusNoContent)
 }
+
+func (h *Handler) closeBody(rq *http.Request) {
+	err := rq.Body.Close()
+	if err != nil {
+
```
crabtree: It is worth to log that fact ;)

pkosiec: Right 😄 I missed that somehow

PR #697, opened 2020-01-22
Add retries to DB calls

polskikiel:
**Description**

- Added retries to DB calls

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/director/oauth/client.go
+++ b/components/kyma-environment-broker/internal/director/oauth/client.go
@@ -13,7 +13,6 @@ import (
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/pkg/errors"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -58,12 +57,10 @@ func (c *oauthClient) WaitForCredentials() error {
 			Namespace: c.secretNamespace,
 			Name:      c.secretName,
 		}, secret)
-		switch {
-		case apierrors.IsNotFound(err):
+		// it fails on connection-refused error on first call and it restarts our application.
+		if err != nil {
 			log.Warnf("secret %s not found", c.secretName)
 			return false, nil
-		case err != nil:
-			return false, errors.Wrapf(err, "while waiting for secret %s", c.secretName)
 		}
 		return true, nil
 	})
--- a/components/kyma-environment-broker/internal/storage/dbsession/read.go
+++ b/components/kyma-environment-broker/internal/storage/dbsession/read.go
@@ -22,7 +22,7 @@ func (r readSession) GetInstanceByID(instanceID string) (internal.Instance, dber
 
 	if err != nil {
 		if err == dbr.ErrNotFound {
-			return internal.Instance{}, dberr.NotFound("Cannot find Instance for instanceID:'%s", instanceID)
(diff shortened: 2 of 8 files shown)


components/kyma-environment-broker/internal/director/oauth/client.go
```diff
@@ -58,12 +57,9 @@ func (c *oauthClient) WaitForCredentials() error {
 			Namespace: c.secretNamespace,
 			Name:      c.secretName,
 		}, secret)
-		switch {
-		case apierrors.IsNotFound(err):
+		if err != nil {
```
polskikiel: it fails on `connection-refused` error on first call and it restarts our application.

polskikiel: we still need to add initialDelayTimeout to our chart to completely fix this issue.

components/kyma-environment-broker/internal/storage/driver/postsql/instance.go
```diff
@@ -18,30 +21,50 @@ func NewInstance(sess dbsession.Factory) *Instance {
...
 func (s *Instance) GetByID(instanceID string) (*internal.Instance, error) {
 	sess := s.NewReadSession()
-	instance, err := sess.GetInstanceByID(instanceID)
-	if err != nil {
-		return nil, errors.Wrapf(err, "while getting instance by ID %s", instanceID)
-	}
-
-	return &instance, nil
+	instance := &internal.Instance{}
+	err := wait.Poll(defaultRetryInterval, defaultRetryTimeout, func() (bool, error) {
```
mszostok: tbh I'm not sure if this retry logic should be placed here. IMO it should be in upper layer  or just some wrapper around that storage which will be responsible for retries and with such approach u will also implement that only once instead of copying that retrie functionality for each function. But let's agree that for now it's good enough, and we will change it in the future 

same thing for retries errors: In current implementation we do not retry on "NotFound" error but in general we are eventually consistent system, so doing that on not found error should be also ok.

components/kyma-environment-broker/internal/storage/dbsession/write.go
```diff
@@ -41,6 +41,9 @@ func (ws writeSession) UpdateInstance(instance internal.Instance) dberr.Error {
 		Set("provisioning_parameters", instance.ProvisioningParameters).
 		Exec()
 	if err != nil {
+		if err == dbr.ErrNotFound {
+			return dberr.NotFound("Cannot find Instance for instanceID:'%s", instance.InstanceID)
```
mszostok:
```suggestion
			return dberr.NotFound("Cannot find Instance for instanceID:'%s'", instance.InstanceID)
```

components/kyma-environment-broker/internal/storage/dbsession/read.go
```diff
@@ -21,7 +21,7 @@ func (r readSession) GetInstanceByID(instanceID string) (internal.Instance, dber
 		LoadOne(&instance)
 
 	if err != nil {
-		if err != dbr.ErrNotFound {
+		if err == dbr.ErrNotFound {
 			return internal.Instance{}, dberr.NotFound("Cannot find Instance for instanceID:'%s", instanceID)
```
mszostok:
```suggestion
			return internal.Instance{}, dberr.NotFound("Cannot find Instance for instanceID:'%s'", instanceID)
```

mszostok:
update chart image

components/kyma-environment-broker/internal/storage/driver/postsql/instance.go
polskikiel: good suggestion, I will add a todo comment

Issue #659, opened 2020-01-16
Application API for eventing configuration should return whole URL for eventing


crabtree:
**Description**

Current implementation returns only the base URL part for the eventing service. It is on the application responsibility to produce a valid path where the events should be sent. The compass API should return the whole URL instead of only the base path. 

The format of the URL: `https://{EVENTING_SERVICE_BASE_URL}/{APP_NAME}/v1/events`

**Reasons**

The reasoning for the change is that the eventing URL format can be changed and if on the application site there is some concatenation done it will make developers change the application code. If the whole URL will be returned from the Compass API the change in the application code will not be required.

When querying for the runtime eventing configuration the API should behave as it is right now, so return only the base URL for eventing service.

PR #704, opened 2020-01-22
Fix app event url

dbadura:
**Description**

- fix application event URL

**Related issue(s)**

See #659 

bump on kyma:
https://github.com/kyma-project/kyma/pull/6962


components/director/internal/model/eventing.go
```diff
@@ -1,29 +1,80 @@
...
 	return &ApplicationEventingConfiguration{
 		EventingConfiguration: EventingConfiguration{
-			DefaultURL: url,
+			DefaultURL: validUrl,
 		},
+	}, nil
+}
+
+func NewEmptyApplicationEventingConfig() (*ApplicationEventingConfiguration, error) {
+	emptyURL, err := validateURL("")
```
kfurgol: I would think about renaming the `validateURL` function. Looking at this particular line I would expect that it returns a bool value. Maybe even split its functionality between `createURL` and `validateURL` functions? Just an idea

components/director/internal/domain/runtime/fixtures_test.go
```diff
@@ -234,10 +235,11 @@ func fixGQLSystemAuth(id string, auth *graphql.Auth) *graphql.SystemAuth {
 	}
 }
 
-func fixModelRuntimeEventingConfiguration(url string) *model.RuntimeEventingConfiguration {
+func fixModelRuntimeEventingConfiguration(t *testing.T, rawURL string) *model.RuntimeEventingConfiguration {
+	validURl := fixValidURL(t, rawURL)
 	return &model.RuntimeEventingConfiguration{
 		EventingConfiguration: model.EventingConfiguration{
-			DefaultURL: url,
+			DefaultURL: validURl,
```
kfurgol:
```suggestion
			DefaultURL: validURL,
```

components/director/internal/domain/application/resolver.go
```diff
@@ -741,10 +744,14 @@ func (r *Resolver) EventingConfiguration(ctx context.Context, obj *graphql.Appli
 	if obj == nil {
 		return nil, errors.New("Application cannot be empty")
 	}
-
-	appID, err := uuid.Parse(obj.ID)
+	tenantID, err := tenant.LoadFromContext(ctx)
 	if err != nil {
-		return nil, errors.Wrap(err, "while parsing application ID as UUID")
+		return nil, errors.New("Error while loading tenant from context")
```
kfurgol:
```suggestion
		return nil, errors.New("error while loading tenant from context")
```

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

components/director/internal/domain/application/resolver.go
```diff
@@ -741,10 +744,14 @@ func (r *Resolver) EventingConfiguration(ctx context.Context, obj *graphql.Appli
...
+	}
+
+	app := r.appConverter.ToModel(obj, tenantID)
+	if app == nil {
+		return nil, errors.New("Application cannot be empty")
```
kfurgol:
```suggestion
		return nil, errors.New("application cannot be empty")
```

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

components/director/internal/model/eventing.go
dbadura: I think `createURL` it's ok to return object and err.

components/director/internal/domain/application/converter.go
```diff
@@ -140,11 +141,26 @@ func (c *converter) CreateInputGQLToJSON(in *graphql.ApplicationRegisterInput) (
 	return string(appInput), nil
 }
 
+func (c *converter) ToModel(obj *graphql.Application, tenantID string) *model.Application {
```
pkosiec: Rename to GraphQLToModel, as it is in other parts of our codebase

tests/director/api/eventing_test.go
```diff
@@ -5,21 +5,28 @@ import (
...
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
 
 	gcli "github.com/machinebox/graphql"
 	"github.com/stretchr/testify/require"
 )
 
+const runtimeEventURL = "https://%s"
```
pkosiec: add `Format` suffix to the name

tests/director/api/eventing_test.go
```diff
@@ -5,21 +5,28 @@ import (
...
+const appEventURL = "https://%s/%s/v1/events"
```
pkosiec: add `Format` suffix to the name

components/director/internal/model/eventing.go
pkosiec: `parseURL`?

components/director/internal/model/eventing.go
```diff
@@ -1,29 +1,80 @@
...
 	return &RuntimeEventingConfiguration{
 		EventingConfiguration: EventingConfiguration{
-			DefaultURL: url,
+			DefaultURL: validUrl,
 		},
-	}
+	}, nil
 }
 
+const ApplicationEventingURLScheme = "https://%s/%s/v1/events"
```
pkosiec: I would not hardcode `https`. Can we get it from `url.URL`?

components/director/internal/model/eventing.go
```diff
@@ -1,29 +1,80 @@
...
+		},
+	}, nil
+}
+
+func validateURL(rawURL string) (url.URL, error) {
+	validUrl, err := url.Parse(rawURL)
+	if err != nil {
+		return url.URL{}, err
+	}
+	if validUrl == nil {
```
pkosiec: is there really a situation when error is nil and validURL is also nil?

components/director/internal/model/eventing.go
```diff
@@ -1,29 +1,80 @@
...
+
 	return &ApplicationEventingConfiguration{
 		EventingConfiguration: EventingConfiguration{
-			DefaultURL: url,
+			DefaultURL: validUrl,
 		},
+	}, nil
+}
+
+func NewEmptyApplicationEventingConfig() (*ApplicationEventingConfiguration, error) {
```
pkosiec: Why dp we need a separate function for creating empty eventing config?

components/director/internal/model/eventing.go
```diff
@@ -1,29 +1,80 @@
...
 type ApplicationEventingConfiguration struct {
 	EventingConfiguration
 }
 
-func NewApplicationEventingConfiguration(url string) *ApplicationEventingConfiguration {
+func NewApplicationEventingConfiguration(runtimeEventURL url.URL, appName string) (*ApplicationEventingConfiguration, error) {
+	var appEventURL string
+
+	if runtimeEventURL.Host == "" {
+		appEventURL = ""
```
pkosiec: I think it's too complicated. Why?

components/director/internal/model/eventing.go
```diff
@@ -1,29 +1,80 @@
...
+	} else {
+		appEventURL = fmt.Sprintf(ApplicationEventingURLScheme, runtimeEventURL.Host, appName)
+	}
+
+	validUrl, err := validateURL(appEventURL)
```
pkosiec: Can we parse the URL and then add the path? It would be simpler approach, right?

components/director/internal/domain/eventing/service.go
```diff
@@ -62,15 +62,20 @@ func (s *service) CleanupAfterUnregisteringApplication(ctx context.Context, appI
...
-func (s *service) SetForApplication(ctx context.Context, runtimeID uuid.UUID, appID uuid.UUID) (*model.ApplicationEventingConfiguration, error) {
+func (s *service) SetForApplication(ctx context.Context, runtimeID uuid.UUID, app model.Application) (*model.ApplicationEventingConfiguration, error) {
 	tenantID, err := tenant.LoadFromContext(ctx)
 	if err != nil {
 		return nil, errors.Wrapf(err, "while loading tenant from context")
 	}
 
+	appID, err := uuid.Parse(app.ID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "while parsing application ID: %s", app.ID)
```
pkosiec: Do we need it? Our DB will return error, right? I don't think we need this kind of manual check now; it's the matter of wrapping DB error

components/director/internal/model/eventing.go
dbadura: I look into the code deeper and there is no such situation.
I can remove this check.

components/director/internal/domain/eventing/service.go
dbadura: Yes, because in internal methods we use UUID type.

components/director/internal/model/eventing.go
dbadura: It's fixture.

components/director/internal/model/eventing.go
pkosiec: Alright, if validUrl can be nil, then keep it for safety reason.

components/director/internal/model/eventing.go
pkosiec: But why you cannot simply call `NewApplicationEventingConfiguration(url.URL{}, ""`?

PR #716, opened 2020-01-23
Converter graphql <-> REST types 

aszecowka:
Implemented converter Application Registry types <==> Director types. 

Not addressed in this PR:
- CSRF
- Documentation fields mapping

**Related issue(s)**

### DIFF
--- a/components/connectivity-adapter/internal/appregistry/service/converter.go
+++ b/components/connectivity-adapter/internal/appregistry/service/converter.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -10,7 +11,7 @@ import (
 
 const (
 	// fields that exist in Application Registry but don't exist in Director GraphQL API, are stored in labels with special prefix
-	prefixUnmappedFields          = "compatibility_unmapped_fields_"
+	prefixUnmappedFields          = "compatibility_unmapped_fields/"
 	unmappedFieldIdentifier       = prefixUnmappedFields + "identifier"
 	unmappedFieldShortDescription = prefixUnmappedFields + "shortDescription"
 )
@@ -47,12 +48,11 @@ func (c *converter) DetailsToGraphQLInput(deprecated model.ServiceDetails) (grap
 			if outApi.Spec == nil {
 				outApi.Spec = &graphql.APISpecInput{}
 			}
+
 			if strings.ToLower(deprecated.Api.ApiType) == "odata" {
 				outApi.Spec.Type = graphql.APISpecTypeOdata
-				outApi.Spec.Format = graphql.SpecFormatJSON // this field is required, so we arbitrary chose JSON
 			} else {
 				outApi.Spec.Type = graphql.APISpecTypeOpenAPI // quite brave assumption that it will be OpenAPI
-				outApi.Spec.Format = graphql.SpecFormatJSON
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


components/connectivity-adapter/internal/appregistry/service/converter.go
```diff
@@ -1,9 +1,18 @@
...
+
 	"github.com/kyma-incubator/compass/components/connectivity-adapter/internal/appregistry/model"
 	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
-	"github.com/kyma-incubator/compass/tests/director/pkg/ptr"
+)
+
+const (
+	// fields that exist in Application Registry but don't exist in Director GraphQL API, are stored in labels with special prefix
+	prefixUnmappedFields          = "compatibility_unmapped_fields_"
+	unmappedFieldIdentifier       = prefixUnmappedFields + "identifier"
```
pkosiec: Did we discuss our convention of naming label keys? I remember something like `prefix/name`, but I don't know whether this is our agreed approach or not. Anyway, this is just an open question.

components/connectivity-adapter/internal/appregistry/service/converter.go
```diff
@@ -12,25 +21,420 @@ func NewConverter() *converter {
...
+
+		if deprecated.Api.ApiType != "" {
+			if outApi.Spec == nil {
+				outApi.Spec = &graphql.APISpecInput{}
+			}
+			if strings.ToLower(deprecated.Api.ApiType) == "odata" {
+				outApi.Spec.Type = graphql.APISpecTypeOdata
+				outApi.Spec.Format = graphql.SpecFormatJSON // this field is required, so we arbitrary chose JSON
+			} else {
+				outApi.Spec.Type = graphql.APISpecTypeOpenAPI // quite brave assumption that it will be OpenAPI
```
pkosiec: > quite brave assumption

tenor-188709088

What about XML with ODATA? I'm not 100% sure, but I believe that App Registry saved the spec regardless the format and type, and then Headless CMS (Rafter) detected the type and uploaded proper spec. Please confirm it with corresponding teams.

If that was the case, unfortunately we need to do the same 🙁 Of course not in the same PR, but still, we need to detect type and format.

components/connectivity-adapter/internal/appregistry/service/converter.go
```diff
@@ -12,25 +21,420 @@ func NewConverter() *converter {
...
+			{
+				Spec: &graphql.EventSpecInput{
+					Data: ptrClob(graphql.CLOB(deprecated.Events.Spec)),
+				},
+			},
+		}
+	}
+
+	if deprecated.Documentation != nil {
+		// TODO later
```
pkosiec: All these `TODOs` will be done in following PRs, right?

components/connectivity-adapter/internal/appregistry/service/converter_test.go
```diff
@@ -0,0 +1,858 @@
...
+	sut := NewConverter()
+
+	for name, tc := range map[string]testCase{
+		"simple attributes": {
+			given: graphql.ApplicationExt{
+				Application: graphql.Application{
+					ID:          "id",
+					Name:        "name",
+					Description: ptrStringOrNilForEmpty("description"),
+					//TODO IntegrationSystemID
```
pkosiec: I think we don't need to think about IntegrationSystemID at all, because it's already provided in labels (for filtering purposes).

components/connectivity-adapter/internal/appregistry/service/converter.go
aszecowka: yes

components/connectivity-adapter/internal/appregistry/service/converter.go
aszecowka: changed

PR #730, opened 2020-01-27
Add label name:appname while creating, updating application

dbadura:
**Description**

- Add label `name:appname` while creating, updating application

**Related issue(s)**

See #642

### DIFF
--- a/components/director/internal/domain/application/service.go
+++ b/components/director/internal/domain/application/service.go
@@ -19,6 +19,7 @@ import (
 
 const (
 	intSysKey = "integration-system-id"
+	nameKey   = "name"
 )
 
 //go:generate mockery -name=ApplicationRepository -output=automock -outpkg=automock -case=underscore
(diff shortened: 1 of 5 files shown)


components/director/internal/domain/application/service.go
```diff
@@ -265,6 +266,8 @@ func (s *service) Create(ctx context.Context, in model.ApplicationRegisterInput)
 	if in.IntegrationSystemID != nil {
 		in.Labels[intSysKey] = *in.IntegrationSystemID
 	}
+	//This is workaround for connectivity adapter https://github.com/kyma-incubator/compass/tree/master/docs/investigations/rest-adapters
```
pkosiec: I don't know if this comment is necessary. I don't know if this is a workaround either 😄 We want to filter via labels, right? So it makes sense to put all properties there

dbadura: Removed : )

PR #734, opened 2020-01-27
Connectivity adapter: renewal and revocation added

akgalwas:
**Description**

- Certificates revocation and renewal implemented
- Getting application name and event base url from Director implemented


akgalwas:
Notes:
- implementation doesn't contain calling Director in signingRequestInfo handle as it requires Connectivity Adapter to be registered as integration system
- Connector Service tests were used to initially check if contract is not broken, however, they require modification to be used as acceptance test (there is a need to create application is Director prior to generating/renewing certificates)

components/connectivity-adapter/cmd/main.go
```diff
@@ -26,23 +26,31 @@ func main() {
...
+	err := connector.RegisterHandler(connectorRouter, cfg.Connector, cfg.AppRegistry.DirectorEndpoint)
 	if err != nil {
-		exitOnError(err, "Failed to init Connector handler")
+		return nil, err
 	}
 
-	log.Printf("Listening on %s", cfg.Address)
-	err = http.ListenAndServe(cfg.Address, router)
-	exitOnError(err, fmt.Sprintf("while listening on %s", cfg.Address))
+	return connectorRouter, nil
```
aszecowka: then `/v1/health` is not accessible anymore

components/connectivity-adapter/cmd/main.go
```diff
@@ -7,14 +7,14 @@ import (
...
-	connector "github.com/kyma-incubator/compass/components/connectivity-adapter/internal/connector"
+	connector "github.com/kyma-incubator/compass/components/connectivity-adapter/internal/connectorservice"
 	"github.com/kyma-incubator/compass/components/connectivity-adapter/pkg/health"
 	"github.com/pkg/errors"
 	"github.com/vrischmann/envconfig"
 )
 
 type config struct {
-	Address string `envconfig:"default=127.0.0.1:8080"`
+	APIAddress string `envconfig:"default=127.0.0.1:8080"`
```
pkosiec: Do we need to rename it? Then we would need to update docs as well and helm charts values

akgalwas: Done

components/connectivity-adapter/cmd/main.go
akgalwas: Done

PR #736, opened 2020-01-27
Implement e2e provisioning test

polskikiel:
**Description**

- Implemented the e2e provisioning test
  - It tests the kyma runtime provisioning scenario
  - It is deployed via TestSuite (Octopus)
  - It fakes the external dependencies in this scenario
  - The test ends with success after a successful call to the dashboard URL.
  - It executes cleanup logic after the test to delete created runtime.

**Related issue(s)**


tests/e2e/provisioning/Makefile
```diff
@@ -0,0 +1,12 @@
...
+BUILDPACK = eu.gcr.io/kyma-project/test-infra/buildpack-golang-toolbox:v20190913-65b55d1
+SCRIPTS_DIR = $(realpath $(shell pwd)/../../..)/scripts
+
+include $(SCRIPTS_DIR)/generic_make_go.mk
+
+VERIFY_IGNORE := /vendor\|/automock\|/testdata
+
+verify:: vet
+
+test-local: ;
```
jasiu001: Is it correct?

tests/e2e/provisioning/pkg/client/broker/broker_client.go
```diff
@@ -0,0 +1,229 @@
...
+			return false, nil
+		}
+		switch domain.LastOperationState(response.State) {
+		case domain.Failed:
+			c.log.Info("Provisioning failed")
+			return true, errors.New("provisioning failed")
+		case domain.Succeeded:
+			c.log.Infof("Runtime Provisioning succeeded!")
+			return true, nil
+		default:
```
jasiu001: handle also `domain.InProgress` state, if in some reason there will be an empty response ("{}") then the state will be an empty string, we can add log about it just in case (in the default case)

tests/e2e/provisioning/pkg/client/broker/broker_client.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	c.log.Infof("Successfully send deprovision request, got operation ID %s", response.Operation)
+	return nil
+}
+
+func (c *Client) AwaitProvisioningSucceeded(operationID string) error {
+	lastOperationURL := fmt.Sprintf("%s%s/%s/last_operation?operation=%s", c.brokerConfig.URL, instancesURL, c.runtimeID, operationID)
+
+	response := lastOperationResponse{}
+	err := wait.Poll(time.Minute, c.brokerConfig.ProvisionTimeout, func() (bool, error) {
```
jasiu001: I think more than one minute can be set as interval ;)

tests/e2e/provisioning/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.12.5-alpine3.9 as builder
+
+ENV SRC_DIR=/go/src/github.com/kyma-incubator/compass/tests/e2e/provisioning
+
+WORKDIR $SRC_DIR
+COPY . $SRC_DIR
+
+RUN CGO_ENABLED=0 GOOS=linux go test -c ./test
```
mszostok:
```suggestion
RUN CGO_ENABLED=0 GOOS=linux go test -ldflags="-s -w" -c ./test
```

with `-ldflags="-s -w"` you will reduce ~15mb on that binary

tests/e2e/provisioning/test/e2e_provisioning_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func Test_E2E_Provisioning(t *testing.T) {
+	ts := newTestSuite(t)
+	operationID, err := ts.brokerClient.ProvisionRuntime()
+	require.NoError(t, err)
+	defer ts.TearDown()
+
+	err = ts.brokerClient.AwaitProvisioningSucceeded(operationID)
```
mszostok: **nit:** I like API where I can explicit set the timeout for such "await" and I do not need to go into impl details. As a result it is more readable for me.

tests/e2e/provisioning/test/e2e_provisioning_test.go
```diff
@@ -0,0 +1,25 @@
...
+	require.NoError(t, err)
+
+	dashboardURL, err := ts.brokerClient.FetchDashboardURL()
+	require.NoError(t, err)
+
+	err = ts.kymaClient.CallDashboardURL(dashboardURL)
```
mszostok:
```suggestion
	err = ts.kymaClient.AssertRedirectedToUAA(dashboardURL)
```

define your expectation instead what is your action

chart/compass/charts/kyma-environment-broker/templates/tests/e2e-provisioning.yaml
```diff
@@ -0,0 +1,109 @@
...
+            value: "3e64ebae-38b5-46a0-b1ed-9ccee153a0ae"
+          command:
+          - "bash"
+          args:
+            - "-c"
+            - |
+              MAX_RETRIES=60
+
+              # copied from registration-job
+              function wait_for_access_to_api_server() {
```
jasiu001: if you have `entrypoint.sh` bash file is not easier to put all bash code in one place?

tests/e2e/provisioning/scripts/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+res=0
```
jasiu001: My IDE says "Add shebang line" :)

tests/e2e/provisioning/pkg/client/broker/dto.go
```diff
@@ -0,0 +1,24 @@
+package broker
+
+type ersContext struct {
```
mszostok: we cannot use `ers` name, maybe sth like `InputContext`?

tests/e2e/provisioning/pkg/client/broker/dto.go
```diff
@@ -0,0 +1,24 @@
...
+	TenantID        string `json:"tenant_id"`
```
mszostok: This filed **should not be used anywhere** because we do not have currently any contract on that filed, so to make it simpler please remove it.

tests/e2e/provisioning/pkg/client/broker/broker_client.go
```diff
@@ -0,0 +1,229 @@
...
+	azurePlanID = "4deee563-e5ec-4731-b9b1-53b42d855f0c"
+
+	instancesURL = "/v2/service_instances"
+)
+
+func (c *Client) ProvisionRuntime() (string, error) {
+	c.log.Infof("Provisioning Runtime [ID: %s, NAME: %s]", c.runtimeID, c.clusterName)
+	requestByte, err := c.prepareProvisionDetails()
+	if err != nil {
+		return "", errors.Wrap(err, "while marshalling request body")
```
mszostok:
```suggestion
		return "", errors.Wrap(err, "while preparing provision details")
```

you do not know why this operation failed ;) what's more in current impl you will end up with log:
```
"while marshalling request body: while marshalling request body:  <error_msg>"
```

tests/e2e/provisioning/test/test_suite.go
```diff
@@ -0,0 +1,75 @@
...
+func newTestSuite(t *testing.T) *Suite {
+	cfg := &Config{}
+	err := envconfig.InitWithPrefix(cfg, "APP")
+	require.NoError(t, err)
+
+	client := newHTTPClient(true)
+
+	log := logrus.New()
+	runtimeID := uuid.New().String()
+	kymaClient := kyma.NewClient(*client, log)
```
mszostok: add field for each logger to assign them to given domain e.g.:
`log.WithField("service", "kyma-client")`

as a results logs will be more readable, for example right now you have such log statement: 
`c.log.Warn(errors.Wrap(err, "while executing request").Error())`

with additional field you will know in which domain this error occurred :) 

same pattern you can also used for each method e.g.:

```go
func (c *Client) ProvisionRuntime() (string, error) {
	pLog := c.log.WithField("action", "provision-runtime")

	pLog.Infof("Provisioning Runtime [ID: %s, NAME: %s]", c.runtimeID, c.clusterName)
	requestByte, err := c.prepareProvisionDetails()
	if err != nil {
		return "", errors.Wrap(err, "while marshalling request body")
	}
	pLog.Infof("Provisioning parameters: %v", string(requestByte))

	// ...
}
```

tests/e2e/provisioning/pkg/client/kyma/kyma_client.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Kyma console URL won't redirect us to the UUA logging page, to achieve that we must call dex with a set of parameters
+// state and nonce params are faked
+func (c *Client) buildTargetURL(dashboardURL string) string {
+	domain := strings.Split(strings.Split(dashboardURL, "console.")[1], "/")[0]
+	params := "&response_type=id_token%20token&scope=audience%3Aserver%3Aclient_id%3Akyma-client%20audience%3" +
+		"Aserver%3Aclient_id%3Aconsole%20openid%20profile%20email%20groups&state=5a4f3d15&nonce=79bf9c29"
+
+	return fmt.Sprintf("https://dex.%s/auth?client_id=console&redirect_uri=%s%s", domain, dashboardURL, params)
```
jasiu001: maybe something more readable :) 
```
u := url.URL{
  Scheme: "https",
  Host: fmt.Sprintf("dex.%s/auth", domain),
}
q := u.Query()
q.Set("client_id", "console")
q.Set("redirect_uri", dashboardURL)
q.Set("response_type", "id_token")
q.Set("scope", "audience")
(...)
u.RawQuery = q.Encode()

return u.String()
```

tests/e2e/provisioning/test/test_suite.go
```diff
@@ -0,0 +1,75 @@
...
+	brokerClient  *broker.Client
+	kymaClient    *kyma.Client
+	runtimeClient *runtime.Client
+}
+
+func (ts *Suite) TearDown() {
+	ts.log.Info("Cleaning up...")
+	err := ts.runtimeClient.TearDown()
+	assert.NoError(ts.t, err)
+	err = ts.brokerClient.DeprovisionRuntime()
```
mszostok: we do not want to wait until Deprovision will be succeeded?

tests/e2e/provisioning/test/test_suite.go
```diff
@@ -0,0 +1,75 @@
...
+
+	log           logrus.FieldLogger
+	brokerClient  *broker.Client
+	kymaClient    *kyma.Client
+	runtimeClient *runtime.Client
+}
+
+func (ts *Suite) TearDown() {
+	ts.log.Info("Cleaning up...")
+	err := ts.runtimeClient.TearDown()
```
mszostok: hm.. `TearDown` on `runtimeClient` looks quite strange. Can we just say 
```
ts.runtimeClient.EnsureUAAInstanceRemoved
```

You already did sth similar for `ts.brokerClient.DeprovisionRuntime()`

tests/e2e/provisioning/pkg/client/kyma/kyma_client.go
```diff
@@ -0,0 +1,69 @@
+package kyma
```
mszostok: IMO this code should be moved to runtime or runtime here. Having runtime and Kyma does not make any sense

tests/e2e/provisioning/pkg/client/kyma/kyma_client.go
mszostok: same comment

tests/e2e/provisioning/pkg/client/kyma/kyma_client.go
```diff
@@ -0,0 +1,69 @@
...
+
+	c.log.Infof("Calling the dashboard URL: %s", targetURL)
+	resp, err := c.client.Get(targetURL)
+	if err != nil {
+		return errors.Wrapf(err, "while calling dashboard '%s'", dashboardURL)
+	}
+	defer c.warnOnError(resp.Body.Close())
+
+	if resp.StatusCode != http.StatusFound {
+		return errors.Errorf("expected status code: %d, got %d", http.StatusFound, resp.StatusCode)
```
mszostok: reading body msg can be helpful for debuging

tests/e2e/provisioning/pkg/client/runtime/runtime_client.go
```diff
@@ -0,0 +1,145 @@
...
+
+	graphCli "github.com/machinebox/graphql"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// accountIDKey is a header key name for request send by graphQL client
+const accountIDKey = "tenant"
```
mszostok: accountIDKey ? why not `tenantHeaderName`?

tests/e2e/provisioning/pkg/client/runtime/runtime_client.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (c *Client) setRuntimeConfig() error {
+	response, err := c.fetchRuntimeConfig()
+	if err != nil {
+		return errors.Wrap(err, "while fetching runtime config")
+	}
+
+	runtimeConfig := *response.Result.RuntimeConfiguration.Kubeconfig
+	runtimeConfigFile := "/tmp/runtime.yaml"
```
mszostok: use https://golang.org/pkg/io/ioutil/#example_TempFile is more safe

tests/e2e/provisioning/pkg/client/runtime/runtime_client.go
```diff
@@ -0,0 +1,145 @@
...
+	err = ioutil.WriteFile(runtimeConfigFile, []byte(runtimeConfig), 0200)
+	if err != nil {
+		return errors.Wrap(err, "while creating runtime kubeconfig file")
+	}
+	err = c.setClientConfig(runtimeConfigFile)
```
mszostok: it should be just create client. It's quite messy with this setting client and then using that in `ensureInstanceRemoved`.  there is no sense of having that in struct in current impl. What's more it is error prone because using `ensureInstanceRemoved` without calling first `setClientConfig` will cause panic

tests/e2e/provisioning/Makefile
ksputo: it needs to be override to skip `go test ./...` while calling verify

tests/e2e/provisioning/pkg/client/kyma/kyma_client.go
polskikiel: I removed a kyma package and moved that struct after renaming it to `dashboard_checker` to the runtime package.

tests/e2e/provisioning/pkg/client/kyma/kyma_client.go
polskikiel: I discussed this with goats and they believe we don't have a better option to assert if UUA works fine. There was one idea to use some http client with js interpreter to just call dashboardURL and behave like a browser but i didn't find anything interesting.

I changed this code accordingly to your proposition.

tests/e2e/provisioning/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+# contains bash and kubectl
+FROM linkyard/kubectl:1.14.2
```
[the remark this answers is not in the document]
polskikiel: Our first functions which provision a runtime has retries which works like ping if the apiserver is not reachable. I've also added `sleep 20` to the container args

tests/e2e/provisioning/scripts/entrypoint.sh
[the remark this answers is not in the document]
polskikiel: We don't use bash anymore

tests/e2e/provisioning/pkg/client/broker/dto.go
polskikiel: removed

tests/e2e/provisioning/test/test_suite.go
polskikiel: we will wait for that

tests/e2e/provisioning/internal/director/client.go
```diff
@@ -0,0 +1,137 @@
...
+	var response successResponse
+	err := dc.graphQLClient.Run(context.Background(), req, &response)
+	if err != nil {
+		return nil, TemporaryError{fmt.Sprintf("while requesting to director client: %s", err)}
+	}
+	return &response, nil
+}
+
+func (dc *Client) setToken() error {
+	lock.Lock()
```
mszostok: why you need that? Currently it is not used

tests/e2e/provisioning/internal/director/client.go
```diff
@@ -0,0 +1,137 @@
...
+	authorizationKey := "Authorization"
+
+	for i := 0; i < reqAttempt; i++ {
+		err := dc.setToken()
+		if err != nil {
+			lastError = err
+			log.Errorf("cannot set token to director client (attempt %d): %s", i, err)
+			continue
+		}
+		req.Header.Add(authorizationKey, fmt.Sprintf("Bearer %s", dc.token.AccessToken))
```
mszostok: nit: updating token should be done based on returned response

tests/e2e/provisioning/internal/director/client.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		// do not wrap error, because type of error (TemporaryError) is important
+		return "", err
+	}
+
+	log.Info("Extract the RuntimeID from the response")
+	return dc.getIDFromRuntime(&response.Result), nil
+}
+
+func (dc *Client) callDirector(req *machineGraph.Request) (*successResponse, error) {
```
mszostok: let's discuss that implementation f2f :)

tests/e2e/provisioning/internal/director/error.go
```diff
@@ -0,0 +1,15 @@
+package director
```
mszostok: this functionality is not used, so why we added it?

tests/e2e/provisioning/pkg/client/broker/broker_client.go
```diff
@@ -0,0 +1,257 @@
...
+	resp, err := c.client.Do(request)
+	if err != nil {
+		return errors.Wrapf(err, "while executing request URL: %s", url)
+	}
+
+	err = json.NewDecoder(resp.Body).Decode(responseBody)
+	if err != nil {
+		return errors.Wrapf(err, "while decoding body")
+	}
+	c.warnOnError(resp.Body.Close())
```
mszostok: this should be moved as defer statement right after success `client.Do`

tests/e2e/provisioning/pkg/client/broker/broker_client.go
```diff
@@ -0,0 +1,257 @@
...
+
+func (c *Client) executeRequest(method, url string, body io.Reader, responseBody interface{}) error {
+	request, err := http.NewRequest(method, url, body)
+	if err != nil {
+		return errors.Wrap(err, "while creating request for provisioning")
+	}
+	request.SetBasicAuth(c.brokerConfig.Auth.Username, c.brokerConfig.Auth.Password)
+	request.Header.Set("X-Broker-API-Version", "2.14")
+
+	resp, err := c.client.Do(request)
```
mszostok: shouldn't you also check the status codes?

PR #737, opened 2020-01-27
Connectivity middleware

dbadura:
**Description**

- add middlewere which get application name from path, fetch from director and put it into context

**Related issue(s)**

See #642


components/connectivity-adapter/internal/appregistry/appdetails/middleware.go
```diff
@@ -1,3 +1,74 @@
...
+	"github.com/kyma-incubator/compass/components/connectivity-adapter/pkg/gqlcli"
+	"github.com/kyma-incubator/compass/components/connectivity-adapter/pkg/reqerror"
+	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
+	"github.com/kyma-incubator/compass/tests/director/pkg/gql"
+	gcli "github.com/machinebox/graphql"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+const appNameVar = "app-name"
```
aszecowka: appNamePathVariable

components/connectivity-adapter/internal/appregistry/appdetails/middleware.go
```diff
@@ -1,3 +1,74 @@
...
+const nameKey = "name"
+
+//go:generate mockery -name=GraphQLRequestBuilder -output=automock -outpkg=automock -case=underscore
+type GraphQLRequestBuilder interface {
```
aszecowka: why your dependency returns graphql query? IMO you should GraphQL should be hidden from middleware and GetApplicationByName (not Applications) should return concrete application

components/connectivity-adapter/internal/appregistry/service/handler.go
```diff
@@ -159,7 +161,31 @@ func (h *Handler) Get(writer http.ResponseWriter, request *http.Request) {
...
+
+	ctx := request.Context()
+	app, err := appdetails.LoadFromContext(ctx)
+	if err != nil {
+		wrappedErr := errors.Wrap(err, "while getting service from context")
+		h.logger.Error(wrappedErr)
+		reqerror.WriteError(writer, wrappedErr, apperrors.CodeInternal)
+		return
+	}
+
```
aszecowka: this method is too long - it will be removed soon, just log application from context.

Issue #675, opened 2020-01-20
Required tenant for global operations


PK85:
**Description**

To create Integration System user must have only proper scope. It was like that till 1.8 Kyma, since 1.9 tenant is required.

**AC**
- remove a need of providing tenant
- check if it works for integrationSystem and tenants queries

**Decision needed**
- if tenant provided we should reject or check and pass or reject?

PR #739, opened 2020-01-27
Allow empty tenant in claims validation

crabtree:
**Description**

- Allow empty tenant in claims validation
- Integration System tests now call API without a tenant ID in the context 

**Related issue(s)**

See also #675


PK85:
To be sure, that change allows empty tenants just in case of integration system right? Apps and runtimes still require that.

crabtree:
> To be sure, that change allows empty tenants just in case of integration system right? Apps and runtimes still require that.

The change just disables the check whether the tenant ID is in the claims of the token in the request to the director. The actual check is done in each of the services that handle the queries and mutations, like here https://github.com/kyma-incubator/compass/blob/master/components/director/internal/domain/application/service.go#L150

Issue #723, opened 2020-01-27
Add test coverage for optional components functionality


mszostok:
**Description**

Add test coverage for optional components functionality introduced in this pull request: https://github.com/kyma-incubator/compass/pull/670

mszostok:
done

PR #749, opened 2020-01-29
Add tests coverage for the opt components functionality

mszostok:
**Description**

- Add tests coverage for the opt components functionality

**Related issue(s)**

- https://github.com/kyma-incubator/compass/issues/723

### DIFF
--- a/components/kyma-environment-broker/internal/runtime/provider.go
+++ b/components/kyma-environment-broker/internal/runtime/provider.go
@@ -24,11 +24,17 @@ const (
 type ComponentsListProvider struct {
 	kymaVersion                      string
 	managedRuntimeComponentsYAMLPath string
+	httpClient                       HTTPDoer
+}
+
+type HTTPDoer interface {
+	Do(req *http.Request) (*http.Response, error)
 }
 
 // NewComponentsListProvider returns new instance of the ComponentsListProvider
 func NewComponentsListProvider(kymaVersion string, managedRuntimeComponentsYAMLPath string) *ComponentsListProvider {
 	return &ComponentsListProvider{
+		httpClient:                       http.DefaultClient,
 		kymaVersion:                      kymaVersion,
 		managedRuntimeComponentsYAMLPath: managedRuntimeComponentsYAMLPath,
 	}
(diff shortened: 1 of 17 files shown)


components/kyma-environment-broker/internal/runtime/provider_test.go
```diff
@@ -0,0 +1,174 @@
...
+				managedRuntimeComponentsYAMLPath: path.Join("testdata", "managed-runtime-components.yaml"),
+				httpErrMessage:                   "Not Found",
+			},
+			expErrMessage: "while getting open source Kyma components list: got unexpected status code, want 200, got 404, url: https://storage.googleapis.com/kyma-development-artifacts/master-123123/kyma-installer-cluster.yaml, body: Not Found",
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			// given
+			fakeHTTPClient := newTestClient(t, tc.given.httpErrMessage, http.StatusNotFound)
```
piotrmiskiewicz: why didn't you use "net/http/httptest" package?

components/kyma-environment-broker/internal/runtime/provider.go
```diff
@@ -24,11 +24,17 @@ const (
...
+	httpClient                       HTTPDoer
+}
+
+type HTTPDoer interface {
+	Do(req *http.Request) (*http.Response, error)
 }
 
 // NewComponentsListProvider returns new instance of the ComponentsListProvider
-func NewComponentsListProvider(kymaVersion string, managedRuntimeComponentsYAMLPath string) *ComponentsListProvider {
+func NewComponentsListProvider(client HTTPDoer, kymaVersion string, managedRuntimeComponentsYAMLPath string) *ComponentsListProvider {
```
piotrmiskiewicz: if there is really no way to do it without injecting http doer please create separate contructor for tests

components/kyma-environment-broker/internal/runtime/provider.go
```diff
@@ -61,11 +67,15 @@ func (r ComponentsListProvider) AllComponents() ([]v1alpha1.KymaComponent, error
 func (r ComponentsListProvider) getOpenSourceKymaComponents() (comp []v1alpha1.KymaComponent, err error) {
```
piotrmiskiewicz: what about pointer receiver?

components/kyma-environment-broker/internal/runtime/provider_test.go
mszostok: the httptest gives you only http server. To use that you will need to inject the URL to provider and in current impl it is a provider responsibility to do that, so I can also assert that 
```go
assert.Equal(t, tc.expectedRequestURL, fakeHTTPClient.RequestURL)
```

So better is to use client and assert real request data

components/kyma-environment-broker/internal/runtime/provider.go
mszostok: I've added dedicated setter available only for tests

PR #750, opened 2020-01-29
Add liveness probe to KEB

ksputo:
**Description**

- Added liveness probe to KEB at `/healthz` endpoint

**Related issue(s)**


chart/compass/charts/kyma-environment-broker/templates/deployment.yaml
```diff
@@ -122,6 +122,14 @@ spec:
...
               containerPort: {{ .Values.broker.port }}
               protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: {{ .Values.broker.statusPort }}
+            failureThreshold: 3
+            periodSeconds: 90
+            timeoutSeconds: 3
+            initialDelaySeconds: 60
```
piotrmiskiewicz: I'd make these values lower (initial delay and period)

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -79,6 +82,9 @@ func main() {
 	logger.RegisterSink(lager.NewWriterSink(os.Stdout, lager.DEBUG))
 	logger.RegisterSink(lager.NewWriterSink(os.Stderr, lager.ERROR))
 
+	logger.Info("Registering healthz endpoint for health probes")
+	go health.NewHandler(fmt.Sprintf("%s:%s", cfg.Host, cfg.StatusPort)).Handle()
```
mszostok: **hit:** IMO "Handle" does not say that it will start server, better for me will be "Server". I will also hide starting the go routine with sth like "ServeAsync". Last thing is to maybe hide the `Sprintf` function inside constructor

`healthz.NewHandler(cfg.Host, cfg.StatusPort).ServeAsync()`

The last problem that I have is that you are ignoring the `http.ListenAndServe`, e.g. we can get error when this port will be already allocated. I'm not quite sure if this should be ignored or not 🤔

components/kyma-environment-broker/internal/health/health.go
```diff
@@ -0,0 +1,28 @@
...
+	return &HealthHandler{Address: addr}
+}
+
+func (h *HealthHandler) Handle() {
+	healthRouter := mux.NewRouter()
+	healthRouter.HandleFunc("/healthz", livenessHandler())
+	http.ListenAndServe(h.Address, healthRouter)
+}
+
+func livenessHandler() func(w http.ResponseWriter, req *http.Request) {
```
mszostok:
```suggestion
func livenessHandler() func(w http.ResponseWriter, _ *http.Request) {
```

**nit:** if you are not using given arg then it makes sense to use underscore so it will be more descriptive that his arg is ignored. Same suggestion for below function

Issue #584, opened 2019-12-18
Tenant storage and Business tenant to compass tenant mapping service


PK85:
**Description**
This ticket is about to make Compass Components to work with external tenants and then internally find a mapping to internal one. It must also work opposite direction when Director calls external TokenService with external tenant. 

**AC**
- allow to store tenants and open only query API to get external tenants. About exposed API discuss with Hasselhoff team to make it nice for UI, see: https://github.com/kyma-project/console/issues/1486
- creation, upgrades and deletion and also mapping between external/internal tenants will be done internally 
- only admin user sees all tenants from query API

dbadura:
WIP: 
1. Create database table with mapping between internal and external tenant (external tenant can be whatever) and we should keep something like external tenant provider name to distinguish tenants.
2. Tenant mapping service use this table to map external tenant to internal tenant
3. When doing request to external token service, we will use external tenant.
4. Add integration tests, which will:
  * register something with external tenant 
  * check if compass will call external TokenService

aszecowka:
After dicussion with KK, in the first iteration we can provide graphQL queries which returns the same data as are already stored in the configmap:
```
  tenants:
    - name: default
      id: 3e64ebae-38b5-46a0-b1ed-9ccee153a0ae
    - name: foo
      id: 1eba80dd-8ff6-54ee-be4d-77944d17b10b
    - name: bar
      id: 9ca034f1-11ab-5b25-b76f-dc77106f571d

```

crabtree:
I'm missing the information where the actual implementation of tenants resolver will be done.  We also need to think about what scopes need to be assigned to call this query.

PR #792, opened 2020-02-04
Migrate db

kjaksik:
**Description**

- Add constraints to `tenant_id`

**Related issue(s)**
#584


components/schema-migrator/migrations/director/202001311508_migrate_tenants.up.sql
```diff
@@ -0,0 +1,29 @@
+
+insert into business_tenant_mappings(id)
+select tenant_id from applications a2 union select tenant_id from runtimes r2;
+update business_tenant_mappings set external_tenant = id;
+update business_tenant_mappings set external_name = 'Tenant';
```
dbadura: Add Where condition, because you will change all `extenal_name` and `external_tenant`.

components/schema-migrator/migrations/director/202001311508_migrate_tenants.down.sql
```diff
@@ -0,0 +1,23 @@
+
+alter table api_definitions
```
dbadura: Apply upper case on all SQLs

PR #800, opened 2020-02-06
Bootstrap Packages GraphQL API mocks

pkosiec:
**Description**

- Bootstrap final Packages GraphQL API

**Related issue(s)**
#803


components/director/pkg/graphql/schema.graphql
```diff
@@ -871,5 +995,16 @@ type Mutation {
...
 	deleteDefaultEventingForApplication(appID: String!): ApplicationEventingConfiguration! @hasScopes(path: "graphql.mutation.deleteDefaultEventingForApplication")
+	"""
+	When APIInstanceAuth is not in pending state, the operation returns error.
+	
+	When used without error, the status of pending auth is set to success.
+	"""
+	setAPIInstanceAuthForPackage(packageID: ID!, authID: ID!, in: AuthInput! @validate): APIInstanceAuth! @hasScopes(path: "graphql.mutation.setAPIInstanceAuthForPackage")
+	deleteAPIInstanceAuthForPackage(packageID: ID!, authID: ID!): APIInstanceAuth! @hasScopes(path: "graphql.mutation.deleteAPIInstanceAuthForPackage")
+	requestAPIInstanceAuthForPackage(packageID: ID!, in: APIInstanceAuthRequestInput!): APIInstanceAuth! @hasScopes(path: "graphql.mutation.requestAPIInstanceAuthForPackage")
+	addPackageDefinition(applicationID: ID!, in: PackageDefinitionCreateInput! @validate): PackageDefinition! @hasScopes(path: "graphql.mutation.addPackageDefinition")
```
aszecowka: to discuss: `PackageDefinition` vs `Package`

components/director/pkg/graphql/schema.graphql
```diff
@@ -533,6 +620,31 @@ type OneTimeTokenForRuntime implements OneTimeToken {
 	rawEncoded: String
 }
 
+type PackageDefinition {
+	id: ID!
+	name: String!
+	description: String
+	authRequestJSONSchema: JSONSchema
```
aszecowka: to discuss `authRequestJSONSchema` --> `instanceParametersJSONSchema`

components/director/pkg/graphql/schema.graphql
```diff
@@ -533,6 +620,31 @@ type OneTimeTokenForRuntime implements OneTimeToken {
...
+	auth(id: ID!): APIInstanceAuth
+	auths: [APIInstanceAuth!]!
+	"""
+	When defined, all Auth requests fallback to defaultAuth.
+	"""
+	defaultAuth: Auth
```
aszecowka: if `defaultAuth` applies only to API, maybe it should be called: `apisDefaultAuth`

components/director/pkg/graphql/schema.graphql
```diff
@@ -783,6 +895,10 @@ type Mutation {
 	"""
 	addAPIDefinition(applicationID: ID!, in: APIDefinitionInput! @validate): APIDefinition! @hasScopes(path: "graphql.mutation.addAPIDefinition")
 	"""
+	Temporary name before doing breaking change
```
aszecowka: mark that `addAPIDefinition` will be changed and at the end we will have only `addAPIDefinitionToPackage`.
The same for `addEventDefinition`

components/director/pkg/graphql/schema.graphql
```diff
@@ -315,6 +360,38 @@ type APIDefinitionPage implements Pageable {
 	totalCount: Int!
 }
 
+type APIInstanceAuth {
```
aszecowka: maybe `APIAuthPerRuntime` or `ApiAuthOnRuntime`?

components/director/pkg/graphql/schema.graphql
```diff
@@ -533,6 +620,31 @@ type OneTimeTokenForRuntime implements OneTimeToken {
 	rawEncoded: String
 }
 
+type PackageDefinition {
+	id: ID!
+	name: String!
+	description: String
+	authRequestJSONSchema: JSONSchema
+	auth(id: ID!): APIInstanceAuth
+	auths: [APIInstanceAuth!]!
```
aszecowka: `auths` --> maybe `runtimeAuths`?

components/director/pkg/graphql/schema.graphql
pkosiec: what do you think about `APIInstanceAuthRequestJSONSchema`?

components/director/pkg/graphql/schema.graphql
pkosiec: hmm, we have API and Event Definitions, so... However, I'm fine with the change.

components/director/pkg/graphql/schema.graphql
pkosiec: Hmm... okay, but then the `auths` field would need to be called `apisAuths`. It is fine for me, but I feel the naming would be too complicated. apisAuths would suggest that auth is created per single API Definition, not per Package. 

I would keep `defaultAuth` and `auths`, as they are on Package level.

components/director/pkg/graphql/schema.graphql
pkosiec: I agree that the APIInstanceAuth is not great, but `Runtime` in the name would be confusing. We will have auth per Service Instance, which means multiple auths on a single Runtime. The name with `Runtime` suggests otherwise.

I was thinking about Usage/Instance names, but I don't have any better alternative for now.

components/director/pkg/graphql/schema.graphql
pkosiec: As in the comment above, runtime would be not proper word for that. But instance/usage... or even `apiInstanceAuths` or `apiUsageAuths`... I have mixed feelings for all combinations...

components/director/pkg/graphql/schema.graphql
pkosiec: okay, then `authRequestInputSchema` will be the final name

Issue #774, opened 2020-01-31
Connectivity Adapter doesn't call Director to obtain Application name and Events base URL on signingRequest/info endpoint


akgalwas:
**Description**

signingRequest/info endpoint doesn't return correct paths to Event Service and Application Registry. According to [Kyma documentation](https://github.com/kyma-project/kyma/blob/master/docs/application-connector/08-02-get-client-certificate.md), management/info is the recommended way for obtaining services urls. But there could be clients that do not adhere to this guidelines and use signingRequest/info to get urls fore accessing Application Registry and Event Service.  

**Expected result**

Application Registry and Event Service urls returned from signingRequest/info are the same as urls returned from management/info. Mind that signingRequests/info is secured with token so in order to call Director, Connectivity Adapter should be registered as an integration system.
 

**Actual result**

```
{
    "api": {
        "certificatesUrl": "https://adapter-gateway.kyma.local/v1/applications/certificates",
        "eventsInfoUrl": "/<system Auth ID>/v1/events/subscribed",
        "eventsUrl": "/<system Auth ID>/v1/events",
        "infoUrl": "https://adapter-gateway-mtls.kyma.local/v1/applications/management/info",
        "metadataUrl": "https://adapter-gateway-mtls.kyma.local/<system Auth ID>/v1/metadata/services"
    },
    "certificate": {
        "extensions": "",
        "key-algorithm": "rsa2048",
        "subject": "O=Org,OU=OrgUnit,L=locality,ST=province,C=DE,CN=<system Auth ID>"
    },
    "csrUrl": "https://adapter-gateway.kyma.local/v1/applications/certificates?token=<token>"
}
```

Please notice, that metadataUrl, eventsUrl and eventsInfo url contain system auth ID whereas the application name should be returned. Additionally eventsUrl and eventsInfo base path is empty.

**Steps to reproduce**

1) Generated token from UI
2) Call the following endpoint with the token
`curl https://adapter-gateway.kyma.local/v1/applications/signingRequests/info\?token`

franpog859:
Before closing the issue make sure that both the implementation and the documentation are provided!

PR #801, opened 2020-02-06
Correct Connectivity Adapter signingRequest/info response

franpog859:
**Description**

- Modify tenant mapping in Director to support one time token flow
- Use Application name in the `signingRequest/info` endpoint response URLs
- Fill URLs corresponding to events in the `signingRequest/info` endpoint response

**Related issue(s)**

See the issue #774


components/connectivity-adapter/internal/connectorservice/model/model.go
```diff
@@ -71,14 +71,46 @@ type TokenResponse struct {
 	Token string `json:"token"`
 }
 
-func MakeCSRURL(newToken, connectivityAdapterBaseURL string) string {
+func MakeCSRInfoResponseWrapper(
```
pkosiec: what do you think about `NewCSRInfoResponse` prefix instead of `MakeCSRInfoResponseWrapper `?

components/connectivity-adapter/internal/connectorservice/director/client.go
```diff
@@ -31,8 +28,6 @@ type client struct {
 }
 
 func (c client) GetApplication(systemAuthID string) (schema.ApplicationExt, apperrors.AppError) {
-
-	logrus.Info("System auth ID: " + systemAuthID)
```
pkosiec: Please keep the log, it may be useful to debug some problems

franpog859: I think `systemAuthID` could be considered as sensitive data. I wanted to delete it then. You could always add it, rebuild an image and use it to debug your new implementation

components/connectivity-adapter/internal/connectorservice/model/model.go
franpog859: Yeah, the code is redundant. I will fix it!

components/connectivity-adapter/internal/connectorservice/model/model.go
```diff
@@ -71,14 +71,46 @@ type TokenResponse struct {
...
+	applicationName,
+	newToken,
+	connectivityAdapterBaseURL,
+	connectivityAdapterMTLSBaseURL,
+	eventServiceBaseURL string,
+	certInfo CertInfo) interface{} {
+
+	return makeCSRInfoResponse(
```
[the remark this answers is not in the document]
franpog859: Yeah, the code is redundant. I will fix it!

components/connectivity-adapter/internal/connectorservice/model/model.go
franpog859: Sounds cool. I'll change it ^^

PR #824, opened 2020-02-11
Group user login

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**


components/director/cmd/director/main.go
```diff
@@ -164,21 +165,25 @@ func configureLogger() {
...
+func getTenantMappingHanderFunc(transact persistence.Transactioner, staticUsersSrc string, staticGroupsSrc string, scopeProvider *scope.Provider) (func(writer http.ResponseWriter, request *http.Request), error) {
 	uidSvc := uid.NewService()
 	authConverter := auth.NewConverter()
 	systemAuthConverter := systemauth.NewConverter(authConverter)
 	systemAuthRepo := systemauth.NewRepository(systemAuthConverter)
 	systemAuthSvc := systemauth.NewService(systemAuthRepo, uidSvc)
-	staticUsersRepo, err := tenantmapping.NewStaticUserRepository(staticUsersSrc)
-	if err != nil {
-		return nil, errors.Wrap(err, "while creating StaticUser repository instance")
+	staticUsersRepo, errUsers := tenantmapping.NewStaticUserRepository(staticUsersSrc)
```
crabtree: I'd go with:
```go
	staticUsersRepo, err := tenantmapping.NewStaticUserRepository(staticUsersSrc)
	if err != nil {
                return nil, errors.Wrap(err, "while creating StaticUser repository instance")
	}
	staticGroupsRepo, err := tenantmapping. NewStaticGroupRepository(staticGroupsSrc)
	if err != nil {
		return nil, errors.Wrap(err, "while creating StaticGroup repository instance")
	}
```

components/director/internal/tenantmapping/mapper_for_user.go
```diff
@@ -6,55 +6,110 @@ import (
...
 
 type mapperForUser struct {
-	staticUserRepo StaticUserRepository
-	tenantRepo     TenantRepository
+	staticUserRepo  StaticUserRepository
+	staticGroupRepo StaticGroupRepository
+	tenantRepo      TenantRepository
+}
+
+func stringInSlice(a string, list []string) bool {
```
crabtree: Please move this func to `pkg/str` as this can be reused in different places

components/director/internal/tenantmapping/mapper_for_user.go
```diff
@@ -6,55 +6,110 @@ import (
...
+const tenantGroupPrefix = "tenantID="
+
+// GetGroupScopesAndTenant get all scopes from group array, without duplicates
+func GetGroupScopesAndTenant(groups []StaticGroup) (string, string) {
+	var filteredScopesArray []string
+	var tenantID string
+
+	for _, group := range groups {
+		if strings.HasPrefix(group.GroupName, tenantGroupPrefix) {
+			// group name is tenantID
```
crabtree: you don't need such comments, you can extract this 'if logic' to some func with a descriptive name

components/director/internal/tenantmapping/mapper_for_user.go
```diff
@@ -6,55 +6,110 @@ import (
...
+			tenantID = strings.Replace(group.GroupName, tenantGroupPrefix, "", 1)
+		} else {
+			// group name is
+			for _, scope := range group.Scopes {
```
crabtree: That does not look very efficient. Maybe you could work on a map and just check if the key is already in the map?

components/director/internal/tenantmapping/mapper_for_user.go
```diff
@@ -6,55 +6,110 @@ import (
...
 	}
 
-	externalTenantID, err = reqData.GetExternalTenantID()
-	if err != nil {
-		if !apperrors.IsKeyDoesNotExist(err) {
-			return ObjectContext{}, errors.Wrap(err, "while fetching external tenant")
+	if len(staticGroups) > 0 {
+		// proceed with group scopes flow
+		scopes, externalTenantID = GetGroupScopesAndTenant(staticGroups)
+		log.Infof("Decided to use groups copes %s\n", scopes)
```
crabtree: If you really need this info, then fix the typo ;)

components/director/internal/tenantmapping/mapper_for_user.go
```diff
@@ -6,55 +6,110 @@ import (
...
+	} else {
```
crabtree: What if the user is a static user but also sends you groups that are not assigned to any scopes?

components/director/internal/tenantmapping/mapper_for_user.go
```diff
@@ -6,55 +6,110 @@ import (
...
+		}
 	}
 
 	tenantMapping, err := m.tenantRepo.GetByExternalTenant(ctx, externalTenantID)
 	if err != nil {
 		return ObjectContext{}, errors.Wrapf(err, "while getting external tenant mapping [ExternalTenantId=%s]", externalTenantID)
 	}
 
-	if !hasValidTenant(staticUser.Tenants, tenantMapping.ExternalTenant) {
+	if len(userGroups) <= 0 && !hasValidTenant(staticUser.Tenants, tenantMapping.ExternalTenant) {
```
crabtree: Why you check here the user groups length? And the length of the slice should never be less than 0.

components/director/internal/tenantmapping/mapper_for_user_test.go
```diff
@@ -45,10 +47,14 @@ func TestMapperForUserGetObjectContext(t *testing.T) {
 		staticUserRepoMock := getStaticUserRepoMock()
 		staticUserRepoMock.On("Get", username).Return(staticUser, nil).Once()
 
+		//staticGroup := []tenantmapping.StaticGroup{}
```
crabtree: Please remove the unused code.

components/director/internal/tenantmapping/mapper_for_user_test.go
```diff
@@ -314,3 +320,8 @@ func getStaticUserRepoMock() *automock.StaticUserRepository {
 	repo := &automock.StaticUserRepository{}
 	return repo
 }
+
+func getStaticGroupRepoMock() *automock.StaticGroupRepository {
```
crabtree: Where this func is used?

components/director/internal/tenantmapping/model.go
```diff
@@ -126,6 +128,34 @@ func (d *ReqData) GetScopes() (string, error) {
 	return "", apperrors.NewKeyDoesNotExistError(ScopesKey)
 }
 
+// GetGroups returns group name or empty string if there's no group
+func (d *ReqData) GetGroups() ([]string, error) {
+	if groupsVal, ok := d.Body.Extra[GroupsKey]; ok {
+
+		groupsArray, arrayConvertedOk := groupsVal.([]interface{})
```
crabtree: Not sure, but maybe you can try to convert to []string directly?

components/director/internal/tenantmapping/model.go
```diff
@@ -126,6 +128,34 @@ func (d *ReqData) GetScopes() (string, error) {
...
+			groupString, stringConversionErr := str.Cast(group)
+			if stringConversionErr != nil {
+				log.Infof("%s skipped because string conversion failed", group)
+				continue
+			}
+
+			groupsFiltered = append(groupsFiltered, groupString)
+		}
+
+		if !arrayConvertedOk || len(groupsFiltered) <= 0 {
```
crabtree: again, the slice length should not be less than zero

components/director/internal/tenantmapping/mapper_for_user.go
parostatkiem: Please let me know if the provided update is efficient enough.

components/director/internal/tenantmapping/mapper_for_user.go
parostatkiem: That's a nice thing to discuss! 🗣

components/director/internal/tenantmapping/mapper_for_user_test.go
parostatkiem: Tests are not finished yet

components/director/internal/tenantmapping/mapper_for_user.go
akucharska: We are taking either group or user. So if someone is logging with static user (but also has some random group without scopes), we are treating him as static group user without scopes

components/director/internal/tenantmapping/model.go
akucharska: It doesn't allow to convert directly to []string.

components/director/internal/tenantmapping/mapper_for_user_test.go
akucharska: in tests

components/director/internal/tenantmapping/model.go
crabtree: Ok, I see. How about this implementation?

```go
	userGroups := []string{}
	groupsVal, ok := d.Body.Extra[GroupsKey]
	if !ok {
		return userGroups
	}

	if groupsArray, ok := groupsVal.([]interface{}); ok {
		for _, group := range groupsArray {
			groupString, err := str.Cast(group)
			if err != nil {
				log.Infof("%+v skipped because string conversion failed", group)
				continue
			}

			userGroups = append(userGroups, groupString)
		}
	}

	return userGroups
```

components/director/internal/tenantmapping/model_test.go
```diff
@@ -238,3 +238,44 @@ func TestReqData_GetScopes(t *testing.T) {
...
+		groups := reqData.GetUserGroups()
+
+		require.Equal(t, expectedGroups, groups)
+	})
+
+	t.Run("returns empty array when groups value is specified in Extra map in a non-string format", func(t *testing.T) {
+		reqData := ReqData{
+			Body: ReqBody{
+				Extra: map[string]interface{}{
+					ScopesKey: []byte{1, 2, 3},
```
crabtree: According to the name of the test, shouldn't this use `GroupsKey`?

PR #829, opened 2020-02-11
Fix for pairing issue

akgalwas:
**Description**

- Added normalization in Connector query to allow providing CSR with white characters

### DIFF
--- a/components/connectivity-adapter/internal/connectorservice/connector/query.go
+++ b/components/connectivity-adapter/internal/connectorservice/connector/query.go
@@ -1,6 +1,10 @@
 package connector
 
-import "fmt"
+import (
+	"fmt"
+	"strings"
+	"unicode"
+)
 
 type queryProvider struct{}
 
@@ -23,7 +27,7 @@ func (qp queryProvider) signCSR(csr string) string {
 		caCertificate
 		clientCertificate
 	}
-    }`, csr)
+    }`, removeWhiteChars(csr))
 }
 
 func (qp queryProvider) token(application string) string {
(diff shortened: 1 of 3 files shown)


components/connectivity-adapter/internal/connectorservice/connector/query.go
```diff
@@ -39,5 +43,14 @@ func (qp queryProvider) revoke() string {
 	return `mutation {
     result: revokeCertificate 
 	}`
-	//
+}
+
+func removeWhiteChars(s string) string {
+	return strings.Map(func(r rune) rune {
```
pkosiec: You can replace it with
```go
strings.Replace(s, " ", "", -1)
```
We will save a few lines of code 🙂

akgalwas: EC sends csr string containing \n and \r so that we need to replace all white characters (spaces, tabs, new lines etc.). 
That is true that removeWhiteChars could be implemented with Replace function, however, the imlementation should take care of different types of white characters.

Here are the possible implementations:
```
func removeWhiteChars1(s string) string{
   s1 := strings.ReplaceAll(s, " ", "")
   s1 = strings.ReplaceAll(s1, "\n", "")
   s1 = strings.ReplaceAll(s1, "\r", "")
   s1 = strings.ReplaceAll(s1, "\t", "")

  return s1
}

func removeWhiteChars2(s string) string {
  whiteChars := []rune{' ', '\n', '\r', '\t'}
  s1 := s
  for _, c := range whiteChars {
     s1 = strings.ReplaceAll(s1, string(c), "")	   	
  }
  
  return s1
}
```

Nice thing about `unicode.IsSpace` is that it is capable of detecting all types of white characters. According to the documentation:

> IsSpace reports whether the rune is a space character as defined by Unicode's White Space property; in the Latin-1 space this is '\t', '\n', '\v', '\f', '\r', ' ', U+0085 (NEL), U+00A0 (NBSP).

I think the approach with Map function in quite readable and meets requirements.

PR #850, opened 2020-02-18
Adapting provisioning broker endpoint to operations

jasiu001:
**Description**

Provisioning operation proposal


components/kyma-environment-broker/internal/process/manager.go
```diff
@@ -0,0 +1,45 @@
...
+
+func NewManager() *Manager {
+	return &Manager{}
+}
+
+func (m *Manager) AddStep(step Step) {
+	m.steps = append(m.steps, step)
+}
+
+func (m *Manager) Execute(operationID string) (error, time.Duration) {
```
piotrmiskiewicz: we need some logging stuff here

components/kyma-environment-broker/internal/process/provisioning/runtime_status.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil, 1 * time.Minute
+	}
+
+	return nil, 0
+}
+
+func (s *RuntimeStatusStep) checkInstanceOutdated(instance *internal.Instance) (error, time.Duration) {
+	addTime := instance.CreatedAt.Add(delayInstanceTime)
+	subTime := time.Now().Sub(addTime)
+
```
piotrmiskiewicz: you can use time.Since(instance.CreatedAt)

piotrmiskiewicz: let's check if the operation is working too long at the beginning of processing - as a first step or just before loop over steps

components/kyma-environment-broker/internal/process/manager.go
```diff
@@ -0,0 +1,45 @@
...
+	operation, err := m.operationStorage.GetProvisioningOperationByID(operationID)
+	if err != nil {
+		return nil, 10 * time.Minute
+	}
+
+	for _, step := range m.steps {
+		err, when := step.Run(operation)
```
piotrmiskiewicz: consider returning modified operation as a result of step.Run() because the step could modify the operation. On the other hand the Run method could get value instead of pointer

components/kyma-environment-broker/internal/process/provisioning/create_runtime.go
```diff
@@ -0,0 +1,111 @@
...
+		log.Errorf("[%s] Call to provision failed: %s", operation.ID, err)
+		return nil, 1 * time.Minute
+	}
+	if resp.RuntimeID == nil {
+		// TODO: return error or time.Duration ?
+		return nil, 10 * time.Minute
+	}
+
+	operation.ProvisionerOperationID = *resp.ID
+	_, err = s.operationStorage.UpdateProvisioningOperation(*operation)
```
piotrmiskiewicz: it is very important to store the ID. consider retries here (for example  wait.PollImmediate)

components/kyma-environment-broker/internal/process/provisioning/create_runtime.go
```diff
@@ -0,0 +1,111 @@
...
+	pp, err := operation.GetProvisioningParameters()
+	if err != nil {
+		return errors.Wrap(err, "while getting provisioning parameters from operation"), 0
+	}
+	rawParameters, err := json.Marshal(pp.Parameters)
+	if err != nil {
+		return errors.Wrap(err, "while marshaling instance parameters"), 0
+	}
+
+	input, err := s.createProvisionInput(pp)
```
piotrmiskiewicz: in the near future we will need to pass here also other fields from ProvisioningOperation instance

components/kyma-environment-broker/internal/process/provisioning/create_runtime.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return s.operationManager.OperationFailed(operation, "invalid operation data - cannot create provisioning input")
+	}
+
+	var provisionerResponse gqlschema.OperationStatus
+	if operation.ProvisionerOperationID == "" {
+		provisionerResponse, err := s.provisionerClient.ProvisionRuntime(pp.ErsContext.GlobalAccountID, requestInput)
+		if err != nil {
+			log.Errorf("call to provisioner failed: %s", err)
+			return operation, 1 * time.Minute, nil
```
piotrmiskiewicz: let's retry after few seconds

components/kyma-environment-broker/internal/process/provisioning/create_runtime.go
```diff
@@ -0,0 +1,163 @@
...
+			log.Errorf("cannot save operation ID from provisioner")
+			return operation, 1 * time.Minute, nil
+		}
+	}
+
+	if provisionerResponse.RuntimeID == nil {
+		provisionerResponse, err = s.provisionerClient.RuntimeOperationStatus(pp.ErsContext.GlobalAccountID, operation.ProvisionerOperationID)
+		if err != nil {
+			log.Errorf("call to provisioner about operation status failed: %s", err)
+			return operation, 5 * time.Minute, nil
```
piotrmiskiewicz: maybe 1 or 2 minutes?

components/kyma-environment-broker/internal/process/provisioning/create_runtime.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+	if provisionerResponse.RuntimeID == nil {
+		return operation, 5 * time.Minute, nil
```
piotrmiskiewicz: 1 or 2 minutes?

components/kyma-environment-broker/internal/broker/instance_create.go
```diff
@@ -52,79 +55,120 @@ func NewProvision(cfg Config, instancesStorage storage.Instances, builderFactory
...
+	// add new operation to queue
+	b.queue.Add(operation.ID)
+
+	return domain.ProvisionedServiceSpec{
+		IsAsync:       true,
+		OperationData: operation.ID,
+	}, nil
+}
+
+func (b *ProvisionEndpoint) validate(details domain.ProvisionDetails) (internal.ERSContext, internal.ProvisioningParametersDTO, error) {
```
piotrmiskiewicz: it is not only about validating, it also extracts so the name 'validate' does not describe what happens

Issue #806, opened 2020-02-07
Implement GraphQL API for Package Instance Auths


pkosiec:
**Description**

Implement GraphQL API for Package Instance Auths.

Mutations:
- [x] RequestPackageInstanceAuthCreation @tgorgol 
- [x] RequestPackageInstanceAuthDeletion @pkosiec 
- [x] SetPackageInstanceAuth @tgorgol 
- [x] DeletePackageInstanceAuth @pkosiec 

Resolvers:
- [x] Package.InstanceAuth @kfurgol 
- [x] Package.InstanceAuths @kfurgol 

**Related issues**
See also https://github.com/kyma-incubator/compass/issues/429

PR #851, opened 2020-02-19
Validate Package inputs

pkosiec:
**Description**

**Related issue(s)**
#806

### DIFF
--- a/components/director/pkg/graphql/package_validation.go
+++ b/components/director/pkg/graphql/package_validation.go
@@ -2,6 +2,7 @@ package graphql
 
 import (
 	"fmt"
+
 	validation "github.com/go-ozzo/ozzo-validation"
 	"github.com/go-ozzo/ozzo-validation/is"
 	"github.com/kyma-incubator/compass/components/director/pkg/inputvalidation"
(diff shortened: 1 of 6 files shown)


components/director/pkg/graphql/package_validation.go
```diff
@@ -1,11 +1,54 @@
...
+	if i.Auth == nil && i.Status == nil {
+		return errors.New("at least one field (Auth or Status) has to be provided")
+	}
+
+	if i.Auth != nil && i.Status != nil && i.Status.Condition != PackageInstanceAuthSetStatusConditionInputSucceeded {
+		return fmt.Errorf("status condition has to be equal to %s when the auth is provided", PackageInstanceAuthSetStatusConditionInputSucceeded)
+	}
+
+	return validation.ValidateStruct(&i,
+		validation.Field(&i.Status, validation.NilOrNotEmpty),
```
tgorgol: We're missing some validation rules for `PackageInstanceAuthStatusInput`. It seems like in the current approach we're not validating if status message and reason are set when the status condition is "FAILED".

Issue #814, opened 2020-02-07
KEB manages logging tenants


piotrmiskiewicz:
**Description**

Logging tenant must be created only one per tenant. Creating logging teant is a long running operation.

AC:
 - logging system client 
 - fake client
 - provide a service (with storage) which allows registering logging tenant if not yet created
 - DB migration script
 - provisioning operation step(s) which interacts with LMS  

The LMS API contract is changing - the request certificate call will not contain certID, the ID will be generated and provided by LMS

piotrmiskiewicz:
What is missing:
 - The LMS contract will be changed, the RequestCertificate payload won't contain certId, the ID will be in the response

piotrmiskiewicz:
https://github.com/kyma-incubator/compass/pull/850

PR #857, opened 2020-02-20
Lms integration

piotrmiskiewicz:
**Description**

- LMS client
- LMS Tenant management

**Related issue(s)**
Related: https://github.com/kyma-incubator/compass/issues/814


components/kyma-environment-broker/internal/broker/instance_create.go
```diff
@@ -172,3 +172,7 @@ func (b *ProvisionEndpoint) handleExistingOperation(operation *internal.Provisio
 	log := fmt.Sprintf("provisioning operation with InstanceID %s already exist", operation.InstanceID)
 	return domain.ProvisionedServiceSpec{}, apiresponses.NewFailureResponse(err, http.StatusConflict, log)
 }
+
+func process() {
```
piotrmiskiewicz: not needed

components/kyma-environment-broker/internal/lms/cert_step.go
```diff
@@ -0,0 +1,170 @@
...
+
+	pp, err := operation.GetProvisioningParameters()
+
+	s.inputCreator.SetOverrides("logging", []*gqlschema.ConfigEntryInput {
+		{Key: "fluent-bit.conf.Service.Flush", Value: "30"},
+		{Key: "fluent-bit.conf.Output.Elasticsearch.enabled", Value: "true"},
+		{Key: "fluent-bit.backend.es.host", Value: host},
+		{Key: "fluent-bit.backend.es.port", Value: "443"},
+		// use cluster name provided by ERS as a prefix
+		{Key: "fluent-bit.backend.es.logstash_prefix", Value: pp.Parameters.Name},
```
piotrmiskiewicz: can we have there whitespaces? should be checked

chart/compass/charts/kyma-environment-broker/templates/secrets.yaml
```diff
@@ -26,4 +26,14 @@ data:
   username: {{ .Values.serviceManager.username | b64enc | quote }}
   password: {{ .Values.serviceManager.password | b64enc | quote }}
 {{- end }}
```
jasiu001: this "end" should be remove, the new "end" is at the end of file

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -104,6 +107,15 @@ func main() {
...
 	}
 
+	logs := logrus.New()
+
+	// LMS
+	lmsClient := lms.NewClient(cfg.LMS, logs)
+	lmsTenantStorage := db.LMSTenants()
+	lmsTenantManager := lms.NewTenantManager(lmsTenantStorage, lmsClient, logs.WithField("service", "lmsTenantManager"))
+	lmsProvideTenantStep := provisioning.NewProvideLmsTenantStep(lmsTenantManager, db.Operations())
+	lmsCertStep := provisioning.NewLmsCertificatesStep(lmsClient, db.Operations())
```
jasiu001: nit: maybe move step creating to lines with other steps?

components/kyma-environment-broker/internal/lms/client.go
```diff
@@ -0,0 +1,398 @@
...
+	samlTenant string
+
+	log logrus.FieldLogger
+}
+
+const (
+	EnvironmentDev  Environment = "dev"
+	EnvironmentProd Environment = "prod"
+
+	ClusterTypeHA         ClusterType = "ha"
```
jasiu001: unused, will it be implemented in the future?

components/kyma-environment-broker/internal/lms/client.go
```diff
@@ -0,0 +1,398 @@
...
+}
+
+func NewClient(cfg Config, log logrus.FieldLogger) Client {
+	if cfg.FakeClient {
+		return NewFakeClient(15 * time.Second)
+	} else {
+		return &client{
+			url:         cfg.URL,
+			clusterType: cfg.ClusterType,
+			environment: Environment(cfg.Environment),
```
jasiu001: `cfg.Environment` is already `Environment` type

components/kyma-environment-broker/internal/lms/client.go
```diff
@@ -0,0 +1,398 @@
...
+	Token      string
+	SamlTenant string
+
+	Disabled   bool
+	FakeClient bool
+}
+
+func NewClient(cfg Config, log logrus.FieldLogger) Client {
+	if cfg.FakeClient {
+		return NewFakeClient(15 * time.Second)
```
jasiu001: Why is this condition here? fake or not client is controlled by ENV?

piotrmiskiewicz: we can start KEB with fake lms client, but if you think it is not a good idea (makes the code smell) - I cen remove it. It was created for testing purpose

components/kyma-environment-broker/internal/lms/client.go
```diff
@@ -0,0 +1,398 @@
...
+
+	body, _ := ioutil.ReadAll(resp.Body)
+	// todo: handle resp
+	logrus.Infof("got %s", body)
+
+	// todo: read certId returned from LMS
+	return fmt.Sprintf("%d", certID), privateKey, nil
+}
+
+var oidEmailAddress = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 9, 1}
```
jasiu001: unused?

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -104,6 +107,15 @@ func main() {
 		fatalOnError(err)
 	}
 
+	logs := logrus.New()
+
+	// LMS
+	lmsClient := lms.NewClient(cfg.LMS, logs)
+	lmsTenantStorage := db.LMSTenants()
+	lmsTenantManager := lms.NewTenantManager(lmsTenantStorage, lmsClient, logs.WithField("service", "lmsTenantManager"))
```
jasiu001: IMO: `db.LMSTenants()` can be use directly as an argument

components/kyma-environment-broker/internal/process/provisioning/lms_tenant_step.go
```diff
@@ -0,0 +1,101 @@
...
+
+	pp, err := operation.GetProvisioningParameters()
+	if err != nil {
+		logger.Errorf("Unable to get provisioning parameters", err.Error())
+		return operation, 0, errors.New("unable to get provisioning parameters")
+	}
+	region := s.provideRegion(pp.Parameters.Region)
+	tenant := pp.ErsContext.GlobalAccountID
+
+	name := tenant
```
jasiu001: you can just assign ID to name variable: `name := pp.ErsContext.GlobalAccountID`
`tenat` variable is not use below

components/kyma-environment-broker/internal/process/provisioning/lms_tenant_step.go
```diff
@@ -0,0 +1,101 @@
...
+		operation.Lms.RequestedAt = time.Now()
+	}
+	op, updateErr := s.repo.UpdateProvisioningOperation(operation)
+
+	if updateErr != nil {
+		return operation, time.Second, nil
+	}
+	operation = *op
+
+	if err != nil {
```
jasiu001: to which action this error relates?

components/kyma-environment-broker/internal/process/provisioning/lms_tenant_step.go
```diff
@@ -0,0 +1,101 @@
...
+		logger.Errorf("Unable to request for LMS tenant ID: %s", err.Error())
+		return operation, 5 * time.Second, nil
+	}
+
+	operation.Lms.TenantID = lmsTenantID
+	if operation.Lms.RequestedAt.IsZero() {
+		operation.Lms.RequestedAt = time.Now()
+	}
+	op, updateErr := s.repo.UpdateProvisioningOperation(operation)
```
jasiu001: you can use a helper function `UpdateOperation` from `OperationManager`:
```
operation, repeat := s.operationManager.UpdateOperation(operation)
if repeat != 0 {
   log.Errorf("cannot save LMS tenant")
   return operation, time.Second, nil
}
```

components/kyma-environment-broker/internal/process/provisioning/lms_tenant_step.go
```diff
@@ -0,0 +1,101 @@
+package provisioning
```
jasiu001: missing test, I know the step is simple and short but all our steps has tests :)

components/kyma-environment-broker/internal/process/provisioning/lms_tenant_step.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (s *provideLmsTenantStep) Name() string {
+	return "Create_LMS_Tenant"
+}
+
+func (s *provideLmsTenantStep) Run(operation internal.ProvisioningOperation, logger logrus.FieldLogger) (internal.ProvisioningOperation, time.Duration, error) {
+	if operation.Lms.TenantID != "" {
+		return operation, 0, nil
+	}
+
```
jasiu001: please, consider timeout for the whole step, you can use the helper method `RetryOperation` from `OperationManager`. Call to `ProvideLMSTenantID` may never work out.

components/kyma-environment-broker/internal/process/provisioning/lms_cert_step.go
```diff
@@ -0,0 +1,187 @@
...
+		if time.Since(operation.Lms.RequestedAt) > lmsTimeout {
+			logger.Error("Setting LMS operation failed - tenant provisioning timed out, last error: %s", err.Error())
+			return s.failLmsAndUpdate(operation)
+		}
+		return operation, tenantReadyRetryInterval, nil
+	}
+	host := ti.DNS
+
+	// request certificates
+	pp, err := operation.GetProvisioningParameters()
```
jasiu001: I know this step is not first, but I suggest make `GetProvisioningParameters` first, because if parameters are wrong then all calls to LmsClient are useless :) step one way or another will fail

components/kyma-environment-broker/internal/process/provisioning/lms_tenant_step.go
piotrmiskiewicz: not needed, I'll remove id

components/kyma-environment-broker/internal/process/provisioning/lms_tenant_step.go
piotrmiskiewicz: The step is tested in TestLmsStepsHappyPath (file: lms_cert_step_test.go). This test combines all 2 lms steps so it tests the LMS requirements.

components/kyma-environment-broker/internal/lms/client.go
piotrmiskiewicz: I'll add validation to use it

components/kyma-environment-broker/internal/lms/client.go
piotrmiskiewicz: I'll remove it

PR #871, opened 2020-02-24
Implement requesting deletion of Package Instance Auth

pkosiec:
**Description**

- Implement requesting deletion of Package Instance Auth

**Related issue(s)**
#806


components/director/internal/domain/package/repository.go
```diff
@@ -118,6 +123,34 @@ func (r *pgRepository) GetForApplication(ctx context.Context, tenant string, id
...
 
+func (r *pgRepository) GetByInstanceAuthID(ctx context.Context, tenant string, instanceAuthID string) (*model.Package, error) {
+	var pkgEnt Entity
+
+	persist, err := persistence.FromCtx(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	stmt := fmt.Sprintf(`SELECT %s FROM %s WHERE tenant_id=$1 AND id=(SELECT %s FROM %s WHERE tenant_id=$1 AND id=$2)`,
```
dbadura: What about such SQL:
`SELECT packages-cols FROM PACKAGES AS P JOIN PACKAGE_INSTANCE_AUTHS AS A on A.PACKAGE_ID=p.ID WHERE A.ID=instanceAuthID`
But generally it's ok.

PR #874, opened 2020-02-25
Add Package Instance Auth creation request and auth set mutations

tgorgol:
**Description**

- Add Package Instance Auth creation request and set auth mutations

- Change migrator jobs restart policy to Never (so the logs from failing containers are not discarded)
- Forbid concurrent runs of Tenant Fetcher job

**Related issue(s)**

#806

### DIFF
--- a/components/director/internal/domain/package/fixtures_test.go
+++ b/components/director/internal/domain/package/fixtures_test.go
@@ -3,16 +3,15 @@ package mp_package_test
 import (
 	"database/sql"
 	"database/sql/driver"
-	"encoding/json"
 	"testing"
 	"time"
 
 	mp_package "github.com/kyma-incubator/compass/components/director/internal/domain/package"
+
 	"github.com/kyma-incubator/compass/components/director/internal/model"
 	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
 	"github.com/kyma-incubator/compass/components/director/pkg/pagination"
 	"github.com/kyma-incubator/compass/components/director/pkg/str"
-	"github.com/stretchr/testify/require"
 )
 
 func fixModelAPIDefinition(id string, pkgID *string, name, description string, group string) *model.APIDefinition {
@@ -174,7 +173,7 @@ func fixPackageModel(t *testing.T, name, desc string) *model.Package {
 		ApplicationID:                  appID,
 		Name:                           name,
 		Description:                    &desc,
-		InstanceAuthRequestInputSchema: fixBasicSchema(t),
+		InstanceAuthRequestInputSchema: fixBasicSchema(),
 		DefaultInstanceAuth:            fixModelAuth(),
 	}
 }
@@ -221,7 +220,7 @@ func fixModelPackageCreateInput(t *testing.T, name, description string) model.Pa
 	return model.PackageCreateInput{
 		Name:                           name,
 		Description:                    &description,
-		InstanceAuthRequestInputSchema: fixBasicSchema(t),
+		InstanceAuthRequestInputSchema: fixBasicSchema(),
 		DefaultInstanceAuth:            &authInput,
 	}
 }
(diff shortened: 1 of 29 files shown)


components/director/pkg/graphql/package_validation_test.go
```diff
@@ -622,64 +622,41 @@ func TestPackageInstanceAuthSetInput_Validate(t *testing.T) {
...
 			Value: graphql.PackageInstanceAuthStatusInput{
 				Condition: graphql.PackageInstanceAuthSetStatusConditionInputSucceeded,
+				Message:   str,
+				Reason:    str,
 			},
 			ExpectedValid: true,
 		},
 		{
-			Name: "Failed condition without reason",
+			Name: "No reason",
```
dbadura:
```suggestion
			Name: "No reason provided",
```

components/director/internal/domain/packageinstanceauth/service_test.go
```diff
@@ -232,6 +234,353 @@ func TestService_Delete(t *testing.T) {
...
+			Input: model.PackageInstanceAuthRequestInput{
+				InputParams: str.Ptr("{"),
+			},
+			InputAuth:      modelAuth,
+			InputSchema:    str.Ptr("{\"type\": \"string\"}"),
+			ExpectedOutput: "",
+			ExpectedError:  errors.New(`while validating value { against JSON Schema: {"type": "string"}: unexpected EOF`),
+		},
+		{
+			Name: "Error when invalid input params",
```
dbadura:
```suggestion
			Name: "Error when input doesn't match schema",
```

components/director/internal/domain/package/fixtures_test.go
```diff
@@ -364,30 +363,21 @@ func fixBasicInputSchema() *graphql.JSONSchema {
...
-	sch := fixBasicInputSchema()
-	require.NotNil(t, sch)
-	var obj map[string]interface{}
-
-	err := json.Unmarshal([]byte(*sch), &obj)
-	require.NoError(t, err)
-	var objTemp interface{}
-	objTemp = obj
-	return &objTemp
+func fixBasicSchema() *string {
```
dbadura: Maybe we should validate `fixSchema`, in case that somebody change it.

components/director/internal/domain/packageinstanceauth/service.go
```diff
@@ -105,16 +176,62 @@ func (s *service) Delete(ctx context.Context, id string) error {
...
+		err := instanceAuth.SetDefaultStatus(model.PackageInstanceAuthStatusConditionSucceeded, ts)
+		if err != nil {
+			return errors.Wrapf(err, "while setting Package Instance Auth status to: %s", model.PackageInstanceAuthStatusConditionSucceeded)
+		}
+	}
+
+	return nil
+}
+
+func (s *service) setCreationStatusFromAuth(instanceAuth *model.PackageInstanceAuth, auth *model.Auth) error {
```
pkosiec:
```suggestion
func (s *service) setCreationStatusFromAuth(instanceAuth *model.PackageInstanceAuth, defaultAuth *model.Auth) error {
```

components/director/internal/domain/packageinstanceauth/service.go
```diff
@@ -67,6 +94,50 @@ func (s *service) GetForPackage(ctx context.Context, id string, packageID string
...
+	tnt, err := tenant.LoadFromContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	instanceAuth, err := s.repo.GetByID(ctx, tnt, id)
+	if err != nil {
+		return errors.Wrap(err, "while getting Package Instance Auth")
+	}
+	if instanceAuth == nil {
```
pkosiec: Shouldn't the repo return `notFoundError`?

tgorgol: It does, I added this check to avoid panics in case `nil, nil` is somehow returned.

components/director/internal/domain/package/fixtures_test.go
tgorgol: Looks like this schema isn't even parsed anywhere in this domain, I don't think validation is really necessary.

Issue #470, opened 2019-11-15
DB diagram document is outdated


pkosiec:
**Description**

Currently, DB schema diagram is outdated - for example: missing Integration Systems or Application Template tables.

We need to synchronize these changes: manually or automatically (by generating it from DB).

PR #879, opened 2020-02-26
Retry syncing JWKS while token validation error

pkosiec:
**Description**

- Retry syncing JWKS while token validation error

**Related issue(s)**
#470

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -28,7 +28,7 @@ global:
       version: "PR-798"
     director:
       dir:
-      version: "PR-871"
+      version: "PR-878"
     gateway:
       dir: pr/
       version: "PR-357"
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -28,7 +28,7 @@ global:
       version: "PR-798"
     director:
       dir:
-      version: "PR-878"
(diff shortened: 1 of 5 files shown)


components/director/internal/authenticator/middleware.go
```diff
@@ -24,20 +26,24 @@ const AuthorizationHeaderKey = "Authorization"
...
+	a.mux.Lock()
+	log.Info("Synchronizing JWKS...")
 	jwks, err := FetchJWK(a.jwksEndpoint)
 	if err != nil {
 		return errors.Wrapf(err, "while fetching JWKS from endpoint %s", a.jwksEndpoint)
 	}
 
-	a.Jwks = jwks
+	a.cachedJWKs = jwks
+	a.mux.Unlock()
```
crabtree: Shouldn't you defer this call right after the Lock? Right now when an error happens you return without calling the Unlock.

chart/compass/values.yaml
```diff
@@ -28,7 +28,7 @@ global:
       version: "PR-798"
     director:
       dir:
-      version: "PR-871"
+      version: "PR-878"
```
crabtree: It should be PR-879

PR #905, opened 2020-03-03
Install knative-kafka in Kyma Runtime

anishj0shi:
w.r.t https://github.com/kyma-project/kyma/issues/7421
Dynamic provisioning of Azure Event- hubs from the kyma environment Broker per provisioned kyma-runtime. 

- [x]  Provision Azure Event Hubs using HAP (Hyperscaler Account Pool)
- [x] Inject Secret of an azure event hub namespace to the kyma-runtime as an installation override.
- [x] Inject ConfigMap for switching knative's default Channel to **KafkaChannel**

### Some Follow ups:
- Use a standard Naming Convention to create Azure Resource Group in Kyma Environment Broker https://github.com/kyma-incubator/compass/issues/967
- Create an Azure Event-hub instance in the same location as the provisioned kyma-runtime https://github.com/kyma-incubator/compass/issues/968


components/kyma-environment-broker/internal/event-hub/event_hub_step.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	log.Infof("HAP lookup for credentials to provision cluster for global account ID %s on Hyperscaler %s", pp.ErsContext.GlobalAccountID, hypType)
+
+	// get hyperscaler credentials from HAP
+	credentials, err := p.accountProvider.GardenerCredentials(hypType, pp.ErsContext.GlobalAccountID)
+	if err != nil {
+		errorMessage := fmt.Sprintf("Unable to retrieve Gardener Credentials from HAP lookup: %v", err)
+		return p.retryOperation(operation, errorMessage, gardenerCredentialsRetryInterval, gardenerCredentialsMaxTime, log)
+	}
+	azureCfg, err := azure.GetConfigfromHAPCredentialsAndProvisioningParams(credentials, pp)
```
piotrmiskiewicz: err not handled

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -129,16 +132,20 @@ func main() {
...
 	runtimeStep := provisioning.NewCreateRuntimeStep(db.Operations(), db.Instances(), provisionerClient)
 	smOverrideStep := provisioning.NewServiceManagerOverridesStep(db.Operations(), cfg.ServiceManager)
 
 	logs := logrus.New()
 	stepManager := process.NewManager(db.Operations(), logs)
 	stepManager.InitStep(initialisation)
+
 	stepManager.AddStep(1, resolveCredentialsStep)
-	stepManager.AddStep(2, smOverrideStep)
+	stepManager.AddStep(2, provisionAzureEventHub)
```
piotrmiskiewicz: is it also valid for GCP provider?

anishj0shi: Hi PM, 
yes currently this is the default eventing system for managed offering. so it's applicable for all.

piotrmiskiewicz: So if I'm spinning Kyma Runtime in GCP azure event hub also will be provisioned, right?

anishj0shi: yup, till we come up with an alternative for GCP. or something generic for both ;)

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
nachtmaar: done

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func extractEndpoint(accessKeys *eventhub.AccessKeys) string {
+	endpoint := strings.Split(*accessKeys.PrimaryConnectionString, ";")[0]
+	endpoint = strings.TrimPrefix(endpoint, "Endpoint=sb://")
+	endpoint = strings.TrimSuffix(endpoint, "/")
+	return endpoint
+}
+
+func appendPort(endpoint string, port int64) string {
```
k15r: you should be good with `port int`

components/kyma-environment-broker/internal/event-hub/event_hub_step_test.go
```diff
@@ -0,0 +1,445 @@
...
+	return &fakeHyperscalerProvider{
+		client: client,
+		err:    nil,
+	}
+}
+
+func NewFakeHyperscalerProviderError() azure.HyperscalerProvider {
+	return &fakeHyperscalerProvider{
+		client: nil,
+		err:    fmt.Errorf("ups ... "),
```
k15r: I assume you do not expect this to happen

components/kyma-environment-broker/internal/event-hub/event_hub_step_test.go
```diff
@@ -0,0 +1,445 @@
...
+
+	// when
+	op, _, err = step.Run(op, fixLogger())
+	require.NoError(t, err)
+	provisionRuntimeInput, err := op.InputCreator.Create()
+	require.NoError(t, err)
+
+	// then
+	allOverridesFound := ensureOverrides(t, provisionRuntimeInput)
+	require.True(t, allOverridesFound[componentNameKnativeEventing], "overrides for %s were not found", componentNameKnativeEventing)
```
k15r: `assert.True` here would print both failures

components/kyma-environment-broker/internal/event-hub/event_hub_step_test.go
```diff
@@ -0,0 +1,445 @@
...
+
+func ensureOperationIsNotRepeated(t *testing.T, err error, when time.Duration, op internal.ProvisioningOperation) {
+	t.Helper()
+	require.NotNil(t, err)
+}
+
+// ensureOverrides ensures that the overrides for
+// - the kafka channel controller
+// - and the default knative channel
+// is set
```
k15r:
```suggestion
// are set
```

components/kyma-environment-broker/internal/event-hub/azure/env.go
```diff
@@ -0,0 +1,34 @@
...
+	config.clientID = clientID
+	config.clientSecret = clientSecret
+	config.tenantID = tenantID
+	config.subscriptionID = subscriptionID
+	config.location = location
+	return config, nil
+}
+
+func GetConfigfromHAPCredentialsAndProvisioningParams(credentials hyperscaler.Credentials, parameters internal.ProvisioningParameters) (*Config, error) {
+	region := "westeurope"
```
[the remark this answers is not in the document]
anishj0shi: there's a follow up: https://github.com/kyma-incubator/compass/issues/968

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -129,16 +132,20 @@ func main() {
...
 	stepManager.AddStep(10, runtimeStep)
+	stepManager.AddStep(3, smOverrideStep)
```
jasiu001: nit: please keep order :)

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
```diff
@@ -0,0 +1,209 @@
+package event_hub
```
jasiu001: all provisioning step should be inside `/internal/process/provisioning/` dir

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		// if the parameters are incorrect, there is no reason to retry the operation
+		// a new request has to be issued by the user
+		log.Errorf("Aborting after failing to get valid operation provisioning parameters: %v", err)
+		return p.operationManager.OperationFailed(operation, "invalid operation provisioning parameters")
+	}
+	log.Infof("HAP lookup for credentials to provision cluster for global account ID %s on Hyperscaler %s", pp.ErsContext.GlobalAccountID, hypType)
+
+	// get hyperscaler credentials from HAP
+	credentials, err := p.accountProvider.GardenerCredentials(hypType, pp.ErsContext.GlobalAccountID)
```
jasiu001: `GardenerCredentials` method underneath fetch secret or update it. Those operations could be temporary, so maybe `time.Duration` here should be returned? Also, it is good practice to not run steps forever (when you return `time.Duration`) so at the beginning of step `operation.UpdatedAt` field can be checked how long operation is in progress state.
Another question is, it is necessary step to make provisioning, if not step can return success with proper log error.

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		// retrying might solve the issue while communicating with azure, e.g. network problems etc
+		errorMessage := fmt.Sprintf("Failed to persist Azure Resource Group [%s] with error: %v", groupName, err)
+		return p.retryOperation(operation, errorMessage, time.Minute, time.Minute*30, log)
+	}
+	log.Printf("Persisted Azure Resource Group [%s]", groupName)
+
+	// create EventHubs Namespace
+	eventHubsNamespace := pp.Parameters.Name
+	eventHubNamespace, err := namespaceClient.CreateNamespace(p.context, azureCfg, groupName, eventHubsNamespace)
```
jasiu001: to be sure, if the method will be called second time with the same params, update operation will be run and no error will be return?

components/kyma-environment-broker/internal/event-hub/azure/namespace_test.go
```diff
@@ -0,0 +1,23 @@
...
+func TestGetResourceGroup(t *testing.T) {
+
+	resourceGroup := "my-resourcegroup"
+	eventHubNamespaceID := fmt.Sprintf("/subscriptions/35d42578-34d1-486d-a689-012a8d514c19/resourceGroups/%s/providers/Microsoft.EventHub/namespaces/nachtmaar-test-2", resourceGroup)
+	nameSpaceFixture := eventhub.EHNamespace{
+		Response: autorest.Response{},
+		ID:       &eventHubNamespaceID,
+	}
+	parsedResourceGroup := GetResourceGroup(nameSpaceFixture)
+	if parsedResourceGroup != resourceGroup {
```
jasiu001: you can just use `assert.Equal(t, resourceGroup, parsedResourceGroup)` from "github.com/stretchr/testify/assert" library

components/kyma-environment-broker/internal/process/operation.go
```diff
@@ -19,17 +19,21 @@ func NewOperationManager(storage storage.Operations) *OperationManager {
 	return &OperationManager{storage: storage}
 }
 
+// OperationSucceeded marks the operation as failed and only repeats it if there is a storage error
```
jasiu001: "marks the operation as failed" ? :)

components/kyma-environment-broker/internal/event-hub/event_hub_step_test.go
```diff
@@ -0,0 +1,445 @@
...
+	provisionRuntimeInput, err := op.InputCreator.Create()
+	require.NoError(t, err)
+
+	// then
+	allOverridesFound := ensureOverrides(t, provisionRuntimeInput)
+	require.True(t, allOverridesFound[componentNameKnativeEventing], "overrides for %s were not found", componentNameKnativeEventing)
+	require.True(t, allOverridesFound[componentNameKnativeEventingKafka], "overrides for %s were not found", componentNameKnativeEventingKafka)
+}
+
+func Test_StepProvisionParametersError(t *testing.T) {
```
jasiu001: please use table test if it is possible

components/kyma-environment-broker/internal/event-hub/event_hub_step_test.go
nachtmaar: I expect this to happen, because it triggers the edge case in the eventhub step.

It is used in the test [here](https://github.com/anishj0shi/compass/blob/00ea98a7139d8d0c954ee8899f43dd2ab1a22bb5/components/kyma-environment-broker/internal/event-hub/event_hub_step_test.go#L222)

It is testing the following lines:

```go
	namespaceClient, err := p.hyperscalerProvider.GetClient(azureCfg)
	if err != nil { <- it produces this error
		// internal error, repeating doesn't solve the problem
		errorMessage := fmt.Sprintf("Failed to create Azure EventHubs client: %v", err)
		return p.operationManager.OperationFailed(operation, errorMessage)
	}
```

k15r: well then lets use some proper message here :D

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// get EventHubs Namespace secret
+	accessKeys, err := namespaceClient.GetEventhubAccessKeys(p.context, *resourceGroup.Name, *eventHubNamespace.Name, authorizationRuleName)
+	if err != nil {
+		// retrying might solve the issue while communicating with azure, e.g. network problems etc
+		errorMessage := fmt.Sprintf("Unable to retrieve access keys to azure event-hub namespace: %v", err)
+		return p.retryOperation(operation, errorMessage, time.Minute, time.Minute*30, log)
+	}
+	kafkaEndpoint := extractEndpoint(&accessKeys)
+	kafkaEndpoint = appendPort(kafkaEndpoint, kafkaPort)
```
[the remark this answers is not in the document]
nachtmaar: Good point, but doesn't work I think.
I wanted to create an instance of the struct (provide host and port) and call .String(), but the struct doesn't have a port entry.

> type URL struct {
>     Scheme     string
>     Opaque     string    // encoded opaque data
>     User       *Userinfo // username and password information
>     Host       string    // host or host:port
>     Path       string    // path (relative paths may omit leading slash)
>     RawPath    string    // encoded path hint (see EscapedPath method); added in Go 1.5
>     ForceQuery bool      // append a query ('?') even if RawQuery is empty; added in Go 1.7
>     RawQuery   string    // encoded query values, without '?'
>     Fragment   string    // fragment for references, without '#'
> }

components/kyma-environment-broker/internal/event-hub/azure/iam.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil {
+			return nil, err
+		}
+		a = autorest.NewBearerAuthorizer(token)
+
+	default:
+		return a, fmt.Errorf("invalid grant type specified")
+	}
+
+	return a, err
```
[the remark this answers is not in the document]
nachtmaar: Removed the switch and the enum. Currently we only need this particular OAuth Grant.

components/kyma-environment-broker/internal/process/operation.go
nachtmaar: Ah, copy paste. Thank you!

components/kyma-environment-broker/internal/event-hub/azure/namespace_test.go
nachtmaar: Method is not required anymore. Thanks for the suggestion 👍

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
nachtmaar: Good catch!

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
nachtmaar: Good catch. I didn't see the last call in the implementation. You are right, it makes sense to retry.

Concerning UpdatedAt, we implemented `retryOperation` which makes use of this field. Moving this to a common package could help others.

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
anishj0shi: if the method is called for the second time with the same parameter, it will not return any error. it would just try to update it.  
ref: https://github.com/Azure/azure-sdk-for-go/blob/f1b3b9b7d4962c94c77d16ce75f9a8e981eb2488/services/eventhub/mgmt/2017-04-01/eventhub/namespaces.go#L132

components/kyma-environment-broker/internal/event-hub/azure/group.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
+)
+
+// getGroupsClient gets a client for handling of Azure ResourceGroups
+func getGroupsClient(config *Config) (resources.GroupsClient, error) {
+	client := resources.NewGroupsClient(config.subscriptionID)
+
+	authorizer, err := GetResourceManagementAuthorizer(config)
+	if err != nil {
+		return resources.GroupsClient{}, fmt.Errorf("failed to initialize authorizer with error: %v", err)
```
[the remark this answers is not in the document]
nachtmaar: I had the same idea, but CI is preventing us from doing so. 


See this pipeline: https://storage.googleapis.com/kyma-prow-logs/pr-logs/pull/kyma-incubator_compass/905/pre-compass-components-kyma-environment-broker/1239571009451855873/build-log.txt
```
internal/event-hub/azure/client.go:27:15: Errorf format %w has unknown verb w
```

Compass is not using `%w` yet in the code. As soon as they do so, we can change it :)

components/kyma-environment-broker/internal/event-hub/azure/group.go
```diff
@@ -0,0 +1,24 @@
...
+	}
+	client.Authorizer = authorizer
+
+	if err = client.AddToUserAgent(config.userAgent); err != nil {
+		return resources.GroupsClient{}, fmt.Errorf("failed to add use agent [%s] with error: %v", config.userAgent, err)
```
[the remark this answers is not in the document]
nachtmaar: Won't do. See comment above. But thanks for the suggestion 👍

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
```diff
@@ -0,0 +1,209 @@
...
+	log logrus.FieldLogger) (internal.ProvisioningOperation, time.Duration, error) {
+
+	hypType := hyperscaler.Azure
+
+	// parse provisioning parameters
+	pp, err := operation.GetProvisioningParameters()
+	if err != nil {
+		// if the parameters are incorrect, there is no reason to retry the operation
+		// a new request has to be issued by the user
+		log.Errorf("Aborting after failing to get valid operation provisioning parameters: %v", err)
```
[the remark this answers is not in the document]
nachtmaar: Good question!

The problem only appears if the provisioning parameters cannot be unmarshalled (e.g. invalid json). The request to the KymaEnvironmentBroker (KEB) is send via the UI later, so I hope the UI will display a reasonable error message. 

The step operation is async, so ideally the provisioning parameters (which are part of the HTTP PUT request), are validated before the async operation started and give the user a 4XX status code. But to be honest, I don't know if this is the case. 

Since the code `operation.GetProvisioningParameters()` is used heavily throughout KEB, I would leave it to the authors of KEB to implement fast user feedback :)

components/kyma-environment-broker/internal/event-hub/event_hub_step.go
nachtmaar: Done. Furthermore, I moved the azure package to `/internal/hyperscaler`. Let me know if you have a different suggestion :)

components/kyma-environment-broker/internal/event-hub/event_hub_step_test.go
nachtmaar: all ups are more detailed now :-D

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -0,0 +1,208 @@
...
+	// get EventHubs Namespace secret
+	accessKeys, err := namespaceClient.GetEventhubAccessKeys(p.context, *resourceGroup.Name, *eventHubNamespace.Name, authorizationRuleName)
+	if err != nil {
+		// retrying might solve the issue while communicating with azure, e.g. network problems etc
+		errorMessage := fmt.Sprintf("Unable to retrieve access keys to azure event-hub namespace: %v", err)
+		return p.retryOperation(operation, errorMessage, time.Minute, time.Minute*30, log)
+	}
+	kafkaEndpoint := extractEndpoint(accessKeys)
+	kafkaEndpoint = appendPort(kafkaEndpoint, kafkaPort)
+	kafkaPassword := *accessKeys.PrimaryConnectionString
```
mszostok: please check the pointer as such expression can lead to panic

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -129,15 +132,19 @@ func main() {
...
 	smOverrideStep := provisioning.NewServiceManagerOverridesStep(db.Operations(), cfg.ServiceManager)
 
 	logs := logrus.New()
 	stepManager := process.NewManager(db.Operations(), logs)
 	stepManager.InitStep(initialisation)
+
 	stepManager.AddStep(1, resolveCredentialsStep)
-	stepManager.AddStep(2, smOverrideStep)
+	stepManager.AddStep(2, provisionAzureEventHub)
+	stepManager.AddStep(3, smOverrideStep)
```
mszostok: Getting the SM overrides are not connected with azure hub, so they be done in parallel, am I right @jasiu001?

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -0,0 +1,208 @@
...
+
+	return operation, 0, nil
+}
+
+func (p *ProvisionAzureEventHubStep) retryOperationOnce(operation internal.ProvisioningOperation, wait time.Duration) (internal.ProvisioningOperation, time.Duration, error) {
+	return operation, wait, nil
+}
+
+// retryOperation retries an operation for at maxTime in retryInterval steps
+func (p *ProvisionAzureEventHubStep) retryOperation(operation internal.ProvisioningOperation, errorMessage string, retryInterval time.Duration, maxTime time.Duration, log logrus.FieldLogger) (internal.ProvisioningOperation, time.Duration, error) {
```
mszostok: @jasiu001 IMO those functions are quite generic and should be provided by the framework itself, WDYT?

jasiu001: I agree it could be used in other places/steps. Maybe the good place for this kind of method is `/internal/process/operation.go`

components/kyma-environment-broker/cmd/broker/main.go
jasiu001: in general, we do not support parallel steps yet but if you mean the same weight for step, you are right it could be the same weight (e.g. 2)

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
nachtmaar: That sounds good! I think it is easier to read than just returning 3 arguments. I will place both methods to operation.go then.

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -0,0 +1,208 @@
...
+	log.Infof("Retrying for %s in %s steps", maxTime.String(), retryInterval.String())
+	if dur < maxTime {
+		return operation, retryInterval, nil
+	}
+	log.Errorf("Aborting after %s of failing retries", maxTime.String())
+	return p.operationManager.OperationFailed(operation, errorMessage)
+}
+
+func extractEndpoint(accessKeys eventhub.AccessKeys) string {
+	endpoint := strings.Split(*accessKeys.PrimaryConnectionString, ";")[0]
```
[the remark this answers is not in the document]
nachtmaar: Good catch! The azure-sdk uses pointers everywhere 😂

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
nachtmaar: We should have a follow up PR then to make use of these methods.

E.g. [here for RetryOperation](https://github.com/kyma-incubator/compass/blob/b13d9c2a0a4b6a2368f8cd7edef0112a92eb51f5/components/kyma-environment-broker/internal/process/provisioning/resolve_creds.go#L78
) and [here for retryOnce](https://github.com/kyma-incubator/compass/blob/b13d9c2a0a4b6a2368f8cd7edef0112a92eb51f5/components/kyma-environment-broker/internal/process/provisioning/resolve_creds.go#L102)

components/kyma-environment-broker/cmd/broker/main.go
nachtmaar: So let's use the same weight, then we are future-proof if parallel steps feature is coming 😎

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
nachtmaar: Done. I have the logger as method argument which does not look too nice. But since the logger comes the steps Run method and is not inside the steps structure for all steps which we have atm, it is the easiest solution for now.

components/kyma-environment-broker/internal/hyperscaler/azure/iam.go
```diff
@@ -0,0 +1,47 @@
+package azure
+
+import (
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+)
+
+var (
+	armAuthorizer autorest.Authorizer
```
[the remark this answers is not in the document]
nachtmaar: Totally agree.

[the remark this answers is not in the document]
nachtmaar: Done. Thanks for the comment. It was even confusing for me ;)

components/kyma-environment-broker/internal/hyperscaler/azure/iam.go
```diff
@@ -0,0 +1,47 @@
...
+func GetResourceManagementAuthorizer(config *Config) (autorest.Authorizer, error) {
+	if armAuthorizer != nil {
+		return armAuthorizer, nil
+	}
+
+	environment, err := config.Environment()
+	if err != nil {
+		return nil, err
+	}
+	a, err := getAuthorizerForResource(config, environment.ResourceManagerEndpoint)
```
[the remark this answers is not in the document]
nachtmaar: Done. Thanks for the comment. It was even confusing for me ;)

components/kyma-environment-broker/internal/hyperscaler/azure/group.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
+)
+
+// getGroupsClient gets a client for handling of Azure ResourceGroups
+func getGroupsClient(config *Config) (resources.GroupsClient, error) {
+	client := resources.NewGroupsClient(config.subscriptionID)
+
+	authorizer, err := GetResourceManagementAuthorizer(config)
+	if err != nil {
+		return resources.GroupsClient{}, fmt.Errorf("failed to initialize authorizer with error: %v", err)
```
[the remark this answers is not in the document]
nachtmaar: Done

components/kyma-environment-broker/internal/hyperscaler/azure/iam.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func GetResourceManagementAuthorizer(config *Config) (autorest.Authorizer, error) {
+	if armAuthorizer != nil {
+		return armAuthorizer, nil
+	}
+
+	environment, err := config.Environment()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
nachtmaar: Done

PR #911, opened 2020-03-04
Add migration with default commerce Application Template

tgorgol:
**Description**

- Add migration with default commerce Application Template
- Move graphqlizer to director


components/schema-migrator/migrations/director/202003041400_add_commerce_template.up.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO app_templates ("id", "name", "description", "application_input", "placeholders", "access_level")
+    VALUES ('00000000-0000-0000-0000-000000000001', 'SAP Commerce Cloud', 'SAP Commerce Cloud v2', '{"name": "{{name}}", "labels": {"compass/legacy": true}, "packages": null, "webhooks": null, "documents": null, "description": "SAP Commerce Cloud v2", "providerName": "SAP", "apiDefinitions": null, "healthCheckURL": null, "eventDefinitions": null, "integrationSystemID": null}', '[{"Name": "name", "Description": "Application''s name"}]', 'GLOBAL');
```
pkosiec: Hmm, do we really want to include commercial product in open source project? That's open question, if everyone is fine with that then I can also live with that.

components/director/internal/domain/apptemplate/graphqlizer_test.go
```diff
@@ -0,0 +1,81 @@
...
+		{
+			Name: "Success when string",
+			Input: graphql.Labels{
+				"bar": "test",
+			},
+			Expected:      "{bar:\"test\",}",
+			ExpectedError: nil,
+		},
+		{
+			Name: "Success when bool",
```
pkosiec: What about case for number?

components/director/internal/domain/apptemplate/graphqlizer.go
```diff
@@ -268,8 +263,31 @@ func (g *Graphqlizer) VersionInputToGQL(in graphql.VersionInput) (string, error)
 	}`)
 }
 
+func (g *Graphqlizer) marshal(obj interface{}) string {
```
pkosiec: Can we use imported graphqlizer from tests? Or can we move graphqlizer to director and then import it in tests?
Please also let known corresponding teams which use copied Graphqlizer.

Issue #904, opened 2020-03-03
Labels in Application Template work only with array values


tgorgol:
**Description**
When creating Application Template with Label that has value type different than array, Director can't convert the model to graphql string to return created entity.

**Expected result**
Application Template should be created successfully and returned in the response.

**Actual result**
```
{
  "errors": [
    {
      "message": "while converting application template to graphql: while graphqlising application create input: while graphqlising application create input: while executing template: template: tmpl:10:13: executing \"tmpl\" at <LabelsToGQL .Labels>: error calling LabelsToGQL: while executing template: template: tmpl:4:23: executing \"tmpl\" at <$v>: range can't iterate over true",
      "path": [
        "createApplicationTemplate"
      ]
    }
  ],
  "data": null
}
```

**Steps to reproduce**
Try creating following Application Template:
```graphql
  createApplicationTemplate(in: {
    name: "example-app-template"
    accessLevel: GLOBAL
    applicationInput: {
      name: "example-app"
      labels: {
      	bug: true
    	}
    }
  }) {
    applicationInput
  }
```

**Troubleshooting**
Looks like the bug is in `graphqlizer.go` file, `LabelsToGQL` function expects the label value to alway be an array. It shouldn't.

PK85:
fixed, closing

Issue #845, opened 2020-02-14
GQL client doesn't allow to send slash in label name


polskikiel:
**Description**

We use Provisioner's ProvisionRuntime mutation to create a runtime that allows us to send labels.
In the runtime agent, we use a pattern for labels keys `prefix/key`, to apply that we would like to set labels on runtime:
- **broker/instance_id**
- **global/subaccount_id**

unfortunately, the graphql client doesn't allow to send special characters like `- / :` as a parameter name.

**Expected result**

We have labels pattern which can be applied using the ProvisionRuntime mutation.

**Actual result**

Providing labels with a slash doesn't work. KEB would need to follow the pattern from runtime agent and label the runtime after it's provisioned.

**Steps to reproduce**

Try to send labels using the ProvisionRuntime mutation using the gql playground:
![image](https://user-images.githubusercontent.com/21159554/74532538-bc558580-4f2f-11ea-81ba-18ca4c6a67a2.png)

pkosiec:
Looks like a bug indeed, because of GraphQL limits for special characters in field keys. In the meantime you can use `setRuntimeLabel` mutations to set your labels.

polskikiel:
@pkosiec We don't want to do that because KEB doesn't know if the runtime was registered in the Director and it may lead to errors. We also need these runtimes labeled just after it starts to provision.

polskikiel:
We can work around that issue by using different label keys without slashes and change it after it will be implemented. So it'd be:
- **broker_instance_id**
- **global_subaccount_id**

Szymongib:
It is not possible in Director itself so I guess it is not an issue for the Provisioner.

pkosiec:
Yup, the problem is in Director.

kfurgol:
We decided to change the contract on labels.
 Label keys now will accept only alphanumeric symbols and an underscore. 

New AC: 
- Change validation of label keys to support only alphanumeric symbols and an underscore. 
- Check if any component(eg. Provisioner) which is using Director API is compliant with this change, make necessary changes if needed in those components and tests
- Notify dependant teams

PR #912, opened 2020-03-05
Allow to use special characters in label key

kfurgol:
**Description**

- Labels are now passed as a JSON in Director, KEB and Provisioner

**Related issue(s)**
#845
#904


kfurgol:
AC has changed: https://github.com/kyma-incubator/compass/issues/845#issuecomment-599478262

PR #937, opened 2020-03-09
Align KEB chart with the one from Kyma repo

mszostok:
**Description**

- Align KEB chart with the one from Kyma repo


chart/compass/charts/kyma-environment-broker/templates/tests/e2e-provisioning.yaml
```diff
@@ -65,7 +65,7 @@ spec:
             - name: APP_TENANT_ID
               value: "{{ .Values.global.defaultTenant }}"
             - name: APP_DUMMY_TEST
-              value: "true"
+              value: "false"
```
polskikiel: You can leave it == true, because if anyone set `global.provisioning.enabled` to true it will be executed on the `pre-master-compass-integration` job with the rest of the acceptance tests. With the Dummy test, we won't block the pipelines and then we can just exclude that test definition, or make that test useful for pre-master jobs.

mszostok: thanks for the explanation. I will add such comment under this value so rest of us will also know the reason behind the `false` value here and `true` in Kyma repo.

mszostok:
cannot merge because the  pre-master-compass-integration is still failing... currently the master is broken with istio-patc

PR #939, opened 2020-03-10
Process existing operations on start

polskikiel:
**Description**

- Implemented processing of existing **in progress** operations on start

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/storage/dbsession/read.go
+++ b/components/kyma-environment-broker/internal/storage/dbsession/read.go
@@ -5,6 +5,7 @@ import (
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/storage/dberr"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/storage/postsql"
+	"github.com/pivotal-cf/brokerapi/v7/domain"
 )
 
 type readSession struct {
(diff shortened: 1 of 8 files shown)


components/kyma-environment-broker/internal/storage/dbsession/read.go
```diff
@@ -43,6 +44,26 @@ func (r readSession) GetOperationByID(opID string) (OperationDTO, dberr.Error) {
...
+	var operations []OperationDTO
+
+	_, err := r.session.
+		Select("*").
+		From(postsql.OperationTableName).
+		Where(stateCondition).
+		Where(typeCondition).
+		Load(&operations)
+	if err != nil {
+		if err == dbr.ErrNotFound {
```
piotrmiskiewicz: if there is no any in progress operations, it is better to return empty list. Such case is not an error.

components/kyma-environment-broker/internal/storage/storage_test.go
```diff
@@ -142,6 +144,11 @@ func TestSchemaInitializer(t *testing.T) {
 		err = svc.InsertProvisioningOperation(givenOperation)
 		require.NoError(t, err)
 
+		ops, err := svc.GetOperationsInProgressByType(dbsession.OperationTypeProvision)
+		require.NoError(t, err)
+		assert.Len(t, ops, 1)
+		assert.Equal(t, givenOperation.Operation, ops[0])
```
piotrmiskiewicz: there is an assertProvisioningOperation function

piotrmiskiewicz: please, run the test and fix the build tag and makefile

polskikiel: 👍

polskikiel: about `assertProvisioningOperation`, I created a similar function `assertOperation`, which compares `internal.Operation` structs.

PR #943, opened 2020-03-10
Gateway add logging

dbadura:
**Description**

- add support for auditlog in gateway
- here is PR bump: https://github.com/kyma-project/kyma/pull/7588

**Related issue(s)**


components/gateway/internal/auditlog/service.go
```diff
@@ -0,0 +1,174 @@
...
+type graphqQuery struct {
+	Query string `json:"query"`
+}
+
+//We assume that if request payload start with `mutation` and
+//if any of response errors has path array length equal 1, that means that mutation failed
+func isReadError(response GraphqlResponse, request string) bool {
+	req := strings.TrimSpace(request)
+	mutation := strings.HasPrefix(req, "mutation")
+	if !mutation {
```
dbadura: I'm going to refactor this method
if mutation then
if query then
else try to marshall request

components/gateway/internal/auditlog/client.go
```diff
@@ -0,0 +1,148 @@
...
+type UUIDService interface {
+	Generate() string
+}
+
+//go:generate mockery -name=TimeService -output=automock -outpkg=automock -case=underscore
+type TimeService interface {
+	Now() time.Time
+}
+
+type AuditlogConfig struct {
```
kjaksik: What is going to happen if there's no audit log service? Please make it possible for the gateway to work without it

components/gateway/internal/auditlog/client.go
```diff
@@ -0,0 +1,148 @@
...
+
+	change.Tenant = c.cfg.Tenant
+	change.UUID = c.uuidSvc.Generate()
+
+	payload, err := json.Marshal(&change)
+	if err != nil {
+		return errors.Wrap(err, "while marshalling auditlog payload")
+	}
+
+	fmt.Printf("PAYLOAD TO CONFIGURATION CHANGE AUDITLOG:\n %s\n", payload)
```
kjaksik: I'm not sure if this is neccessary

kjaksik: Maybe you can log the type of audit log but not the whole payload

components/gateway/internal/auditlog/client_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+
+	"github.com/kyma-incubator/compass/components/gateway/internal/auditlog"
+	"github.com/kyma-incubator/compass/components/gateway/internal/auditlog/model"
+)
+
+//TODO:
+//{"attributes":"The 'attributes' property cannot be null or empty"}<nil>+--- PASS: TestLogConfigurationChangeToAuditlog (1.11s)
```
kjaksik:
```suggestion
```

components/gateway/internal/auditlog/client_test.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func fixConfigChangeLog() model.ConfigurationChange {
+	return model.ConfigurationChange{
+		User: "test-user",
+		Object: model.Object{
+			ID: map[string]string{
+				"name":           "Config Change",
+				"externalTenant": "external tenant",
+				"apiConsumer":    "applicaiton",
```
kjaksik:
```suggestion
				"apiConsumer":    "application",
```

components/gateway/internal/auditlog/client.go
```diff
@@ -0,0 +1,147 @@
...
+	http             http.Client
+	configChangeURL  string
+	securityEventURL string
+}
+
+func NewClient(cfg AuditlogConfig, uuidSvc UUIDService, tsvc TimeService) *Client {
+	client := http.Client{
+		Timeout: time.Second * 5,
+	}
+	configChangeURL := cfg.AuditLogURL + ConfigChangeURLPath
```
crabtree: This is not really the safest way of building the URLs

components/gateway/internal/auditlog/client.go
```diff
@@ -0,0 +1,147 @@
...
+	t := c.timeSvc.Now()
+	logTime := t.Format(LogFormatDate)
+	change.Time = logTime
+
+	change.Tenant = c.cfg.Tenant
+	change.UUID = c.uuidSvc.Generate()
+
+	payload, err := json.Marshal(&change)
+	if err != nil {
+		return errors.Wrap(err, "while marshalling auditlog payload")
```
crabtree:
```suggestion
		return errors.Wrap(err, "while marshaling auditlog payload")
```

components/gateway/internal/auditlog/client.go
```diff
@@ -0,0 +1,147 @@
...
+	req.SetBasicAuth(c.cfg.User, c.cfg.Password)
+
+	response, err := c.http.Do(req)
+	if err != nil {
+		return errors.Wrapf(err, "while sending auditlog to: %s", c.configChangeURL)
+	}
+	defer c.closeBody(response.Body)
+
+	if response.StatusCode != http.StatusCreated {
+		fmt.Printf("Got different status code: %d\n", response.StatusCode)
```
crabtree: You should use the logrus library to be consistent with our other components.

components/gateway/internal/auditlog/client.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		fmt.Print(string(output))
+		return errors.Errorf("Write to auditlog failed with status code: %d", response.StatusCode)
+	}
+
+	return nil
+}
+
+func (c *Client) LogSecurityEvent(event model.SecurityEvent) error {
+	t := c.timeSvc.Now()
```
crabtree: This can be moved to some helper func as this part is repeated.

components/gateway/internal/auditlog/client.go
```diff
@@ -0,0 +1,147 @@
...
+
+	req, err := http.NewRequest("POST", c.securityEventURL, bytes.NewBuffer(payload))
+	if err != nil {
+		return errors.Wrap(err, "while creating request")
+	}
+
+	return c.sendAuditLog(req)
+}
+
+func (c *Client) sendAuditLog(req *http.Request) error {
```
crabtree: Why don't you use this function when sending the configuration change events?

components/gateway/internal/auditlog/client_test.go
```diff
@@ -0,0 +1,223 @@
...
+			assert.Equal(t, r.URL.Path, "/audit-log/v2/security-events")
+			inputLog := readSecurityEventRequestBody(t, r)
+			assert.Equal(t, expectedLog, inputLog)
+			w.WriteHeader(http.StatusCreated)
+		}))
+		defer ts.Close()
+
+		cfg := auditlog.AuditlogConfig{User: "user", Password: "password", AuditLogURL: ts.URL, Tenant: testTenant}
+
+		uuidSvc := &automock.UUIDService{}
```
crabtree: Please move those mock declarations to some function and just reuse it.

components/gateway/internal/auditlog/service.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func NewAuditlogSink(logsChannel chan AuditlogMessage) *AuditlogSource {
+	return &AuditlogSource{
+		logsChannel: logsChannel,
+	}
+}
+
+func (sink *AuditlogSource) Log(request, response string, claims proxy.Claims) error {
+	msg := AuditlogMessage{
```
crabtree: Why do you expect to return an error from this func?

components/gateway/internal/auditlog/service.go
```diff
@@ -0,0 +1,191 @@
...
+			Name: "response",
+			Old:  "",
+			New:  "success",
+		})
+
+		err = svc.client.LogConfigurationChange(log)
+		return errors.Wrap(err, "while sending to auditlog")
+	}
+
+	if idx := svc.isInsufficientScopeError(graphqlResponse.Errors); idx != -1 {
```
crabtree: How about changing the approach a little bit. I would name this func hasInsufficeintScopeError and return boolean and the index of the error. And also what if there are two calls with insufficient scopes?

components/gateway/internal/auditlog/service.go
```diff
@@ -0,0 +1,191 @@
...
+		})
+		return errors.Wrap(err, "while sending secuity event to auditlog")
+	}
+
+	isReadErr, err := isReadError(graphqlResponse, request)
+	if err != nil {
+		return errors.Wrap(err, "while checking if error is read error")
+	}
+
+	if isReadErr {
```
crabtree: I think this can be refactored and simplified. What about:
```go
log := svc.createConfigChangeLog(claims, request)
if isReadErr {
  log.Attributes = append(log.Attributes, model.Attribute{
    Name: "response",
    Old:  "",
    New:  "success",
  })
} else {
  log.Attributes = append(log.Attributes, model.Attribute{
    Name: "response",
    Old:  "",
    New:  response,
  })
}
err = svc.client.LogConfigurationChange(log)
return errors.Wrap(err, "while sending configuration change")
```

components/gateway/pkg/proxy/transport.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (t *Transport) getClaims(headers http.Header) (Claims, error) {
+	token := headers.Get("Authorization")
+	if token == "" {
+		return Claims{}, errors.New("invalid bearer token")
+	}
+
+	token = strings.TrimPrefix(token, "Bearer ")
+	token = strings.Split(token, ".")[1]
```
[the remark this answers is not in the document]
crabtree: Or maybe you can use some jwt library to work with the token?

components/gateway/pkg/proxy/transport.go
```diff
@@ -0,0 +1,105 @@
...
+	token = string(tk)
+	err = json.Unmarshal([]byte(token), &claims)
+	if err != nil {
+		return Claims{}, nil
+	}
+
+	return claims, nil
+}
+
+func (t *Transport) closeBody(body io.ReadCloser) {
```
crabtree: How about extracting this func to some httpcommons package?

components/gateway/internal/auditlog/service.go
dbadura: I want to be compatible with `AuditlogService`

components/gateway/internal/auditlog/service.go
dbadura: I am going to save the whole payload

chart/compass/charts/gateway/templates/deployment.yaml
```diff
@@ -32,6 +32,40 @@ spec:
               value: "http://compass-director.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.global.director.port }}"
             - name: APP_CONNECTOR_ORIGIN
               value: "http://compass-connector.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.global.connector.graphql.external.port }}"
+            - name: APP_AUDITLOG_ENABLED
+              value: "{{ .Values.global.auditlog.enabled }}"
+{{ if .Values.global.auditlog.enabled }}
```
crabtree: Please describe this parameters in the readme file.

PR #953, opened 2020-03-11
Global gardener secret config

koala7659:
This PR contains helm chart changes to unify the configuration for gardener kubeconfig secret and configuration used by both Kyma Environement Broker and Provisioner. Following values have been moved to global chart level (file values.yml) and applied in respective deployments.

```
  gardener:
         project: 
         kubeconfigPath:
         kubeconfig:
         secretName: 
         manageSecrets:
```

This PR will also require changes in the global override configuration for example on mps-dev cluster


chart/compass/values.yaml
```diff
@@ -150,6 +150,13 @@ global:
       name: compass-istio-gateway
       namespace: compass-system
 
+  gardener:
+    project: "kyma-dev" # Gardener project connected to SA
+    kubeconfigPath: "/gardener/kubeconfig/kubeconfig"
```
piotrmiskiewicz: this should not be a parameter, let's hardcode it in the chart (it is connected with volume mount)

chart/compass/values.yaml
```diff
@@ -34,15 +34,15 @@ global:
       version: "PR-913"
     provisioner:
       dir:
-      version: "PR-913"
+      version: "PR-953"
```
Szymongib: You are not making any changes in the `components` directory, so the images will not be created. The same is true for KEB and Provisioner tests.

chart/compass/values.yaml
koala7659: I have changed it to be built from configuration variable  .Values.global.gardener.kubeconfigMountPath

chart/compass/values.yaml
koala7659: True. Only charts are updated

crabtree:
The component was moved to the https://github.com/kyma-project/control-plane repository.

PR #964, opened 2020-03-12
Remove istio-init from installation cr

piotrmsc:
**Description**
https://github.com/kyma-project/kyma/pull/7565 istio-init chart was removed from kyma, therefore it should be removed in compass as well as it does not exists in installer anymore.

- Removed istio-init from installation crs

**Related issue(s)**


piotrmsc:
invalid, done in #963

PR #977, opened 2020-03-16
Update gcp version too

rakesh-garimella:
**Description**

- Update gcp k8s version too.
**Related issue(s)**


mszostok:
Here you also need to update the value: https://github.com/kyma-incubator/compass/blob/34587c751df713638a9673d190f1bde9db19a5ec/components/kyma-environment-broker/internal/process/provisioning/create_runtime_test.go#L57

Issue #522, opened 2019-12-01
De-provisioning internal server error


andreaAlkalay:
In the creation of a kyma cluster we have several steps. 
In case a step prior to the provision call fails the de-provision call returns 500 Internal server error.

If the environment instance-id does not exist shouldn't you return a 404 or 200?

Here is a sample:
Url: https://kyma-env-broker.compass-stage.compass.yaas.io/v2/service_instances/54B95508-5B80-437A-A675-3204F7CAA027?service_id=47c9dcbf-ff30-448e-ab36-d3bad66ba281&plan_id=cf3698e9-8833-4c10-8213-c84d2438da07&accepts_incomplete=true
Method: DELETE
Body: 
============= End Request =============
============= Begin Response =============
Status Code: 500
Status Text: Internal Server Error

Thanks,
Andrea.

piotrmiskiewicz:
The OSB API defines Http 410 GONE in case the instance was already deprovisioned.

piotrmiskiewicz:
The scenario withdeprovisioning of failed instance (or not existing instance) must be tested and checked if it is consistent with OSB API

PR #983, opened 2020-03-17
KEB: Return 410 GONE if instance does not exists while deprovisioning

piotrmiskiewicz:
**Description**

- return HTTP 410 while deprovisioning not existing instance

**Related issue(s)**
Related:  #522

### DIFF
--- a/components/kyma-environment-broker/internal/storage/dbsession/write.go
+++ b/components/kyma-environment-broker/internal/storage/dbsession/write.go
@@ -77,7 +77,10 @@ func (ws writeSession) DeleteInstance(instanceID string) dberr.Error {
 		Where(dbr.Eq("instance_id", instanceID)).
 		Exec()
 
-	return dberr.Internal("Failed to delete record from Instance table: %s", err)
+	if err != nil {
+		return dberr.Internal("Failed to delete record from Instance table: %s", err)
+	}
+	return nil
 }
 
 func (ws writeSession) UpdateInstance(instance internal.Instance) dberr.Error {
--- a/components/kyma-environment-broker/internal/storage/dbsession/write.go
+++ b/components/kyma-environment-broker/internal/storage/dbsession/write.go
@@ -3,7 +3,7 @@ package dbsession
 import (
 	"time"
 
-	dbr "github.com/gocraft/dbr"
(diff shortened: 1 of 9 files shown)


components/kyma-environment-broker/internal/storage/dbsession/write.go
```diff
@@ -72,6 +72,14 @@ func (ws writeSession) InsertInstance(instance internal.Instance) dberr.Error {
 	return nil
 }
 
+func (ws writeSession) DeleteInstance(instanceID string) dberr.Error {
+	_, err := ws.deleteFrom(postsql.InstancesTableName).
+		Where(dbr.Eq("instance_id", instanceID)).
+		Exec()
+
+	return dberr.Internal("Failed to delete record from Instance table: %s", err)
```
ksputo: we should not return error if record was successfully deleted for storage
```suggestion
	if err != nil {
		return dberr.Internal("Failed to delete record from Instance table: %s", err)
	}
	return nil
```

components/kyma-environment-broker/internal/storage/driver/postsql/instance.go
```diff
@@ -69,3 +69,8 @@ func (s *Instance) Update(instance internal.Instance) error {
 		return true, nil
 	})
 }
+
+func (s *Instance) Delete(instanceID string) error {
+	sess := s.NewWriteSession()
+	return sess.DeleteInstance(instanceID)
```
ksputo: question: every other interaction with postsql database includes retries - is it okay to not use `wait.Poll` in case of deleting?

components/kyma-environment-broker/internal/storage/dbsession/write.go
piotrmiskiewicz: it is a mistake, of course you are right

components/kyma-environment-broker/internal/storage/driver/postsql/instance.go
piotrmiskiewicz: In general, I don't like idea of having retries on this level, but I could add it here like for other operations. The problem is currently the deprovisioning call is synchronous, so it should not take too much time.

PR #984, opened 2020-03-17
Add validation to label keys

kfurgol:
**Description**

- Added validation for label keys (only alphanumeric characters and underscore allowed)

**Related issue(s)**
#845


components/kyma-environment-broker/internal/director/query.go
```diff
@@ -2,7 +2,7 @@ package director
 
 import "fmt"
 
-const consoleURLLabelKey = "runtime/console_url"
+const consoleURLLabelKey = "runtime_console_url"
```
crabtree: Shouldn't it be `runtime_consoleUrl`, according to the new convention?

components/director/internal/domain/application/service.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	intSysKey = "integration-system-id"
+	intSysKey = "integration_system_id"
```
crabtree: Just minor, don't we want to have a convention that underscores is the separator? In this case, I would go with `integrationSystemId`

kfurgol:
> Just one more overall comment. We have some cases when we use label keys in format {uuid}_defaultEventing. The {uuid} part may contain dashes, for example `e98614ab-f02a-4ab0-ba8b-574996d38212_defaultEventing`. Right now we create those labels from the Director backend so no validation rules are applied, but maybe it is worth discussing.

I see two options now: 
- loosen validation for the label keys allowing the `-` character, but still, it won't be possible to create labels with that character within `registerApplication` and `registerRuntime` mutations, only by `setRuntimeLabel` and `setApplicationLabel`.
- leave it as it is, accepting that our API will return something not fully supported

Issue #990, opened 2020-03-17
Enable the API Packages in SKR


mszostok:
**Description**

On open-source Kyma the API Packages are not enabled by default in version 1.11. The behavior for the  runtime should be different and there the packages should be enabled. This can be done by sending global override `global.enableAPIPackages=true` during the Kyma installation.

**AC**
- KEB sends global `global.enableAPIPackages` override during Kyma installation

PR #991, opened 2020-03-17
[KEB] Add packages support override for SKR

mszostok:
**Description**

- Add packages support override for SKR

**Related issue(s)**

- https://github.com/kyma-incubator/compass/issues/990

### DIFF
--- a/components/kyma-environment-broker/internal/process/provisioning/input/input.go
+++ b/components/kyma-environment-broker/internal/process/provisioning/input/input.go
@@ -3,8 +3,8 @@ package input
 import (
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal"
 	"github.com/kyma-incubator/compass/components/provisioner/pkg/gqlschema"
-
 	"github.com/pkg/errors"
+	"github.com/vburenin/nsync"
 )
 
 const (
@@ -17,8 +17,10 @@ type Config struct {
 }
 
 type RuntimeInput struct {
-	input     gqlschema.ProvisionRuntimeInput
-	overrides map[string][]*gqlschema.ConfigEntryInput
+	input           gqlschema.ProvisionRuntimeInput
+	mutex           *nsync.NamedMutex
+	overrides       map[string][]*gqlschema.ConfigEntryInput
+	globalOverrides []*gqlschema.ConfigEntryInput
 
 	hyperscalerInputProvider  HyperscalerInputProvider
 	optionalComponentsService OptionalComponentService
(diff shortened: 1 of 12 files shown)


components/kyma-environment-broker/internal/process/provisioning/input/input.go
```diff
@@ -17,8 +17,10 @@ type Config struct {
 }
 
 type RuntimeInput struct {
-	input     gqlschema.ProvisionRuntimeInput
-	overrides map[string][]*gqlschema.ConfigEntryInput
+	input           gqlschema.ProvisionRuntimeInput
+	mutex           *nsync.NamedMutex
```
piotrmiskiewicz: what is the purpose of using named mutex? You are locking access to a field components by different threads. That's ok, but why NamedMutext. Methods SetOverrides and AppendOverrides shares the same field but use different lock name - it works like different mutexes.

components/kyma-environment-broker/internal/process/provisioning/input/input.go
```diff
@@ -30,13 +32,39 @@ func (r *RuntimeInput) SetProvisioningParameters(params internal.ProvisioningPar
...
+	r.mutex.Lock("SetOverrides")
+	defer r.mutex.Unlock("SetOverrides")
+
 	r.overrides[component] = overrides
 	return r
 }
 
+// AppendOverrides appends overrides for the given components, the existing overrides are preserved.
+func (r *RuntimeInput) AppendOverrides(component string, overrides []*gqlschema.ConfigEntryInput) internal.ProvisionInputCreator {
+	r.mutex.Lock("AppendOverrides")
```
piotrmiskiewicz: SetOverrides and AppendOverrides works with overrides field but uses different mutex name which works like different mutex. In my opinion it should be the same name.

components/kyma-environment-broker/internal/process/provisioning/input/input_test.go
```diff
@@ -90,3 +155,25 @@ func fixKymaComponentList() []v1alpha1.KymaComponent {
...
+	optComponentsSvc.On("ComputeComponentsToDisable", []string(nil)).Return([]string{})
+	optComponentsSvc.On("ExecuteDisablers", mappedComponentList).Return(mappedComponentList, nil)
+	return optComponentsSvc
+}
+
+func assertContainsAllOverrides(t *testing.T, gotOverrides []*gqlschema.ConfigEntryInput, expOverrides ...[]*gqlschema.ConfigEntryInput) {
+	var expected []*gqlschema.ConfigEntryInput
+	for _, o := range expOverrides {
+		expected = append(expected, o...)
+
```
piotrmiskiewicz: unnecessary line

PR #997, opened 2020-03-18
Allow to configure connection pool to the database in Director, Tenan…

aszecowka:
…t Fetcher and Tenant Loader

**Description**

- DB pool configuration for Director: 30 active/2 idle - by default Postgres (embedded and GCP managed) has configured `max_connections = 100`, because of Director, Broker, and Provisioner  uses the same Postgres server, I decided to use 1/3 of available connections. 
Average Grapgl query for runtimes (when 300 runtimes defined) takes 1.93s.
For 50 active/5 idle connections it was only a little bit faster: 1.86s. For max one active and one idle connection, average time was 25s. 
- Tenant Loader and tenant fetcher: 1 active/1 idle because these two components do not perform concurrent requests to DB. 
- ...

**Related issue(s)**


components/director/README.md
```diff
@@ -71,4 +73,4 @@ The Director binary allows to override some configuration parameters. You can sp
 
 ## Usage
 
-Example GraphQL calls can be found [here](examples/README.md)
\ No newline at end of file
+Example GraphQL calls can be found [here](examples/README.md)
```
klaudiagrz:
```suggestion
Find examples of GraphQL [here](examples/README.md).
```

Issue #968, opened 2020-03-15
Create an Azure Event-hub instance in the same location as the provisioned kyma-runtime


anishj0shi:
while creating a service instance in the Kyma Environment Broker, we need to propagate azure's region from the provisioning parameter to the event-hub namespace provisioning step.  as the event-hub namespace has to semantically in the same region to that of provisioned kyma-runtime

anishj0shi:
@k15r  you probably have to be careful in propagating Region from provisioning parameter directly, as it might have regions which are not acceptable regions for azure.

k15r:
well then i guess we need a mapping from 'unacceptable' => 'acceptable'
Who would provide this mapping?

anishj0shi:
As other cloud providers are not supported right now, it just makes sense to validate the request with azure's region list. Having a 1:1 mapping seems unrealistic and almost impossible to maintain.

PR #1006, opened 2020-03-19
use provisioning location for eventhubs

k15r:
**Description**

- use location from provisioning parameters for eventhub region
- do naive mapping from azure to google

**Related issue(s)**

#968


components/kyma-environment-broker/internal/hyperscaler/azure/env.go
```diff
@@ -24,11 +27,37 @@ func GetConfig(clientID, clientSecret, tenantID, subscriptionID, location string
...
+	switch parameters.PlanID {
+	case broker.AzurePlanID:
+		if !azureRegions()[region] {
+			return "", fmt.Errorf("supplied region \"%v\" is not a valid region for Azure", region)
+		}
+
+	case broker.GcpPlanID:
+		if azureRegion, ok := gcpToAzureRegionMapping()[region]; ok {
+			region = azureRegion
+			break
```
nachtmaar: (nit) Why is this break required ?

components/kyma-environment-broker/internal/hyperscaler/azure/env_test.go
```diff
@@ -0,0 +1,107 @@
...
+			name: "valid azure region",
+			args: args{
+				hyperscalerType: hyperscaler.Azure,
+				planID:          broker.AzurePlanID,
+				region:          "westeurope",
+			},
+			want:    "westeurope",
+			wantErr: false,
+		},
+		// TODO: Add test cases.
```
nachtmaar: Leftover?

components/kyma-environment-broker/internal/hyperscaler/azure/env_test.go
```diff
@@ -0,0 +1,107 @@
...
+			}
+
+			got, err := mapRegion(credentials, parameters)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("mapRegion() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if got != tt.want {
+				t.Errorf("mapRegion() got = %v, want %v", got, tt.want)
+			}
```
nachtmaar: (nit): KEB is using `stretchr/testify/`

```suggestion
			if tt.wantErr {
				require.NotNil(t, err, "mapRegion() error = %v, wantErr %v", err, tt.wantErr)
			} else {
				require.Nil(t, err, "mapRegion() error = %v, wantErr %v", err, tt.wantErr)
			}
			assert.Equal(t, got, tt.want, "mapRegion() got = %v, want %v", got, tt.want)
```

components/kyma-environment-broker/internal/hyperscaler/azure/env_test.go
```diff
@@ -0,0 +1,107 @@
+package azure
+
+import (
+	"testing"
+
```
nachtmaar: In case you want the testitify suggestion below:

```suggestion

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"

```

components/kyma-environment-broker/internal/process/provisioning/event_hub_step_test.go
```diff
@@ -19,10 +19,10 @@ import (
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/broker"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/hyperscaler"
-	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/hyperscaler/automock"
+	hyperscalerautomock "github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/hyperscaler/automock"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/hyperscaler/azure"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/process/provisioning/input"
-	inputAutomock "github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/process/provisioning/input/automock"
+	inputautomock "github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/process/provisioning/input/automock"
```
nachtmaar: 👍

components/kyma-environment-broker/internal/hyperscaler/azure/env.go
k15r: otherwise it would require an `else`

components/kyma-environment-broker/internal/hyperscaler/azure/env_test.go
k15r: good catch

components/kyma-environment-broker/internal/hyperscaler/azure/config.go
```diff
@@ -43,3 +43,75 @@ func (c *Config) Environment() (*azure.Environment, error) {
 
 	return c.environment, nil
 }
+
+func azureRegions() map[string]bool {
```
mszostok: use ` map[string]struct{}` instead of `map[string]bool`.

`struct` [occupies zero bytes of storage](https://dave.cheney.net/2014/03/25/the-empty-struct) (here is no need to have a bool flag).

components/kyma-environment-broker/internal/hyperscaler/azure/env_test.go
```diff
@@ -0,0 +1,107 @@
...
+func Test_mapRegion(t *testing.T) {
+	type args struct {
+		hyperscalerType hyperscaler.HyperscalerType
+		planID          string
+		region          string
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    string
```
mszostok: want -> wantRegion

components/kyma-environment-broker/internal/hyperscaler/azure/config.go
```diff
@@ -43,3 +43,75 @@ func (c *Config) Environment() (*azure.Environment, error) {
...
+		"westcentralus":      true,
+		"westeurope":         true,
+		"westindia":          true,
+		"westus":             true,
+		"westus2":            true,
+	}
+}
+
+func gcpToAzureRegionMapping() map[string]string {
+	return map[string]string{
```
mszostok: could you also adjust this enum in plan schema: https://github.com/kyma-incubator/compass/blob/master/components/kyma-environment-broker/internal/broker/plans.go#L63

so just put there your region list from this map:
```
               "enum": ["asia-east1", "asia-east2", "asia-northeast1", "asia-northeast2", "asia-northeast3", "asia-south1", "asia-southeast1", "australia-southeast1", "europe-north1", "europe-west1", "europe-west2", "europe-west3", "europe-west4", "europe-west6", "northamerica-northeast1", "southamerica-east1", "us-central1", "us-east1", "us-east4", "us-west1", "us-west2", "us-west3"]
```

as a result it will be unified, and we also have validation for this plan schema so in `mapRegion` you will don't get anything unexpected :)

piotrmiskiewicz: Be aware the region list (in plans.go) must be valid from Gardener point of view - Gardener must accept the region, so we cannot add an existing GCP region which is not supported by Gardener.

components/kyma-environment-broker/internal/hyperscaler/azure/config.go
k15r: the bool makes the check for membership really nice to read. But if size is of concern here i will change it

components/kyma-environment-broker/internal/hyperscaler/azure/config.go
k15r: so it would be safer to use the list from the schema in my code?

components/kyma-environment-broker/internal/hyperscaler/azure/config.go
k15r: done

components/kyma-environment-broker/internal/hyperscaler/azure/config.go
piotrmiskiewicz: yes, you can use the list from the schema. please, add there (in the schema) a comment regarding usage of region so in the future if someone adds a new region - the gcpToAzureRegionMapping function could be changed.

k15r: will do! 👍

components/kyma-environment-broker/internal/hyperscaler/azure/config.go
```diff
@@ -43,3 +43,49 @@ func (c *Config) Environment() (*azure.Environment, error) {
 
 	return c.environment, nil
 }
+
+// see internal/broker/plans.go for list of available azure regions
+// important" ! filter non eventhub regions out
+
+func azureRegions() map[string]struct{} {
+	return map[string]struct{}{
```
PK85: We need to leave only two regions: westeurope, westus2.

components/kyma-environment-broker/internal/broker/plans.go
```diff
@@ -9,7 +11,198 @@ const (
...
+	}
+
+	bytes, err := json.Marshal(rs)
+	if err != nil {
+		panic(err)
+	}
+	return bytes
+}
+func AzureSchema() []byte {
+	f := new(bool)
```
anishj0shi: if you want pointer to a boolean, there's a helper function in KEB. 

https://github.com/kyma-incubator/compass/blob/2f2e7a8ed3d8c83be03e2198db7ccd4457f73f93/components/kyma-environment-broker/internal/ptr/ptr.go#L3

k15r: a good, to know, but imho we could keep it like this atm. Next refactoring will come and fix this if needed

PR #1015, opened 2020-03-20
Avs post kyma provision

abbi-gaurav:


components/kyma-environment-broker/internal/avs/internal_eval_assistant.go
```diff
@@ -0,0 +1,54 @@
...
+func NewInternalEvalAssistant(avsConfig Config) *InternalEvalAssistant {
+	return &InternalEvalAssistant{avsConfig: avsConfig}
+}
+
+func (iec *InternalEvalAssistant) CreateBasicEvaluationRequest(operations internal.ProvisioningOperation, url string) (*BasicEvaluationCreateRequest, error) {
+	return newBasicEvaluationCreateRequest(operations, iec, iec.avsConfig.GroupId, url)
+}
+
+func (iec *InternalEvalAssistant) SetOverrides(inputCreator internal.ProvisionInputCreator, evaluationId int64) {
+	inputCreator.SetOverrides("avs-bridge", []*gqlschema.ConfigEntryInput{
```
polskikiel: please use supported `AppendOverrides` function. `SetOverrides` has been deprecated, see [here](https://github.com/kyma-incubator/compass/blob/master/components/kyma-environment-broker/internal/model.go#L19).

polskikiel: then I would think to rename also `SetOverrides` function from `EvalAssistant` interface to `AppendOverrides` to be consistent

components/kyma-environment-broker/internal/avs/delegator.go
```diff
@@ -34,15 +27,15 @@ func NewDelegator(avsConfig Config, operationsStorage storage.Operations) *Deleg
...
 }
 
-func (del *Delegator) DoRun(logger logrus.FieldLogger, operation internal.ProvisioningOperation, modelSupplier func(provisioningOperation internal.ProvisioningOperation) (*BasicEvaluationCreateRequest, error)) (internal.ProvisioningOperation, time.Duration, error) {
+func (del *Delegator) DoRun(logger logrus.FieldLogger, operation internal.ProvisioningOperation, evalAssistant EvalAssistant, url string) (internal.ProvisioningOperation, time.Duration, error) {
 	logger.Infof("starting the step")
 
-	if operation.AvsEvaluationInternalId != 0 {
+	if evalAssistant.CheckIfAlreadyDone(operation) {
 		msg := fmt.Sprintf("step has already been finished previously")
 		return del.operationManager.OperationSucceeded(operation, msg)
```
polskikiel: please change this line to:
```
return operation, 0, nil
```
to skip this step.

Currently the
```
OperationSucceeded(operation, msg)
```
ends the whole provisioning process with success

polskikiel: don't mind, we fix it in this PR: https://github.com/kyma-incubator/compass/pull/1019

abbi-gaurav:
@polskikiel , I applied the feedback. please have a look.
I am waiting for Gabor to share the external tester access id so that I can test the post provisioning flow.

components/kyma-environment-broker/internal/avs/delegator.go
polskikiel: please return `return operation, 0, nil` here

abbi-gaurav: ok, from the previous comment, it seems I do not need to change it.

PR #1016, opened 2020-03-20
Add knative-eventing-kafka component

nachtmaar:
**Description**

- Follow up PR on https://github.com/kyma-incubator/compass/pull/905
- The code in the PR already creates the installation-overrides, now we can enable the component such that we have Kafka based eventing in Kyma

**How to Test**
```shell
$ kyma test run knative-eventing-kafka -w
```

**Related issue(s)**

### DIFF
--- a/chart/compass/charts/kyma-environment-broker/values.yaml
+++ b/chart/compass/charts/kyma-environment-broker/values.yaml
@@ -48,6 +48,8 @@ additionalRuntimeComponents:
     namespace: "kyma-system"
   - name: "compass-runtime-agent"
     namespace: "compass-system"
+  - name: "knative-eventing-kafka"
+    namespace: "knative-eventing"
 
 kymaVersion: "1.10.0"
 
--- a/chart/compass/charts/kyma-environment-broker/values.yaml
+++ b/chart/compass/charts/kyma-environment-broker/values.yaml
@@ -51,7 +51,7 @@ additionalRuntimeComponents:
   - name: "knative-eventing-kafka"
     namespace: "knative-eventing"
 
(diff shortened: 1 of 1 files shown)


chart/compass/charts/kyma-environment-broker/values.yaml
```diff
@@ -48,8 +48,10 @@ additionalRuntimeComponents:
     namespace: "kyma-system"
   - name: "compass-runtime-agent"
     namespace: "compass-system"
+  - name: "knative-eventing-kafka"
+    namespace: "knative-eventing"
 
-kymaVersion: "1.10.0"
+kymaVersion: "master-99c5dca2"
```
anishj0shi: I'm not sure we should propose this SHA change. :p

nachtmaar: It is used to set the `APP_KYMA_VERSION` environment variable in the KEB deployment. We need to update it to have the resources for knative-eventing-kafka inside the installer image. We added the component files in this [PR](https://github.com/kyma-project/kyma/pull/7486). So we need a newer Kyma version than 1.10. We could also use 1.11 as soon as it is out.

mszostok:
```suggestion
kymaVersion: "1.11.0-rc2"
```

I've tested that with rc2 and more or less it's working

mszostok:
please also update this chart on Kyma repo. Our pipelines are executed only when Kyma repo is updated: https://github.com/kyma-project/kyma/tree/master/resources/compass/charts/kyma-environment-broker

PR #1020, opened 2020-03-21
Add GQL types and operations for automatic scenarios assignment

crabtree:
**Description**

- Add GQL types and operations for automatic scenarios assignment

**Related issue(s)**


components/director/internal/domain/root_resolver.go
```diff
@@ -250,6 +251,14 @@ func (r *queryResolver) Tenants(ctx context.Context) ([]*graphql.Tenant, error)
 	return r.tenant.Tenants(ctx)
 }
 
+func (r *queryResolver) AutomaticScenarioAssignment(ctx context.Context, scenarioName string) (*graphql.AutomaticScenarioAssignment, error) {
+	return nil, errors.New("not implemented")
```
kfurgol: Maybe it would be good to return some mock data instead of `not implemented` error? Open question

crabtree: I would go with the implementation of the mocks in the actual scenarios assignment domain layer (in the scenarios assignment dedicated resolver)

chart/compass/charts/director/config.yaml
```diff
@@ -15,6 +15,8 @@ graphql:
     integrationSystems: ["integration_system:read"]
     viewer: []
     tenants: ["tenant:read"]
+    automaticScenarioAssignments: ["scenario_assignment:read"]
```
tgorgol: Maybe we could think about a different name for this scope because there might be an assumption that the user without `scenario_assignment:write` scope can't assign scenarios, while it's still possible for him by setting the label. I was thinking about something that highlights the `automatic` part, eg. `automatic_scenario_assignment` or just `automatic_scenario`?

crabtree: fair enough will go with `automatic_scenario_assignment`

Issue #1024, opened 2020-03-23
Implement data model for AutomaticScenarioAssignment


crabtree:
**Description**

Implement a data model including:
- database migrations
- entities
- models

Remember that AutomaticScenarioAssignment is scoped to a specific tenant.

**Reasons**

To smoothly work in parallel on the ApplicationScenarioAssignment topics we need this as a prerequisite.

**Attachments**

PR #1035, opened 2020-03-23
Create AutomaticScenarioAssignement migrations and entity

kjaksik:
**Description**

- create migrations
- create entity struct

**Related issue(s)**
#1024


components/schema-migrator/migrations/director/202003231306_add_automatic_scenario_assignement.up.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE automatic_scenario_assignments (
+    scenario VARCHAR(128),
+    tenant_id UUID NOT NULL,
+    key VARCHAR(256) NOT NULL,
+    value JSONB);
```
aszecowka: I am wondering if the value should be of type JSONB, or just pure string. We discussed that we are going to support only simple values. @crabtree what do you think?

crabtree: I'd go with simple string here.

components/director/internal/domain/scenarioassignement/entity.go
```diff
@@ -0,0 +1,8 @@
+package scenarioassignement
+
+type Entity struct {
+	Scenario string `db:"scenario"`
+	TenantID string `db:"tenant_id"`
+	Key      string `db:"key"`
```
crabtree: How about:
```suggestion
	SelectorKey      string `db:"selector_key"`
```

components/director/internal/domain/scenarioassignement/entity.go
```diff
@@ -0,0 +1,8 @@
...
+	Value    string `db:"value"`
```
crabtree: How about:
```suggestion
	SelectorValue    string `db:"selector_value"`
```

Issue #942, opened 2020-03-10
Provisioning of custom components


janmedrek:
**Description**

In current state, Provisioner accepts ComponentConfigurationInput of following model:
```
&gqlschema.ComponentConfigurationInput{ 
 			Component: component.Name, 
 			Namespace: component.Namespace, 
}
```

With the recent addition to Kyma Installer, there is now a possibility to define the URL Source of a component to be installed. We would like to introduce this functionality in Provisioner, this would require changing the input model to the following structure:
```
&gqlschema.ComponentConfigurationInput{ 
 			Component: component.Name, 
 			Namespace: component.Namespace, 
 			Source:
 			  URL: component.Source.URL,
}
```

**Acceptance Criteria**
- Provisioner accepts URL Source as a parameter for Component Configuration
- URL Sources are propagated to KymaInstallation CR in a provisioned cluster

**Subtasks**
- [x] Provisioner schema changed
- [x] Database schema updated
- [x] Input for documentation provided
- [ ] Documentation updated

PR #1046, opened 2020-03-24
Enable component custom sources in Provisioner schema

franpog859:
**Description**

- Enable component custom sources feature in Provisioner schema

GraphQL mutation change example:
```gql
mutation {
   provisionRuntime(
      config: {
         kymaConfig: {
            components: [
               {
                  component: "cluster-essentials"
                  namespace: "kyma-system"
                  sourceURL: "github.com/kyma-project/kyma.git//resources/cluster-essentials" 
               }
...
```

**Related issue(s)**

See the issue #942


components/provisioner/internal/provisioning/input_converter_test.go
```diff
@@ -498,6 +501,11 @@ func fixKymaGraphQLConfigInput() *gqlschema.KymaConfigInput {
 					fixGQLConfigEntryInput("test.config.key2", "value2", util.BoolPtr(false)),
 				},
 			},
+			{
+				Component: rafterComponent,
+				Namespace: kymaSystemNamespace,
+				SourceURL: util.StringPtr(rafterSourceURL),
+			},
```
Szymongib: Can we also add a component with the source URL to resolver integration tests? That way we will verify it works fine with DB,

franpog859: All right!

franpog859:
@Szymongib sure! I talked with the team responsible for the nested field in the Installer CR. It was done "just in case" so I think the simple version could be better here, in the Provisioner API

PR #1055, opened 2020-03-24
Implement async deprovisioning in KEB

polskikiel:
**Description**

- Implemented async deprovisioning in KEB
   - moved db structs to separate package dbmodel
   - grouped steps definition in main to array per operation type
   - processing deprovisioning operations at start 

**Related issue(s)**


components/kyma-environment-broker/internal/avs/delegator.go
```diff
@@ -21,14 +20,14 @@ const (
 )
 
 type Delegator struct {
-	operationManager *process.OperationManager
+	operationStorage storage.Provisioning
```
polskikiel: couldn't use `OperationManager` because of the imports loop with provisioning package

components/kyma-environment-broker/internal/broker/instance_deprovision.go
```diff
@@ -46,20 +54,35 @@ func (b *DeprovisionEndpoint) Deprovision(ctx context.Context, instanceID string
...
+		logger.Errorf("cannot create new operation: %s", err)
+		return domain.DeprovisionServiceSpec{}, errors.New("cannot create new operation")
 	}
-
-	err = b.instancesStorage.Delete(instanceID)
+	err = b.operationsStorage.InsertDeprovisioningOperation(operation)
 	if err != nil {
-		logger.Errorf("unable to delete instance: %s", err)
-		return domain.DeprovisionServiceSpec{}, apiresponses.NewFailureResponse(err, http.StatusInternalServerError, fmt.Sprintf("could not delete instance, instanceID %s", instanceID))
+		logger.Errorf("cannot save operations: %s", err)
```
piotrmiskiewicz: cannot save operation

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -2,10 +2,17 @@ package main
 
 import (
 	"context"
+
+	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/process/deprovisioning"
+	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/process/provisioning"
+	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/storage/dbsession/dbmodel"
```
jasiu001: should be moved to proper section below

components/kyma-environment-broker/internal/avs/delegator.go
piotrmiskiewicz: It is strange, why only the AvS suffers? what is the cycle (which packages)?

components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
```diff
@@ -0,0 +1,117 @@
...
+	switch status.State {
+	case gqlschema.OperationStateSucceeded:
+		repeat, err := s.removeInstance(instance.InstanceID)
+		if err != nil || repeat != 0 {
+			return operation, repeat, err
+		}
+		return s.operationManager.OperationSucceeded(operation, msg)
+	case gqlschema.OperationStateInProgress:
+		return operation, 2 * time.Minute, nil
+	case gqlschema.OperationStatePending:
```
piotrmiskiewicz: what is the difference between pending and in progress?
next question: how long does the deprovisioning take? if it is 5 min - retries must not be bigger than half a minute in my opinion. if it takes 1 minute usually, lets retry after 15 sec

components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+}
+
+func (s *InitialisationStep) checkRuntimeStatus(operation internal.DeprovisioningOperation, log logrus.FieldLogger) (internal.DeprovisioningOperation, time.Duration, error) {
+	if time.Since(operation.UpdatedAt) > CheckStatusTimeout {
+		log.Infof("operation has reached the time limit: updated operation time: %s", operation.UpdatedAt)
+		return s.operationManager.OperationFailed(operation, fmt.Sprintf("operation has reached the time limit: %s", CheckStatusTimeout))
+	}
+
+	instance, err := s.instanceStorage.GetByID(operation.InstanceID)
```
jasiu001: you fetch instance in Run method

components/kyma-environment-broker/internal/process/provisioning/initialisation.go
```diff
@@ -48,7 +47,7 @@ func NewInitialisationStep(os storage.Operations, is storage.Instances, pc provi
 }
 
 func (s *InitialisationStep) Name() string {
-	return "Initialization"
+	return "Provision_Init"
```
jasiu001: Provision_Initialization, it's used to log, it doesn't have to be short

components/kyma-environment-broker/internal/avs/delegator.go
jasiu001: I'm not sure but probably because all steps have logic inside `process/provisioning` dir, and this step all logic has outside. In `process/provisioning` this step has only initiation.

piotrmiskiewicz: rename manger to (ProvisioningManager|DeprovisioningManager)

components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
polskikiel: it takes about 15mins if it's successful so I will set retries to 1min

components/kyma-environment-broker/internal/avs/delegator.go
polskikiel: reverted

Issue #1023, opened 2020-03-23
Implement mutation for deleting AutomaticScenarioAssignment type for given scenario name


crabtree:
**Description**

Implement mutation for deleting AutomaticScenarioAssignment type for a given scenario. In order to provide the ability to remove the AutomaticScenarioAssignment entity by the scenario name we need to implement the mutation `deleteAutomaticScenarioAssignmentForScenario(scenarioName: String!): AutomaticScenarioAssignment!`, that for a given scenario name will delete the `AutomaticScenarioAssignment`.

**Attachments**

Issue #1031, opened 2020-03-23
Implement the flow for automatic scenario assignment based on the defined selector for runtimes


crabtree:
**Description**

Based on the selectors defined by the AutomaticScenarioAssignment entities we need to implement the mechanism that will automatically assign the runtimes to scenarios according to the selectors. 

- When new runtime is registered the mechanism should be triggered and assign this runtime to the appropriate scenarios based on AutomaticScenarioAssignment entities @kfurgol 
- When new AutomaticScenarioAssignment entity is created / modified / deleted all existing runtimes need to be verified whether there are changes required for the scenarios assignment @pkosiec @dbadura 
- When the scenarios assignment for a given runtime is modified it needs to be verified. When the scenario assignment by the automatic mechanism is removed it needs to be restored as it has higher priority @kfurgol 
- When changing the labels on the runtime we need to recalculate the AutomaticScenarioAssignment rules for that runtime. It may happen that the given selector will no longer be valid for a particular runtime so the scenarios assigned by this selector should be removed. @kfurgol 
- prepare integration test for ASA engine @kjaksik 

Please keep in mind that all AutomaticScenarioAssignment entities are bound to the tenant.

**Reasons**

**Attachments**

crabtree:
Additional requirement:
- When changing the labels on the runtime we need to recalculate the AutomaticScenarioAssignment rules for that runtime. It may happen that the given selector will no longer be valid for a particular runtime so the scenarios assigned by this selector should be removed.

PR #1070, opened 2020-03-25
Mutation for deleting AutomaticScenarioAssignment type for a given scenario

crabtree:
**Description**

- The mutation for deleting AutomaticScenarioAssignment type for a given scenario

**Related issue(s)**

See also #1023


components/director/internal/domain/scenarioassignment/fixtures_test.go
```diff
@@ -19,10 +20,13 @@ const (
 )
 
 func fixModel() model.AutomaticScenarioAssignment {
-	return model.AutomaticScenarioAssignment{
+	return fixModelWithScenarioName(scenarioName)
+}
+
+func fixGQLAutomaticScenarioAssignment() graphql.AutomaticScenarioAssignment {
```
pkosiec: If my PR comes first, then you can switch to `fixGQL` function; in the other case I will cleanup the fixtures here

components/director/internal/domain/scenarioassignment/service.go
```diff
@@ -65,3 +66,18 @@ func (s *service) GetForScenarioName(ctx context.Context, scenarioName string) (
 	}
 	return sa, nil
 }
+
+// DeleteForScenarioName deletes the assignment for a given scenario in a scope of a tenant
+func (s *service) DeleteForScenarioName(ctx context.Context, scenarioName string) error {
```
pkosiec: I assume the additional logic will be implemented here: https://github.com/kyma-incubator/compass/issues/1031
Is that right?

crabtree: Yes, this is only a basic operation. The #1031 issue covers business logic.

PR #1074, opened 2020-03-25
fix auditlog jwt message

dbadura:
**Description**

- oauth flow need to pass *user* as '$USER' value and *tenant* as '$PROVIDER' value. Those values are automatically filled by auditlog service.
- Add more information about the request to `securit-event`
- bump with PR image: https://github.com/kyma-project/kyma/pull/7806

**Related issue(s)**


dbadura:
It would be nice to have information of who called director with insufficient scopes.

components/gateway/cmd/main.go
```diff
@@ -105,16 +105,14 @@ func initAuditLogsSvc(done chan bool) AuditogService {
...
-	httpClient, tenant := initAuditlogClient(auditlogCfg)
-	uuidSvc := uuid.NewService()
-	timeSvc := &time.TimeService{}
+	httpClient, msgFactory := initAuditlogClient(auditlogCfg)
 
-	auditlogClient, err := auditlog.NewClient(auditlogCfg, httpClient, uuidSvc, timeSvc, tenant)
+	auditlogClient, err := auditlog.NewClient(auditlogCfg, httpClient)
 	exitOnError(err, "Error while creating auditlog client from cfg")
 
 	auditlogMsgChannel := make(chan auditlog.AuditlogMessage)
```
kjaksik: It would be nice to avoid stuttering in the auditlog types

components/gateway/internal/auditlog/service.go
```diff
@@ -122,21 +134,13 @@ func (svc *Service) parseResponse(response string) (model.GraphqlResponse, error
 	return graphqResponse, nil
```
kjaksik: Typo here :)

components/gateway/cmd/main.go
```diff
@@ -105,16 +105,14 @@ func initAuditLogsSvc(done chan bool) AuditogService {
 	err := envconfig.InitWithPrefix(&auditlogCfg, "APP")
 	exitOnError(err, "Error while loading auditlog cfg")
 
-	httpClient, tenant := initAuditlogClient(auditlogCfg)
-	uuidSvc := uuid.NewService()
-	timeSvc := &time.TimeService{}
+	httpClient, msgFactory := initAuditlogClient(auditlogCfg)
```
aszecowka:
```
	httpClient, msgFactory := initAuditlogClient(auditlogCfg)

	auditlogClient, err := auditlog.NewClient(auditlogCfg, httpClient)
```
don't you think that it looks strange? You initAuditLog, and then create `audiitlogClient`

components/gateway/cmd/main.go
```diff
@@ -105,16 +105,14 @@ func initAuditLogsSvc(done chan bool) AuditogService {
...
 
-	auditlogClient, err := auditlog.NewClient(auditlogCfg, httpClient, uuidSvc, timeSvc, tenant)
+	auditlogClient, err := auditlog.NewClient(auditlogCfg, httpClient)
 	exitOnError(err, "Error while creating auditlog client from cfg")
 
-	auditlogMsgChannel := make(chan auditlog.AuditlogMessage)
+	auditlogMsgChannel := make(chan auditlog.Message)
```
aszecowka: make this channel buffered, thanks to that your producer won't be blocked by the consumer. 
In addition to that, modify this code to avoid deadlocks:
```
func (sink *Sink) Log(request, response string, claims proxy.Claims) error {
	msg := Message{
		Request:  request,
		Response: response,
		Claims:   claims,
	}
	sink.logsChannel <- msg
	return nil
}
```
There can be a situation that no one reads from the channel, and even it will be buffered channel, at some moment this operation will block everything. Sending to the channel should be done with some timeout (pseudocode):
```
select {
  case sink.logsChannel <- msg:
 case <- time.After(time.Second*5):
   return errors.New("oh no, something is really wrong")
}
```

components/gateway/cmd/main.go
```diff
@@ -124,30 +122,40 @@ func initAuditLogsSvc(done chan bool) AuditogService {
...
+
+			return auditlog.NewBasicAuthClient(basicCfg), auditlog.BasicAuthMessageFactory("proxy", basicCfg.Tenant, uuidSvc, timeSvc)
+		}
+	case auditlog.OAuth:
+		{
+			ctx := context.Background()
+
+			var oauthCfg auditlog.OAuthConfig
+			err := envconfig.InitWithPrefix(&oauthCfg, "APP")
+			exitOnError(err, "while loading oauth config from envs")
```
aszecowka: IMHO `exitOnError` should be used only in the main function, other functions should just return errors

components/gateway/internal/auditlog/factory.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"time"
+
+	"github.com/kyma-incubator/compass/components/gateway/internal/auditlog/model"
+)
+
+const (
+	LogFormatDate  = "2006-01-02T15:04:05.999Z"
+	UserVariable   = "$USER"
```
aszecowka: please add somewhere the purpose of this consts: that they are automatically filled by external service, they are used probably only in OAuth flow. I am also wondering, if these variables should not be configurable (with current values as a default), I can imagine situation that External Audit Service is changing the name of those variables, and then we have to build a new image of our component, instead of just changing env variables.

components/gateway/internal/auditlog/service_test.go
```diff
@@ -221,6 +223,41 @@ func TestAuditlogService_LogConfigurationChange(t *testing.T) {
 	})
 
 }
+func TestSink_ChannelStuck(t *testing.T) {
+	//GIVEN
+	timeout := time.Millisecond * 100
```
aszecowka: why this test is so complicated:
- sink.Log is synchronous function AFAIK, so you don't have to create goroutine
- a channel can be not buffered - then if you will send only one message and don't read from the channel you will get an error
- please check if we do not have a resource leak (if channel should not be closed at the end of the test)

components/gateway/cmd/main.go
```diff
@@ -37,6 +39,11 @@ type HTTPTransport interface {
 	RoundTrip(req *http.Request) (resp *http.Response, err error)
 }
 
+const (
```
aszecowka: these constants seem to be unused

components/gateway/README.md
```diff
@@ -13,36 +13,43 @@ Gateway binary allows you to override some configuration parameters. You can spe
...
-| **APP_AUDITLOG_CHANNEL_SIZE**    | The auditlog Message Channel size                                              | `100`            | 
-| **APP_AUDITLOG_CHANNEL_TIMEOUT** | The auditlog Message Channel timeout                                           | `5s`             | 
+| Name                             | Description                                                                       |   Default value  | 
+| -------------------------------- | --------------------------------------------------------------------------------- | ---------------- | 
+| **APP_AUDITLOG_URL**             | The URL under which the audit log service is available                            |      None        | 
+| **APP_AUDITLOG_CONFIG_PATH**     | The path for logging configuration changes                                        |      None        | 
+| **APP_AUDITLOG_SECURITY_PATH**   | The path for logging security events                                              |      None        | 
+| **APP_AUDITLOG_AUTH_MODE**       | The audit log authorization mode. The possible values are `basic` and `oauth`.    |      None        |  
 
+Gateway sends audit log messages to the local channel.
```
aszecowka: what about rephrasing it to:
Gateway process AuditLog messages asynchronously using configurable Go channel.

components/gateway/internal/auditlog/service_test.go
```diff
@@ -225,38 +224,16 @@ func TestAuditlogService_LogConfigurationChange(t *testing.T) {
...
 func TestSink_ChannelStuck(t *testing.T) {
 	//GIVEN
-	timeout := time.Millisecond * 100
-	ctx, _ := context.WithTimeout(context.TODO(), timeout*2)
-	channelSize := 5
-	messageSize := channelSize + 1
-	chanMsg := make(chan auditlog.Message, channelSize)
-	done := make(chan error)
-	sink := auditlog.NewSink(chanMsg, timeout)
+	chanMsg := make(chan auditlog.Message)
```
aszecowka: this test is so simple right now :) Good job. Consider renaming test name, for example: `TestSink_TimeoutOnWrite`

dbadura: Yes,  the the test is shorter.

components/gateway/README.md
```diff
@@ -8,36 +8,48 @@ Gateway is a component that exposes a service through which it forwards the inco
...
-| **APP_AUDITLOG_SECURITY_PATH**   | The path for logging security events                                            |
-| **APP_AUDITLOG_AUTH_MODE**       | The audit log authorization mode. The possible values are `basic` and `oauth`.  |
+| Name                             | Description                                                                       |   Default value  | 
+| -------------------------------- | --------------------------------------------------------------------------------- | ---------------- | 
+| **APP_AUDITLOG_URL**             | The URL under which the audit log service is available                            |      None        | 
+| **APP_AUDITLOG_CONFIG_PATH**     | The path for logging configuration changes                                        |      None        | 
+| **APP_AUDITLOG_SECURITY_PATH**   | The path for logging security events                                              |      None        | 
+| **APP_AUDITLOG_AUTH_MODE**       | The audit log authorization mode. The possible values are `basic` and `oauth`.    |      None        |
```
klaudiagrz: I guess there must be any default value here 🤔 What is the authorization method if the user does not specify any?

components/gateway/README.md
```diff
@@ -8,36 +8,48 @@ Gateway is a component that exposes a service through which it forwards the inco
...
-| **APP_AUDITLOG_AUTH_MODE**       | The audit log authorization mode. The possible values are `basic` and `oauth`.  |
+| Name                             | Description                                                                       |   Default value  | 
+| -------------------------------- | --------------------------------------------------------------------------------- | ---------------- | 
+| **APP_AUDITLOG_URL**             | The URL under which the audit log service is available                            |      None        | 
+| **APP_AUDITLOG_CONFIG_PATH**     | The path for logging configuration changes                                        |      None        | 
+| **APP_AUDITLOG_SECURITY_PATH**   | The path for logging security events                                              |      None        | 
+| **APP_AUDITLOG_AUTH_MODE**       | The audit log authorization mode. The possible values are `basic` and `oauth`.    |      None        |  
 
-If you set **APP_AUDITLOG_AUTH_MODE** to `basic`, you are required to pass the following values:
+Gateway process AuditLog messages asynchronously using configurable Go channel.
```
klaudiagrz:
```suggestion
Gateway processes audit log messages asynchronously using the configurable Go channel.
```

components/gateway/README.md
```diff
@@ -8,36 +8,48 @@ Gateway is a component that exposes a service through which it forwards the inco
...
+| Name                             | Description                                                   |  
+| -------------------------------- | ------------------------------------------------------------- |  
+| **APP_AUDITLOG_USER**            | The username to the audit log service                         |
+| **APP_AUDITLOG_PASSWORD**        | The password to the audit log service                         |
+| **APP_AUDITLOG_TENANT**          | The tenant for which audit logs are created                   |
+
+If you set **APP_AUDITLOG_AUTH_MODE** to `oauth`, you must specify the following environment variables:
+
+| Name                              | Description                                                     |   Default value  | Required|
+| --------------------------------- | --------------------------------------------------------------- |----------------- |---------|
```
klaudiagrz:
```suggestion
| --------------------------------- | --------------------------------------------------------------- |----------------- |:---------:|
```

components/gateway/README.md
dbadura: The component won't start.

PR #1086, opened 2020-03-26
Label def scenario depends on assignments

aszecowka:
**Description**

- Update on `scenarios` label definition performs an additional check to ensure, that AutomaticScenarioAssignment scenarios matched a new definition.
- fixed enormous bug in the JSON Schema validator, ValidationResult.Error was always not nil
- simplified JSON Schema Validator tests

**Related issue(s)**


components/director/internal/domain/labeldef/service.go
```diff
@@ -155,3 +165,53 @@ func (s *service) validateExistingLabelsAgainstSchema(ctx context.Context, schem
...
+		}
+		if res.Error != nil {
+			return errors.Wrapf(res.Error, "Scenario Assignment [scenario=%s] is not valid against a new schema", used)
+		}
+	}
+	return nil
+}
+
+func (s *service) fetchScenariosFromAssignments(ctx context.Context, tenantID string) ([]string, error) {
+	m := make(map[string]struct{})
```
crabtree: To discuss, but maybe it is worth implementing a service method that will return you what you need directly from the database instead of doing this in the source code?

PR #1104, opened 2020-03-31
Document Automatic Scenario Assignment feature

aszecowka:
**Description**

- ...
- ...
- ...

**Related issue(s)**


docs/compass/03-automatic-scenario-assignment.md
```diff
@@ -0,0 +1,213 @@
...
+        "owner": "warehouse-admin@mycompany.com",
+        "scenarios": [
+          "WAREHOUSE"
+        ]
+      }
+    }
+  }
+}
+```
+[//]: # TODO: currently runtime is also assigned to the DEFAULT scenario, but should it?
```
pkosiec, line 100: I don't think it should. It would be great to write it down:
If automatic scenario assignment is specified and new runtime is matched by selector, DEFAULT scenario is not assigned (when this feature is turned on).

Obviously only when we agree that this is how it should work.

docs/compass/03-automatic-scenario-assignment.md
```diff
@@ -0,0 +1,213 @@
+# Automatic Scenario Assignment
+
+To connect and group your Applications and Runtimes, assign them to the same scenario.
```
pkosiec, line 3: I would say that the first sentence is not needed

docs/compass/03-automatic-scenario-assignment.md
```diff
@@ -0,0 +1,213 @@
...
+Automatic Scenario Assignment feature allows defining condition when a Scenario is automatically assigned to the Runtime.
```
pkosiec, line 4: We can specify what the condition means - a label selector). Then we could have here 2 sentences, which is good enough for an introduction.

docs/compass/03-automatic-scenario-assignment.md
```diff
@@ -0,0 +1,213 @@
+# Automatic Scenario Assignment
+
```
pkosiec, line 2: Add Introduction header

docs/compass/03-automatic-scenario-assignment.md
```diff
@@ -0,0 +1,213 @@
...
+   deleteAutomaticScenarioAssignmentForScenario(scenarioName: String!): AutomaticScenarioAssignment 
+   deleteAutomaticScenarioAssignmentsForSelector(selector: LabelSelectorInput!): [AutomaticScenarioAssignment!]! 
+}
+```
+When creating Assignment, the following conditions are checked:
+* for given scenario, at most one Assignment exists
+* given scenario has to exists
+* selector's value type is a string
+
+Director API contains queries for fetching all assignments, assignments for the given Scenario and the given label selector.
```
pkosiec, line 36: I would recommend adding subheaders:
`### Queries`, `### Mutations` etc.

docs/compass/03-automatic-scenario-assignment.md
pkosiec: I would not limit the feature to runtimes. We could rephrase it and write in another sentence that only Runtimes are supported for now.

docs/compass/03-automatic-scenario-assignment.md
```diff
@@ -0,0 +1,213 @@
...
+  "data": {
+    "runtimes": {
+      "data": [
+        {
+          "id": "b5e1bf58-e8bb-4bde-a9c0-87650b0909c0",
+          "name": "warehouse-runtime-1",
+          "labels": {
+            "owner": "warehouse-admin@mycompany.com",
+            "scenarios": [
+              "DEFAULT"
```
pkosiec, line 136: hmm, should we replace deleted auto-assigned scenario with DEFAULT? Or should we leave it empty?

docs/compass/03-automatic-scenario-assignment.md
```diff
@@ -0,0 +1,213 @@
...
+    }
+  }
+}
+```
+
+TODO: in the implementation, WAREHOUSE scenario is not removed
+
+The same situation occurs if you modify or remove the label `createdBy` for the Runtime.
+
+Assignment Runtime to scenarios occurs also when we define Assignment.
```
pkosiec, line 150: Please add subheaders to the Example section. Here we can start another subsection

docs/compass/03-automatic-scenario-assignment.md
klaudiagrz: I agree this might be helpful.

docs/compass/03-automatic-scenario-assignment.md
```diff
@@ -0,0 +1,213 @@
...
+   selector: Label!
+}
+
+type Label {
+   key: String!
+   value: Any!
+}
+
+```
+If the Runtime is labeled with a label that matches the label's selector, the Runtime is assigned to the given Scenario.
```
klaudiagrz, line 21:
```suggestion
If a Runtime is labeled with a label that matches the value of the **selector** parameter, the Runtime is assigned to the given Scenario.
```

docs/compass/03-automatic-scenario-assignment.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+Director API contains the following mutations for managing Automatic Scenario Assignments.
```
klaudiagrz, line 24:
```suggestion
Director API contains the following mutations for managing Automatic Scenario Assignments:
```

docs/compass/03-automatic-scenario-assignment.md
klaudiagrz, line 36:
```suggestion
Director API contains queries that allow you to fetch all assignments, fetch assignments for the given Scenario, and fetch assignments for the given label selector:
```

docs/compass/03-automatic-scenario-assignment.md
aszecowka: according to @klaudiagrz such header is not necessary (it should be used only in the README files)

PR #1106, opened 2020-03-31
Define liveness and readiness probes for Director

aszecowka:
**Description**

- liveness probe checks the connection with DB every 10 seconds, with an initial delay set to 50s (configurable). Delay is set, because we don't want to restart pod when a connection with DB is not yet established (see persistence.go `waitForPersistance`)
- there is no point for configuring readinessProbe, because `waitForPersistence` blocks exposing any API until we get a connection to DB. 
- `periodSeconds` and `timeoutSeconds` are configurable, but their values are the same as default values for livenessProbe

**Related issue(s)**


pkosiec:
> there is no point for configuring readinessProbe, because waitForPersistence blocks exposing any API until we get a connection to DB.

I think it makes sense to configure it - otherwise our container is treated as "ready" but "wait for persistence" operation can be still ongoing. It is useful especially when we use "wait for all pods ready" script on CI.

> The kubelet uses readiness probes to know when a container is ready to start accepting traffic. A Pod is considered ready when all of its containers are ready. One use of this signal is to control which Pods are used as backends for Services. When a Pod is not ready, it is removed from Service load balancers.

components/director/internal/healthz/handler.go
```diff
@@ -1,17 +1,33 @@
...
+}
+
 // NewHTTPHandler returns function which handles healtz calls
-func NewHTTPHandler(log *logrus.Logger) func(writer http.ResponseWriter, request *http.Request) {
+func NewHTTPHandler(p Pinger, log *logrus.Logger) func(writer http.ResponseWriter, request *http.Request) {
 	return func(writer http.ResponseWriter, request *http.Request) {
+		err := p.PingContext(request.Context())
+		if err != nil {
+			writer.WriteHeader(http.StatusInternalServerError)
+			_, err = writer.Write([]byte(fmt.Sprintf("error: %v", err)))
```
pkosiec, line 23: I'm wondering if we should log this error as well. Technically we can get k8s events and gather issues from there, but... We can have all in one place (Director logs) as well. What do you think?

chart/compass/charts/director/templates/deployment.yaml
```diff
@@ -89,6 +89,20 @@ spec:
...
+                  port: {{.Values.deployment.args.containerPort }}
+                  path: "/healthz"
+              initialDelaySeconds: {{ .Values.deployment.liveness.initialDelaySeconds }}
+              timeoutSeconds: {{ .Values.deployment.liveness.timeoutSeconds }}
+              periodSeconds: {{.Values.deployment.liveness.periodSeconds }}
+          readinessProbe:
+              httpGet:
+                  port: {{.Values.deployment.args.containerPort }}
+                  path: "/ready"
+              initialDelaySeconds: {{ .Values.deployment.readiness.initialDelaySeconds }}
```
dbadura, line 103: Where can I find this value?

aszecowka:
@pkosiec 
`waitForPersistence` pings DB every second and when it finishes it returns a connection to DB or in case of failures it returns errors and we exit with panic.
This method blocks exposing any API, only after we get connection to DB we start listening on incoming connections.

But without readiness Probe, Kubernetes treats such service as ready for accepting connections from the very beginning.

IMO the best option is to provide another endpoint `/ready`, which just returns HTTP 200. 
This endpoint is configured after `waitForPersistence` so thanks to that a pod will be marked as a ready, only when a connection to DB is established. 
This endpoint should be called frequently, to mark Pod ready as soon as possible, to not prolong director installation time. 
IMO it does not make sense to use healthz endpoint, because every 2 s (according to our current configuration) we will ping the DB.

pkosiec:
@aszecowka I agree. Another `/ready` endpoint sounds like the best option.

components/director/cmd/director/main.go
```diff
@@ -133,8 +133,11 @@ func main() {
 
 	mainRouter.HandleFunc(cfg.RuntimeMappingEndpoint, runtimeMappingHandlerFunc)
 
-	log.Infof("Registering Healthz endpoint...")
-	mainRouter.HandleFunc("/healthz", healthz.NewHTTPHandler(log.StandardLogger()))
+	log.Infof("Registering readiness endpoint...")
+	mainRouter.HandleFunc("/ready", healthz.NewReadinessHandler())
```
pkosiec, line 137: What do you think about `/readyz`?

https://github.com/kubernetes/kubernetes/pull/78458/files

chart/compass/charts/director/templates/deployment.yaml
aszecowka: good catch, I remove this configuration option

PR #1108, opened 2020-04-01
Refactor Scenario Assignment service to use Assignment engine

pkosiec:
**Description**

- Refactor Scenario Assignment service to use Assignment engine

**Related issue(s)**
#1031


components/director/internal/domain/scenarioassignment/engine_service.go
```diff
@@ -0,0 +1,30 @@
+package scenarioassignment
+
+import "github.com/kyma-incubator/compass/components/director/internal/model"
+
+type assignmentEngineService struct{}
```
aszecowka, line 5: package is called `scenarioassignment`, your type `assignmentEngineService`, maybe we can rename it to just `engine`?

components/director/internal/domain/scenarioassignment/service.go
```diff
@@ -27,16 +27,25 @@ type ScenariosDefService interface {
 	GetAvailableScenarios(ctx context.Context, tenantID string) ([]string, error)
 }
 
-func NewService(repo Repository, scenarioDefSvc ScenariosDefService) *service {
+//go:generate mockery -name=AssignmentEngineService -output=automock -outpkg=automock -case=underscore
+type AssignmentEngineService interface {
+	EnsureScenarioAssignedToRuntimesMatchingSelector(in model.AutomaticScenarioAssignment) error
```
aszecowka, line 32: minor: maybe those method names will be simpler:
```
	EnsureUpToDateLabels
	RemoveAssignedScenario
	RemoveAssignedScenarios
```

components/director/internal/domain/scenarioassignment/service.go
```diff
@@ -50,14 +59,20 @@ func (s *service) Create(ctx context.Context, in model.AutomaticScenarioAssignme
 		return model.AutomaticScenarioAssignment{}, err
 	}
 	err = s.repo.Create(ctx, in)
-	switch {
-	case err == nil:
-		return in, nil
```
aszecowka, line 55: on `err == nil` you can do nothing and keep `switch` statement

components/director/internal/domain/scenarioassignment/service.go
```diff
@@ -126,12 +141,22 @@ func (s *service) List(ctx context.Context, pageSize int, cursor string) (*model
 	return s.repo.List(ctx, tnt, pageSize, cursor)
 }
 
-func (s *service) DeleteForSelector(ctx context.Context, selector model.LabelSelector) error {
+func (s *service) DeleteManyForSameSelector(ctx context.Context, in []*model.AutomaticScenarioAssignment) error {
```
aszecowka, line 144: why on the resolver level we have to call our service twice?
```
	assignments, err := r.svc.ListForSelector(ctx, selector)
	if err != nil {
		return nil, errors.Wrapf(err, "while getting the Assignments for selector [%v]", selector)
	}

	err = r.svc.DeleteManyForSameSelector(ctx, assignments)
	if err != nil {
		return nil, errors.Wrapf(err, "while deleting the Assignments for selector [%v]", selector)
	}
```
Then it looks strange that we ask the same service for some data, and on the `Delete` method we validate again that data.

components/director/internal/domain/scenarioassignment/service.go
aszecowka: or even we can have the only one `Remove` method and call it in the loop.

pkosiec: > or even we can have the only one Remove method and call it in the loop.

I thought about it, but ideally I would like to do it on database level (with one query) or at least embed the loop in engine itself.

components/director/internal/domain/scenarioassignment/service.go
pkosiec: Unfortunately the `ListForSelector` is needed to return all assignments as a result of the resolver.

I had two options:
- pass the listed assignments to the service again
- query database twice :/

I picked the second option. Do you have any other ideas?

components/director/internal/domain/scenarioassignment/service.go
pkosiec: Yup, but I think in this way it is simpler. the `switch` looked good when it was at the end of the method. Of course that's just a personal preference.

PR #1111, opened 2020-04-01
Add documentation for deprovisioning

polskikiel:
**Description**

- Add documentation for deprovisioning

**Related issue(s)**


docs/kyma-environment-broker/03-04-add-deprovisioning-step.md
```diff
@@ -0,0 +1,158 @@
...
+## Add deprovisioning step
+
+To add a new deprovisioning step, follow these steps:
+
+1. Create a new file in [this](https://github.com/kyma-incubator/compass/blob/master/components/kyma-environment-broker/internal/process/deprovisioning) directory. 
+
+2. Implement this interface in your deprovisioning step:
+
+    ```go
+    type Step interface {
```
anishj0shi, line 20: I'm wondering, why there's a different step interface? doesn't it make sense if we add another method in current step interface's lifecycle?

polskikiel: Maybe that's an overhead right now, but I'd like to have this separated because we don't exactly know how many steps will be implemented in the future and how both structs - `ProvisionOperation` and `DeprovisionOperation` will differ. What's more not every ProvisioningOperation needs a deprovisioning callback.

Issue #1013, opened 2020-03-20
Azure event bus fails if the namespace length is less than 6


rakesh-garimella:
**Description**

Azure evnet hub fails with following message
`Error occured: Failed to persist Azure EventHubs Namespace [rg-38] with error: eventhub.NamespacesClient#CreateOrUpdate: Invalid input: autorest/validation: validation failed: parameter=namespaceName constraint=MinLength value=\"rg-38\" details: value length must be greater than or equal to 6, Retrying for 30m0s in 1m0s steps" instanceID=rg-38 operation=628725fb-e65b-41a0-a7a7-7a1a5d6eb3ca step="Provision Azure Event Hubs"`

nachtmaar:
In this case retrying really makes no sense. We should do validation on the name which is based on the provisioning parameters name, in the http put call to KEB. Moreover, we should print the error which causes the retry [here](https://github.com/kyma-incubator/compass/blob/b0fcdff99d61381abe2782c7e95f7415cb6bc74e/components/kyma-environment-broker/internal/process/operation.go#L64). Otherwise the error will only be shown after the last retry. @rakesh-garimella already did it in the above log line.

PR #1112, opened 2020-04-01
Validate length of name in parameter

sayanh:
**Description**

- Validate length of the name in parameter

**Related issue(s)**

See #1013

### DIFF
--- a/components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
+++ b/components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
@@ -69,11 +69,6 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
 	}
 	log.Infof("HAP lookup for credentials to provision cluster for global account ID %s on Hyperscaler %s", pp.ErsContext.GlobalAccountID, hypType)
 
-	// validate length of name parameter
-	if err = validateLengthOfName(pp.Parameters.Name); err != nil {
-		return p.operationManager.OperationFailed(operation, err.Error())
-	}
-
 	// get hyperscaler credentials from HAP
 	credentials, err := p.accountProvider.GardenerCredentials(hypType, pp.ErsContext.GlobalAccountID)
 	if err != nil {
@@ -139,13 +134,6 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
 	operation.InputCreator.SetOverrides(componentNameKnativeEventingKafka, getKafkaChannelOverrides(kafkaEndpoint, k8sSecretNamespace, "$ConnectionString", kafkaPassword, kafkaProvider))
 
 	return operation, 0, nil
-}
-
-func validateLengthOfName(name string) error {
-	if len(name) < 6 {
-		return fmt.Errorf("length of name [%s] in parameter must be greater than or equal to 6", name)
-	}
-	return nil
 }
 
 func extractEndpoint(accessKeys eventhub.AccessKeys) string {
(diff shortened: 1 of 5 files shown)


components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -68,6 +68,11 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
 	}
 	log.Infof("HAP lookup for credentials to provision cluster for global account ID %s on Hyperscaler %s", pp.ErsContext.GlobalAccountID, hypType)
 
+	// validate length of name parameter
+	if err = validateLengthOfName(pp.Parameters.Name); err != nil {
```
nachtmaar, line 72: Can't we move the validation to the synchronous http request instead of having it in the asynchronous step? That way, the user would get direct feedback.

components/kyma-environment-broker/internal/process/provisioning/event_hub_step_test.go
```diff
@@ -434,6 +444,22 @@ func fixProvisioningOperation(t *testing.T) internal.ProvisioningOperation {
...
 
+func fixProvisioningOperationWithInvalidName(t *testing.T) internal.ProvisioningOperation {
+	op := internal.ProvisioningOperation{
+		Operation: internal.Operation{},
+		ProvisioningParameters: `{
+			"plan_id": "4deee563-e5ec-4731-b9b1-53b42d855f0c",
+			"parameters": {
+                "name": "foo",
+        		"components": [],
+				"region": "westeurope"
```
nachtmaar, line 455: (nit) indention is wrong:

```suggestion
			"parameters": {
				"name": "foo",
				"components": [],
				"region": "westeurope"
```

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
anishj0shi: +1 to @nachtmaar , can we propose this in broker schema?

sayanh: Making this part of the synchronous step sounds good. How about adding this part of already existing validations [here](https://github.com/kyma-incubator/compass/blob/master/components/kyma-environment-broker/internal/broker/instance_create.go#L116)?

sayanh: I think, making this validation as a part of the struct will make things simple and concise.
```Name                        string   `json:"name" validate:"required,min=6"```

anishj0shi: I'd say, we'd also need to set the constraint on the graphQl layer over here: https://github.com/kyma-incubator/compass/blob/master/components/kyma-environment-broker/internal/broker/plans.go#L155

anishj0shi: > Making this part of the synchronous step sounds good. How about adding this in the part of already existing validations here? 
will this not throw runtime panic in case of unmarshalling? @sayanh

mszostok: Hi, I'm ok with adding that to the schema as @anishj0shi proposed 👍

sayanh: Changed in JSON schema.

components/kyma-environment-broker/internal/process/provisioning/event_hub_step_test.go
sayanh: Fixed

mszostok:
@nachtmaar I see that those requirements are quite resonable, so we can als specify them as a regex pattern for the `name` field in schema

sayanh:
@nachtmaar @mszostok I created https://github.com/kyma-incubator/compass/issues/1136 to fine-tune the name validation holistically. I'd prefer to solve it in that issue and not stretch the scope of this bug.

Issue #924, opened 2020-03-06
KEB: deleting service instance returns 500 error


pbochynski:
**Description**

It is not possible to successfully delete a service instance. Delete request returns operation id that is not persisted. The subsequent request returns the internal error (500). 

**Expected result**

The delete request returns OK

**Actual result**

```
"Deprovision instanceID:" "468099BF-7A41-4704-9A19-EF659A92D076"
{"timestamp":"1583488852.922296286","source":"kyma-env-broker","message":"kyma-env-broker.deprovision.unknown-error","log_level":2,"data":{"correlation-id":"1edf3a54-b3cb-4838-b549-23773759a7d3","error":"instance not found","instance-id":"468099BF-7A41-4704-9A19-EF659A92D076","session":"77"}}
{"timestamp":"1583488852.922296286","source":"kyma-env-broker","message":"kyma-env-broker.deprovision.unknown-error","log_level":2,"data":{"correlation-id":"1edf3a54-b3cb-4838-b549-23773759a7d3","error":"instance not found","instance-id":"468099BF-7A41-4704-9A19-EF659A92D076","session":"77"}}
127.0.0.1 - - [06/Mar/2020:10:00:52 +0000] "DELETE /v2/service_instances/468099BF-7A41-4704-9A19-EF659A92D076?service_id=47c9dcbf-ff30-448e-ab36-d3bad66ba281&plan_id=4deee563-e5ec-4731-b9b1-53b42d855f0c&accepts_incomplete=true HTTP/1.1" 500 37
```

**Steps to reproduce**

Create the instance and delete it.

pbochynski:
It still doesn't work in case provisioning failed or instance is already deleted. See example:
```
time="2020-04-01T10:13:23Z" level=info msg="Deprovisioning triggered, details: {PlanID:4deee563-e5ec-4731-b9b1-53b42d855f0c ServiceID:47c9dcbf-ff30-448e-ab36-d3bad66ba281 Force:false}" instanceID=468099BF-7A41-4704-9A19-EF659A92D076
{"timestamp":"1585736003.662887573","source":"kyma-env-broker","message":"kyma-env-broker.deprovision.instance-missing","log_level":2,"data":{"correlation-id":"f62bf91e-2f74-4af5-859f-1c207eb8e20e","error":"instance does not exist","instance-id":"468099BF-7A41-4704-9A19-EF659A92D076","session":"212"}}
[...]

PR #1116, opened 2020-04-02
Save instance to storage on provision call

polskikiel:
**Description**

- Save instance to storage on provision call

**Related issue(s)**

fixes https://github.com/kyma-incubator/compass/issues/924


components/kyma-environment-broker/internal/process/provisioning/create_runtime.go
```diff
@@ -97,7 +84,7 @@ func (s *CreateRuntimeStep) Run(operation internal.ProvisioningOperation, log lo
 		return operation, 1 * time.Minute, nil
 	}
 
-	err = s.instanceStorage.Insert(internal.Instance{
+	err = s.instanceStorage.Update(internal.Instance{
```
piotrmiskiewicz, line 87: it is better to get instance, update field(s) which must be updated and save it (operation update)

PR #1122, opened 2020-04-02
Allow disabling default scenario assignment

tgorgol:
**Description**

- Add feature flag allowing disabling default scenario assignment
- Allow removing last scenario from application (by deleting scenarios label)


components/director/internal/domain/application/service.go
```diff
@@ -256,14 +257,11 @@ func (s *service) Create(ctx context.Context, in model.ApplicationRegisterInput)
...
 		return "", err
 	}
+	s.scenariosService.AddDefaultScenarioIfEnabled(&in.Labels)
 
-	if _, ok := in.Labels[model.ScenariosKey]; !ok {
-		if in.Labels == nil {
-			in.Labels = map[string]interface{}{}
-		}
-		in.Labels[model.ScenariosKey] = model.ScenariosDefaultValue
+	if in.Labels == nil {
```
pkosiec, line 262: Hmm, if we need to initialize labels anyway, can we do it before `AddDefaultScenarioIfEnabled`? Then you won't need to use pointer on `map[string]interface{}` as you will only modify the existing `map[string]interface{}`
https://play.golang.org/p/Q6vrAmmJWR6

pkosiec: BTW Why here we always initialize `in.Labels`, but in Runtime service not?

components/director/README.md
```diff
@@ -45,30 +45,31 @@ $ PGPASSWORD=pgsql@12345 psql -U postgres -W -h 127.0.0.1 -f <(cat components/sc
...
+| **APP_DB_SSL**                               | `disable`                       | Database SSL mode (disable / enable)                           |
+| **APP_DB_MAX_OPEN_CONNECTIONS**              | `2`                             | The maximum number of open connections to the database         |                                                      
+| **APP_DB_MAX_IDLE_CONNECTIONS**              | `2`                             | The maximum number of connections in the idle connection pool  |
+| **APP_API_ENDPOINT**                         | `/graphql`                      | The endpoint for GraphQL API                                   |
+| **APP_PLAYGROUND_API_ENDPOINT**              | `/graphql`                      | The endpoint of GraphQL API for the Playground                 |
+| **APP_TENANT_MAPPING_ENDPOINT**              | `/tenant-mapping`               | The endpoint of Tenant Mapping Service                         |
+| **APP_SCOPES_CONFIGURATION_FILE**            |                                 | The path for scopes configuration file                         |
```
klaudiagrz, line 62: Please add None in fields where there are no default values.

components/director/README.md
```diff
@@ -45,30 +45,31 @@ $ PGPASSWORD=pgsql@12345 psql -U postgres -W -h 127.0.0.1 -f <(cat components/sc
...
+| -------------------------------------------- | ------------------------------- | -------------------------------------------------------------- |
+| **APP_ADDRESS**                              | `127.0.0.1:3000`                | The address and port for the service to listen on              |
+| **APP_DB_USER**                              | `postgres`                      | Database username                                              |
+| **APP_DB_PASSWORD**                          | `pgsql@12345`                   | Database password                                              |
+| **APP_DB_HOST**                              | `localhost`                     | Database host                                                  |
+| **APP_DB_PORT**                              | `5432`                          | Database port                                                  |
+| **APP_DB_NAME**                              | `postgres`                      | Database name                                                  |
+| **APP_DB_SSL**                               | `disable`                       | Database SSL mode (disable / enable)                           |
```
klaudiagrz, line 56:
```suggestion
| **APP_DB_SSL**                               | `disable`                       | Database SSL mode. The possible values are `disable` and `enable`.                          |
```

components/director/README.md
```diff
@@ -45,30 +45,31 @@ $ PGPASSWORD=pgsql@12345 psql -U postgres -W -h 127.0.0.1 -f <(cat components/sc
...
+| **APP_API_ENDPOINT**                         | `/graphql`                      | The endpoint for GraphQL API                                   |
+| **APP_PLAYGROUND_API_ENDPOINT**              | `/graphql`                      | The endpoint of GraphQL API for the Playground                 |
+| **APP_TENANT_MAPPING_ENDPOINT**              | `/tenant-mapping`               | The endpoint of Tenant Mapping Service                         |
+| **APP_SCOPES_CONFIGURATION_FILE**            |                                 | The path for scopes configuration file                         |
+| **APP_SCOPES_CONFIGURATION_FILE_RELOAD**     | `1m`                            | The period when the scopes configuration file is reloaded      |
+| **APP_JWKS_ENDPOINT**                        | `file://hack/default-jwks.json` | The path for JWKS                                              |
+| **APP_JWKS_SYNC_PERIOD**                     | `5m`                            | The period when the JWKS is synced                             |
+| **APP_ONE_TIME_TOKEN_URL**                   |                                 | The endpoint for fetching one time token                       |
```
klaudiagrz, line 66:
```suggestion
| **APP_ONE_TIME_TOKEN_URL**                   |                                 | The endpoint for fetching a one-time token                       |
```

components/director/README.md
```diff
@@ -45,30 +45,31 @@ $ PGPASSWORD=pgsql@12345 psql -U postgres -W -h 127.0.0.1 -f <(cat components/sc
...
-| APP_JWKS_SYNC_PERIOD                     | `5m`                            | The period when the JWKS is synced                            |
-| APP_ONE_TIME_TOKEN_URL                   |                                 | The endpoint for fetching one time token                      |
-| APP_CONNECTOR_URL                        |                                 | The endpoint of Connector                                     |
-| APP_OAUTH20_CLIENT_ENDPOINT              |                                 | The endpoint for managing OAuth 2.0 clients                   |
-| APP_OAUTH20_PUBLIC_ACCESS_TOKEN_ENDPOINT |                                 | The public endpoint for fetching OAuth 2.0 access token       |
-| APP_STATIC_USERS_SRC                     |                                 | The path for static users configuration file                  |
-| APP_LEGACY_CONNECTOR_URL                 |                                 | The URL of the legacy Connector signing request info endpoint |
+| ENV                                          | Default                         | Description                                                    |
```
klaudiagrz, line 48:
```suggestion
| Environment variable                                          | Default                         | Description                                                    |
```

components/director/internal/domain/application/service.go
tgorgol: Because in Application creation we're always adding `name` and `integrationSystemID` labels. I put the initilization right above them because it's tied with them, not with adding scenarios (which won't initialize labels if no scenario has to assigned - that happens in runtime creation).

PR #1125, opened 2020-04-02
Reprocess deprovisioning after fail

polskikiel:
**Description**

- Reprocess deprovisioning when there is an existing operation in DB.

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/broker/instance_deprovision.go
+++ b/components/kyma-environment-broker/internal/broker/instance_deprovision.go
@@ -5,8 +5,9 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/google/uuid"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal"
+
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/storage"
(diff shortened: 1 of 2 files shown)


components/kyma-environment-broker/internal/broker/instance_deprovision.go
```diff
@@ -54,29 +54,43 @@ func (b *DeprovisionEndpoint) Deprovision(ctx context.Context, instanceID string
...
+				return domain.DeprovisionServiceSpec{}, errors.New("cannot update existing operation")
+			}
+		case domain.InProgress:
+			// return existing operation
+			return domain.DeprovisionServiceSpec{
+				IsAsync:       true,
+				OperationData: existingOperation.ID,
+			}, nil
+		}
+	case dberr.IsNotFound(errStorage):
```
jasiu001, line 80: it doesn't have to be the separate case, return existing operation in the previous case and do not change that.

components/kyma-environment-broker/internal/broker/instance_deprovision.go
```diff
@@ -54,29 +54,43 @@ func (b *DeprovisionEndpoint) Deprovision(ctx context.Context, instanceID string
...
+		switch existingOperation.State {
+		case domain.Failed:
+			// reprocess operation again
+			operationID = existingOperation.ID
+			existingOperation.State = domain.InProgress
+			_, err = b.operationsStorage.UpdateDeprovisioningOperation(*existingOperation)
+			if err != nil {
+				return domain.DeprovisionServiceSpec{}, errors.New("cannot update existing operation")
+			}
+		case domain.InProgress:
```
jasiu001, line 73: use default, success operation also can be returned

components/kyma-environment-broker/internal/broker/instance_deprovision.go
```diff
@@ -54,29 +54,43 @@ func (b *DeprovisionEndpoint) Deprovision(ctx context.Context, instanceID string
...
-	operation, err := internal.NewDeprovisioningOperationWithID(operationID, instanceID)
-	if err != nil {
-		logger.Errorf("cannot create new operation: %s", err)
-		return domain.DeprovisionServiceSpec{}, errors.New("cannot create new operation")
-	}
-	err = b.operationsStorage.InsertDeprovisioningOperation(operation)
-	if err != nil {
-		logger.Errorf("cannot save operation: %s", err)
-		return domain.DeprovisionServiceSpec{}, errors.New("cannot save operation")
+		switch existingOperation.State {
```
jasiu001, line 64: it is not just simpler to add one condition:
```go
	case existingOperation != nil && !dberr.IsNotFound(errStorage):
		if existingOperation.State == domain.Failed {
			operationID = existingOperation.ID
			existingOperation.State = domain.InProgress
			_, err = b.operationsStorage.UpdateDeprovisioningOperation(*existingOperation)
			if err != nil {
				return domain.DeprovisionServiceSpec{}, errors.New("cannot update existing operation")
			}
		}
		return domain.DeprovisionServiceSpec{
			IsAsync:       true,
			OperationData: existingOperation.ID,
		}, nil
	}
```

PR #1129, opened 2020-04-03
Sort listed runtimes by name

sjanota:
**Description**

- Sort runtimes by name so the order is more natural and predictable to the cockpit user

**Related issue(s)**


crabtree:
Please update the `chart/compass/values.yaml` file with the Director image from your PR.

PR #1132, opened 2020-04-03
Adjust LMS timeouts

piotrmiskiewicz:
**Description**

- adjust LMS timeout for cert generation
- change output to forward
- direct dump provisioner mutation to stdout
- send globalaccount_id to LMS
- change tenant name normalization

Related: #814

### DIFF
--- a/components/kyma-environment-broker/internal/provisioner/client.go
+++ b/components/kyma-environment-broker/internal/provisioner/client.go
@@ -2,7 +2,6 @@ package provisioner
 
 import (
 	"context"
-	"crypto/tls"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -10,7 +9,6 @@ import (
 	schema "github.com/kyma-incubator/compass/components/provisioner/pkg/gqlschema"
 	gcli "github.com/machinebox/graphql"
 	"github.com/pkg/errors"
-	"net/http/httputil"
 )
 
 // accountIDKey is a header key name for request send by graphQL client
@@ -37,17 +35,7 @@ type client struct {
 }
 
 func NewProvisionerClient(endpoint string, queryDumping bool) Client {
-	httpClient := &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
-			},
-		},
-	}
-
-	httpclient := httputil.NewClientConn()
-
-	graphQlClient := gcli.NewClient(endpoint, gcli.WithHTTPClient(httpClient))
+	graphQlClient := gcli.NewClient(endpoint, gcli.WithHTTPClient(http.DefaultClient))
 	if queryDumping {
 		graphQlClient.Log = func(s string) {
 			fmt.Println(s)
@@ -128,7 +116,7 @@ func (c *client) ReconnectRuntimeAgent(accountID, runtimeID string) (string, err
 type GCPRuntimeStatus struct {
 	LastOperationStatus     *schema.OperationStatus         `json:"lastOperationStatus"`
 	RuntimeConnectionStatus *schema.RuntimeConnectionStatus `json:"runtimeConnectionStatus"`
-	RuntimeConfiguration    struct {
+	RuntimeConfiguration struct {
 		ClusterConfig         *schema.GCPConfig  `json:"clusterConfig"`
 		KymaConfig            *schema.KymaConfig `json:"kymaConfig"`
 		Kubeconfig            *string            `json:"kubeconfig"`
@@ -172,7 +160,7 @@ func (c *client) executeRequest(req *gcli.Request, respDestination interface{}) 
 	}
 
 	wrapper := &graphQLResponseWrapper{Result: respDestination}
-	err := c.graphQLClient.Run(context.Background(), req, wrapper)
+	err := c.graphQLClient.Run(context.TODO(), req, wrapper)
 	if err != nil {
 		return errors.Wrap(err, "Failed to execute request")
 	}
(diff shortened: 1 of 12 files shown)


components/kyma-environment-broker/internal/provisioner/client.go
```diff
@@ -25,16 +30,34 @@ type Client interface {
...
+
+	queryDumping bool
 }
 
-func NewProvisionerClient(endpoint string, queryLogging bool) Client {
+func NewProvisionerClient(endpoint string, queryDumping bool) Client {
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
```
mszostok, line 44: we do not want to make that configurable?

components/kyma-environment-broker/internal/provisioner/client.go
```diff
@@ -25,16 +30,34 @@ type Client interface {
...
 	queryProvider queryProvider
 	graphqlizer   Graphqlizer
+
+	queryDumping bool
 }
 
-func NewProvisionerClient(endpoint string, queryLogging bool) Client {
+func NewProvisionerClient(endpoint string, queryDumping bool) Client {
+	httpClient := &http.Client{
+		Transport: &http.Transport{
```
mszostok, line 42: What about the timeout? Right now you have a zero values and
>  Zero means no timeout.

so use `http.DefaultTransport.(*http.Transport).Clone()` and then change the values that you want to
+ add timeout to client itself


------
**EDIT:**

I just discover that we have a pkg for that 
```
httputil.NewClient(30, cfg.Director.SkipCertVerification)
```

but this package we also have a bug about default timeouts in transport layer. Please fix it as described above :)

components/kyma-environment-broker/internal/provisioner/client.go
```diff
@@ -25,16 +30,34 @@ type Client interface {
...
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		},
+	}
+
+	graphQlClient := gcli.NewClient(endpoint, gcli.WithHTTPClient(httpClient))
+	if queryDumping {
+		graphQlClient.Log = func(s string) {
+			fmt.Println(s)
```
mszostok, line 52: can we use logger here? We already have a 2 approaches for dumping info to stdout, adding 3rd one is not a good idea IMO

components/kyma-environment-broker/internal/provisioner/client.go
```diff
@@ -25,16 +30,34 @@ type Client interface {
...
+		}
+	}
+
 	return &client{
-		graphQLClient: graphql.NewGraphQLClient(endpoint, true, queryLogging),
+		graphQLClient: graphQlClient,
 		queryProvider: queryProvider{},
 		graphqlizer:   Graphqlizer{},
+		queryDumping:  queryDumping,
```
mszostok, line 60: why you store that?

components/kyma-environment-broker/internal/provisioner/client.go
```diff
@@ -132,9 +155,27 @@ func (c *client) RuntimeOperationStatus(accountID, operationID string) (schema.O
...
+	if reflect.ValueOf(respDestination).Kind() != reflect.Ptr {
+		return errors.New("destination is not of pointer type")
+	}
+
+	type graphQLResponseWrapper struct {
+		Result interface{} `json:"result"`
+	}
+
+	wrapper := &graphQLResponseWrapper{Result: respDestination}
+	err := c.graphQLClient.Run(context.Background(), req, wrapper)
```
mszostok, line 175: here we should have `context.TODO()` as this request should be canceled on shutdown

components/kyma-environment-broker/internal/provisioner/graphqlizer.go
```diff
@@ -139,6 +141,11 @@ func (g *Graphqlizer) UpgradeClusterConfigToGraphQL(in gqlschema.UpgradeClusterI
 }
 
 func (g *Graphqlizer) KymaConfigToGraphQL(in gqlschema.KymaConfigInput) (string, error) {
+	for _, c := range in.Components {
```
mszostok, line 144: personally I do not like when method is changing the input param when the method name does not say that. 

Maybe you better will to add just a quote function to tmpl engine in https://github.com/kyma-incubator/compass/pull/1132/files#diff-3b004e25742ce2ba4a362869881893e9R193
```go
fm["strQuote"] = strconv.Quote
```

then you can use `{{ .Value | strQuote }}`


working example: https://play.golang.org/p/Pu58JEhFCCg

mszostok, line 144: you change that so do we have a uni test for that? Because I see that contract was changed and unit tests were not adjusted which is quite odd

components/kyma-environment-broker/internal/provisioner/client.go
piotrmiskiewicz: My idea is to dump the request as it is - it is more readable. It is hard to see in the log what is going to the Provisioner. Anyway, we should not log the request because of sensitive data there, so in short time we need to change it/disable. I've create an issue: https://github.com/kyma-incubator/compass/issues/1169

If you still don't like to dump directly to stdout, I can revert this change.

mszostok: If you created an issue, and we will address that in the near future then it is ok for me 👍

components/kyma-environment-broker/internal/provisioner/graphqlizer.go
piotrmiskiewicz: contract was not changed, but the multiline string was not tested, I've added a test

components/kyma-environment-broker/internal/provisioner/client.go
piotrmiskiewicz: It seems we don't need this flag

components/kyma-environment-broker/internal/provisioner/client.go
piotrmiskiewicz: I've changed httputil

Issue #1017, opened 2020-03-20
Automatic status update for Application and Runtime entities


crabtree:
**Description**

Status of Application and Runtime entities should be automatically set to `CONNECTED` when the given Application/Runtime communicates with the Director API. The communication can be direct or through the Integration System. Thanks to the tenant-mapping-handler the context of a request contains the required information.

In case the user will modify the status manually, it can be replaced by the automatically computed one.

**Reasons**

We need that feature to know if application or runtime is connected to compass. Without that functionality all entities are marked as INITIAL only.

**AC**

1. Status of Application and Runtime entities must be automatically set to `CONNECTED` when the given Application/Runtime communicates with the Director API. Regardless of the previous status.
**Example**:  Application register package (status: CONNECTED); Runtime ask for applications (status: CONNECTED)

2. If IntegrationSystem communicates with the Director API it manages status of Runtimes and Applications and automatic set to `CONNECTED` feature is disabled (must not happen).
**Example**: Integration System sets Application/Runtime status to "FAILED" (status: FAILED); IS register Application packages without status attriubute (status: stays as it was)

3. In case of User flow update, status is not set to `CONNECTED`, stays as it was. In case the user will modify the status manually, it can be replaced by the automatically computed one.
**Example**:  User creates Application or Runtime by hand (status: as user set); User set status to application definition managed by application (status: changed to user one, when external application will update application definition status changes to `CONNECTED`)

PK85:
Check status:
1) :ok:

PR #1133, opened 2020-04-03
Add automatic status update feature

kjaksik:
**Description**

- Automatically set the application/runtime status to `CONNECTED` when it calls the director

**Related issue(s)**
#1017


components/director/internal/statusupdate/handler.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/kyma-incubator/compass/components/director/pkg/persistence"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/kyma-incubator/compass/components/director/internal/consumer"
+)
+
+type update struct {
+	table    Table
```
aszecowka, line 18: why table is a part of struct?  it can be a local variable in Handler method

components/director/internal/statusupdate/handler.go
```diff
@@ -0,0 +1,120 @@
...
+				u.table = runtimesTable
+			default:
+				next.ServeHTTP(w, r.WithContext(ctx))
+				return
+			}
+
+			tx, err := u.transact.Begin()
+			if err != nil {
+				u.writeError(w, errors.Wrap(err, "while opening transaction").Error(), http.StatusInternalServerError)
+				return
```
aszecowka, line 67: updating status is only auxiliary functionality. Because of that, even if we got an error I suggest continuing handling request. IMO, we should always call `next.ServeHTTP`, do not write errors in the response (just log them). @crabtree @pkosiec what do you think?

components/director/internal/statusupdate/handler.go
```diff
@@ -0,0 +1,120 @@
...
+
+			ctxWithDB := persistence.SaveToContext(ctx, tx)
+
+			isConnected, err := u.repo.IsConnected(ctxWithDB, consumerInfo.ConsumerID, string(u.table))
+			if err != nil {
+				u.writeError(w, errors.Wrap(err, "while checking status").Error(), http.StatusInternalServerError)
+				return
+			}
+
+			if isConnected {
```
aszecowka, line 79: seems to be simpler:
```
			if !isConnected {
				err = u.repo.UpdateStatus(ctxWithDB, consumerInfo.ConsumerID, string(u.table))
				if err != nil {

					u.writeError(w, errors.Wrap(err, "while updating status").Error(), http.StatusInternalServerError)
					return
				}
			}
```

components/director/internal/statusupdate/handler_test.go
```diff
@@ -0,0 +1,226 @@
...
+		Name             string
+		TxFn             func() (*persistenceautomock.PersistenceTx, *persistenceautomock.Transactioner)
+		RepoFn           func() *automock.StatusUpdateRepository
+		Request          *http.Request
+		ExpectedStatus   int
+		ExpectedResponse string
+		MockHandler      http.Handler
+	}{
+		{
+			Name: "Success when integration system",
```
aszecowka, line 46: what does it mean success? If you name this test, like "In case of Integration System, do nothing and execute next handler"  I would learn more just from the test name about the functionality.

components/director/internal/statusupdate/handler_test.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func createRequestWithClaims(t *testing.T, id string, consumerType consumer.ConsumerType) *http.Request {
+	req := http.Request{}
+	apiConsumer := consumer.Consumer{ConsumerID: id, ConsumerType: consumerType}
+	ctxWithConsumerInfo := consumer.SaveToContext(context.TODO(), apiConsumer)
+	return req.WithContext(ctxWithConsumerInfo)
+}
+
+func testHandler(t *testing.T, consumerID string, consumerType consumer.ConsumerType) http.HandlerFunc {
```
aszecowka, line 215: minor: fixNextHandler or dummyNextHandler

components/director/internal/statusupdate/handler_test.go
```diff
@@ -0,0 +1,226 @@
...
+			RepoFn: func() *automock.StatusUpdateRepository {
+				repo := automock.StatusUpdateRepository{}
+				repo.On("IsConnected", mock.Anything, testID, applicationsTable).Return(false, nil)
+				repo.On("UpdateStatus", mock.Anything, testID, applicationsTable).Return(testErr)
+				return &repo
+			},
+			Request:          createRequestWithClaims(t, testID, consumer.Application),
+			ExpectedStatus:   http.StatusInternalServerError,
+			ExpectedResponse: "{\"errors\":[{\"message\":\"while updating status: test\"}]}\n",
+			MockHandler:      testHandler(t, testID, consumer.Application),
```
aszecowka, line 183: minor: MockNextHandler, or DummyNextHandler

tests/director/gateway-integration/automatic_status_update_test.go
```diff
@@ -0,0 +1,182 @@
...
+		appOauthCredentialData, ok := appAuth.Auth.Credential.(*graphql.OAuthCredentialData)
+		require.True(t, ok)
+		require.NotEmpty(t, appOauthCredentialData.ClientSecret)
+		require.NotEmpty(t, appOauthCredentialData.ClientID)
+
+		t.Log("Issue a Hydra token with Client Credentials")
+		accessToken := getAccessToken(t, appOauthCredentialData, "application:read")
+		oauthGraphQLClient := gql.NewAuthorizedGraphQLClientWithCustomURL(accessToken, fmt.Sprintf("https://compass-gateway-auth-oauth.%s/director/graphql", domain))
+
+		t.Log("Requesting Viewer as Application")
```
aszecowka, line 125: this probably is not necessary, getApplication will be enough

components/director/internal/statusupdate/repo_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+		//WHEN
+		res, err := repo.IsConnected(ctx, testID, "applications")
+
+		//THEN
+		require.NoError(t, err)
+		assert.True(t, res)
+	})
+
+	t.Run("Success for runtimes", func(t *testing.T) {
```
dbadura, line 45: Do we need cases for `runtimes`?
How the differ from cases for `applications`?

The same question for `Update Query`

tests/director/gateway-integration/automatic_status_update_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
+	"github.com/kyma-incubator/compass/tests/director/pkg/gql"
+	"github.com/kyma-incubator/compass/tests/director/pkg/idtokenprovider"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAutomaticStatusUpdate(t *testing.T) {
+	domain := os.Getenv("DOMAIN")
```
dbadura, line 21: Do we need all those `os` calls in tests?
Can we for example use `testing.M`?

tests/director/gateway-integration/automatic_status_update_test.go
```diff
@@ -0,0 +1,182 @@
...
+			StatusCondition: &status,
+		}
+		appUpdated, err := updateApplicationWithinTenant(t, ctx, dexGraphQLClient, tenant, app.ID, appUpdateInput)
+		require.NoError(t, err)
+
+		t.Log("Ensure the status condition")
+		assert.Equal(t, graphql.ApplicationStatusConditionFailed, appUpdated.Status.Condition)
+	})
+
+	t.Run("Test status update as Integration System", func(t *testing.T) {
```
dbadura, line 60: I wonder if tests for IS and users are needed?
Because we test the flow in unit tests.

tests/director/gateway-integration/automatic_status_update_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+		t.Log("Issue a Hydra token with Client Credentials")
+		accessToken := getAccessToken(t, appOauthCredentialData, "application:read")
+		oauthGraphQLClient := gql.NewAuthorizedGraphQLClientWithCustomURL(accessToken, fmt.Sprintf("https://compass-gateway-auth-oauth.%s/director/graphql", domain))
+
+		t.Log("Requesting Viewer as Application")
+		viewer := graphql.Viewer{}
+		req := fixGetViewerRequest()
+
+		assert.Equal(t, graphql.ApplicationStatusConditionInitial, app.Status.Condition)
```
dbadura, line 129: Why do you check here the status?
I think this assert is in the middle of nowhere.

The same situation is for runtime test.

components/director/internal/statusupdate/handler_test.go
```diff
@@ -0,0 +1,226 @@
...
+			ExpectedStatus:   http.StatusOK,
+			ExpectedResponse: "OK",
+			MockHandler:      testHandler(t, testID, consumer.Application),
+		},
+		{
+			Name: "Success when runtime",
+			TxFn: txGen.ThatSucceeds,
+			RepoFn: func() *automock.StatusUpdateRepository {
+				repo := automock.StatusUpdateRepository{}
+				repo.On("IsConnected", mock.Anything, testID, runtimesTable).Return(false, nil)
```
dbadura, line 86: We can use `txtest.CtxWithDBMatcher`

components/director/internal/statusupdate/handler.go
```diff
@@ -0,0 +1,120 @@
...
+type Table string
+
+//go:generate mockery -name=StatusUpdateRepository -output=automock -outpkg=automock -case=underscore
+type StatusUpdateRepository interface {
+	UpdateStatus(ctx context.Context, id, table string) error
+	IsConnected(ctx context.Context, id, table string) (bool, error)
+}
+
+const (
+	applicationsTable Table = "applications"
```
aszecowka, line 32: handler should not be aware of table names, it will be better to have separate methods for app and runtimes (IsApplicationConnected) or introduce some enum.

tests/director/gateway-integration/automatic_status_update_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+		t.Logf("Registered Integration System with [id=%s]", intSys.ID)
+		defer unregisterIntegrationSystem(t, ctx, dexGraphQLClient, testConfig.DefaultTenant, intSys.ID)
+
+		t.Log("Request Client Credentials for Integration System")
+		intSysOauthCredentialData := requestClientCredentialsForIntegrationSystem(t, ctx, dexGraphQLClient, testConfig.DefaultTenant, intSys.ID)
+
+		t.Log("Issue a Hydra token with Client Credentials")
+		accessToken := getAccessToken(t, intSysOauthCredentialData, "application:write")
+		oauthGraphQLClient := gql.NewAuthorizedGraphQLClientWithCustomURL(accessToken, fmt.Sprintf("https://compass-gateway-auth-oauth.%s/director/graphql", testConfig.Domain))
```
dbadura, line 68: You can create this URL before all tests.

PR #1134, opened 2020-04-03
Add runtime info handler in KEB

mszostok:
**Description**

- Add runtime info handler in KEB


components/kyma-environment-broker/internal/appinfo/dto.go
```diff
@@ -0,0 +1,30 @@
...
+		SubAccountID      string    `json:"subaccountId"`
+		ServiceInstanceID string    `json:"serviceInstanceId"`
+		ServiceClassID    string    `json:"serviceClassId"`
+		ServiceClassName  string    `json:"serviceClassName"`
+		ServicePlanID     string    `json:"servicePlanId"`
+		ServicePlanName   string    `json:"servicePlanName"`
+		Status            DTOStatus `json:"status"`
+	}
+
+	DTOStatus struct {
```
piotrmiskiewicz, line 18: please, use one convention, DTO as suffix or prefix, see: RuntimeDTO and DTOStatus

components/kyma-environment-broker/internal/storage/dbsession/read.go
```diff
@@ -1,18 +1,42 @@
...
+
+	"github.com/gocraft/dbr"
 	"github.com/pivotal-cf/brokerapi/v7/domain"
 )
 
 type readSession struct {
 	session *dbr.Session
 }
 
+func (r readSession) FindAllInstancesJoinedWithProvisionOperation(prct ...predicate.Predicate) ([]internal.InstanceWithOperation, dberr.Error) {
```
piotrmiskiewicz, line 20: it seems it will also join with deprovisioning operation, is that ok?

components/kyma-environment-broker/internal/appinfo/dto.go
mszostok: agree, fixed 👍

components/kyma-environment-broker/internal/storage/dbsession/read.go
mszostok: yes, check the testdata result there is also a test for that :)

PR #1141, opened 2020-04-06
Log error message in all retries

nachtmaar:
**Description**

- Log error message in `RetryOperation` with every retry
- `RetryOperation` only logged the error message then the last retry failed. If the whole operation (inclusive retries) takes long time (e.g. 1 hour), then a developer should not have to wait so long to see the error. Especially, if it is non recoverable error.

**Related issue(s)**

https://github.com/kyma-incubator/compass/issues/1013#issuecomment-601706654


nachtmaar:
ToDO:

- [ ] do the same change in provision_operation.go

PR #1147, opened 2020-04-07
Ensure every runtime has "name" label

sjanota:
**Description**

- [x] Add the "name" label to every newly created runtime
- [x] Add migration to add this label to already existing runtimes 
- [x] Adjust integration tests

**Related issue(s)**


components/schema-migrator/migrations/director/202007041321_add_name_labels_to_runtimes.up.sql
```diff
@@ -0,0 +1,10 @@
+
+INSERT INTO labels
+    (id ,tenant_id, runtime_id, "key", "value")
+SELECT uuid_in(md5(random()::text || clock_timestamp()::text)::cstring), unnamed.tenant_id, unnamed.runtime_id, 'name', to_json(unnamed.runtime_name)
```
dbadura, line 4: You can use `uuid` extension which should be in PostgreSQL or use pg-crypto
https://www.postgresqltutorial.com/postgresql-uuid/

sjanota: I had to bump psql image as the old one don't have the `uuid-ossp` extension

sjanota:
Generic mechanism for fitlering json labels can be tricky. Closing this one in favor. of simpler option: https://github.com/kyma-incubator/compass/pull/1158

PR #1151, opened 2020-04-07
Implement deprovisioning for Avs

abbi-gaurav:


components/kyma-environment-broker/internal/process/deprovisioning/avs_evaluations.go
```diff
@@ -0,0 +1,51 @@
...
+func (ars *AvsEvaluationRemovalStep) Run(deProvisioningOperation internal.DeprovisioningOperation, logger logrus.FieldLogger) (internal.DeprovisioningOperation, time.Duration, error) {
+	if deProvisioningOperation.AVSExternalEvaluationDeleted && deProvisioningOperation.AVSInternalEvaluationDeleted {
+		logger.Infof("Both internal and external evaluations have been deleted")
+		return deProvisioningOperation, 0, nil
+	}
+
+	provisioningOperation, err := ars.operationsStorage.GetProvisioningOperationByInstanceID(deProvisioningOperation.InstanceID)
+	if err != nil {
+		logger.Errorf("error while getting provisioning deProvisioningOperation from storage")
+		return deProvisioningOperation, time.Second * 10, nil
```
polskikiel, line 41: We don't want to fail or timeout the de-provisioning operation if some step is constantly failing. So our idea is to implement a function in the `internal/process/deprovision_operation.go`:
```
// RetryOperationWithoutFail retries an operation for at maxTime in retryInterval steps and omits the operation if retrying failed
func (om *DeprovisionOperationManager) RetryOperationWithoutFail(operation internal.DeprovisioningOperation, description string, retryInterval time.Duration, maxTime time.Duration, log logrus.FieldLogger) (internal.DeprovisioningOperation, time.Duration, error) {
	since := time.Since(operation.UpdatedAt)

	log.Infof("Retrying for %s in %s steps", maxTime.String(), retryInterval.String())
	if since < maxTime {
		return operation, retryInterval, nil
	}
	// update description to track failed steps
	updatedOperation, repeat := om.update(operation, domain.InProgress, description)
	if repeat != 0 {
		return updatedOperation, repeat, nil
	}

	log.Errorf("Omitting after %s of failing retries", maxTime.String())
	return updatedOperation, 0, nil
}
```
Using this function, you can execute a given number of retries for the error paths. In case your step exceeds the timeout, the deprovisioning operation won't fail and the processing queue will move to the next step.

components/kyma-environment-broker/internal/model.go
```diff
@@ -78,9 +78,12 @@ type ProvisioningOperation struct {
 	AVSEvaluationExternalId int64 `json:"avs_evaluation_external_id"`
 }
 
-// DeprovisioningOperation holds all information about provisioning operation
+// DeprovisioningOperation holds all information about de-provisioning operation
 type DeprovisioningOperation struct {
 	Operation `json:"-"`
+
+	AVSInternalEvaluationDeleted bool `json:"avs_internal_evaluation_deleted"`
```
polskikiel, line 85: I'd create separate structs for AVSFlags and AVSIDs under one struct AVS which I'd use here. reason: look comment below

components/kyma-environment-broker/internal/avs/delegator.go
```diff
@@ -112,3 +116,66 @@ func responseBody(resp *http.Response) string {
...
+	}
+
+	httpClient, err := del.clientHolder.getClient(logger)
+	if err != nil {
+		return err
+	}
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return err
```
polskikiel, line 161: you can add context to the errors above using `errors.Wrap()`

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -218,6 +220,10 @@ func main() {
 			weight: 10,
 			step:   deprovisioning.NewRemoveRuntimeStep(db.Operations(), db.Instances(), provisionerClient),
 		},
+		{
+			weight: 1,
```
polskikiel, line 224: I'd prefer to have these steps sorted ascending by weight

components/kyma-environment-broker/internal/process/deprovisioning/avs_evaluations.go
abbi-gaurav: @polskikiel , it would be nice if all these are conventions are the part of the documentation.
The implementation so far was based on the documentation https://github.com/kyma-incubator/compass/blob/master/docs/kyma-environment-broker/03-03-runtime-provisioning-and-deprovisioning.md.

This could save time and reduce unnecessary wasted effort.

abbi-gaurav:
@polskikiel , please have a look. I have applied the changes

components/kyma-environment-broker/internal/process/deprovisioning/avs_evaluations.go
polskikiel: @abbi-gaurav You're right, I will update the documentation and add such a function in separate PR. Then you can just rebase and use it in your step.

polskikiel: https://github.com/kyma-incubator/compass/pull/1164

PR #1152, opened 2020-04-08
Remove assigned scenario

dbadura:
**Description**

- add remove scenario in engine

**Related issue(s)**

### DIFF
--- a/components/director/internal/domain/scenarioassignment/engine.go
+++ b/components/director/internal/domain/scenarioassignment/engine.go
@@ -38,20 +38,24 @@ func (e *engine) EnsureScenarioAssigned(ctx context.Context, in model.AutomaticS
 	return e.upsertMergedScenarios(ctx, labels, in.ScenarioName, e.uniqueScenarios)
 }
 
-func (e *engine) uniqueScenarios(scenarios []interface{}, newScenario string) ([]string, error) {
-	set := make(map[string]struct{})
-	set[newScenario] = struct{}{}
-
-	for _, scenario := range scenarios {
-		output, ok := scenario.(string)
-		if !ok {
-			return nil, errors.New("scenario is not a string")
-		}
-		set[output] = struct{}{}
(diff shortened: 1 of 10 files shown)


components/director/internal/domain/label/repository.go
```diff
@@ -173,6 +173,31 @@ func (r *repository) DeleteByKey(ctx context.Context, tenant string, key string)
 
 	return nil
 }
+func (r *repository) GetRuntimeScenariosWhereRuntimesLabelsMatchSelector(ctx context.Context, tenantID, selectorKey, selectorValue string) ([]model.Label, error) {
```
pkosiec, line 176: What's the difference between this method and `GetRuntimeScenariosWhereLabelsMatchSelector`?

components/director/internal/domain/scenarioassignment/engine.go
```diff
@@ -66,16 +70,32 @@ func (e *engine) upsertMergedScenarios(ctx context.Context, labels []model.Label
...
+		}
+		set[output] = struct{}{}
+	}
+
+	return str.MapToSlice(set), nil
 }
 
-func (engine) RemoveAssignedScenarios(in []*model.AutomaticScenarioAssignment) error {
-	// TODO: Implement it
+func (e *engine) scenariosDiff(scenarios []interface{}, diffScenario string) ([]string, error) {
```
pkosiec, line 88: this is basically appending new Scenario, right?
IF so:

```suggestion
func (e *engine) appendScenario(scenarios []interface{}, scenario string) ([]string, error) {
```

components/director/internal/domain/label/repository.go
dbadura: It's leftover. I will remove it.

components/director/internal/domain/scenarioassignment/engine.go
dbadura: Nope, It will remove the scenario.
I will rename it to `removeScenario`

pkosiec: okay

PR #1158, opened 2020-04-09
Add searchPhrase to runtimes query

sjanota:
**Description**

- Add additional `searchPhrase` to `runtimes` query to allow searching by name.

**Related issue(s)**


sjanota:
We're dropping the support for search on runtimes view for now.

PR #1153, opened 2020-04-08
Runtime upgrade in provisioner

Szymongib:
**Description**

Changes proposed in this pull request:

- Add mutation to upgrade Kyma Runtimes

**Related issue(s)**
#647

Issue #647, opened 2020-01-14
Runtime upgrades with Compass Provisioner


janmedrek:
**Description**

The goal of this task is to have the first (simplest) version of mutation that allows the user to upgrade his Kyma Runtime to the newer version.

The current assumptions are:
- no additional migration actions are needed (they will be handled manually)
- upgrade increases Kyma version by one release (no automated subsequent upgrades in cases such as 1.5 -> 1.9 upgrade)

**Acceptance Criteria**
- first version of API designed and implemented
- short tutorial with commands and exemplary calls created 
- prepare and implement a pipeline for testing runtime upgrades

**Subtasks**
- [x] Do research and prepare a plan for implementation, discuss with the team - @Szymongib 
- [x] Upgrade in Installation SDK - @Maladie 
- [x] Prepare API, change Provisioner
- [x] Create a tutorial / guide / docs - @majakurcius 
- [x] Implement tests 

Follow up issue for pipelines integration: https://github.com/kyma-project/kyma/issues/8080

PR #1175, opened 2020-04-14
Document Runtime upgrade and rollback with Runtime Provisioner for devs

majakurcius:
**Description**

Add developer documentation on: 
- the `upgradeRuntime` mutation performing Runtime upgrade by one release. 
- the `rollBackUpgradeOperation` mutation performing upgrade rollback.

**Related issue(s)**
#647 

**Related PR(s)**
#1153

### DIFF
--- a/docs/internal/runtime-upgrade.md
+++ b/docs/internal/runtime-upgrade.md
@@ -1,0 +1,208 @@
+# Runtime upgrade with the Runtime Provisioner
+
+## Upgrade Runtime
+
+Runtime Provisioner allows to upgrade Kyma Runtime using the GraphQL API. To upgrade a Runtime to version 1.11.0 with the standard Runtime components use the following mutation:
+
+```graphql
+mutation {
+  upgradeRuntime(id: "{RUNTIME_ID}", input: {
+    kymaConfig: { 
+        version: "1.11.0",
+        configuration: [
+            {
+                key: "a.config.key"
+                value: "a.config.value"
+            }
+        ]
+        components: [
+          {
+            component: "cluster-essentials"
+            namespace: "kyma-system"
+          }
+          {
+            component: "testing"
+            namespace: "kyma-system"
+          }
+          {
+            component: "istio"
+            namespace: "istio-system"
+          }
+          {
+            component: "xip-patch"
+            namespace: "kyma-installer"
+          }
+          {
+            component: "istio-kyma-patch"
+            namespace: "istio-system"
+          }
+          {
+            component: "knative-serving-init"
+            namespace: "knative-serving"
+          }
+          {
+            component: "knative-serving"
+            namespace: "knative-serving"
+          }
+          {
+            component: "knative-eventing"
+            namespace: "knative-eventing"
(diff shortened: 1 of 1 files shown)


docs/internal/runtime-upgrade.md
```diff
@@ -0,0 +1,208 @@
+# Runtime upgrade with the Runtime Provisioner
+
+## Upgrade Runtime
```
klaudiagrz, line 3: Do we need this heading? I'd remove it and go on with the content.

docs/internal/runtime-upgrade.md
```diff
@@ -0,0 +1,208 @@
...
+
+Runtime Provisioner allows to upgrade Kyma Runtime using the GraphQL API. To upgrade a Runtime to version 1.11.0 with the standard Runtime components use the following mutation:
```
klaudiagrz, line 5:
```suggestion
Runtime Provisioner allows you to upgrade Kyma Runtime using the GraphQL API. To upgrade a Runtime to the 1.11.0 version with the standard Runtime components, use the following mutation:
```

docs/internal/runtime-upgrade.md
```diff
@@ -0,0 +1,208 @@
...
+}
+```
+
+The Kyma configuration passed to the upgrade mutation overrides the previous configuration. The components and parts of the configuration not specified in the new version are removed.
+
+Similarly to provisioning, the upgrade operation is asynchronous and returns the [Runtime Operation Status](../provisioner/08-03-runtime-operation-status.md).
+
+## Roll back the latest upgrade
+
+The Runtime Provisioner API allows to roll back the Kyma configuration to the one from before the latest upgrade. This mutation does not affect the cluster itself in any way, it only affects the database. In the case a cluster is rolled back manually, for example restored from a backup, rollback on the database must be performed to keep the data in sync. To roll back an upgrade, the upgrade must be the last operation performed by the Runtime Provisioner on the given Runtime. In other words, you cannot perform a rollback if other operations have been performed on the Runtime after the upgrade.
```
klaudiagrz, line 136:
```suggestion
The Runtime Provisioner API allows you to roll back the Kyma configuration to the one from before the latest upgrade. This mutation does not affect the cluster itself in any way, it only affects the database. In the case a cluster is rolled back manually, for example restored from a backup, a rollback on the database must be performed to keep the data in sync. To roll back an upgrade, the upgrade must be the last operation performed by the Runtime Provisioner on the given Runtime. In other words, you cannot perform a rollback if other operations have been performed on the Runtime after the upgrade. 
```

docs/internal/runtime-upgrade.md
majakurcius: I think we do, cos there's another section on Rollback later.

docs/internal/runtime-upgrade.md
majakurcius: I'll add the comma, but I'll stick to the original version description as the one in the suggestion AFAIK is not really how it's written in the IT context.

Issue #1103, opened 2020-03-31
Insecure SQL queries in Director


pkosiec:
**Description**

Currently DB queries in Director are insecured, as they are built with `fmt.Sprintf` and string builder.
We need to convert them all to parametrized queries.

Most of them are located in `repo` helpers, but we also have custom queries like in applications for runtime query or `apiruntimeauth` domain.

Our findings:
- Create in `repo` is safe
- List has `additionalConditions` parameter (which is string)
- Conditions mechanism in `repo` helpers is unsafe.

pkosiec:
I went through all repositories in Director. These are my findings.


## No issues

- all custom DB queries in repositories - user input is escaped with pq.QuoteLiteral.
- Filtering in list:
`components/director/internal/domain/label/query.go` - user input is escaped with pq.QuoteLiteral ("github.com/lib/pq")
- repo: Creator, Upserter, Updater use jmoiron/sqlx package which use parametrized queries
- repo: ExistQuerier, Deleter, Getter use parametrized queries


## To improve:
repo/common.go - GetQueryPart(idx int)  -> 
quote values for IN operation - DONE in PR: https://github.com/kyma-incubator/compass/pull/1187

repo/list.go + repo/list_pageable - additionalConditions is string, where is no quoting by default; this is the problematic function: `buildSelectStatement` from `common.go`

kfurgol:
I've come to the same conclusion as @pkosiec, nothing to add from my side.

PR #1187, opened 2020-04-16
Fix vulnerabilities in repository helpers

pkosiec:
**Description**

- Fix vulnerabilities in repository helpers

**Related issue(s)**
#1103


components/director/internal/repo/conditions.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/lib/pq"
+)
+
+type ConditionOp string
+
+const (
```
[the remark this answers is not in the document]
pkosiec: I just moved the code, but you're right - I've refactored it to follow the OCP principle.

components/director/internal/repo/query_builder.go
```diff
@@ -0,0 +1,111 @@
...
+	singleKeyExistsOp, singleKeyExistsOpPlaceholder,
+}
+
+var reverseTempReplace = []string{
+	anyKeyExistsOpPlaceholder, anyKeyExistsOp,
+	allKeysExistOpPlaceholder, allKeysExistOp,
+	singleKeyExistsOpPlaceholder, singleKeyExistsOp,
+}
+
+// sqlx doesn't detect ?| and ?& operators properly
```
pkosiec, line 83: I needed to do a temporary replace, because sqlx.Rebind doesn't detect operators.

So we could end up with this:
```
SELECT id, tenant_id, name, description, status_condition, status_timestamp, creation_timestamp FROM public.runtimes WHERE tenant_id = $1 AND id IN (SELECT "runtime_id" FROM public.labels WHERE "runtime_id" IS NOT NULL AND "tenant_id" = $2 AND "key" = $3 AND "value" **$4|** array[$5]) ORDER BY creation_timestamp ASC
```

The only solution for that would be switching to `db.NamedExec` and use named parameters (switch from `[]interface{}` in helpers to `map[string]interface{}`, but it would be too time consuming to do it - to generate/ensure param keys are unique etc. We can think about it in future.

components/director/internal/repo/conditions.go
```diff
@@ -0,0 +1,114 @@
+package repo
+
+import (
+	"fmt"
+	"strings"
+)
+
+type Condition interface {
+	// GetQueryPart returns formatted string that will be included in the SQL query for a given condition
+	GetQueryPart() string
```
dbadura, line 10: I think about returns here also the `args` as `interfacae{}` and have only one method `GetCondition`

pkosiec: Hmm, but anyway `getAllArgs` and `writeEnumeratedConditions` use only query or arguments. I would need to refactor these functions as well, so maybe I will stop here with further refactoring :)

PR #1190, opened 2020-04-17
Configures monitoring with OIDC authentication

jasiu001:
**Description**

- functionality to register ServiceProvider on IAS platform
- adding a label with GrafanaURL to runtime
- deregistration ServiceProvider from IAS platform


chart/compass/charts/kyma-environment-broker/templates/secrets.yaml
```diff
@@ -29,10 +29,21 @@ data:
...
 type: Opaque
 data:
   token: {{ .Values.lms.token | b64enc | quote }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "{{ .Values.ias.secretName }}"
+  labels:
+  {{ include "kyma-env-broker.labels" . | indent 4 }}
```
ksputo, line 44: Remove indent here:
```suggestion
{{ include "kyma-env-broker.labels" . | indent 4 }}
```

components/kyma-environment-broker/internal/ias/assertion_attribute.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return attributes
+}
+
+// GenerateAssertionAttribute remove not existing in template attributes, leaves existing
+func (a *AssertionAttributeDeliver) GenerateAssertionAttribute(serviceProvider ServiceProvider) []AssertionAttribute {
+	var newAssertionAttributes []AssertionAttribute
+	aats := a.assertionAttributesKeys()
+
+	for key, _ := range aats {
```
mszostok, line 48: `for in for` always scares me. Also, this map with bools was problematic to understand at the beginning
WDYT about this? IMO is shorter and more readable

```go
	defaults := a.tmplDeepCopy()
	// overrides defaults with given attr
	for _, overrideAtr := range serviceProvider.AssertionAttributes {
		if _, found := defaults[overrideAtr.AssertionAttribute]; found {
			defaults[overrideAtr.AssertionAttribute] = overrideAtr
		}
	}

	// convert to slice
	var newAssertionAttributes []AssertionAttribute
	for _, atr := range defaults {
		newAssertionAttributes = append(newAssertionAttributes, atr)
	}
```

just replace `assertionAttributesKeys` function with:
```go
func (a AssertionAttributeDeliver) tmplDeepCopy() map[string]AssertionAttribute {
	cpy := map[string]AssertionAttribute{}
	for k, v := range a.assertionAttributesTemplate {
		cpy[k] = v
	}
	return cpy
}
```

components/kyma-environment-broker/internal/ias/assertion_attribute_test.go
```diff
@@ -0,0 +1,46 @@
...
+			},
+			{
+				AssertionAttribute: "should_be_removed",
+				UserAttribute:      "shouldBeRemoved",
+			},
+		},
+	})
+
+	// then
+	assert.ElementsMatch(t, []AssertionAttribute{
```
mszostok, line 28: `assert.ElementsMatch` nice 👍 I didn't know that

components/kyma-environment-broker/internal/ias/client.go
```diff
@@ -0,0 +1,231 @@
...
+)
+
+func NewClient(cli *http.Client, cfg ClientConfig) *Client {
+	return &Client{
+		config:     cfg,
+		httpClient: cli,
+	}
+}
+
+func (c Client) CloseBodyError() error {
```
mszostok, line 53: it is not used anywhere. IMO we should get rid of it as it's quite not convenient.  The better option is to just use the append error, see my example in `do` method

components/kyma-environment-broker/internal/ias/client.go
```diff
@@ -0,0 +1,231 @@
...
+
+	return c.call(pathAccess, payload)
+}
+
+func (c *Client) GetCompany() (*Company, error) {
+	company := &Company{}
+	request := &Request{Method: http.MethodGet, Path: PathCompanyGlobal}
+
+	response, err := c.do(request)
+	defer c.closeResponseBody(response)
```
mszostok, line 80: imo this needs to be replaced with:
```go
	defer func() {
		err = multierror.Append(err, errors.Wrap(c.closeResponseBody(response), "while trying to close body reader")).ErrorOrNil()
	}()
```

and `closeResponseBody` should  look like that:
```go
func (c *Client) closeResponseBody(response *http.Response) error {
	if response.Body == nil {
		return nil
	}
	return response.Body.Close()
}
```

as a result client does not have to play with calling another method like `CloseBodyError()`.,
```go
c, err := GetCompany()
if err != nil {
}
if err := c.CloseBodyError(); err != nil {
}
```
additionally, you also didnt use that method at all. What's more you didnt have any cleanup for CloseBodyError, so if such error occurred  once then it was always there which is quite misleading if you are executing e.g. `GetCompany` few times.

components/kyma-environment-broker/internal/ias/builder.go
```diff
@@ -0,0 +1,46 @@
...
+
+func NewBundleBuilder(httpClient *http.Client, config Config) BundleBuilder {
+	return &Builder{
+		httpClient: httpClient,
+		config:     config,
+	}
+}
+
+func (b *Builder) NewBundle(identifier string) Bundle {
+	client := NewClient(b.httpClient, ClientConfig{
```
mszostok, line 39: why we always creating a new client? IMO such thing should be thread save

components/kyma-environment-broker/internal/ias/client_test.go
```diff
@@ -0,0 +1,313 @@
...
+						Action:    "Allow",
+						Group:     "admins",
+						GroupType: "cloud",
+					},
+				},
+				DefaultAction: "Allow",
+			},
+		},
+	}
+	err := client.SetAuthenticationAndAccess(serviceProviderID, auth)
```
mszostok, line 136: with `SetFoo` method such test does not make sense because we do not assert if sth was really changed. Could you rewrite those test so it will not pass if someone will really change the contract?

components/kyma-environment-broker/internal/process/provisioning/ias_type.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return 0, nil
+}
+
+func (s *IASType) handleError(operation internal.ProvisioningOperation, err error, log logrus.FieldLogger, msg string) (time.Duration, error) {
+	log.Errorf("%s: %s", msg, err)
+	switch {
+	case kebError.IsTemporaryError(err):
+		if time.Since(operation.UpdatedAt) > setIASTypeTimeout {
```
mszostok, line 57: can we just use `return s.operationManager.RetryOperation(` here?

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -218,16 +228,24 @@ func main() {
...
+			disabled: cfg.IAS.Disabled,
+		},
 		{
 			weight: 10,
 			step:   deprovisioning.NewRemoveRuntimeStep(db.Operations(), db.Instances(), provisionerClient),
 		},
 	}
 	for _, step := range deprovisioningSteps {
-		deprovisionManager.AddStep(step.weight, step.step)
+		if !step.disabled {
```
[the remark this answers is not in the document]
jasiu001: when all steps will be done (and their status will be clear) then we can think about creating one env for all steps or some config map

components/kyma-environment-broker/internal/process/provisioning/initialisation.go
```diff
@@ -21,12 +22,16 @@ import (
 const (
 	// the time after which the operation is marked as expired
 	CheckStatusTimeout = 3 * time.Hour
+
+	// label key used to send to director
+	grafanaURLLabel = "operator_grafanaUrl"
```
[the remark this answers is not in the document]
jasiu001: make sense, I just copied it from the task, changed to snake_case

components/kyma-environment-broker/internal/director/client.go
```diff
@@ -101,18 +125,41 @@ func (dc *Client) callDirector(req *machineGraph.Request) (*successResponse, err
...
+	return &response, nil
 }
 
-func (dc *Client) call(req *machineGraph.Request) (*successResponse, error) {
-	var response successResponse
-	err := dc.graphQLClient.Run(context.Background(), req, &response)
-	if err != nil {
-		return nil, kebError.AsTemporaryError(err, "while requesting to director client")
+func (dc *Client) setLabelsInDirector(req *machineGraph.Request) (*runtimeLabelResponse, error) {
+	var response runtimeLabelResponse
```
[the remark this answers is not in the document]
jasiu001: It requires deeper refactoring, after the discussion, I will create a separate task for that

components/kyma-environment-broker/internal/director/client_test.go
```diff
@@ -304,10 +304,57 @@ func TestClient_GetConsoleURL(t *testing.T) {
...
+
+	client := NewDirectorClient(oc, qc)
+	client.token = oauth.Token{
+		AccessToken: "1234xyza",
+		Expiration:  time.Now().Unix() + 999,
+	}
+
+	request := createGraphQLLabelRequest(client, accountID, runtimeID, labelKey, labelValue)
+
+	qc.On("Run", context.Background(), request, mock.AnythingOfType("*director.runtimeLabelResponse")).Run(func(args mock.Arguments) {
```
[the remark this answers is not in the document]
jasiu001: right 👍

components/kyma-environment-broker/internal/ias/assertion_attribute.go
jasiu001: done.

components/kyma-environment-broker/internal/ias/client.go
jasiu001: removed

components/kyma-environment-broker/internal/process/provisioning/ias_type.go
jasiu001: it is because `RetryOperation` after reached timeout marks operation as failed, I decided to not mark the whole operation as failed just because this last sub-step failed (SKR works, dasbordURL is set) and only left an error log about failed this part of provisioning.

docs/kyma-environment-broker/03-03-runtime-provisioning-and-deprovisioning.md
```diff
@@ -16,6 +16,7 @@ The provisioning process contains the following steps:
 | Resolve_Target_Secret                  | Hyperscaler Account Pool | Provides the name of a Gardener Secret that contains  Hypescaler account credentials used during cluster provisioning.                                | @koala7659       |
 | AVS_Configuration_Step                 | AvS                      | Sets up external and internal monitoring of Kyma Runtime.                                      | @abbi-guarav     |
 | Create_LMS_Tenant                      | LMS                      | Requests a tenant in the LMS system or provides a tenant ID if it was created before.                                                              | @piotrmiskiewicz |
+| IAS_Registration                       | Identity Authentication Service | Registers new ServiceProvider on IAS, generates client ID and secret and inserts to grafana overrides. Step is not required and can be disabled. | @jasiu001 |
```
klaudiagrz, line 19:
```suggestion
| IAS_Registration                       | Identity Authentication Service | Registers a new ServiceProvider on IAS, generates client ID and Secret, and inserts them to Grafana overrides. This step is not required and can be disabled. | @jasiu001 |
```

docs/kyma-environment-broker/03-03-runtime-provisioning-and-deprovisioning.md
```diff
@@ -31,9 +32,10 @@ Any deprovisioning step shouldn't block the entire deprovisioning operation. Use
 The deprovisioning process contains the following steps:
 | Name                         | Domain         | Status      | Description                                                                            | Owner     |
 |------------------------------|----------------|-------------|----------------------------------------------------------------------------------------|-----------|
-| Deprovision_Initialization   | Deprovisioning | Done        | Initialize the `DeprovisioningOperation` instance with data fetched from the `ProvisioningOperation`. | @jasiu001 |
+| Deprovision_Initialization   | Deprovisioning | Done        | Initialize the `DeprovisioningOperation` instance with data fetched from the `ProvisioningOperation`. | @polskikiel |
```
klaudiagrz, line 35:
```suggestion
| Deprovision_Initialization   | Deprovisioning | Done        | Initializes the `DeprovisioningOperation` instance with data fetched from the `ProvisioningOperation`. | @polskikiel |
```

docs/kyma-environment-broker/03-03-runtime-provisioning-and-deprovisioning.md
```diff
@@ -31,9 +32,10 @@ Any deprovisioning step shouldn't block the entire deprovisioning operation. Use
...
 | Deprovision Azure Event Hubs | Event Hub      | In progress | Deletes the Azure Event Hub Namespace.                                                  | @montaro  |
-| Remove_Runtime               | Deprovisioning | Done        | Triggers deprovisioning of a Runtime in the Runtime Provisioner.                                            | @jasiu001 |
+| IAS_Deregistration           | Identity Authentication Service | Done | Removes ServiceProvider from IAS | @jasiu001 |
```
klaudiagrz, line 37:
```suggestion
| IAS_Deregistration           | Identity Authentication Service | Done | Removes the ServiceProvider from IAS. | @jasiu001 |
```

components/kyma-environment-broker/internal/process/provisioning/ias_type.go
mszostok: if this is the way how it should work the :+1:

components/kyma-environment-broker/internal/process/deprovisioning/ias_deregistration_test.go
```diff
@@ -0,0 +1,43 @@
...
+	operation := internal.DeprovisioningOperation{
+		Operation: internal.Operation{
+			InstanceID: iasInstanceID,
+		},
+	}
+
+	step := NewIASDeregistrationStep(memoryStorage.Operations(), bundleBuilder)
+
+	// when
+	_, repeat, err := step.Run(operation, logrus.New())
```
mszostok, line 38: Please use `logger.NewLogDummy()`. Because right now it produces logs which only mess the test output. If you want to assert that given msg was logged then use `logger.NewLogSpy()`.  The same suggestion for all others occurrences

components/kyma-environment-broker/internal/ias/client.go
```diff
@@ -0,0 +1,235 @@
...
+	return c.call(c.serviceProviderPath(spID), payload)
+}
+
+func (c *Client) SetAuthenticationAndAccess(spID string, payload AuthenticationAndAccess) error {
+	pathAccess := fmt.Sprintf(PathAccess, spID)
+
+	return c.call(pathAccess, payload)
+}
+
+func (c *Client) GetCompany() (*Company, error) {
```
mszostok, line 72:
```suggestion
func (c *Client) GetCompany() (_ *Company,  err error) {
```

the one important thing is to declare error in signature so the defer statement will work properly. 

Same for other places

PR #1197, opened 2020-04-20
Use unique names for Event Hubs namespace and resource group

Abd4llA:
**Description**

Azure Event Hubs namespace names are globally unique. So we need to make sure that we're using
a unique namespace name each time we're provisioning a Kyma runtime.

- Use a unique name for Azure Event Hubs namespace and resource group

### DIFF
--- a/components/kyma-environment-broker/internal/hyperscaler/azure/client.go
+++ b/components/kyma-environment-broker/internal/hyperscaler/azure/client.go
@@ -51,16 +51,17 @@ func (nc *AzureClient) CreateNamespace(ctx context.Context, azureCfg *Config, gr
 }
 
 func (nc *AzureClient) CheckNamespaceAvailability(ctx context.Context, name string) (bool, error) {
-	//TODO Validate name string
+	if name == "" {
+		return false, errors.New("Name cannot be empty")
+	}
 	res, err := nc.eventhubNamespaceClient.CheckNameAvailability(ctx, eventhub.CheckNameAvailabilityParameter{Name: &name})
 	if err != nil {
 		return false, errors.Wrapf(err, "Failed to check Event Hubs namespace availability with name: %s", name)
 	}
 	if res.NameAvailable == nil {
 		return false, errors.Errorf("Failed to check Event Hubs namespace availability with name: %s. Received no response using Azure client.", name)
-	} else {
-		return *res.NameAvailable, nil
 	}
+	return *res.NameAvailable, nil
 }
 
 func (nc *AzureClient) createAndWait(ctx context.Context, resourceGroupName string, namespaceName string, parameters eventhub.EHNamespace) (result eventhub.EHNamespace, err error) {
--- a/components/kyma-environment-broker/internal/hyperscaler/azure/client.go
+++ b/components/kyma-environment-broker/internal/hyperscaler/azure/client.go
@@ -74,19 +74,19 @@ func (nc *AzureClient) CheckResourceGroupAvailability(ctx context.Context, name 
 	}
 	group, err := nc.resourcegroupClient.CheckExistence(ctx, name)
 	if err != nil {
-		return false, errors.Wrapf(err, "Failed to check Azure resource group availability with name: %s. Error " +
+		return false, errors.Wrapf(err, "Failed to check Azure resource group availability with name: %s. Error "+
 			"while using Azure client.", name)
 	}
-	switch group.StatusCode{
+	switch group.StatusCode {
 	case 404:
 		available = true
 	case 204:
 		available = false
 	default:
-		return false, errors.Errorf("Failed to check Azure resource group availability with name: %s." +
+		return false, errors.Errorf("Failed to check Azure resource group availability with name: %s."+
 			" Unexpected API response code %d", name, group.StatusCode)
 	}
-	return available , nil
+	return available, nil
 }
 
 func (nc *AzureClient) NamespaceExists(ctx context.Context, resourceGroupName string, namespaceName string, tags Tags) (bool, error) {
@@ -102,28 +102,28 @@ func (nc *AzureClient) NamespaceExists(ctx context.Context, resourceGroupName st
 	if err != nil {
 		return false, errors.Wrapf(err, "Failed to check Azure Event Hubs namespace availability with name: %s.", namespaceName)
 	}
-	if ! exists {
+	if !exists {
 		return false, nil
 	}
 
 	res, err := nc.eventhubNamespaceClient.ListByResourceGroup(ctx, resourceGroupName)
 	if err != nil {
-		return false, errors.Errorf("Failed to check Event Hubs namespace availability with name: %s in resource" +
+		return false, errors.Errorf("Failed to check Event Hubs namespace availability with name: %s in resource"+
 			"group %s. Error while using Azure client. %v", namespaceName, resourceGroupName, err)
 	}
 	if res.Response().StatusCode != 200 && res.Response().StatusCode != 201 {
-		return false, errors.Errorf("Failed to check Event Hubs namespace availability with name: %s in resource" +
+		return false, errors.Errorf("Failed to check Event Hubs namespace availability with name: %s in resource"+
 			"group %s. Unexpected API response code. Expected 2XX but received %d", namespaceName, resourceGroupName,
 			res.Response().StatusCode)
 	}
 	for res.NotDone() {
 		namespaces := res.Values()
-		for _, ns := range namespaces{
+		for _, ns := range namespaces {
 			exists = matchWithTags(namespaceName, tags, *ns.Name, ns.Tags)
 		}
 		err := res.NextWithContext(ctx)
 		if err != nil {
-			return false, errors.Errorf("Failed to check Event Hubs namespace availability with name: %s in resource" +
+			return false, errors.Errorf("Failed to check Event Hubs namespace availability with name: %s in resource"+
 				"group %s. Error while listing namespaces. %v", namespaceName, resourceGroupName, err)
 		}
 	}
@@ -139,18 +139,18 @@ func (nc *AzureClient) ResourceGroupExists(ctx context.Context, name string, tag
 	if err != nil {
 		return false, errors.Wrapf(err, "Failed to check Azure resource group availability with name: %s.", name)
 	}
-	if ! exists {
+	if !exists {
 		return false, nil
 	}
 	// Now we are safe to assume resource group exists, so any errors returned are not related to resource group not
 	// existing
 	group, err := nc.resourcegroupClient.Get(ctx, name)
 	if err != nil {
-		return false, errors.Wrapf(err, "Failed to check Azure resource group availability with name: %s. Error " +
+		return false, errors.Wrapf(err, "Failed to check Azure resource group availability with name: %s. Error "+
 			"while using Azure client.", name)
 	}
 	if group.StatusCode != 200 {
-		return false, errors.Errorf("Failed to check Azure resource group availability with name: %s." +
+		return false, errors.Errorf("Failed to check Azure resource group availability with name: %s."+
 			" Unexpected API response code. Expected 2XX but received %d", name, group.StatusCode)
 	}
 	return matchWithTags(name, tags, *group.Name, group.Tags), nil
--- a/components/kyma-environment-broker/internal/hyperscaler/azure/client.go
+++ b/components/kyma-environment-broker/internal/hyperscaler/azure/client.go
@@ -15,6 +15,7 @@ type AzureInterface interface {
 	CreateNamespace(ctx context.Context, azureCfg *Config, groupName, namespace string, tags Tags) (*eventhub.EHNamespace, error)
 	// CheckNamespaceAvailability check an Event Hubs namespace name availability
 	CheckNamespaceAvailability(ctx context.Context, name string) (bool, error)
+	CheckResourceGroupAvailability(ctx context.Context, name string) (bool, error)
 	NamespaceExists(ctx context.Context, resourceGroupName string, namespaceName string, tags Tags) (bool, error)
 	ResourceGroupExists(ctx context.Context, name string, tags Tags) (bool, error)
 }
@@ -66,13 +67,45 @@ func (nc *AzureClient) CheckNamespaceAvailability(ctx context.Context, name stri
 	return *res.NameAvailable, nil
 }
 
+func (nc *AzureClient) CheckResourceGroupAvailability(ctx context.Context, name string) (bool, error) {
+	var available bool
+	if name == "" {
+		return false, errors.New("Resource group name cannot be empty")
+	}
+	group, err := nc.resourcegroupClient.CheckExistence(ctx, name)
+	if err != nil {
+		return false, errors.Wrapf(err, "Failed to check Azure resource group availability with name: %s. Error " +
+			"while using Azure client.", name)
+	}
+	switch group.StatusCode{
+	case 404:
+		available = true
+	case 204:
+		available = false
+	default:
+		return false, errors.Errorf("Failed to check Azure resource group availability with name: %s." +
+			" Unexpected API response code %d", name, group.StatusCode)
+	}
+	return available , nil
+}
+
 func (nc *AzureClient) NamespaceExists(ctx context.Context, resourceGroupName string, namespaceName string, tags Tags) (bool, error) {
+	exists := false
 	if namespaceName == "" {
 		return false, errors.New("Namespace name cannot be empty")
 	}
 	if resourceGroupName == "" {
 		return false, errors.New("Resource group name cannot be empty")
 	}
+	exists, err := nc.CheckResourceGroupAvailability(ctx, resourceGroupName)
+
+	if err != nil {
+		return false, errors.Wrapf(err, "Failed to check Azure Event Hubs namespace availability with name: %s.", namespaceName)
+	}
+	if ! exists {
+		return false, nil
+	}
+
 	res, err := nc.eventhubNamespaceClient.ListByResourceGroup(ctx, resourceGroupName)
 	if err != nil {
 		return false, errors.Errorf("Failed to check Event Hubs namespace availability with name: %s in resource" +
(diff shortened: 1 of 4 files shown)


components/kyma-environment-broker/internal/hyperscaler/azure/client.go
```diff
@@ -46,6 +50,19 @@ func (nc *AzureClient) CreateNamespace(ctx context.Context, azureCfg *Config, gr
...
 
+func (nc *AzureClient) CheckNamespaceAvailability(ctx context.Context, name string) (bool, error) {
+	//TODO Validate name string
+	res, err := nc.eventhubNamespaceClient.CheckNameAvailability(ctx, eventhub.CheckNameAvailabilityParameter{Name: &name})
+	if err != nil {
+		return false, errors.Wrapf(err, "Failed to check Event Hubs namespace availability with name: %s", name)
+	}
+	if res.NameAvailable == nil {
+		return false, errors.Errorf("Failed to check Event Hubs namespace availability with name: %s. Received no response using Azure client.", name)
+	} else {
```
k15r, line 61: this else is unnecessary
make return *res.NameAvailable, nil the default

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -6,16 +6,18 @@ import (
...
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/hyperscaler/azure"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/process"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/ptr"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/storage"
 	"github.com/kyma-incubator/compass/components/provisioner/pkg/gqlschema"
+
+	"github.com/Azure/azure-sdk-for-go/services/eventhub/mgmt/2017-04-01/eventhub"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
```
k15r, line 20: nit: use `goimports -local github.com/kyma-project/kyma,github.com/kyma-incubator -w <pathtofile>` for some tool assisted ordering of the imports

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -96,29 +97,35 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+	// Generating a unique name for resource group and namespace with operation ID as suffix if available
+	uniqueName, err := p.getUniqueNSName(operation.ID, azureClient, log)
+	if err != nil {
+		return p.operationManager.OperationFailed(operation, fmt.Sprintf("Can't find a unique name for Event Hubs namespace. Error:%v", err))
+	}
+
 	// create Resource Group
-	groupName := pp.Parameters.Name
+	groupName := uniqueName
 	// TODO(nachtmaar): use different resource group name https://github.com/kyma-incubator/compass/issues/967
```
k15r, line 108: is this comment still valid?

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -140,6 +147,41 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+// getUniqueNSName will try to use suffix to generate a unique name for Event Hubs namespace, or generate a random one
+// if it already exists
+func (p *ProvisionAzureEventHubStep) getUniqueNSName(suffix string, azureClient azure.AzureInterface,
+	log logrus.FieldLogger) (string, error) {
+	uniqueName := fmt.Sprintf("skr-%s", suffix)
+	u, err := azureClient.CheckNamespaceAvailability(p.context, uniqueName)
+	if err != nil {
+		log.Errorf("Error while calling Azure Event Hubs client. Can't check namespace name availability. "+
+			"Error: %v", err)
+		return "", errors.Errorf("Error while calling Azure Event Hubs client. Can't check namespace name "+
```
k15r, line 159: nit: build message once, use twice

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -140,6 +147,41 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+	attempts := 10
+	for !u && attempts > 0 {
+		newName := fmt.Sprintf("skr-%s", uuid.New().String())
+		log.Info("Namespace %s already exists. Checking new name %s.", uniqueName, newName)
+		uniqueName = newName
+		u, err = azureClient.CheckNamespaceAvailability(p.context, uniqueName)
+		if err != nil {
+			log.Errorf("Error while calling Azure Event Hubs client. Can't check namespace name availability. "+
+				"Error: %v", err)
+			return "", errors.Errorf("Error while calling Azure Event Hubs client. Can't check namespace name "+
```
k15r, line 172: create once, use twice

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -140,6 +147,41 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+	uniqueName := fmt.Sprintf("skr-%s", suffix)
+	u, err := azureClient.CheckNamespaceAvailability(p.context, uniqueName)
+	if err != nil {
+		log.Errorf("Error while calling Azure Event Hubs client. Can't check namespace name availability. "+
+			"Error: %v", err)
+		return "", errors.Errorf("Error while calling Azure Event Hubs client. Can't check namespace name "+
+			"availability. Error: %v", err)
+	}
+
+	attempts := 10
```
k15r, line 163: nit: lets make the number of retries a costant, or make it configurable

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -140,6 +147,41 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+				"Error: %v", err)
+			return "", errors.Errorf("Error while calling Azure Event Hubs client. Can't check namespace name "+
+				"availability. Error: %v", err)
+		}
+		attempts--
+	}
+
+	if !u {
+		log.Errorf("Can't find a unique name for Event Hubs namespace. Failed after 10 attempts.")
+		return "", errors.New("Can't find a unique name for Event Hubs namespace. Failed after 10 attempts.")
```
k15r, line 180: ... and then use the constant/ configured value here
btw: did i mention: create message once, use twice :D

components/kyma-environment-broker/internal/hyperscaler/azure/client.go
```diff
@@ -46,6 +50,19 @@ func (nc *AzureClient) CreateNamespace(ctx context.Context, azureCfg *Config, gr
 	return &ehNamespace, err
 }
 
+func (nc *AzureClient) CheckNamespaceAvailability(ctx context.Context, name string) (bool, error) {
+	//TODO Validate name string
```
k15r, line 54: curious: how and why

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -140,6 +147,41 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+	if err != nil {
+		log.Errorf("Error while calling Azure Event Hubs client. Can't check namespace name availability. "+
+			"Error: %v", err)
+		return "", errors.Errorf("Error while calling Azure Event Hubs client. Can't check namespace name "+
+			"availability. Error: %v", err)
+	}
+
+	attempts := 10
+	for !u && attempts > 0 {
+		newName := fmt.Sprintf("skr-%s", uuid.New().String())
```
k15r, line 165: in case the id skr-<suffix> ns is already used we dont use <suffix> anymore. Would it make sense to still use it in the new unique name to make it easier to map it to the cluster?
In case the newName would be to long you could easily use something other than UUID... i think we dont need the universal uniqueness at this point

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
Abd4llA: GoLand's import sorting is used for the code here. When I used the cmd line as you mentioned, it didn't produce any modifications for me. am I missing something?

components/kyma-environment-broker/internal/hyperscaler/azure/client.go
Abd4llA: validation of empty name added

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -140,6 +147,39 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+
+	if !u {
+		msg := fmt.Sprintf("Can't find a unique name for Event Hubs namespace. Failed after %d attempts.",
+			retryCount)
+		log.Error(msg)
+		return "", errors.New(msg)
+	}
+	return uniqueName, nil
+}
+
```
nachtmaar, line 182: (nit) Calls to CheckNamespaceAvailability and error creation could be moved to an inner function to avoid some repetition and improved readability. 
Please use whatever version you prefer. Just an idea :)

```suggestion
func (p *ProvisionAzureEventHubStep) getUniqueNSName(suffix string, azureClient azure.AzureInterface,
	log logrus.FieldLogger) (string, error) {
	checkAvailability := func(uniqueName string) (bool, error) {
		isUnique, err := azureClient.CheckNamespaceAvailability(p.context, uniqueName)
		if err != nil {
			msg := "Error while calling Azure Event Hubs client. Can't check namespace name availability. Error: %v"
			log.Errorf(msg, err)
			return false, errors.Errorf(msg, err)
		}
		return isUnique, nil
	}

	uniqueName := fmt.Sprintf("skr-%s", suffix)
	isUnique, err := checkAvailability(uniqueName)
	if err != nil {
		return "", err
	}

	for i := retryCount; !isUnique && i > 0; i-- {
		newName := fmt.Sprintf("skr-%s", uuid.New().String())
		log.Info("Namespace %s already exists. Checking new name %s.", uniqueName, newName)
		uniqueName = newName
		isUnique, err = checkAvailability(uniqueName)
		if err != nil {
			return "", err
		}
	}

	if !isUnique {
		msg := fmt.Sprintf("Can't find a unique name for Event Hubs namespace. Failed after %d attempts.",
			retryCount)
		log.Error(msg)
		return "", errors.New(msg)
	}
	return uniqueName, nil
}
```

nachtmaar: (nit) Isn't this a perfect fit for a unit test? 😎

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -96,29 +98,34 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
 		azure.TagOperationID:  &operation.ID,
 	}
 
+	// Generating a unique name for resource group and namespace with operation ID as suffix if available
+	uniqueName, err := p.getUniqueNSName(operation.ID, azureClient, log)
```
nachtmaar, line 102: thinking ... should we use `instance.id` instead? I am not sure about the relation of operation to instance. Does an instance always have only one operation?

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
k15r: > (nit) Calls to CheckNamespaceAvailability and error creation could be moved to an inner function to avoid some repetition and improved readability.

Or we get rid off the 2 different ways to generate names (as humans cannot use them anyway) and reduce to just one function ... no need for a closure here

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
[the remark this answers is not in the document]
Abd4llA: > mhh ... isn't this a typical [TOCTTOU](https://de.wikipedia.org/wiki/Time-of-Check-to-Time-of-Use-Problem) problem? Instead of checking, we could just generate the first name, try it, check for a conflict on azure side (some 4xx) and repeat with other names if required.

You are correct that indeed a race condition can happen between time of check and time of creation, however the design of the Azure SDK does not return a clear status code/error for this condition (unless you wanna parse an error string and our client breaks in case Azure changes the error message, which I'm really not in favour of).

 Generally, TOCTTOC problems are critical on the service side because say if it was vulnerable it can cause two namespaces to be created with the same name.

However, in our case, there is no harm if a race occurred between TOC and TOC 😄 . All the logic of checking NS availability is just to handle the very corner case of the operation ID which is a (UUID) collision, something that is supposedly a very low probability but safer to handle.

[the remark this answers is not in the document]
Abd4llA: > If `getUniqueNSName` fails because of network problems, shouldn't we retry the operation then? There might be a chance that the network problem goes away after a few minutes.
> 
> This is at least how an error during `CreateNamespace` is treated atm.

You are right. Done.

components/kyma-environment-broker/internal/hyperscaler/azure/client.go
```diff
@@ -46,6 +52,81 @@ func (nc *AzureClient) CreateNamespace(ctx context.Context, azureCfg *Config, gr
...
+			"while using Azure client. %v", name, err)
+	}
+	if group.StatusCode != 200 && group.StatusCode != 201 {
+		return false, errors.Errorf("Failed to check Azure resource group availability with name: %s." +
+			" Unexpected API response code. Expected 2XX but received %d", name, group.StatusCode)
+	}
+	return matchWithTags(name, tags, *group.Name, group.Tags), nil
+}
+
+func matchWithTags(nameA string, tagsA Tags, nameB string, tagsB Tags) bool {
```
nachtmaar, line 117: FYI: for deprovisioning we only use the `instanceID` tag to retrieve the `resource group`. There is also a filter available when listing resources in the sdk. See [deprovisioning PR compass#1065](https://github.com/kyma-incubator/compass/blob/b379256c0602d4bf69590937c44e1e8a606489ca/components/kyma-environment-broker/internal/hyperscaler/azure/client.go#L96)

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -140,6 +139,77 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+		return false, errors.Wrapf(err, "Can't check Azure resource group %s.", resourcesName)
+	}
+	// we are using the same name for resource group and namespace name
+	if exists {
+		exists, err = azureClient.NamespaceExists(p.context, resourcesName, resourcesName, tags)
+		if err != nil {
+			return false, errors.Wrapf(err, "Can't check Azure Event Hubs namespace %s.", resourcesName)
+		}
+	}
+	return exists, nil
```
k15r, line 154: i really like early returns
```suggestion
	exists, err := azureClient.ResourceGroupExists(p.context, resourcesName, tags)
	if err != nil {
		return false, errors.Wrapf(err, "Can't check Azure resource group %s.", resourcesName)
	}
	// we are using the same name for resource group and namespace name
	if ! exists {
	         return false, nil
	}
	exists, err = azureClient.NamespaceExists(p.context, resourcesName, resourcesName, tags)
	if err != nil {
		return false, errors.Wrapf(err, "Can't check Azure Event Hubs namespace %s.", resourcesName)
	}
	return exists, nil
```

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -140,6 +139,77 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+			available, err := azureClient.CheckResourceGroupAvailability(p.context, uniqueName)
+			if err != nil {
+				msg := "Error while calling Azure Event Hubs client. Can't check resource group availability. Error: %v"
+				log.Errorf(msg, err)
+				return "", errors.Errorf(msg, err)
+			}
+			if available {
+				return uniqueName, nil
+			}
+		}
```
k15r, line 204:
```suggestion
		if ! available {
		   continue
		}
		available, err := azureClient.CheckResourceGroupAvailability(p.context, uniqueName)
		if err != nil {
			msg := "Error while calling Azure Event Hubs client. Can't check resource group availability. Error: %v"
			log.Errorf(msg, err)
			return "", errors.Errorf(msg, err)
		}
		if available {
			return uniqueName, nil
		}
```

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -140,6 +139,77 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
...
+	if _, err = azureClient.CreateResourceGroup(p.context, azureCfg, groupName, tags); err != nil {
+		// retrying might solve the issue while communicating with azure, e.g. network problems etc
+		return "", errors.Errorf("Failed to create Azure Resource Group [%s] with error: %v", groupName, err)
+	}
+	log.Info("created Azure Resource Group [%s]", groupName)
+
+	// create EventHubs Namespace
+	namespaceName := uniqueName
+	if _, err = azureClient.CreateNamespace(p.context, azureCfg, groupName, namespaceName, tags); err != nil {
+		return "", errors.Errorf("Failed to create Azure Event Hubs namespace [%s] with error: %v", namespaceName, err)
```
k15r, line 175: i guess we would have to cleanup the ResourceGroup in this case?

marcobebway:
Closing this as we decided to use the InstanceID for the names: https://github.com/kyma-incubator/compass/pull/1304#discussion_r422883106.

PR #1200, opened 2020-04-20
Provisioner sets subaccountID annotation based on region

Maladie:
**Description**

- Provisioner can now get subbacountID from ConfigMap
- Installation timeout set to 24h

**Related issue(s)**


chart/compass/charts/provisioner/templates/deployment.yaml
```diff
@@ -102,12 +98,17 @@ spec:
...
             - name: APP_LOG_LEVEL
               value: "{{ .Values.logs.level }}"
-          volumeMounts:
-            - mountPath: /gardener/kubeconfig
-              name: gardener-kubeconfig
-              readOnly: true
+        {{if .Values.gardener.tenantConfigMapName }}
+        volumeMounts:
+          - mountPath: /gardener/tenant
+            name: gardener-tenant-config
```
Szymongib, line 104: I would suggest something like `gardener-audit-log-tenant-config`

chart/compass/charts/provisioner/values.yaml
```diff
@@ -13,6 +13,8 @@ gardener:
   project: "" # Gardener project connected to SA
   kubeconfigPath: "/gardener/kubeconfig/kubeconfig"
   kubeconfig: "" # Base64 encoded Gardener SA key
+  tenantConfigPath: "/gardener/tenant/config"
+  tenantConfigMapName: ""
```
Szymongib, line 17: I think in default installation `tenantConfigPath` should be empty as we want `shouldEnableAuditLogs` to evaluate to false. I would also rename those variables to highlight that that those are audit log tenants, as there are other tenants in the system.

components/provisioner/internal/gardener/provisioner.go
```diff
@@ -106,7 +112,43 @@ func newDeprovisionOperation(id, runtimeId, message string, state model.Operatio
...
+
+func (g *GardenerProvisioner) getAuditLogTenant(region string) (string, error) {
+	file, err := os.Open(g.auditLogTenantConfigPath)
+
+	if err != nil {
+		// As ConfigMap is optional, if file not exists we return empty string and no error
+		if os.IsNotExist(err) {
+			logrus.Infof("Audit logs can't be enabled. File in path %s not exists", g.auditLogTenantConfigPath)
+			return "", nil
+		}
```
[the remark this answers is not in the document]
Maladie: It was necessary to perform some manual tests. It can be changed now.

PR #1207, opened 2020-04-21
Keb configure provisioning timeout

piotrmiskiewicz:
**Description**

- provisioning timeout is configurable
- fluentbit configuration overrides change in FILTER section: change cluster_name to subaccount_id

### DIFF
--- a/chart/compass/charts/kyma-environment-broker/values.yaml
+++ b/chart/compass/charts/kyma-environment-broker/values.yaml
@@ -38,6 +38,9 @@ resources: {}
 provisioner:
   URL: "http://compass-provisioner.compass-system.svc.cluster.local:3000/graphql"
   dumpRequests: "false"
+
+  # Defines how long should the Kyma Environment Broker checks the status of the provisioning in the Provisioner.
+  # The Provisioner timeout is defined in chart/compass/charts/provisioner/values.yaml
   timeout: "12h"
   gardener:
     # name of the secret with kubeconfig to the gardener cluster
--- a/chart/compass/charts/kyma-environment-broker/values.yaml
+++ b/chart/compass/charts/kyma-environment-broker/values.yaml
@@ -38,6 +38,7 @@ resources: {}
 provisioner:
   URL: "http://compass-provisioner.compass-system.svc.cluster.local:3000/graphql"
   dumpRequests: "false"
+  timeout: "12h"
   gardener:
     # name of the secret with kubeconfig to the gardener cluster
     secretName: "gardener"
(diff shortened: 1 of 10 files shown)


chart/compass/charts/kyma-environment-broker/values.yaml
```diff
@@ -38,6 +38,7 @@ resources: {}
 provisioner:
   URL: "http://compass-provisioner.compass-system.svc.cluster.local:3000/graphql"
   dumpRequests: "false"
+  timeout: "12h"
```
mszostok, line 41: hm.. I was thought that here we are storing values directly related to "provisioner" component. This timeout is not sent to provisioner it's only about our timeout per step, right? Provisioner for installation has it own timeout https://github.com/kyma-incubator/compass/blob/master/chart/compass/charts/provisioner/templates/deployment.yaml#L81-L82

For me it's quite confusing.

btw. should we also take care of the installation timeout on provisoner side? because having 24h per step in our KEB is not enough

piotrmiskiewicz: it is about provisioner! it is how long should we check the status in the provisioner.

piotrmiskiewicz: there will be (or it is merged) also change in the provisioner

mszostok: IMO we should add a comment above this timeout field to explicitly said that it is not send to provisoner and provisoner has it’s own timeout which should also be taken into account because if I will not know the provisioner part then I will thought that it is about provionser timeout. Just check the name ‘proviosner.timeout’ - it is really confusing

components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
```diff
@@ -39,7 +39,7 @@ func NewInitialisationStep(os storage.Operations, is storage.Instances, pc provi
 }
 
 func (s *InitialisationStep) Name() string {
-	return "Deprovision_Initialization"
+	return "c"
```
mszostok, line 42: this needs to be reverted

components/kyma-environment-broker/internal/process/provisioning/lms_cert_step.go
```diff
@@ -173,7 +173,7 @@ func (s *lmsCertStep) Run(operation internal.ProvisioningOperation, logger logru
 [FILTER]
         Name record_modifier
         Match *
-        Record cluster_name %s
+        Record subaccount_id %s
```
mszostok, line 176: pr which is also changing that: https://github.com/kyma-incubator/compass/pull/1218/files so maybe it should be aligned between them

piotrmiskiewicz: I've added a comment there yesterday.

PR #1208, opened 2020-04-21
Adding support for sending Audit Logs from Connector   

koala7659:
Following code modification have been made:

1. Adding tenantMappingService (Director hydrator) and claims map into connector oathkeeper rules to enrich Connector requests
2. Using Gateway for sending Audit Logs for connector requests
3. Modification of connector upstream value for connectivity-adapter


components/connectivity-adapter/internal/connectorservice/api/certificates.go
```diff
@@ -48,10 +48,8 @@ func (ch *certificatesHandler) SignCSR(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	contextLogger.Info("Generating certificate")
```
akgalwas, line 51: I think we should not remove this log entry.

components/connectivity-adapter/internal/connectorservice/connector/client.go
```diff
@@ -18,15 +17,11 @@ type client struct {
 	timeout       time.Duration
 }
 
-func NewClient(compassConnectorAPIURL string, timeout time.Duration) (Client, error) {
-	gqlExternalAPIClient := graphql.NewClient(compassConnectorAPIURL, graphql.WithHTTPClient(&http.Client{}))
-
-	client := &client{
-		gqlAPIClient: gqlExternalAPIClient,
-		timeout:      timeout,
+func NewClient(gqlClient *graphql.Client, timeout time.Duration) Client {
```
akgalwas, line 20: timeout argument is no longer used (timeout value hardcoded)

components/connectivity-adapter/internal/connectorservice/api/certificates.go
koala7659: Ok, fixed. My mistake

components/connectivity-adapter/internal/connectorservice/connector/client.go
koala7659: That's correct

PR #1234, opened 2020-04-23
Exposing resources in Helm charts

lumi017:
**Description**

- Exposing resources in Helm charts

**Related issue(s)**


chart/compass/charts/kyma-environment-broker/templates/deployment.yaml
```diff
@@ -154,7 +154,7 @@ spec:
             timeoutSeconds: 3
             initialDelaySeconds: 20
           resources:
```
mszostok, line 156: **suggestion**
if we're already using `nindent` then IMO it will be more readable to have it in one line.

```yaml
resources: {{- toYaml .Values.resources | nindent 12 }}
```

PR #1238, opened 2020-04-24
Add External Services Mock chart

tgorgol:
**Description**

- Add External Services Mock chart

### DIFF
--- a/chart/compass/charts/external-services-mock/templates/auditlog-configmap.yaml
+++ b/chart/compass/charts/external-services-mock/templates/auditlog-configmap.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.global.auditlog.configmap.name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Chart.Name }}
+    release: {{ .Release.Name }}
+data:
+  auditlog-oauth-user-var: "mock-user"
+  auditlog-oauth-tenant-var: "mock-tenant"
+  auditlog-url: "https://compass-external-services-mock.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.service.port }}"
+  auditlog-config-path: "audit-log/v2/configuration-changes"
+  auditlog-security-path: "audit-log/v2/security-events"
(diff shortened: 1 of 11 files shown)


chart/compass/values.yaml
```diff
@@ -257,7 +262,14 @@ global:
       maxOpenConnections: 1
       maxIdleConnections: 1
 
-  enableCompassDefaultScenarioAssignment: true
+  auditlog:
+    configmap:
```
pkosiec, line 266: `configMap`

chart/compass/values.yaml
```diff
@@ -257,7 +262,14 @@ global:
       maxOpenConnections: 1
       maxIdleConnections: 1
 
-  enableCompassDefaultScenarioAssignment: true
+  auditlog:
```
pkosiec, line 265: `auditLog`?

chart/compass/charts/gateway/values.yaml
```diff
@@ -10,9 +10,5 @@ gateway:
   enabled: true
   manageCerts: true
   auditlog:
-    enabled: false
-    authMode: "basic"
-    configmap:
-      name: "compass-gateway-auditlog-config"
-    secret:
-      name: "compass-gateway-auditlog-secret"
+    enabled: true # TODO: Switch to false after testing
```
pkosiec, line 13: I think this value shouldn't be here anymore - we should use global value

chart/compass/charts/gateway/values.yaml
```diff
@@ -10,9 +10,5 @@ gateway:
...
+    authMode: "oauth" # TODO: Switch to basic after testing
```
pkosiec, line 14: I wonder if we shouldn't move it also to global, under `auditLog`

chart/compass/charts/external-services-mock/templates/auditlog-configmap.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: {{ .Values.global.auditlog.configMapName }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Chart.Name }}
+    release: {{ .Release.Name }}
+data:
+  auditlog-oauth-user-var: "mock-user"
+  auditlog-oauth-tenant-var: "mock-tenant"
+  auditlog-url: "http://compass-external-services-mock.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.service.port }}"
```
dbadura, line 12: `auditlog-oauth-tenant-var` and `auditlog-oauth-user-var` is not needed.
We have default values, which are checked in `external-services-mock`.

chart/compass/charts/gateway/values.yaml
pkosiec: In future it would be good to move it to global values

PR #1240, opened 2020-04-24
Set runtime label with kibana url

piotrmiskiewicz:
**Description**

- introduce SetLabel method in the InputCreator which allows every step to add a label
- set label with lmsURL


jasiu001:
lgtm

PR #1242, opened 2020-04-24
Add scripts for temporary job for Compass with audit log mock enabled

pkosiec:
**Description**

- Add scripts for temporary job for Compass with audit log mock enabled

### DIFF
--- a/installation/scripts/testing-auditlog-temp.sh
+++ b/installation/scripts/testing-auditlog-temp.sh
@@ -1,0 +1,113 @@
+#!/usr/bin/env bash
+ROOT_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+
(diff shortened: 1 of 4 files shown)


installation/scripts/testing-auditlog-temp.sh
```diff
@@ -0,0 +1,113 @@
...
+
+${kc} get clustertestsuites.testing.kyma-project.io > /dev/null 2>&1
+if [[ $? -eq 1 ]]
+then
+   echo "ERROR: script requires ClusterTestSuite CRD"
+   exit 1
+fi
+
+matchNames=$(cat <<-END
+    matchNames:
```
dbadura, line 26: Should you add  new matchNames: `compass-director-external-services-integration`?

pkosiec: I didn't add it yet, because there is no such test. I think the test author should include it

PR #1263, opened 2020-04-29
Make gardener k8s version parametrized

mszostok:
**Description**

- Make gardener k8s version parametrized


chart/compass/charts/kyma-environment-broker/templates/deployment.yaml
```diff
@@ -172,6 +172,8 @@ spec:
               value: {{ .Values.gardener.project }}
             - name: APP_GARDENER_KUBECONFIG_PATH
               value: {{ .Values.gardener.kubeconfigPath }}
+            - name: APP_GARDENER_KUBERNETES_VERSION
```
piotrmiskiewicz, line 175: are you sure it is APP_GARDENER_KUBERNETES_VERSION ?

mszostok: thanks, good catch 👍

PR #1267, opened 2020-04-30
Fix KEB steps behavior

jasiu001:
**Description**

- marking the parameter `SubAccountID` as required in broker create instance action
- handle 404 response code for EDP client during the deletion process
- handle 404 response code for IAS client during the deletion process

### DIFF
--- a/components/kyma-environment-broker/internal/ias/client.go
+++ b/components/kyma-environment-broker/internal/ias/client.go
@@ -41,6 +41,7 @@ type (
 		Path    string
 		Body    io.Reader
 		Headers map[string]string
+		Delete  bool
 	}
 )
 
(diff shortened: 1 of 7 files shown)


components/kyma-environment-broker/internal/ias/client.go
```diff
@@ -41,6 +41,7 @@ type (
 		Path    string
 		Body    io.Reader
 		Headers map[string]string
+		Delete  bool
```
mszostok, line 44: maybe we can have consistent naming with edp and also use `allowNotFound`

jasiu001: I thought about this parameter more like a request method (Delete: true), I need to recognize if it is delete action, for IAS API it is just PUT method on `/service/sps/delete` endpoint.

PR #1270, opened 2020-05-04
Build Compass components with Go 1.14

NickyMateev:
**Description**

Some unit tests make error assertions which are breaking on Go 1.14.X due to missing quotation signs.

- Adapt unit tests with missing quotes in certain error test expectations

Reference breaking change in Go: https://github.com/golang/go/issues/37630

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -15,28 +15,28 @@ global:
       path: eu.gcr.io/kyma-project/incubator
     connector:
       dir:
-      version: "PR-1162"
+      version: "PR-1270"
     connectivity_adapter:
       dir:
-      version: "PR-1208"
+      version: "PR-1270"
     pairing_adapter:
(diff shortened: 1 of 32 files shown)


pkosiec:
@NickyMateev Hmm, the tests won't pass as still on Prow we use older Go while running tests:
https://github.com/kyma-project/test-infra/blob/master/templates/config.yaml#L129

There are two options:
- Modify the image tag globally for all Compass components - it would mean that all Compass components (and all Compass teams) will need to switch to Go 1.14. We can try to push it, but it would mean fixing this kind of issues in all Compass components. Hopefully there's not that many issues
- Duplicate the config and use the 1.14 image tag only for some components (like Director).

I would recommend going with the first option, however it can take a while.

Also, we need to update Dockerfiles for each Compass component (to use alpine with Go 1.14 to build components).

PR #1272, opened 2020-05-04
Provide Make targets for Minikube local development

NickyMateev:
**Description**

When developing and testing **Compass** components as a whole in a **Minikube** local environment developers will want to have a way to quickly integrate new changes and then assert them/test then. Manual intervention of Helm Charts and recreation of **Compass** each time a new change is made to the codebase is too slow to yield fast feedback to the developer.

A way to get faster feedback is to build a new Docker image of the **Compass** component being worked on and updating the corresponding deployment with the new image.

This change aims to automate this process by providing two new Make targets to the generic Makefile so that every **Compass** component can make use of them. Only the Schema Migrator component, being a one-time Job, doesn't make much sense and doesn't really fit to use use of these two targets, so it has been excluded from ability to run them.

- `build-to-minikube` target - builds the corresponding **Compass** component Docker image to the Minikube's Docker Registry.
- `deploy-on-minikube` target - updates the corresponding **Compass** component's Deployment with the latest image and restarts the deployment.
- The Pairing Adapter Deployment container has been renamed to be equal to it's Chart Name to comply with the conventions of all the other **Compass** components which do that. This way the Make targets can work for the Pairing Adapter as well.

### DIFF
--- a/scripts/generic_make_go.mk
+++ b/scripts/generic_make_go.mk
@@ -36,8 +36,6 @@ DIRS_TO_IGNORE = go list ./... | grep "$(VERIFY_IGNORE)"
 DEPLOYMENT_NAME="compass-"$(COMPONENT_NAME)
 # NAMESPACE defines the namespace into which the component is deployed
 NAMESPACE="compass-system"
-# TEMP_FILE_PATH is used to store the old deployment resource state during redeployment of said deployment
-TEMP_FILE_PATH=$(COMPONENT_DIR)/$(COMPONENT_NAME)-tmp.yml
 
 # Base docker configuration
 DOCKER_CREATE_OPTS := -v $(LOCAL_DIR):$(WORKSPACE_LOCAL_DIR):delegated --rm -w $(WORKSPACE_COMPONENT_DIR) $(BUILDPACK)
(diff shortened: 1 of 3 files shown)


scripts/generic_make_go.mk
```diff
@@ -88,6 +99,10 @@ docker-create-opts:
 MOUNT_TARGETS = build resolve ensure dep-status check-imports imports check-fmt fmt errcheck vet generate pull-licenses gqlgen
 $(foreach t,$(MOUNT_TARGETS),$(eval $(call buildpack-mount,$(t))))
 
+# Builds new docker image inside Minikube's Docker Registry
```
pkosiec, line 102:
```suggestion
# Builds new Docker image into Minikube's Docker Registry
```

scripts/generic_make_go.mk
```diff
@@ -164,3 +179,8 @@ exec:
 	@docker run $(DOCKER_INTERACTIVE) \
     		-v $(COMPONENT_DIR):$(WORKSPACE_COMPONENT_DIR):delegated \
     		$(DOCKER_CREATE_OPTS) bash
+
+# Redeploys component in Minikube cluster by replacing old deployment with the newly built one
```
pkosiec, line 183:
```suggestion
# Sets locally built image for a given component in Minikube cluster 
```

PR #1276, opened 2020-05-04
Annotations based on seeds instead of regions

janmedrek:
**Description**

- Changed Gardener audit log mapping to be based on seeds instead of regions


components/provisioner/internal/gardener/shoot_reconciler.go
```diff
@@ -264,3 +284,62 @@ func (r *ProvisioningOperator) setDeprovisioningFinished(cluster model.Cluster,
...
+func setAuditConfig(shoot *gardener_types.Shoot, policyConfigMapName, subAccountId string) {
+	if shoot.Spec.Kubernetes.KubeAPIServer == nil {
+		shoot.Spec.Kubernetes.KubeAPIServer = &gardener_types.KubeAPIServerConfig{}
+	}
+
+	shoot.Spec.Kubernetes.KubeAPIServer.AuditConfig = &gardener_types.AuditConfig{
+		AuditPolicy: &gardener_types.AuditPolicy{
+			ConfigMapRef: &v12.ObjectReference{Name: policyConfigMapName},
+		},
+	}
```
Szymongib, line 297: Correct me if I am wrong, but from what I understand the Config Map will not change during Shoot lifetime, so we can set it during the provisioning. I think it would be a little cleaner to not to modify spec in our controller.

components/provisioner/internal/gardener/shoot_reconciler.go
```diff
@@ -91,6 +103,14 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
+	seed := getSeed(shoot)
+	if seed != "" {
+		err := r.enableAuditLogs(&shoot, r.auditLogsCMName, seed)
+		if err != nil {
+			log.Errorf("Failed to enable audit logs for %s shoot: %s", shoot.Name, err.Error())
+		}
+	}
```
Szymongib, line 112: I think this code should be called after checking if `shouldReconcileShoot`, so that each Provisioner handles only Shoots from its database.

components/provisioner/internal/gardener/shoot_reconciler.go
```diff
@@ -264,3 +284,62 @@ func (r *ProvisioningOperator) setDeprovisioningFinished(cluster model.Cluster,
...
+}
+
+func (r *Reconciler) enableAuditLogs(shoot *gardener_types.Shoot, policyConfigMapName, seed string) error {
+	logrus.Info("Enabling audit logs")
+	tenant, err := r.getAuditLogTenant(seed)
+	if err != nil {
+		return err
+	}
+
+	if tenant != "" {
```
Szymongib, line 309: I think we could also check the current audit log tenant on Shoot and update only if they are not equal.
We can also flip the if statement to make it a little more clear.
```go
if tenant == "" {
    logrus.Warnf("Cannot enable audit logs. Tenant for region %s is empty", seed)
    return nil
} else if tenant == currentTenant {
    logrus.Debugf("Seed %s did not changed, skipping annotating with Audit Log Tenant", seed)
    return nil
}

logrus.Infof("Modifying Audit Log Tenant") // We can log old one and new one but I am not sure how confidential are those

return r.provisioningOperator.updateShoot(*shoot, func(s *gardener_types.Shoot) {
			setAuditConfig(s, policyConfigMapName, tenant)
		})
```

janmedrek: I would skip logging it, added info for which shoot the tenant was changed (this is absolutely sufficient).

PR #1284, opened 2020-05-05
Bump Provisioner image version

janmedrek:
**Description**

- bump Provisioner image version

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -36,7 +36,7 @@ global:
       version: "PR-1216"
     provisioner:
       dir:
-      version: "PR-1244"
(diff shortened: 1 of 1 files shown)


chart/compass/values.yaml
```diff
@@ -60,7 +60,7 @@ global:
         version: "PR-1156"
       provisioner:
         dir:
-        version: "PR-1193"
```
Szymongib, line 63: This is an image of tests `global.images.tests.provisioner` :)

janmedrek: That was dumb, indeed.
> "Mind that such missteps are the exception, and not the rule."

Issue #1136, opened 2020-04-06
Fine tune naming validation for resourcegroup/event hubs


sayanh:
**Description**

Right now, the length of the cluster has a minimum length validation. The name of the cluster is used to create Resource groups and Event hubs.
 The rules are as follows:
`Eventhub` related naming rules -> https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsofteventhub
`Resourcegroup` naming rules -> https://docs.microsoft.com/bs-latn-ba/rest/api/resources/deployments/validate#uri-parameters

**Reasons**

If naming validation fails in KEB, the asynchronous step of event-hub creation fails. Once failed, it goes for a retry which never succeeds. This is not a great experience. The validation check should happen right at the beginning and fail fast before even starting the async processes.

marcobebway:
Blocked until we decide on the naming.

PR #1304, opened 2020-05-07
Use the InstanceID as a unique name for the EventHubs Resource Group and Namespace

marcobebway:
**Description**

- Use the InstanceID as a unique name for the EventHubs Resource Group and Namespace.

**Related issue(s)**

- https://github.com/kyma-incubator/compass/issues/1136.

### DIFF
--- a/components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
+++ b/components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
@@ -83,7 +83,7 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
 	}
 
 	// create hyperscaler client
-	namespaceClient, err := p.EventHub.HyperscalerProvider.GetClient(azureCfg, log)
+	azureClient, err := p.EventHub.HyperscalerProvider.GetClient(azureCfg, log)
 	if err != nil {
 		// internal error, repeating doesn't solve the problem
 		errorMessage := fmt.Sprintf("Failed to create Azure EventHubs client: %v", err)
@@ -98,9 +98,8 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
 	}
 
 	// create Resource Group
-	groupName := pp.Parameters.Name
-	// TODO(nachtmaar): use different resource group name https://github.com/kyma-incubator/compass/issues/967
-	resourceGroup, err := namespaceClient.CreateResourceGroup(p.EventHub.Context, azureCfg, groupName, tags)
+	groupName := operation.InstanceID
+	resourceGroup, err := azureClient.CreateResourceGroup(p.EventHub.Context, azureCfg, groupName, tags)
 	if err != nil {
 		// retrying might solve the issue while communicating with azure, e.g. network problems etc
 		errorMessage := fmt.Sprintf("Failed to persist Azure Resource Group [%s] with error: %v", groupName, err)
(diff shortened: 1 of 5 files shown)


components/kyma-environment-broker/internal/model.go
```diff
@@ -99,16 +99,15 @@ type InstanceWithOperation struct {
...
-
 	// following fields are not stored in the storage
+	Operation    `json:"-"`
 	InputCreator ProvisionInputCreator `json:"-"`
 
-	Avs AvsLifecycleData `json:"avs"`
+	// following fields are serialized to JSON and stored in the storage
+	Lms                    LMS              `json:"lms"`
+	Avs                    AvsLifecycleData `json:"avs"`
+	EventHubGeneratedName  string           `json:"eventhub_generated_name"`
```
anishj0shi, line 109: how are we ensuring this parameter is persisted in storage?

marcobebway: It is part of the `ProvisioningOperation` struct which is [persisted](https://github.com/kyma-incubator/compass/blob/f4b3be5e9567db4a473096e46f2254979ac5928e/components/kyma-environment-broker/internal/process/provisioning/create_runtime.go#L67) to the storage every time the provisioning manager gets invoked by the worker queue and calls [execute](https://github.com/kyma-incubator/compass/blob/f4b3be5e9567db4a473096e46f2254979ac5928e/components/kyma-environment-broker/internal/process/provisioning/manager.go#L62).

anishj0shi: But, correct me if I'm wrong, not all the fields in this struct are persisted, right?
there must be a schema definition which controls this. did you cross check with the DB?

marcobebway: Yes, I already checked that in the underlying storage. Fields with exported json names are stored in the database. Fields with `json:"-"` are skipped.

nachtmaar: Schema is json ;) We can persist the update ourselves using `UpdateOperation` in the `OperationManager`. Doing so would prevent a not updated database then the step never reaches the queue/scheduler.

mszostok: I have one problem with it.

**Steps:**
1. A UUID: `operation.EventHubGeneratedName = uuid.New().String()`  is generated in 
`event_hub_step.go`
2. `EventHubGeneratedName` used to `ResourceGroup`
3. `EventHubGeneratedName` used to created `Namespace`
4. The `EventHubGeneratedName` is persisted on `OperationFailed` or `OperationSucceeded`


and my fears are connected with step 4 because we can have a situation when there will be an error while saving that to db but you already created resources with that id and you will lose the correlation.  


so my proposition is to use the `InstanceID`. This is a globally unique non-empty string send by Platform during provision call. This ID is used for future requests (bind and deprovision), so the Service Broker uses it to correlate the resource it creates.

You do not have to generate your own ID as this one is already available and can be used. This will simplify the logic and also this id is saved before the steps are executed so it is always present in db. What's more, it's even remembered by Platform which triggered that provisioning request, so even if our db will blow up we will be able to correlate that resource.

anishj0shi: Hi @mszostok , we initially thought to go ahead with instance id, but we saw some redundancies amongst the 3 MPS environment we have, at the time of testing. so for example in the broker request for MPS-DEV we raise a provisioning request with instance_id *aj-test1* it create an event-hub with the name *aj-test1*, now we the same request was triggered in MPS-STAGE, it will create a provisioning operation because there's no entry for the key *aj-test1* but this will fail in creating azure event hub. because these names have to unique at azure data centre level. hence we have to always ensure this name is unique independent of the provisioning parameter.

components/kyma-environment-broker/internal/process/provisioning/event_hub_step.go
```diff
@@ -109,17 +114,19 @@ func (p *ProvisionAzureEventHubStep) Run(operation internal.ProvisioningOperatio
 	log.Printf("Persisted Azure Resource Group [%s]", groupName)
 
 	// create EventHubs Namespace
-	eventHubsNamespace := pp.Parameters.Name
-	eventHubNamespace, err := namespaceClient.CreateNamespace(p.EventHub.Context, azureCfg, groupName, eventHubsNamespace, tags)
```
anishj0shi, line 113: @marcobebway can we use the instance_id instead of name ?

marcobebway: Done.

PK85:
@marcobebway if you switch to service instance ID please remove validation where SKR name was used and you

marcobebway:
> @marcobebway if you switch to service instance ID please remove validation where SKR name that was used only for Event Hub Step.

Done.

PR #1305, opened 2020-05-07
Improve Logging for Retry Operation in KEB

anishj0shi:
Currently when a retry operation is triggered. it's really difficult to know, why it was triggered on the first place. because the errormessage passed in the function will only be used if the operation is failed. 
Introducing additional log to specify that a retry operation was triggered due to an error message.

### DIFF
--- a/components/kyma-environment-broker/internal/process/deprovision_operation.go
+++ b/components/kyma-environment-broker/internal/process/deprovision_operation.go
@@ -72,6 +72,7 @@ func (om *DeprovisionOperationManager) RetryOperationOnce(operation internal.Dep
 func (om *DeprovisionOperationManager) RetryOperation(operation internal.DeprovisioningOperation, errorMessage string, retryInterval time.Duration, maxTime time.Duration, log logrus.FieldLogger) (internal.DeprovisioningOperation, time.Duration, error) {
 	since := time.Since(operation.UpdatedAt)
 
+	log.Infof("Retry Operation was triggered with message: %s", errorMessage)
 	log.Infof("Retrying for %s in %s steps, error: %s", maxTime.String(), retryInterval.String(), errorMessage)
 	if since < maxTime {
 		return operation, retryInterval, nil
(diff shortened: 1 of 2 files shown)


components/kyma-environment-broker/internal/process/deprovision_operation.go
```diff
@@ -72,6 +72,7 @@ func (om *DeprovisionOperationManager) RetryOperationOnce(operation internal.Dep
 func (om *DeprovisionOperationManager) RetryOperation(operation internal.DeprovisioningOperation, errorMessage string, retryInterval time.Duration, maxTime time.Duration, log logrus.FieldLogger) (internal.DeprovisioningOperation, time.Duration, error) {
 	since := time.Since(operation.UpdatedAt)
 
+	log.Infof("Retry Operation was triggered with message: %s", errorMessage)
 	log.Infof("Retrying for %s in %s steps, error: %s", maxTime.String(), retryInterval.String(), errorMessage)
```
piotrmiskiewicz, line 76: the existing log statement is already logging errorMessage

anishj0shi: @piotrmiskiewicz it doesn't mention the error message till the operation is failed. once failed then it logs. it would be nice to see the reason behind retry as well, right?

piotrmiskiewicz: I agree with the idea of this PR. You are absolutely right! In the line 76 we are already logging the errorMessage. This is the reason you don't need to add line 75 or I'm missing something?

anishj0shi: @piotrmiskiewicz when i was debugging, i saw there were some operations which calls retry for like 30 mins, now this RetryOperation will not log the error message till 30 mins, I thought it would be a good idea to have an intermediate log to prompt user that why this RetryOperation was triggered on the first place

PR #1322, opened 2020-05-11
sparated the script.configMapName

lauraanddola:


pkosiec:
Hi, it is already applied on master:
https://github.com/kyma-incubator/compass/blob/master/chart/compass/values.yaml#L92

That's why I'm closing this PR 🙂 Cheers!

PR #1327, opened 2020-05-11
Read component order as *int

Szymongib:
**Description**

- Read component order as *int

**Related issue(s)**

### DIFF
--- a/components/provisioner/internal/provisioning/persistence/dbsession/read.go
+++ b/components/provisioner/internal/provisioning/persistence/dbsession/read.go
@@ -3,8 +3,9 @@ package dbsession
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/kyma-incubator/compass/components/provisioner/internal/util"
 	"sort"
+
+	"github.com/kyma-incubator/compass/components/provisioner/internal/util"
 
 	dbr "github.com/gocraft/dbr/v2"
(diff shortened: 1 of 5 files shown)


components/provisioner/internal/provisioning/persistence/dbsession/read.go
```diff
@@ -71,15 +73,15 @@ func (r readSession) GetCluster(runtimeID string) (model.Cluster, dberrors.Error
...
+		return model.Cluster{}, dberr.Append("Cannot get Provider config for runtimeID: %s", runtimeID)
 	}
 	cluster.ClusterConfig = providerConfig
 
-	kymaConfig, err := r.getKymaConfig(runtimeID, cluster.ActiveKymaConfigId)
-	if err != nil {
-		return model.Cluster{}, dberrors.NotFound("Cannot find Kyma config for runtimeID: %s", runtimeID)
+	kymaConfig, dberr := r.getKymaConfig(runtimeID, cluster.ActiveKymaConfigId)
+	if dberr != nil {
+		return model.Cluster{}, dberr.Append("Cannot get Kyma config for runtimeID: %s", dberr.Error())
```
franpog859, line 84: Printing runtimeID would be more helpful here than dberr.Error()

Szymongib:
> The `component_order` column has no NOT NULL constraint. Is it the reason for this change?

Yes

PR #1342, opened 2020-05-13
exclude any logs from LMS resulting from keycloak-gatekeeper

a-thaler:
**Description**

Kyma is using the keycloak-gatekeeper in some places for verification of a authenticated user request. The logs created by the proxy will contain the email address of the user and should not get collected by LMS.
All these containers are named "kcproxy" derived from the former project name "keycloak-proxy", excluding that container name pattern for the LMS config will solve the problem.

**Related issue(s)**


rakesh-garimella:
As per template [here](https://github.com/kyma-project/kyma/blob/master/resources/logging/charts/fluentbit/templates/_helpers.tpl#L131) it should not add any whitespace after comma so this looks good. We had issues when there was whitespace around comma.

Issue #1236, opened 2020-04-24
MVP monitoring on KEB


PK85:
**Description**
This issue is about to:
- define metrics that KEB will expose
- show KEB metrics dashoard in the MPS grafana 
- will be an input for configuring alerting

**Reasons**

We're blind right now.

**AC**
- expose technical metrics:
  - CPU / Memory consumption
  - active goroutines
  - active DB connections number in connection pool
  - avarage time of provisioning
- expose business metrics:
  - number of service instances
    - status of provisioning
    - status of deprovisioning
  - status from all steps per service instance
    - provisioning sucess/failure 
    - deprovisioning sucess/failure

piotrmiskiewicz:
I'd add:
 - how long does it take to provision instance (per provider: gcp, azure)

ebensom:
I'd add theses metrics, whoich would be used for alerting:

- `keb_provisioning_result{runtime=, operation_id=,...}` : 0 if failed, 1 if successful, 2 if in progress
- `keb_provisioning_step_result{runtime=xx,operation_id=,step_name=,} ` : `0` if failed, `1` if successful, `2` if in progress

The same metrics for deprovisioning.

ebensom:
@mszostok Quick question, do you plan to expose HAP usage metrics also?

PR #1351, opened 2020-05-15
KEB Operations metrics

piotrmiskiewicz:
**Description**

- KEB: operations metrics

**Related issue(s)**
Related: #1236


components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -184,9 +186,26 @@ func main() {
 	bundleBuilder := ias.NewBundleBuilder(httpClient, cfg.IAS)
 	iasTypeSetter := provisioning.NewIASType(bundleBuilder, cfg.IAS.Disabled)
 
+	// metrics collectors
```
mszostok, line 189: IMO the main is going to be bigger and bigger. Can we just hide all those 16 lines of code to 1 ine? Sth like

```go

eventPublisher := metrics.RegisterAll()

// then in use that in managers, e.g.
provisionManager := provisioning.NewManager(db.Operations(), eventPublisher, logs.WithField("provisioning", "manager"))
	
```

IMO defining all those metrics gives us more mess instead of readability

components/kyma-environment-broker/internal/event/application_broker.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// ApplicationEventBroker implements a simple event broker which allows to send event across the application.
+type ApplicationEventBroker struct {
+	mu sync.Mutex
+
+	handlers map[reflect.Type][]Handler
+}
+
+func NewApplicationEventBroker() *ApplicationEventBroker {
```
mszostok, line 22: what about sth like `NewPubSub`? just to avoid words like `application`, `broker` which can be misleading because of our domain (we already have Application, we already have broker). 

IMO `NewPubSub` exactly says what will be returned

components/kyma-environment-broker/internal/event/application_broker.go
```diff
@@ -0,0 +1,48 @@
...
+	return &ApplicationEventBroker{
+		handlers: make(map[reflect.Type][]Handler),
+	}
+}
+
+func (b *ApplicationEventBroker) Publish(ctx context.Context, ev interface{}) {
```
mszostok, line 28: IMO using `Publish` method should be non-blocking. Why do we want to wait until all subscribers will handle that event? It's a little bit against that pattern

piotrmiskiewicz: done

components/kyma-environment-broker/internal/event/application_broker.go
piotrmiskiewicz: done

components/kyma-environment-broker/cmd/broker/main.go
piotrmiskiewicz: refactored

components/kyma-environment-broker/internal/process/deprovisioning/manager_test.go
```diff
@@ -9,9 +9,15 @@ import (
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/storage"
 
+	"context"
```
mszostok, line 12: **nit:** please sort import

components/kyma-environment-broker/internal/process/provisioning/manager.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal"
 	"github.com/kyma-incubator/compass/components/kyma-environment-broker/internal/storage"
 
+	"context"
```
mszostok, line 10: **nit:** please sort import

mszostok:
Things that we discussed f2f and can be commented under issue:

- document follow-up task as this was  only MVP (e.g. instance per state, db metrics, etc.)
- add small documentation that describes what kind of metrics are exposed by KEB and which one have also Grafana dashboard.

Issue #1355, opened 2020-05-18
KEB: Deprovisioning got stuck


piotrmiskiewicz:
**Description**

When the deprovisioning operation is not present in the DB, the processing got stuck: 
```
Cannot fetch operation from storage: while getting operation by ID: Operation with id 2f3133dd-cdb3-499a-8731-5a3bdfe21ed5 not exist
```

Investigate, how it is possible the operation dissapeared.

PR #1356, opened 2020-05-18
Reprocess deprovisioning in KEB with correct existing operation ID

piotrmiskiewicz:
**Description**

- use correct operation ID in case of reprocessing deprovisioning operation

**Related issue(s)**
Resolves: #1355


components/kyma-environment-broker/internal/broker/instance_deprovision.go
```diff
@@ -76,7 +76,7 @@ func (b *DeprovisionEndpoint) Deprovision(ctx context.Context, instanceID string
 
 			logger.Info("Reprocessing failed deprovisioning of runtime")
 
-			b.queue.Add(operationID)
+			b.queue.Add(existingOperation.ID)
```
piotrmiskiewicz, line 79: this change fixes the bug

PR #1360, opened 2020-05-18
Extend documentation about statuses in the Director

klaudiagrz:
**Description**

- Extend documentation about statuses in the Director
- Fix docs numeration

**Related issue(s)**


docs/director/03-statuses.md
```diff
@@ -0,0 +1,17 @@
+# Application and Runtime status
+
+Applications and Runtimes in the Director use the `Status` field to inform the user about their condition. The possible conditions are:
+- `INITIAL` - used for newly created Applications and Runtimes that haven't performed any call to the Director
```
majakurcius, line 4:
```suggestion
- `INITIAL` - used for newly created Applications and Runtimes that have not performed any calls to the Director yet
```

It's just a suggestions, but the original version is okay too.

docs/director/03-statuses.md
```diff
@@ -0,0 +1,17 @@
...
+- `PROVISIONING` - used for Runtimes that are being provisioned
+- `CONNECTED` - used for connected Applications and Runtimes
+- `FAILED` - used for Applications and Runtimes whose connection with the Director failed
```
majakurcius, line 7:
```suggestion
- `FAILED` - used for Applications and Runtimes whose connection with the Director failed
```

I was wondering: `connection` or `attempt to connect`?

docs/director/03-statuses.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Automatic status update
+
+When a given Application or Runtime communicates with the Director API for the first time, the Director automatically sets the entity's status to `CONNECTED`.
+
+The communication can be either direct or through the Integration System.
+- If the Integration System registers Applications and Runtimes in the Director, it manages the statuses of Runtimes and Applications on its own.
+- If you register Applications and Runtimes directly in the Director, you can manage the statuses manually.
+
+In both ways of communication, the statuses of Applications and Runtimes are automatically set to `CONNECTED` every time they connect with the Director API, regardless their previous condition.
```
majakurcius, line 17:
```suggestion
In both ways of communication, the statuses of Applications and Runtimes are automatically set to `CONNECTED` every time they connect with the Director API, regardless of their previous condition.
```
See [this](https://forum.wordreference.com/threads/regardless-or-regardless-of.3453541/).

docs/director/03-statuses.md
```diff
@@ -0,0 +1,17 @@
...
+- `CONNECTED` - used for connected Applications and Runtimes
+- `FAILED` - used for Applications and Runtimes whose connection with the Director failed
+
+## Automatic status update
+
+When a given Application or Runtime communicates with the Director API for the first time, the Director automatically sets the entity's status to `CONNECTED`.
+
+The communication can be either direct or through the Integration System.
+- If the Integration System registers Applications and Runtimes in the Director, it manages the statuses of Runtimes and Applications on its own.
+- If you register Applications and Runtimes directly in the Director, you can manage the statuses manually.
```
majakurcius, line 15: `directly in the Director` sound a bit funny, but no better alternative comes to my mind either...
`right` or `explicitly`? But they don't sound great. Let's leave it like this if nothing better comes to mind :)

docs/director/03-statuses.md
klaudiagrz: Good catch, thanks! :)

docs/director/03-statuses.md
klaudiagrz: Changed 👍

Issue #1325, opened 2020-05-11
POC Error handling in Director


PK85:
**Description**
POC that will be used in: https://github.com/kyma-incubator/compass/issues/1326

**AC**
- Find format in which we want to return errors that will fulfill all kind of errors (e.g. tenant or entity not found, etc.)
- Define and implement minimal set of error types (e.g. TENANT_NOT_FOUND, ENTITY_NOT_FOUND, etc.)

PR #1366, opened 2020-05-19
Error handling POC

dbadura:
**Description**

POC of error handling in Director, which contains:
-  minimal custom error building system 
- graphql error presenter, which add one attribute `error_code` to the `errors` section in response
- middleware which look for internal errors and not handled errors by our implmenetation in errors stack. If found print the stack on console and return internal error to the user.
- simple SQL mapper
- special mutation `testError`, which trigger our custom errors.

Also I implemented basic handling in following cases:
- Invalid UUID in query, e.g `application(id)`
- invalid data in `registerApplication`
- not exisiting application
- add package to not exist application (currently I return Detailed error) I think we should check if such application exist, before putting the package.

Feel free to suggest changes, ask questions to improve implementation and indicate not covered cases .

**Related issue(s)**

See #1325


dbadura:
Add `Unsuficient Scopes` Error code.

dbadura:
I don't like the idea of using `iota`. What do you think about replace `iota` with similar http errors?

dbadura:
I would also add `panic handler`

components/director/pkg/customerrors/error_codes.go
```diff
@@ -0,0 +1,13 @@
+package customerrors
+
+type ErrorCode int
+
+const (
+	UnhandledError ErrorCode = iota
+	InternalError
+	NotFound
+	NotUnique
+	TenantNotFound
```
kjaksik, line 10: Please, add a `TenantNotExisting` error. It happens when user passes a tenant, that doesn't have a mapping in Director

components/director/pkg/customerrors/error_codes.go
```diff
@@ -0,0 +1,13 @@
...
+	InvalidData
+	ConstaintVolation
```
kjaksik, line 12:
```suggestion
	ConstraintViolation
```

tgorgol:
> I don't like the idea of using `iota`. What do you think about replace `iota` with similar http errors?

I agree that we shouldn't use iota, to avoid a situation where someone changes order of consts and breaks the contract. But I wouldn't try replicating the HTTP response codes, just use our numbers.

tgorgol:
> I would also add [panic handler](https://gqlgen.com/reference/errors/#the-panic-handler)

gqlgen recovers from panics by default, but we could use custom one to attach the UUID I mentioned earlier.

docs/internal/proposals/error-handling/error-handling.md
```diff
@@ -0,0 +1,48 @@
...
+* panics
+* postgreSQL, except errors like NotFound, NotUnique
+
+To deal with the errors we introduced custom errors, which contains error codes.
+
+## Proposed custom errors and theirs error codes (the error codes can change):
+
+| Error type           | Error code  |                            Description                                                      |
+|----------------------|-------------|---------------------------------------------------------------------------------------------|
+| Internal Error       | 1           | Error which cannot be handler in director                                                   |
```
crabtree, line 27:
```suggestion
| Internal Error       | 1           | Error which cannot be handled in the Director                                                   |
```

docs/internal/proposals/error-handling/error-handling.md
```diff
@@ -0,0 +1,48 @@
...
+In group of internal errors, are errors from:
+* external libraries
+* panics
+* postgreSQL, except errors like NotFound, NotUnique
+
+To deal with the errors we introduced custom errors, which contains error codes.
+
+## Proposed custom errors and theirs error codes (the error codes can change):
+
+| Error type           | Error code  |                            Description                                                      |
```
crabtree, line 25: Why are you mixing two naming conventions here? One with space in between and one without space. I think we should have only one convention.

docs/internal/proposals/error-handling/error-handling.md
```diff
@@ -0,0 +1,48 @@
...
+| Not Found            | 2           | Error indicate that given resource cannot be found and further processing is impossible     |
+| Not Unique           | 3           | Error indicates that given resource is not unique                                           |
+| Tenant Not Found     | 4           | Internal Tenant not found in director                                                       |
+| InvalidData          | 5           | The input data is invalid, error description should be delivered in error message           | 
+| ConstraintViolation  | 6           | Error which indicate that this operation can't happen because referenced resource not exist |
+| Insufficient Scopes  | 7           | Error which indicate insufficient scopes                                                    |
+
+NotFoundError can be triggered in mutation like `addPackage` to not existing application.
+
+ConstraintViolation can be replaced with `NotFound`  error.
```
crabtree, line 37: So should it be replaced? I do not like having two different ways of handling the same error, it would be confusing.

components/director/pkg/customerrors/errors.go
```diff
@@ -0,0 +1,68 @@
...
+func (e Error) Is(err error) bool {
+	if customErr, ok := err.(Error); ok {
+		return e.errorCode == customErr.errorCode
+	} else {
+		return false
+	}
+}
+
+//TODO: We can provide more constructors for different types and use format prepared messages in contructors.
+func NewNotUniqueErr() Error {
```
crabtree, line 40: I think this is the way to go, every error should be constructed by its constructor function.

components/director/pkg/customerrors/builder.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (builder *Builder) InvalidData(msg string) ArgsStep {
+	builder.args["reason"] = msg
+	builder.message = "Invalid input"
+	builder.errorType = InvalidData
+	return builder
+}
+
+func (builder *Builder) NotUnique(msg string) ArgsStep {
```
kjaksik, line 63: Maybe you can add ResourceType to this error too?

components/director/pkg/customerrors/consts.go
```diff
@@ -0,0 +1,27 @@
...
+const (
+	UnhandledError      ErrorType = 10
+	InternalError       ErrorType = 11
+	NotFound            ErrorType = 20
+	NotUnique           ErrorType = 21
+	InvalidData         ErrorType = 22
+	InsufficientScopes  ErrorType = 23
+	ConstraintViolation ErrorType = 24
+	TenantNotFound      ErrorType = 25
+	TenantNotExist      ErrorType = 26
```
kjaksik, line 15: There has been a small update to the tenant errors, please rename it to `TenantIsRequired`

components/director/pkg/customerrors/builder.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+type Builder struct {
+	args      map[string]string
+	errorType ErrorType
+	message   string
+	parentErr error
+}
+
+func NewBuilder() TypeStep {
```
pkosiec, line 38: IMHO this builder shouldn't be exported. We should agree on a single approach to create new errors.

dbadura: You mean `Builder` struct?

dbadura: In some cases, like in `sql` helpers, it's not possible to create `NotFoundErr`, because we don't know what exactly not found.

components/director/pkg/customerrors/sql_mapper.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	switch pqerr.Code {
+	case persistence.UniqueViolation:
+		return NewBuilder().WithStatusCode(NotUnique).Wrap(err).Build()
+	case persistence.ForeignKeyViolation:
+		return NewBuilder().WithStatusCode(ConstraintViolation).WithMessage(pqerr.Detail).Wrap(err).Build()
+	}
+
+	return NewBuilder().InternalError("").Wrap(err).Build()
```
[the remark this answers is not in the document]
tgorgol: Also, as seen in above example, we're not validating if ID is a valid UUID in input validators, so the DB will return an error instead. This error is treated as internal server error while it's clearly user error, we should either catch it here or adjust validators I think.

components/director/pkg/customerrors/presenter.go
```diff
@@ -0,0 +1,54 @@
...
+
+		return &gqlerror.Error{
+			Message:    customErr.Error(),
+			Path:       graphql.GetResolverContext(ctx).Path(),
+			Extensions: map[string]interface{}{"error_code": customErr.errorCode, "error": customErr.errorCode.String()},
+		}
+	}
+	errID := p.uuidService.Generate()
+	fmt.Printf("Not handled error yet, errorID:%s, %v\n", errID, err)
+	return &gqlerror.Error{
```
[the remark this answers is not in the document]
dbadura: I will handle those error in old way but will logged the message to console.

components/director/internal/repo/create.go
```diff
@@ -5,10 +5,9 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/kyma-incubator/compass/components/director/pkg/apperrors"
+	"github.com/kyma-incubator/compass/components/director/pkg/customerrors"
```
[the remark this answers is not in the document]
dbadura: It's temporary package name.

components/director/pkg/customerrors/builder.go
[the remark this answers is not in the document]
dbadura: Ok, I will hide `Builder`

docs/internal/proposals/error-handling/error-handling.md
```diff
@@ -0,0 +1,49 @@
...
+* panics
+* postgreSQL, except errors like NotFound, NotUnique
+
+To deal with the errors we introduced custom errors, which contains error codes.
+
+## Proposed custom errors and theirs error codes (the error codes can change):
+
+| Error type           | Error code  |                            Description                                                            |
+|----------------------|-------------|---------------------------------------------------------------------------------------------------|
+| UnhandledError       | 10          | Error which cannot be handled in director and also is handled in old way                          |
```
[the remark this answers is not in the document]
dbadura: I will remove this error.

components/director/pkg/customerrors/sql_mapper.go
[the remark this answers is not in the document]
dbadura: I think we should catch it earlier.

PR #1394, opened 2020-05-26
Certificate chain order changed

akgalwas:
**Description**

- Certificate chain order changed in Connector

### DIFF
--- a/components/connector/internal/certificates/service.go
+++ b/components/connector/internal/certificates/service.go
@@ -92,10 +92,10 @@ func (svc *certificateService) encodeCertificates(rawCaCertificate, rawClientCer
 			return EncodedCertificateChain{}, err
 		}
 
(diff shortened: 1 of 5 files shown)


components/connector/internal/certificates/service.go
```diff
@@ -92,7 +92,7 @@ func (svc *certificateService) encodeCertificates(rawCaCertificate, rawClientCer
 			return EncodedCertificateChain{}, err
 		}
 
-		caCrtBytes = svc.createCertChain(rootCABytes, caCrtBytes)
+		caCrtBytes = svc.createCertChain(caCrtBytes, rootCABytes)
 	}
```
koala7659, line 96: Very small suggestion (you can safely ignore it). Maybe it could be be worthy to rename createCertChain function and its arguments to reflect that it adds next derived certificate into the chain? For example appendNewDerivedCertificate(derivedCrt, chain []byte).

Issue #1384, opened 2020-05-24
Enhance tenants query with information about tenant usage


PK85:
**Description**

We'd like to know if given tenant has been already used by customer.

**Reasons**

Director has a lot of tenants and MPS Operator likes to see tenants that are in use.

**AC**

- enhance *tenants* query with information isInUse true/false

**Technical Solution hint**
- check if tenant has not empty label defitions. If at least one exist it means tenant is in use.

kwiatekus:
Hey.
Filtering of tenants in use should be reflected in the UI.
It seems like a tiny change [here](https://github.com/kyma-project/console/blob/master/compass/src/config/luigi-config/helpers/navigation-helpers.js#L17) - therefore we wont create a dedicated issue for the changes in the ui.
Lets Fix it top to bottom under this issue and include PR to compass UI here.

PK85:
Hi @kwiatekus, please take into account that your tiny change doesn't include the required change in the search details, where user should see inUse, and if typing then all (inUse on top)

pkosiec:
Hi there, we discussed it offline and the final field name will be `initialized. Cheers!

PR #1404, opened 2020-05-28
Add "initialized" metadata field for Tenants query in Director

pkosiec:
**Description**

- Add inUse metadata for Tenants query in Director

**Related issue(s)**
#1384


tests/director/api/tenants_api_test.go
```diff
@@ -28,12 +28,25 @@ func TestQueryTenants(t *testing.T) {
...
+func assertTenants(t *testing.T, in []*graphql.Tenant, actual []*graphql.Tenant) {
+	for _, inTnt := range in {
+		found := false
+		for _, actTnt := range actual {
+			if inTnt.ID != actTnt.ID {
+				continue
+			}
+			found = true
+
+			assert.Equal(t, inTnt.Name, actTnt.Name)
```
pkosiec, line 44: I didn't check the `inUse` field value, as it could change when we ran tests in different order. I see two options here:
1. not checking the `inUse` value at all 
1. in this test create an App in one of the tenants and check if the value is true (it could be already beause of other tests, but we need to make sure the LabelDefinition is created)

I went for the option No. 1, but of course feel free to disagree with me 🙂

tests/director/go.mod
```diff
@@ -20,8 +20,9 @@ require (
 	github.com/vrischmann/envconfig v1.2.0
 	golang.org/x/crypto v0.0.0-20200210222208-86ce3cb69678 // indirect
 	golang.org/x/net v0.0.0-20200202094626-16171245cfb2
-	golang.org/x/sys v0.0.0-20200212091648-12a6c2dcc1e4 // indirect
 	k8s.io/apimachinery v0.17.3 // indirect
 )
 
 replace gopkg.in/yaml.v2 => gopkg.in/yaml.v2 v2.2.8
+
+replace github.com/kyma-incubator/compass/components/director => github.com/pkosiec/compass/components/director v0.0.0-20200528135802-12a6f2924528
```
pkosiec, line 28: It will be changed in a follow-up PR.

components/director/internal/domain/tenant/repository.go
```diff
@@ -87,13 +95,20 @@ func (r *pgRepository) ExistsByExternalTenant(ctx context.Context, externalTenan
...
-	conditions := repo.Conditions{
-		repo.NewEqualCondition(statusColumn, Active),
+	persist, err := persistence.FromCtx(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "while fetching persistence from context")
 	}
 
-	err := r.listerGlobal.ListGlobal(ctx, &entityCollection, conditions...)
+	prefixedFields := strings.Join(str.PrefixStrings(tableColumns, "t."), ", ")
+	query := fmt.Sprintf(`SELECT DISTINCT %s, ld.tenant_id IS NOT NULL AS %s
```
pkosiec, line 104: It was agreed with our UI team to sort the tenants firstly by `inUse` field, and then by `name`.

tests/director/api/tenants_api_test.go
dbadura: What do you think about create 2 additional totally seperate tenants and for one, create runtime or application?

components/director/pkg/str/strings.go
```diff
@@ -38,3 +40,12 @@ func Cast(i interface{}) (string, error) {
 
 	return "", apperrors.NewInvalidStringCastError()
 }
+
+func PrefixStrings(in []string, prefix string) []string {
```
dbadura, line 44: What do you think about using `variadic` arg?
It will be easier nicer to use.

tests/director/api/tenants_api_test.go
kfurgol: Well, I see also option No. 3, adding a new tenant just for that specific use case. But I believe that would be an overkill, so I'm fine with option No. 1

components/director/examples/query-tenants/query-tenants.graphql
```diff
@@ -2,6 +2,8 @@
 query {
   result: tenants {
     id
+    internalID
```
kfurgol, line 5: I just wanted to double-check that it's not a security issue?

components/director/pkg/str/strings.go
pkosiec: I'm not convinced it'll be nicer. We use slice of strings as an input all the times anyway, and also we have to return a slice, so the input and output format is unified.

components/director/internal/domain/tenant/fixtures_test.go
```diff
@@ -82,3 +111,13 @@ func newGraphQLTenant(id, internalID, name string) *graphql.Tenant {
 		Name:       str.Ptr(name),
 	}
 }
+
+func fixModelBusinessTenantMapping() model.BusinessTenantMapping {
+	return model.BusinessTenantMapping{
+		ID:             uuid.New().String(),
+		Name:           "tenant",
+		ExternalTenant: "external-tenant",
+		Provider:       "SAP",
```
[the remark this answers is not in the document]
pkosiec: It's basically moved from other file, but you're right, I will change the provider 😄

components/director/examples/query-tenants/query-tenants.graphql
pkosiec: Well, we do have this field in GraphQL type, so everyone can query it. What I did is including this field in examples, because our approach is to always include all available fields for all examples.

tests/director/api/tenants_api_test.go
```diff
@@ -28,12 +28,25 @@ func TestQueryTenants(t *testing.T) {
...
 	t.Log("Check if tenants were received")
 
-	for _, tenant := range defaultTenants {
-		assert.Contains(t, output, tenant)
-	}
+	assertTenants(t, defaultTenants, output)
 	saveExample(t, getTenantsRequest.Query(), "query tenants")
 }
 
+func assertTenants(t *testing.T, in []*graphql.Tenant, actual []*graphql.Tenant) {
```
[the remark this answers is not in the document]
pkosiec: Yup, I'll move it to `assertions.go`.

components/director/internal/domain/tenant/repository.go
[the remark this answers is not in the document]
pkosiec: Yes, it was agreed to work like this. However, you made a valid point - maybe it would be better to call it "initialized", not "in use". What do you think?

tests/director/api/tenants_api_test.go
pkosiec: That's good idea, I will create 2 tenants (initialized and not initialized), and then I will use them for tests.

components/director/internal/domain/tenant/repository.go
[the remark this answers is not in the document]
kfurgol: IMO `initialized` would work better, since I've associated `inUse` immediately with our `tenants in use` metrics, and it's not the same, so I'd go with that!

PR #1418, opened 2020-06-09
Extend KEB internal Avs to support custom overrides

fhivemind:
Enables Tags and Service configuration overrides for internal Avs configuration of KEB component.

**Description**

- Extend Avs to support Tags and Service configs for internal and external testers
- Handle HTTP requests and responses with defined Tags and Services configs
- Add implementation tests

### DIFF
--- a/components/kyma-environment-broker/internal/avs/client.go
+++ b/components/kyma-environment-broker/internal/avs/client.go
@@ -12,12 +12,12 @@ import (
 
 type clientHolder struct {
 	httpClient        *http.Client
-	avsConfig         *Config
+	avsConfig         Config
 	clientInitialized uint32
 	mu                sync.Mutex
 }
 
-func newClientHolder(avsConfig *Config) *clientHolder {
+func newClientHolder(avsConfig Config) *clientHolder {
 	return &clientHolder{
 		avsConfig: avsConfig,
 	}
--- a/components/kyma-environment-broker/internal/avs/client.go
+++ b/components/kyma-environment-broker/internal/avs/client.go
@@ -12,12 +12,12 @@ import (
 
 type clientHolder struct {
 	httpClient        *http.Client
-	avsConfig         Config
+	avsConfig         *Config
 	clientInitialized uint32
 	mu                sync.Mutex
 }
 
-func newClientHolder(avsConfig Config) *clientHolder {
+func newClientHolder(avsConfig *Config) *clientHolder {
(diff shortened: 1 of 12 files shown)


components/kyma-environment-broker/internal/avs/client.go
```diff
@@ -12,12 +12,12 @@ import (
 
 type clientHolder struct {
 	httpClient        *http.Client
-	avsConfig         Config
+	avsConfig         *Config
```
piotrmiskiewicz, line 15: why did you change this to a pointer?

fhivemind: As described in commit details: 
Switches to pointer-type avs configs across KEB, as in the future the configuration might get updated at runtime (provisioning, deprovisoning) and the changes should be properly delegated.

piotrmiskiewicz: The configuration is set by environment variables. There is no any runtime configuration change right now. The change to a pointer is not enough. What about synchronization? you are sharing data without any locking? is it safe? Let's prepare for a runtime config change if there is an issue for that.

fhivemind: I agree that this functionality is currently underused, as there are no runtime config changes. However, this allows a more robust Evaluation Assistant class-specific configuration. In this case, we wouldn't have to rely on the singular classes but use a more OOP approach (possibility of not only using two EvalAssistants). The runtime changes should then be reflected on all the inherited EvalAssistants, which this is trying to accomplish. To ensure consistency, mutex implementation would be required.
Regarding safety, there is no way of making read-only objects or creating constant pointer passes in Go, hence any implementation using this approach is by its nature -- unsafe. We can switch from using current avs config object to a getter-setter model, rendering the object changes safer, however, this might get a bit longer in case of object extensions.

In any case, thanks for pointing out the possible twisters. What do you think would be the best approach to follow, extending this further or falling back to static config objects?

piotrmiskiewicz: I'd like to see a use case of change the config in the runtime before the implementation. Let's do not implement something which is not defined.
Passing a struct as a value makes a copy which prevents runtime changes. It is something like passing value object and it is ok.

fhivemind:
I have, and in either case, these configurations just extend the current model -- the provisioning or de-provisioning logic didn't change.

PR #1425, opened 2020-06-16
Add a purpose for the SKRs

franpog859:
**Description**

- add `purpose` field to the `provisionRuntime` Provisioner mutation
- add `purpose` column do the `gardener_config` Provisioner table
- update KEB to use the `purpose` filed in the mutation
- define default `purpose` which can be overridden


components/provisioner/pkg/gqlschema/schema.graphql
```diff
@@ -150,6 +151,7 @@ input GardenerConfigInput {                   # Gardener project in which the cl
     machineType: String!                            # Type of node machines, varies depending on the target provider
     region: String!                                 # Region in which the cluster is created
     provider: String!                               # Target provider on which to provision the cluster (Azure, AWS, GCP)
+    purpose: String                                 # Purpose is the purpose class for this cluster
```
koala7659, line 154: Just a question:  I'm not sure, why that this field is nullable. Do we assume KEB can send us the configuration without setting the purpose field?

components/kyma-environment-broker/internal/dto.go
```diff
@@ -27,7 +27,15 @@ func (p ProvisioningParameters) IsEqual(input ProvisioningParameters) bool {
...
 		return false
 	}
-	if !reflect.DeepEqual(p.Parameters, input.Parameters) {
+
+	// TODO: If there are already not resolved operations in the db
+	//       KEB will raise an error "provisioning operation already
+	//       exist" after updating the image on the environments.
+	//       Revert this check after some time, when every operation
+	//       has at least the default Gardener Shoot Purpose to:
+	//       if !reflect.DeepEqual(p.Parameters, input.Parameters) {
```
majakurcius, line 36: We should add a follow-up task for that. Do we have one?

components/provisioner/pkg/gqlschema/schema.graphql
franpog859: KEB should always send this field. This is GraphQL though and the Gardener has a default value for it so I think it's good to leave it optional. This approach also would not provide any problems with synchronization because both calls - with and without the `purpose` field will work

components/kyma-environment-broker/internal/dto.go
franpog859: Sure! I'll create one :D I wanted to hear the Gopher's opinion on this solution first

piotrmiskiewicz:
Please, look how we configure default Kubernetes version:
in the deployment: `APP_PROVISIONING_KUBERNETES_VERSION`
the config in the code: internal/process/provisioning/input/builder.go (type Config)

In the InputBuidlerFactory, configure the default Purposem  just like Kubernetes version:
```provisionInput.ClusterConfig.GardenerConfig.KubernetesVersion = f.config.KubernetesVersion```
The line above copies the default value from KEB configuration into the builder.

In the input.go, there is a method applyProvisioningParameters which applies values set by the client if set or use default one if not set (if the value provided by the client is empty).

File: components/kyma-environment-broker/internal/dto.go (hunk 2 above)
Comment (piotrmiskiewicz) [reply components/kyma-environment-broker/internal/dto.go:36]: should not be changed, discussed

File: components/kyma-environment-broker/internal/provider/azure_provider.go [hunk 3]
```diff
@@ -19,6 +23,7 @@ func (p *AzureInput) Defaults() *gqlschema.ClusterConfigInput {
 			MachineType:    "Standard_D8_v3",
 			Region:         DefaultAzureRegion,
 			Provider:       "azure",
+			Purpose:        &DefaultAzurePurpose,
```
piotrmiskiewicz, line 26: do not specify the purpose here, the line is not needed

components/kyma-environment-broker/internal/provider/gcp_provider.go
```diff
@@ -21,6 +25,7 @@ func (p *GcpInput) Defaults() *gqlschema.ClusterConfigInput {
 			MachineType:    "n1-standard-4",
 			Region:         DefaultGCPRegion,
 			Provider:       "gcp",
+			Purpose:        &DefaultGCPPurpose,
```
piotrmiskiewicz, line 28: do not specify the purpose here, the line is not needed

components/kyma-environment-broker/internal/provider/azure_provider.go
franpog859: Done!

components/kyma-environment-broker/internal/provider/gcp_provider.go
franpog859: Done!

PR #1465, opened 2020-06-29
Add LicenceType field to GardenerConfig

jasiu001:
**Description**

- Add functionality to annotate shoot resource during creating SKR with licence type field


components/provisioner/internal/gardener/provisioner.go
```diff
@@ -63,6 +63,9 @@ func (g *GardenerProvisioner) ProvisionCluster(cluster model.Cluster, operationI
 
 	annotate(shootTemplate, operationIdAnnotation, operationId)
 	annotate(shootTemplate, runtimeIdAnnotation, cluster.ID)
+	if cluster.LicenceType != "" {
```
akgalwas, line 66: We could simplify this design a bit. 
Some facts: 
- LicenceType is a field that is specific to GardenerConfig
- LicenseType annotation could be added in ToShootTemplate
- Annotations for OperationID and RuntimeID are added in ProvisionCluster method since those values are not available in ToShootTemplate

My proposal is to add an annotation in ToShootTemplate implementation. It will have the following implications:
- No need to change model.Cluster 
- No need to add GetLicenceType method on ProviderConfiguration
- No need to change gardener.ProvisionCluster method.

components/provisioner/internal/model/provider.go
```diff
@@ -8,4 +8,5 @@ import (
 type ProviderConfiguration interface {
 	ToHydroformConfiguration(credentialsFileName string) (*types.Cluster, *types.Provider, error)
 	ToShootTemplate(namespace string, accountId string, subAccountId string) (*gardener_types.Shoot, error)
+	GetLicenceType() string
```
akgalwas, line 11: As mentioned in the previous note it is not needed.

components/provisioner/internal/model/gardener_config.go
```diff
@@ -31,6 +33,7 @@ type GardenerConfig struct {
 	MachineType            string
 	Provider               string
 	Purpose                *string
+	LicenceType            string
```
Szymongib, line 36: The nullable field cannot be read to `string` if the value is `NULL`.

Issue #1021, opened 2020-03-23
Custom Headers and Query Parameters cannot contain '-' character


Szymongib:
**Description**

When registering Application to Director user can specify `AdditionalHeaders` and `AdditionalQueryParams` in `DefaultInstanceAuth`.
However, the header and query parameter names can not contain `-` character which in case of headers is quite common, for example, `Content-Type` header.

**Expected result**

Headers with `-` can be registered by Director API.

**Actual result**

Query structure is invalid and therefore rejected by the GraphQL server.

**Steps to reproduce**

Try to register Application with `AdditionalHeaders` that contain `-` like `Content-Type`.

**Troubleshooting**

**Ideas/Solutions**
- make `QueryParams` and `AdditionalHeaders` as JSON Object (String)

**AC**
After change
- adjust Connectivity Adapter
- adjust Agent

PK85:
@kfurgol has checked that issue, and looks that we have design API issue here, as HTTPHeaders are a map in graphQL and that '-' character is not allowed. @crabtree @pkosiec how we can enable those chars here?

akrausesap:
Unfortunately Qualtrics uses such a header: https://api.qualtrics.com/docs/api-key-authentication

PR #1469, opened 2020-06-30
Headers and QueryParams refactoring

crabtree:
**Description**

- Introduce new properties for handling serialized headers and query parameters
- Deprecate existing input properties for headers and query parameters

Changes in the connectivity-adapter component will be introduced in the other PR.

**Related issue(s)**

See also #1021


components/director/pkg/graphql/http_headers.go
```diff
@@ -28,3 +29,25 @@ func (y HttpHeaders) MarshalGQL(w io.Writer) {
...
+
+func (y *HttpHeadersSerialized) Unmarshal() (map[string][]string, error) {
+	var data map[string][]string
+	if y == nil {
+		return data, nil
+	}
+
+	err := json.Unmarshal([]byte(*y), &data)
+
+	return data, err
```
kjaksik, line 43: I think it would be nice to add one of our custom errors here, eg. `InvalidData`. What do you think about that?

components/director/pkg/graphql/query_params.go
```diff
@@ -28,3 +29,25 @@ func (y QueryParams) MarshalGQL(w io.Writer) {
...
+
+func (y *QueryParamsSerialized) Unmarshal() (map[string][]string, error) {
+	var data map[string][]string
+	if y == nil {
+		return data, nil
+	}
+
+	err := json.Unmarshal([]byte(*y), &data)
+
+	return data, err
```
kjaksik, line 43: I think it would be nice to add one of our custom errors here, eg. `InvalidData`. What do you think about that?

PR #1470, opened 2020-06-30
Update KEB chart Avs tester tag config parsing

fhivemind:
**Description**

Changes the way how Avs configuration is being defined and parsed within KEB chart. This was changed due to Kyma Operator inability to override objects. 

Now, KEB Avs configuration uses multiline string which is possible to override.


fhivemind:
Compass component removed from project, closing this.  :(

PR #1480, opened 2020-07-02
Connectivity adapter use serialized Headers and QueryParams API

crabtree:
**Description**

- Connectivity adapter use serialized Headers and QueryParams API

**Related issue(s)**

See also #1021


components/connectivity-adapter/internal/appregistry/service/converter_test.go
```diff
@@ -12,6 +12,9 @@ import (
 )
 
 func TestConverter_DetailsToGraphQLCreateInput(t *testing.T) {
+	additionalQueryParamsSerialized := graphql.QueryParamsSerialized("{\"q1\":[\"a\",\"b\"],\"q2\":[\"c\",\"d\"]}")
```
dbadura, line 15: You can use backtics to escape the json.

PR #1488, opened 2020-07-06
Remove dependency on local postgresql installation in compass run script

### LINKED ISSUE #1487: Compass run script depends on local PostgreSQL installation
## Description

I tried running Compass recently and I saw that in order to run it successfully, I had to install **PostgreSQL** locally because some commands from the Compass `run.sh` script were using PostgreSQL tools.

And since the script is utilizing **Docker**, I thought that it would be a great idea to further extend Docker usage to remove the local dependency on PostgreSQL.

danielgospodinow:
**Description**

- Run **PostgreSQL** initialization scripts directly in the Docker container to remove the dependency on a local PosgreSQL installation. 

**Related issue(s)**
Resolves #1487

### DIFF
--- a/components/director/run.sh
+++ b/components/director/run.sh
@@ -34,6 +34,7 @@ DB_USER="postgres"
 DB_PWD="pgsql@12345"
 DB_NAME="compass"
 DB_PORT="5432"
+DB_HOST="127.0.0.1"
 
 function cleanup() {
     echo -e "${GREEN}Cleanup Postgres container${NC}"
@@ -44,17 +45,19 @@ trap cleanup EXIT
 
 echo -e "${GREEN}Start Postgres in detached mode${NC}"
 docker run -d --name ${POSTGRES_CONTAINER} \
+            -e POSTGRES_HOST=${DB_HOST} \
             -e POSTGRES_USER=${DB_USER} \
             -e POSTGRES_PASSWORD=${DB_PWD} \
             -e POSTGRES_DB=${DB_NAME} \
-            -p 5432:5432 \
+            -e POSTGRES_PORT=${DB_PORT} \
+            -p ${DB_PORT}:${DB_PORT} \
             postgres:${POSTGRES_VERSION}
 
 set +e
 echo '# WAITING FOR CONNECTION WITH DATABASE #'
 for i in {1..30}
 do
-    pg_isready -U "$DB_USER" -h 127.0.0.1 -p "$DB_PORT" -d "$DB_NAME"
+    docker exec ${POSTGRES_CONTAINER} pg_isready -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
     if [ $? -eq 0 ]
     then
         dbReady=true
@@ -72,10 +75,10 @@ set -e
 
 echo -e "${GREEN}Populate DB${NC}"
 
-PGPASSWORD=pgsql@12345 psql -U ${DB_USER} -h 127.0.0.1 -f <(cat ${ROOT_PATH}/../schema-migrator/migrations/director/*.up.sql) ${DB_NAME}
-
-PGPASSWORD=pgsql@12345 psql -U ${DB_USER} -h 127.0.0.1 -f <(cat ${ROOT_PATH}/../schema-migrator/seeds/director/*.sql) ${DB_NAME}
-
+cat ${ROOT_PATH}/../schema-migrator/migrations/director/*.up.sql | \
+    docker exec -i ${POSTGRES_CONTAINER} psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
+cat ${ROOT_PATH}/../schema-migrator/seeds/director/*.sql | \
+    docker exec -i ${POSTGRES_CONTAINER} psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
 
 if [[  ${SKIP_APP_START} ]]; then
         echo -e "${GREEN}Skipping starting application${NC}"


dbadura:
Hmm, something doesn't work, I got the following logs:
```
Start Postgres in detached mode
242c8cdf42b1a97e6ef13e0270a202c75c5f8fa0b1a10d3832571d5b3b0c4a25
# WAITING FOR CONNECTION WITH DATABASE #
: - no response
: - accepting connections
Populate DB
psql: could not connect to server: Connection refused
        Is the server running on host "127.0.0.1" and accepting
        TCP/IP connections on port 5432?
Cleanup Postgres container
test-postgres
```
I added `sleep 0.5` before populating the db and it helped.

dbadura:
Don't forget to sign the CLA.

danielgospodinow:
> Hmm, something doesn't work, I got the following logs:
> 
> ```
> Start Postgres in detached mode
> 242c8cdf42b1a97e6ef13e0270a202c75c5f8fa0b1a10d3832571d5b3b0c4a25
> # WAITING FOR CONNECTION WITH DATABASE #
> : - no response
> : - accepting connections
> Populate DB
> psql: could not connect to server: Connection refused
>         Is the server running on host "127.0.0.1" and accepting
>         TCP/IP connections on port 5432?
> Cleanup Postgres container
> test-postgres
> ```
> 
> I added `sleep 0.5` before populating the db and it helped.

Hmm, strange, I haven't encountered a similar issue. Neither with the new changes nor without them.

kjaksik:
The same for me
```
Start Postgres in detached mode
795a0c3931819f044e284dacc97ae86aae26a84dab512e4088f0dbfc6aca709a
# WAITING FOR CONNECTION WITH DATABASE #
: - no response
: - accepting connections
Populate DB
psql: could not connect to server: Connection refused
	Is the server running on host "127.0.0.1" and accepting
	TCP/IP connections on port 5432?
Cleanup Postgres container
test-postgres
```

dbadura:
> > Hmm, something doesn't work, I got the following logs:
> > ```
> > Start Postgres in detached mode
> > 242c8cdf42b1a97e6ef13e0270a202c75c5f8fa0b1a10d3832571d5b3b0c4a25
> > # WAITING FOR CONNECTION WITH DATABASE #
> > : - no response
> > : - accepting connections
> > Populate DB
> > psql: could not connect to server: Connection refused
> >         Is the server running on host "127.0.0.1" and accepting
> >         TCP/IP connections on port 5432?
> > Cleanup Postgres container
> > test-postgres
> > ```
> > 
> > 
> > I added `sleep 0.5` before populating the db and it helped.
> 
> Hmm, strange, I haven't encountered such an issue. Neither with the new changes nor without them.



> > Hmm, something doesn't work, I got the following logs:
> > ```
> > Start Postgres in detached mode
> > 242c8cdf42b1a97e6ef13e0270a202c75c5f8fa0b1a10d3832571d5b3b0c4a25
> > # WAITING FOR CONNECTION WITH DATABASE #
> > : - no response
> > : - accepting connections
> > Populate DB
> > psql: could not connect to server: Connection refused
> >         Is the server running on host "127.0.0.1" and accepting
> >         TCP/IP connections on port 5432?
> > Cleanup Postgres container
> > test-postgres
> > ```
> > 
> > 
> > I added `sleep 0.5` before populating the db and it helped.
> 
> Hmm, strange, I haven't encountered such an issue. Neither with the new changes nor without them.

What OS do you use?

danielgospodinow:
I'm runnning macOS. Could this be the issue?

kjaksik:
> I'm runnning macOS. Could this be the issue?
We are running macOS too so it shouldn't be the issue then, we will investigate it further

danielgospodinow:
I'll also try to find out what could be the issue.

dbadura:
I noticed that when you exec `pg_isready` inside container you will get status code 0 in second run, while running locally you will get status code equal to 0 at 3 run:
My changes
```bash
    docker exec ${POSTGRES_CONTAINER} bash -c pg_isready -U "$DB_USER" -h "$DB_HOST" -p "$DB_PORT" -d "$DB_NAME"
    EXIT_CODE=$?
    echo "Exit Code: $EXIT_CODE"
    pg_isready -U "$DB_USER" -h "$DB_HOST" -p "$DB_PORT" -d "$DB_NAME"
    LOCAL_EXIT_CODE=$?
    echo "Local Code: $LOCAL_EXIT_CODE"
    if [ $EXIT_CODE -eq 0 ]
          ....
```

```
Exit Code: 2
127.0.0.1:5432 - no response
Local Code: 2
/var/run/postgresql:5432 - accepting connections
Exit Code: 0
127.0.0.1:5432 - no response
Local Code: 2
```

danielgospodinow:
Good observations 👍 but can this really impact something? Because this command is executed in a loop that is broken only when a success status code ( 0 ) is received which implies that the PostgreSQL in the container is up and running correctly.

Also, I didn't manage to reproduce this issue on my machine. 😞

danielgospodinow:
I just caught a slight bug that I've made and I just fixed it. I've used `bash -c 'pg_isready -U "$DB_NAME" ...` but in fact, no such environment variable `$DB_NAME` existed in the container's environment variables.

Can you try and run the script again? Maybe this could've been the issue.

components/director/run.sh
```diff
@@ -56,7 +58,7 @@ echo '# WAITING FOR CONNECTION WITH DATABASE #'
 for i in {1..30}
 do
     # 'exit $?' is a hack to return correct exit codes when docker exec is run with tty (-t).
-    docker exec -it ${POSTGRES_CONTAINER} bash -c 'pg_isready -U "$DB_USER" -h "$DB_HOST" -p "$DB_PORT" -d "$DB_NAME"; exit $?'
+    docker exec -it ${POSTGRES_CONTAINER} bash -c 'pg_isready -U "$POSTGRES_USER" -h "$POSTGRES_HOST" -p "$POSTGRES_PORT" -d "$POSTGRES_DB"; exit $?'
```
dbadura, line 61:
```suggestion
    docker exec ${POSTGRES_CONTAINER} pg_isready -U "$DB_USER" -h "$DB_HOST" -p "$DB_PORT" -d "$DB_NAME"
```

components/director/run.sh
```diff
@@ -75,9 +77,9 @@ set -e
 echo -e "${GREEN}Populate DB${NC}"
 
 cat ${ROOT_PATH}/../schema-migrator/migrations/director/*.up.sql | \
-    docker exec -i ${POSTGRES_CONTAINER} psql -U "$DB_USER" -h "$DB_HOST" -p "$DB_PORT" -d "$DB_NAME"
+    docker exec -i ${POSTGRES_CONTAINER} bash -c 'psql -U "$POSTGRES_USER" -h "$POSTGRES_HOST" -p "$POSTGRES_PORT" -d "$POSTGRES_DB"'
```
dbadura, line 80: I think it's not needed to run this command through `bash -c`.
It can be as it was earlier.

danielgospodinow: Yes, the `bash -c` is something that I don't like either, but it's needed for the hack in order to return the corrent status code of the `docker exec` command. Otherwise, without this hack, `docker exec` always returns 0 regardless of whether the command in the container failed or not.

dbadura: Are you sure that `docker exec` always return 0?
I tried following code in loop:
```bash
    docker exec ${POSTGRES_CONTAINER} ls /bins
    EXIT_CODE=$?
    echo $EXIT_CODE
```
And it always return 2.

```bash
    docker exec ${POSTGRES_CONTAINER} ls /bin
    EXIT_CODE=$?
    echo $EXIT_CODE
```
returns 0

danielgospodinow: You're actually right. It was my fault, It seems that I've mistaken about this status code.

I've refactored as you suggested, can you test if the script is working okay?

components/director/run.sh
```diff
@@ -72,10 +75,10 @@ set -e
 
 echo -e "${GREEN}Populate DB${NC}"
 
-PGPASSWORD=pgsql@12345 psql -U ${DB_USER} -h 127.0.0.1 -f <(cat ${ROOT_PATH}/../schema-migrator/migrations/director/*.up.sql) ${DB_NAME}
-
-PGPASSWORD=pgsql@12345 psql -U ${DB_USER} -h 127.0.0.1 -f <(cat ${ROOT_PATH}/../schema-migrator/seeds/director/*.sql) ${DB_NAME}
-
+cat ${ROOT_PATH}/../schema-migrator/migrations/director/*.up.sql | \
+    docker exec -i ${POSTGRES_CONTAINER} psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
```
dbadura, line 79: I would remove `interactive` mode

components/director/run.sh
```diff
@@ -72,10 +75,10 @@ set -e
...
+cat ${ROOT_PATH}/../schema-migrator/seeds/director/*.sql | \
+    docker exec -i ${POSTGRES_CONTAINER} psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
```
dbadura, line 81: I would remove `interactive` mode

components/director/run.sh
dbadura: Works 😃

components/director/run.sh
danielgospodinow: I've considered that, but I've left it there because of the pipe before the `docker exec`.

https://stackoverflow.com/a/43099210 
>Or you can change it to -t if you want TTY support but don't have it available on the input device. Do this for color formatting of the output in your logs, or for when you later attach to the container with a proper terminal.

dbadura: Yes you're right :)

PR #1504, opened 2020-07-27
Parameterize tenant fetcher job

pankrator:
**Parameterize tenant fetcher job**

- Allow the tenant fetcher to be more precisely configured (query param names, event response fields)
- Tenant fetcher job chart is adjusted that for each job configuration - a new k8s CronJob is created and will be executed on schedule

The default values in the `Values.yaml` of the helm chart are commented, because we don't know how to preconfigure a job. It is an example of how a job can be configured.
I accept any ideas on how to make the helm chart better, as the installer overrides has limitations about overriding arrays and thus limiting the ideas.


components/director/internal/tenantfetcher/service.go
```diff
@@ -59,11 +71,24 @@ func (s Service) SyncTenants() error {
...
 	}
 
+	deleteTenantsMap := make(map[string]model.BusinessTenantMappingInput)
+	for _, ct := range tenantsToDelete {
+		deleteTenantsMap[ct.ExternalTenant] = ct
+	}
+
+	for i := len(tenantsToCreate) - 1; i >= 0; i-- {
+		if _, found := deleteTenantsMap[tenantsToCreate[i].ExternalTenant]; found {
+			tenantsToCreate = append(tenantsToCreate[:i], tenantsToCreate[i+1:]...)
```
NickyMateev, line 88: Is it possible to have a delete event before a create event which would mean that we would still want to have the tenant created (kept in the `tenantsToCreate` map)?

components/director/internal/tenantfetcher/service_test.go
```diff
@@ -61,39 +62,103 @@ func TestService_SyncTenants(t *testing.T) {
 			},
 			APIClientFn: func() *automock.EventAPIClient {
 				client := &automock.EventAPIClient{}
-				client.On("FetchTenantEventsPage", tenantfetcher.CreatedEventsType, 1).Return(fixTenantEventsResponse(tenantEvents, 3, 1), nil).Once()
-				client.On("FetchTenantEventsPage", tenantfetcher.UpdatedEventsType, 1).Return(fixTenantEventsResponse(tenantEvents, 3, 1), nil).Once()
-				client.On("FetchTenantEventsPage", tenantfetcher.DeletedEventsType, 1).Return(fixTenantEventsResponse(tenantEvents, 3, 1), nil).Once()
+				client.On("FetchTenantEventsPage", tenantfetcher.CreatedEventsType, tenantfetcher.QueryParams{
```
NickyMateev, line 65: This `tenantfetcher.QueryParams` variable is created in-place identically throughout the test file - maybe extract it?

chart/compass/values.yaml
```diff
@@ -212,26 +212,71 @@ global:
...
-      tenantUpdated: "127.0.0.1/events?type=updated"
-    fieldMapping:
-      idField: "id"
-      nameField: "name"
-      discriminatorField: ""
-      discriminatorValue: ""
-    dbPool:
-      maxOpenConnections: 1
-      maxIdleConnections: 1
+    # job1:
```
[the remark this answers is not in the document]
pankrator: The reason is that if there is something in here - it will be taken as a default. The problem is that we don't know what will be the default name of the production jobs, so there are only commented examples of two jobs.

components/director/internal/tenantfetcher/client.go
```diff
@@ -28,25 +31,55 @@ type APIConfig struct {
...
 
+type QueryParams map[string]string
+
+type tenantResponseMapper struct {
+	TotalPagesField   string
+	TotalResultsField string
+	EventsField       string
+}
+
+func (trd *tenantResponseMapper) Remap(v map[string]interface{}) TenantEventsResponse {
```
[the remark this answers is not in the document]
pankrator: yes, sure. I will do it

components/director/internal/tenantfetcher/service.go
pankrator: yes, this is possible. There is also an issue in github on this. This will be fixed in a future improvement as it has not happened in prod and it is not expected to have such situation. Of course `expected` is not something to rely on.

pankrator:
> Is there risk of merging this with the helm charts being updated as they are and breaking the any existing MPS deployments based on previous (unadapted) pipeline configurations/overrides?

Yes there is. Before merging this PR - The configuration/overrides have to be updated and immediately after merge this PR. If the configuration is changed, but the code is not deployed this will break the jobs. Not sure though if when you change the configuration whether the pipeline will be triggered.

PR #1529, opened 2020-08-20
Add logs regarding Application connection flow

nyordanoff:
**Description**

The scope of this PR is to add logs in the client certificate Application pairing [flow](https://github.com/kyma-incubator/compass/blob/master/docs/connector/02-01-connector-service.md) as well as in the flows where Application metadata is added.

- Client certificate pairing
    - Logs in the whole Connector internal GraphQL API regarding one-time token issuing
    - Logs in the whole Connector external GraphQL API regarding Application pairing via client certificate/certificate revocation and fetching configuration
- Adding Application metadata
    - Directly through Application registration mutations or registration from application template
    - Explicitly through adding a certain Metadata resource and taking into account all other relatively created resources from that mutation


components/director/internal/domain/api/resolver.go
```diff
@@ -81,24 +82,30 @@ func (r *Resolver) AddAPIDefinitionToPackage(ctx context.Context, packageID stri
 	}
 	defer r.transact.RollbackUnlessCommitted(tx)
 
+	log.Infof("Adding APIDefinition to package %s", packageID)
+
 	ctx = persistence.SaveToContext(ctx, tx)
 
 	convertedIn, err := r.converter.InputFromGraphQL(&in)
 	if err != nil {
+		log.Error("Error occurred while converting GraphQL input to APIDefinition. ", err.Error())
```
KirilKabakchiev, line 91: in director there is https://github.com/kyma-incubator/compass/blob/69ec36ed414a25c7f3a706948efcad19eeb9031b/components/director/cmd/director/main.go#L144 error presenter that i think already logs some errors so i suspect you might be double logging all the errors right now - could you double check on that ?

components/connector/internal/api/token_resolver.go
```diff
@@ -28,28 +28,28 @@ func NewTokenResolver(tokenService tokens.Service) TokenResolver {
 }
 
 func (r *tokenResolver) GenerateApplicationToken(ctx context.Context, appID string) (*externalschema.Token, error) {
-	r.log.Infof("Generating token for %s Application...", appID)
+	r.log.Infof("Generating token for Application with id %s", appID)
 
 	token, err := r.tokenService.CreateToken(appID, tokens.ApplicationToken)
 	if err != nil {
-		r.log.Error(err.Error())
+		r.log.Errorf("Error occurred while creating token for Application with id %s : %s ", appID, err.Error())
```
KirilKabakchiev, line 35: i didnt see an error presenter if connector so if it works fine for director error logging i think adding one here as well would make sure sense instead of logging every error separately ? Also it's worth checking if the currently returned errors contain the additional details you added in the log entry and if not you can also enrich the returned errors whereever it seems necessary

chart/compass/values.yaml
```diff
@@ -23,7 +23,7 @@ global:
       version: "PR-1460"
     director:
       dir:
-      version: "PR-1504"
+      version: "PR-1529"
```
KirilKabakchiev, line 26: There are some other resolvers involved in pairing such as applications, application templates. 
Also you've added some logs to subresolvers such as  api and event definitions but not to others like documents, fetch requests, package instance auths ,etc.

KirilKabakchiev, line 26: you've added logs to some of the resolvers involved in the token/cert resolution such as cert resolver and one time token resolver but there but there is also a lot of branching in tenant mapping handler (director rest api) as part of pairing flow that could benefit from info logging

KirilKabakchiev, line 26: pairing is also possible with client credentials (for apps, int systems and runtimes) - i didnt see new logs there

components/director/internal/domain/api/resolver.go
```diff
@@ -81,24 +82,30 @@ func (r *Resolver) AddAPIDefinitionToPackage(ctx context.Context, packageID stri
 	}
 	defer r.transact.RollbackUnlessCommitted(tx)
 
+	log.Infof("Adding APIDefinition to package %s", packageID)
```
KirilKabakchiev, line 85: you only info log that we are starting something is starting (for example to add api definition to package) and you then log when it has succeeded. Maybe you would make sense to have some more info/debug logs in the specific flows as well and not only on the surface. (for example logging what actually happens in the resolver as well as going into the service and logging any specific branching there as well.

Just to give a few examples - if im adding a package with apis right now, there will be no logs that apis are being added as they are added relatively in the service layer and also there will be no logs whether or not fetch requests are added because this happens even deeper in the service layer

components/director/internal/domain/api/resolver.go
```diff
@@ -156,6 +169,9 @@ func (r *Resolver) DeleteAPIDefinition(ctx context.Context, id string) (*graphql
 		return nil, err
 	}
 	defer r.transact.RollbackUnlessCommitted(tx)
+
+	log.Infof("Deleting APIDefinition %s", id)
```
KirilKabakchiev, line 173: maybe in these logs you could clarify that you are logging the id - "Deleting APIDefinition with id %s"

KirilKabakchiev (changes requested):
I think that some important pairing flows are missing. Also there is chance that this PR introduces double error logging. 

Regarding info/debug logs - In my understanding the idea of having info/debug logs is to provide a brief summary of all the steps during a request. Right now when a query/mutation is triggered only the name of the query/mutation is info logged there is no information in the logs of what happened or didnt happen during this query/mutation

components/director/internal/domain/api/resolver.go
nyordanoff: You are right. The `presenter` logs the error so there is no point logging it again in my logs. I double checked it in all places where I have added logs. 

One minor thing: the `presenter` logs the errors on `info` level. Do you think it is okay that way or should we change it to `error` level?

components/director/internal/domain/api/resolver.go
nyordanoff: Agreed.

Actually, most of the existing logs prior to this PR were in this format without explicitly mentioning `with id`. I followed this approach for consistency reasons but as it was pointed out in the other comments - it is way more useful using the `with id` format.

So, there are still some leftover logs with the old format, I will go through all of them and add it.

components/connector/internal/api/certificate_resolver.go
```diff
@@ -52,17 +52,19 @@ func NewCertificateResolver(
 }
 
 func (r *certificateResolver) SignCertificateSigningRequest(ctx context.Context, csr string) (*externalschema.CertificationResult, error) {
+	r.log.Debug("Authenticating the call for signing the Certificate Signing Request.")
+
 	clientId, err := r.authenticator.Authenticate(ctx)
 	if err != nil {
-		r.log.Errorf(err.Error())
+		r.log.Error("Failed authentication during the signing CSR process. ", err.Error())
```
KirilKabakchiev, line 59: If its not possible to address with this PR, please create a separate issue to properly handle errors and remove the double error handling

components/connector/internal/certificates/service.go
```diff
@@ -44,13 +45,17 @@ func NewCertificateService(
...
+	log.Debugf("Successfully loaded the CSR with Common Name %s", subject.CommonName)
 
 	err = svc.checkCSR(csr, subject)
 	if err != nil {
+		log.Errorf("Error occurred while checking the values of the CSR with Common Name %s", subject.CommonName)
 		return EncodedCertificateChain{}, err
 	}
+	log.Debugf("Successfully checked the values of the CSR with Common Name %s", subject.CommonName)
 
 	return svc.signCSR(csr)
```
KirilKabakchiev, line 60: one more debug log here Successfully signed CSR with common name after signCSR before returning

components/connector/internal/tokens/service.go
```diff
@@ -24,6 +25,8 @@ func NewTokenService(store Cache, generator TokenGenerator) *tokenService {
 }
 
 func (svc *tokenService) CreateToken(clientId string, tokenType TokenType) (string, apperrors.AppError) {
+	log.Infof("Generating token for %s with id %s", tokenType, clientId)
```
KirilKabakchiev, line 28: there is a very similar log before entering this method.. 	

r.log.Infof("Generating one-time token for Application with id %s", appID)

components/connector/internal/tokens/service.go
```diff
@@ -34,6 +37,7 @@ func (svc *tokenService) CreateToken(clientId string, tokenType TokenType) (stri
 		ClientId: clientId,
 	}
 
+	log.Debugf("Storing token for %s with id %s in the cache", tokenType, clientId)
```
KirilKabakchiev, line 40: It would be useful if you could log for how long the token is cached (Caching one-time token for X mins

components/director/internal/domain/application/resolver.go
```diff
@@ -267,6 +269,8 @@ func (r *Resolver) RegisterApplication(ctx context.Context, in graphql.Applicati
 
 	ctx = persistence.SaveToContext(ctx, tx)
 
+	log.Infof("Registering Application with name %s", in.Name)
```
KirilKabakchiev, line 272: `Registering Application with name %s", in.Name` here vs `Application with id %s successfully registered", id` below. Be consistent in related logs. If at the first log location you don't have the id yet, atleast log also the name on the second location

components/director/internal/domain/application/service.go
```diff
@@ -250,18 +258,18 @@ func (s *service) Create(ctx context.Context, in model.ApplicationRegisterInput)
 
 	err = s.labelUpsertService.UpsertMultipleLabels(ctx, appTenant, model.ApplicationLabelableObject, id, in.Labels)
```
KirilKabakchiev, line 259: you could log that we are updating labels and dump the new labels

components/director/internal/domain/application/service.go
```diff
@@ -226,17 +229,22 @@ func (s *service) Create(ctx context.Context, in model.ApplicationRegisterInput)
...
+		return "", errors.Wrapf(err, "Error occurred while creating Application with name %s", in.Name)
 	}
 
+	log.Debugf("Ensuring Scenarios label definition exists for Tenant %s", appTenant)
 	err = s.scenariosService.EnsureScenariosLabelDefinitionExists(ctx, appTenant)
 	if err != nil {
 		return "", err
 	}
+
+	log.Debugf("Checking to add Default scenario if enabled for Application with id %s", id)
```
KirilKabakchiev, line 247: well this does not bring any value. I'd be interested to see if the default scenario is actually added for this app or not. Telling that we will check does not bring much value

components/director/internal/domain/apptemplate/resolver.go
```diff
@@ -152,10 +153,12 @@ func (r *Resolver) CreateApplicationTemplate(ctx context.Context, in graphql.App
 		return nil, err
 	}
 
+	log.Infof("Creating an Application Template with name %s", convertedIn.Name)
 	id, err := r.appTemplateSvc.Create(ctx, convertedIn)
 	if err != nil {
 		return nil, err
 	}
+	log.Infof("Successfully created an Application Template with id %s", id)
```
KirilKabakchiev, line 161: log name here as well to be consistent with the log in 156

components/director/internal/domain/apptemplate/resolver.go
```diff
@@ -184,36 +187,44 @@ func (r *Resolver) RegisterApplicationFromTemplate(ctx context.Context, in graph
 
 	ctx = persistence.SaveToContext(ctx, tx)
 
+	log.Infof("Registering an Application from Application Template with name %s", in.TemplateName)
```
KirilKabakchiev, line 190: can we log any details about the applicaton at this point (id, name?)

components/director/internal/domain/apptemplate/resolver.go
```diff
@@ -184,36 +187,44 @@ func (r *Resolver) RegisterApplicationFromTemplate(ctx context.Context, in graph
...
-		return nil, errors.Wrapf(err, "while validating application input from application template [name=%s]", convertedIn.TemplateName)
+		return nil, errors.Wrapf(err, "while validating application input from Application Template with name %s", convertedIn.TemplateName)
 	}
 
 	appCreateInputModel, err := r.appConverter.CreateInputFromGraphQL(appCreateInputGQL)
 	if err != nil {
 		return nil, errors.Wrap(err, "while converting ApplicationFromTemplate input")
 	}
 
+	log.Infof("Creating an Application from Application Template with name %s", in.TemplateName)
```
KirilKabakchiev, line 222: same as above

components/director/internal/domain/package/service.go
```diff
@@ -251,8 +255,9 @@ func (s *service) createAPIs(ctx context.Context, packageID, tenant string, apis
 
 		err = s.apiRepo.Create(ctx, api)
 		if err != nil {
-			return errors.Wrap(err, "while creating API for application")
+			return errors.Wrapf(err, "while creating APIDefinition with id %s within Package with id %s", apiDefID, packageID)
 		}
+		log.Infof("Successfully created APIDefinition with id %s within Package with id %s", apiDefID, packageID)
```
KirilKabakchiev, line 260: i think log levels are not consistent in a couple of places - here you log related resource creation on info, while during creating related resources of applications (namely webhooks, the same is logged on debug)

components/director/internal/repo/update.go
```diff
@@ -94,6 +96,7 @@ func (u *universalUpdater) unsafeUpdateSingle(ctx context.Context, dbEntity inte
 	if err = persistence.MapSQLError(err, u.resourceType, "while updating single entity"); err != nil {
 		return err
 	}
+	log.Debugf("Executing query: %s", stmtBuilder.String())
```
KirilKabakchiev, line 99: i havent looked at the exact query logged but in general parameters (whatever goes in the place of ? placeholders) should not be logged (normally queries are logged without the parameters/arguments filled in). Or at the very least we should make sure we don't log passwords , keys and other sensitive information.

components/director/internal/domain/application/service.go
nyordanoff: Here when a new Label is created (because we are in creation flow in that function) its `id` is logged along with the respective object type and its `id`. Do you mean to log the key-value pair of each label as well? If so, should it be on `Info` or `Debug` level?

components/director/internal/domain/apptemplate/resolver.go
nyordanoff: The `id` is not available as the Application is not yet created. Regarding the Application `name`, where the log is it isn't yet available. 

However, few lines down the code(line 195) when we get the `ApplicationTemplate` by its name we can access the `ApplicationInputJSON` which holds the app name, unmarshal the JSON and get the name as a string. 

I implemented a "straight-forward" solution, you can have a look.

components/director/internal/domain/package/service.go
nyordanoff: My concern was that if many webhooks/labels are included during Application registration, the logs on `Info` will be filled with not that important information which will make them less readable.

From your point of view, is it ok for them to be on `Info`?

components/director/internal/repo/update.go
nyordanoff: In the actual logs the placeholders are not replaced with their actual values so there is no leak.

components/connector/internal/certificates/service.go
nyordanoff: Sorry, didn't understand exactly what the content of the log should be:

`Successfully signed CSR with common name after signCSR before returning` or `Successfully signed CSR with common name after signCSR` and place it before the return statement?

components/director/internal/domain/package/service.go
KirilKabakchiev: Maybe we keep labels on debug and hooks on info

components/director/internal/domain/application/service.go
KirilKabakchiev: I think labels can be on debug

components/connector/internal/certificates/service.go
KirilKabakchiev: 2nd (method call, handle error, log success, return)

KirilKabakchiev (approved):

PR #1530, opened 2020-08-24
Add automated setup of kyma and compass in local dev

KirilKabakchiev:
**Description**

The following PR introduces an option to run local dev compass with full blown kyma underneath using run.sh script. Previously run.sh only supported running compass in standalone mode, now additional arguments allow running single tenant kyma with embedded compass.

```bash
./run.sh --kyma-release 1.14.0 --kyma-installation full 
```

The command above would run kyma 1.14.0 and install compass from local sources ontop. This allows testing more complex local scenarios that require an actual runtime.

```bash
./run.sh  --kyma-installation minimal 
```

The command above will run compass with minimal required kyma components (such as ory, istio, etc)

Both arguments are optional, `kyma-installation` defaults to `minimal` and `kyma-release` defaults to the value in `KYMA_VERSION` file.


docs/compass/04-01-installation.md
```diff
@@ -133,94 +138,8 @@ Once Compass is installed, Runtime Agent will be configured to fetch the Runtime
...
-      namespace: kyma-installer
-      labels:
-        installer: overrides
-        kyma-project.io/installation: ""
-    data:
-      global.disableLegacyConnectivity: "true"
-    EOF
-    ```
-
-1. In order to reduce memory and CPU usage, from the downloaded `installer-cr-cluster-runtime.yaml.tpl` file, comment out the components you don't want to use, such as `monitoring`, `tracing`, `logging`, and `kiali`.
```
NickyMateev, line 163: This deleted note regarding the exclusion and inclusion of certain components seems reasonable to still exist in one form or another. If not for us, then for newer contributors who might look into streamlining their kyma installation for some reason

installation/resources/kyma/installer-cr-kyma.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: "installer.kyma-project.io/v1alpha1"
```
NickyMateev, line 1: This file and `installation/resources/kyma/installer-overrides-kyma-config-local.yaml` were previously downloaded each time. Now we'll use local copies - is there reason to be concerned that we might miss out on some kyma configuration changes this way or do we belief this configuration should fit our needs for a long time

NickyMateev (commented):
Does it make sense to move the `KYMA_VERSION` file into the new `installation/resources/kyma/` directory as well?

installation/resources/kyma/installer-cr-kyma.yaml
KirilKabakchiev: Im not sure, I was thinking to keep them locally for now as this one is slightly reduced (I removed some components such as kiali, tracing, etc).  Also, we currently have the minimal kyma CR locally, so for consistency reasons i thought we could use the same approach here for the full kyma CR.

Maybe if we see that next kyma releases change the format, etc we should think about downloading them each time (both minimal and full CRs) and also downloading the kyma cli that is used with the correct version (matching the version specified in kyma_version file or as --kyma-release arg). This would probably require doing a docker-in-docker installation so that we don't interfere with any kyma cli already installed on the host.

NickyMateev (dismissed):

nkanchev (approved):

NickyMateev (approved):

crabtree (approved):

PR #1533, opened 2020-09-02
Add status code check to tenant fetcher

PetarTodorovv:
**Description**

Currently, the tenant fetcher does not return an error if the request to the tenant's provider service returns status code that is not expected (200 OK and 204 No Content).
This PR adds a check for the status code from that call and a test for it.


desislavaa (approved):

pankrator (approved):
looks good to me

PR #1536, opened 2020-09-17
Migrate istio RBAC to istio authorization policy

strekm:
**Description**

- Migrate istio RBAC to istio authorization policy

**Related issue(s)**


kubadz (dismissed):

kubadz (approved):

kubadz (dismissed):

ataleksandrov (approved):

chart/compass/charts/connector/templates/service-role-binding.yaml
```diff
@@ -11,27 +11,22 @@ metadata:
...
-  roleRef:
-    kind: ServiceRole
-    name: {{ template "fullname" . }}-internal
----
+    - from:
+      - source:
+          principals:
+          - {{ tpl .Values.istio.rbac.connectorInternalApi.compassDirector . }}
+          - {{ tpl .Values.istio.rbac.connectorInternalApi.compassConnectorTests . }}
+    - to:
```
KirilKabakchiev, line 25: It looks to me (but i might be wrong) that this is more restrictive that what was previously defined. Right now we will restrict access to all connector apis (external and internal) to be accessible only by the director and compass tests service acccounts.
So im wondering with this change will apps that want to pair with compass and obtain CSR configuration details from connector be able to request it ? (we have two cases - the app can be running in the same cluster in different namespace (or in the same namespace as well i guess for test purposes?) or outside of the cluster altogether

chart/compass/charts/connector/templates/service-role-binding.yaml
```diff
@@ -11,27 +11,22 @@ metadata:
...
+      - operation:
+          ports:
```
KirilKabakchiev, line 27: as we are not using services for rules now, do we need to keep both clusterip services (connector internal service and connector external service?) or can one be removed ?

chart/compass/charts/connector/templates/service-role-binding.yaml
```diff
@@ -11,27 +11,22 @@ metadata:
...
+          - "{{ .Values.global.connector.graphql.internal.port }}"
```
KirilKabakchiev, line 28: does it make sense instead of pointing to ports to simply say  methods: ["GET", "POST"] or  paths: ["/*"] or not specifying anything in "to" (in the docs it says if nothing is specified in to, it will match all paths, methods,ports, etc for this workload) or is there any benefit for us to use ports? (This is more a question, out of curiosity)

chart/compass/charts/connector/templates/service-role-binding.yaml
kubadz: Previously, the `compass-connector` application was secured with the [ClusterRbacConfig](https://github.com/kyma-project/kyma/blob/master/resources/core/charts/istio-rbac/templates/rbac-config.yaml) for which the value is defined [here](https://github.com/kyma-project/kyma/blob/master/resources/core/charts/istio-rbac/values.yaml#L8). As you can see it secures only one service - `compass-connector-internal`. For access to that service, the ServiceRole was created which gives an access to all methods and all paths. This ServiceRole was bounded to `cluster.local/ns/{{ .Release.Namespace }}/sa/compass-director ` and `cluster.local/ns/{{ .Release.Namespace }}/sa/compass-connector-tests` serviceaccounts with the ServiceRoleBinding.

Again, please notice that the only service which was secured with RBAC was `compass-connector-internal`. You can find it [here](https://github.com/kyma-incubator/compass/blob/78c43b3f308b41b3e9af08c4f2f3a19b75494660/chart/compass/charts/connector/templates/service.yaml#L27). The ports used by this service are `{{ .Values.global.connector.graphql.internal.port }}`  and `{{ .Values.global.istio.proxy.port }}`, so the access to these ports should be possible only for serviceaccounts mentioned above. And all the remaining ports should be open to anyone.

Now what this new AuthorizationPolicy does is:
1) allow the access to the whole application (any ports) for two defined serviceaccounts with this code:
```
    - from:
      - source:
          principals:
          - {{ tpl .Values.istio.rbac.connectorInternalApi.compassDirector }}
          - {{ tpl .Values.istio.rbac.connectorInternalApi.compassConnectorTests }}
```
2) allow the access to the particular ports for anyone:
```
    - to:
      - operation:
          ports:
          - "{{ .Values.global.connector.graphql.internal.port }}"
          - "{{ .Values.global.istio.proxy.port }}"
          - "{{ .Values.global.connector.validator.port }}"
          - "{{ .Values.global.connector.graphql.external.port }}"
```

Actually, there is a mistake in `2)`, as the only ports that should be present, should be all ports that are not exposed by `compass-connector-internal` service. So the proper list is:
```
          - "{{ .Values.global.connector.validator.port }}"
          - "{{ .Values.global.connector.graphql.external.port }}"
```

Please notice two things:
- In contrary to ServiceRole the new AuthorizationPolicy applies to the workload (Pod), not to the particular service.
- `to` and `from` defines separate rules in this particular case, as both are the separate objects on the list:
```
- from:
   ...
- to:
   ...
```
the situation you described would happen if these were the properties of the same object:
```
- from:
   ...
   to:
   ...
```

I hope that this explanation is good enough. In case of any question, I will be happy to answer them.

chart/compass/charts/connector/templates/service-role-binding.yaml
kubadz: as described in https://github.com/kyma-incubator/compass/pull/1536/files#r495713387, the only ports that should stay here are ones from the external service.

chart/compass/charts/connector/templates/service-role-binding.yaml
kubadz: We have to keep it, because we want to ensure that the access to external ports is allowed for anyone, as described [here](https://github.com/kyma-incubator/compass/pull/1536/files#r495713387)

chart/compass/charts/connector/templates/service-role-binding.yaml
KirilKabakchiev: thanks, great explanation! So its actually two separate rules - one has just a source, saying this source can access to the whole app, the second rule has only destination/operation, saying that any source can access this destination/operation (public apis).

And if i understand correctly, the correction that is needed to remove the internal ports from the second rule's destination so that they are not accessible by everyone.

chart/compass/charts/connector/templates/service-role-binding.yaml
KirilKabakchiev: thanks, my question was actually do we need the clusterIP connector-internal-service ? Can we just put all four ports on connector-service clusterip?

KirilKabakchiev:
@kubadz i simplified the services a bit and removed the extra ports, could you check again? thanks!

chart/compass/charts/connector/templates/service-role-binding.yaml
kubadz: Exactly!

chart/compass/charts/connector/templates/service-role-binding.yaml
kubadz: Oh, you are right. If nothing else than Istio depends on it, then we can put all ports in one service. But I guess that some applications use the service names as addresses (for example: `compass-connector-internal.compass-system.svc.cluster.local` for connecting to internal endpoints), so we would need to know where this should be changed.

Also, it may be only my opinion, but I think that the separation for internal/external ports with services is a good and "foolproof" idea. The probability of making a mistake and for example exposing internal ports by accident is IMO lower when you see that the ports are used by a service called `internal`.

kubadz: Ok, I see that you already did a merge of the two services. Please make sure that `chart/compass/charts/connector/templates/tests/test-connector.yaml` and `chart/compass/charts/director/templates/deployment.yaml` were the only places where `-internal` prefix should be removed.

KirilKabakchiev: i didnt see any other places. My problem with the two services is that after the migration to the new authz rule there is no explicit "binding" that the ports defined a service called internal are actually internal (and there was such a binding with the previous istio cr). Thats why i thought it would make sense

ataleksandrov (approved):

KirilKabakchiev (approved):

PR #1550, opened 2020-10-06
Load certs revocation list config map when changed

pankrator:
**There is a problem when connector tries to load the revocation list on every certificate check. The k8s API server starts to reject and throttle requests when they queue up**

- Watch the revocation list configmap for changes
- Load the revocation list in memory when there are new changes
- Check the certificates against the in memory data


components/connector/internal/revocation/cache.go
```diff
@@ -0,0 +1,39 @@
...
+}
+func (cc *revocationCache) Get() map[string]string {
+	data, found := cc.cache.Get(key)
+	if !found {
+		return make(map[string]string)
+	}
+
+	revocationListData, ok := data.(map[string]string)
+	if !ok {
+		return make(map[string]string)
```
DimitarPetrov, line 35: This is pretty unexpected case, we can think of returning an error or log something in this case since it indicates something wrong internally in the app logic.

components/connector/internal/revocation/loader.go
```diff
@@ -0,0 +1,109 @@
...
+		configMapName:       configMapName,
+		reconnectInternal:   reconnectInternal,
+	}
+}
+
+func (rl *revocationListLoader) Run(ctx context.Context) {
+	rl.startKubeWatch(ctx)
+}
+
+func (rl *revocationListLoader) startKubeWatch(ctx context.Context) {
```
DimitarPetrov, line 41: Is this function needed you can add the logic in Run since Run only invokes this function and does not give any benefit.

components/connector/internal/revocation/loader.go
```diff
@@ -0,0 +1,109 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			log.Println("context cancelled, stop revocation config map watcher")
```
DimitarPetrov, line 45: Why not get logger down to the revocationListLoader and use it. Valid for the whole file.

components/connector/internal/revocation/loader.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (rl *revocationListLoader) Run(ctx context.Context) {
+	rl.startKubeWatch(ctx)
+}
+
+func (rl *revocationListLoader) startKubeWatch(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
```
DimitarPetrov, line 44: This context is a context.Background and it is not cancelled anywhere. Shouldn't we propagate the bootstrap context here?

components/connector/internal/revocation/loader.go
```diff
@@ -0,0 +1,109 @@
...
+				log.Println("Error event is received, stop revocation list configmap watcher and try again...")
+				return false
+			}
+		}
+	}
+}
+
+func connectWatch(manager Manager, configMapName string) (watch.Interface, error) {
+	log.Println("Starting watcher for revocation list configmap changes...")
+	for {
```
DimitarPetrov, line 102: Why is this in a loop when it is invoked in a loop also in the startKubeWatch?

components/connector/internal/revocation/loader.go
```diff
@@ -0,0 +1,109 @@
...
+		default:
+		}
+		watcher, err := connectWatch(rl.manager, rl.configMapName)
+		if err != nil {
+			log.Printf("Could not initialize watcher: %s. Sleep for %s and try again...", err, rl.reconnectInternal.String())
+			time.Sleep(rl.reconnectInternal)
+			continue
+		}
+		log.Println("Waiting for revocation list configmap events...")
+		for {
```
DimitarPetrov, line 56: Why is this for loop needed since there is a outside loop and we can processEvents has loop also. Here in the if bread can be continue and it will bi sufficient?

components/connector/internal/revocation/loader.go
```diff
@@ -0,0 +1,109 @@
...
+				return false
+			}
+			switch ev.Type {
+			case watch.Added:
+				fallthrough
+			case watch.Modified:
+				log.Println("revocation list updated")
+				config, ok := ev.Object.(*v1.ConfigMap)
+				if !ok {
+					log.Println("Unexpected error: object is not configmap. Try again")
```
DimitarPetrov, line 84: Isn't this a configuration issue which should not be silently retrained since it will never succeed. IS there any case that this will be the case if everything around watch is properly configured?

components/connector/internal/revocation/loader.go
```diff
@@ -0,0 +1,109 @@
...
+					return true
```
DimitarPetrov, line 85: This can be continue since you have a for loop here. Also then you will only have false case and can return an error instead of bool which will be more descriptive and the outher loop is unneeded.

components/connector/pkg/oathkeeper/validation_hydrator.go
```diff
@@ -93,12 +93,7 @@ func (tvh *validationHydrator) ResolveIstioCertHeader(w http.ResponseWriter, r *
 		return
 	}
 
-	isCertificateRevoked, err := tvh.revocationList.Contains(hash)
-	if err != nil {
-		tvh.log.Infof("Failed to check if certificate is revoked: %s", err.Error())
-		respondWithAuthSession(w, authSession)
-		return
-	}
+	isCertificateRevoked := tvh.revocationList.Contains(hash)
```
DimitarPetrov, line 96: Not needed variable can be inplace in the if below.

components/connector/internal/revocation/loader.go
```diff
@@ -0,0 +1,109 @@
...
+	revocationListCache Cache
+	manager             Manager
+	configMapName       string
+	reconnectInternal   time.Duration
+}
+
+func NewRevocationListLoader(revocationListCache Cache,
+	manager Manager,
+	configMapName string,
+	reconnectInternal time.Duration,
```
desislavaa, line 27: reconnectInternal -> reconnectInterval

components/connector/config/components.go
```diff
@@ -50,6 +54,11 @@ func InitInternalComponents(cfg Config, k8sClientset kubernetes.Interface) (Comp
 	)
 
 	certLoader := certificates.NewCertificateLoader(certCache, secretsRepository, caSecretName, rootCASecretName)
+	revocationListLoader := revocation.NewRevocationListLoader(revocationListCache,
+		k8sClientset.CoreV1().ConfigMaps(configmapNamespace.Namespace),
+		configmapNamespace.Name,
+		time.Second,
```
DimitarPetrov, line 60: Does it makes sense this to be configurable?

DimitarPetrov (changes requested):

components/connector/internal/revocation/repository.go
```diff
@@ -3,31 +3,37 @@ package revocation
...
-		configListManager: configListManager,
-		configMapName:     configMapName,
+func NewRepository(configListManager Manager, revocationListCache Cache, configMapName string) RevocationListRepository {
+	obj := &revocationListRepository{
+		configListManager:   configListManager,
+		configMapName:       configMapName,
+		revocationListCache: revocationListCache,
 	}
+
+	return obj
```
danielgospodinow, line 36: Is there any specific reason to create an object and then return it instead of just returning a newly constructed object? Just like before.

components/connector/internal/revocation/loader.go
pankrator: There is no bootstrap context in the whole project

components/connector/internal/revocation/loader.go
pankrator: I don't see any logger configured at all, so I decided to leave it as it is. When there is a proper logger implementation this can be done better

DimitarPetrov: Probably we can use logrus just to be consistent in the future, but still... Since there isn't any logging infrastructure I understand that we cannot do much at all. It is ok

components/connector/internal/revocation/loader.go
pankrator: I don't expect it to be any different, but wanted to handle the possibility. We have the option to report something or just print an error message. There is also the possibility to kill the process of course and k8s will start a new one, which will have the same configuration issue

components/connector/internal/revocation/loader.go
pankrator: didn't want to return an error as it doesn't mean there is a problem in general

components/connector/internal/revocation/loader.go
pankrator: leftover. It does not loop any way...

components/connector/internal/revocation/loader.go
DimitarPetrov: Since it is part of the connector itself I agree that it is enough to log it.

components/connector/internal/revocation/loader.go
pankrator: overlooked one of cases in which we don't need reconnect, but yes you are right, it can be handleded in the processEvents

components/connector/internal/revocation/loader.go
DimitarPetrov: I agree that false case is not error at all. Is it reasonable when there isn't any true case this function can be void and if it finishes we just retry in the outher function?

components/connector/config/components.go
pankrator: It is needed because of the tests. It can be made configurable anyway.

components/connector/internal/revocation/loader.go
pankrator: cleaner name as it will show exactly what will be started, not sure if it makes any sense

components/connector/internal/revocation/loader.go
pankrator: I agree on this

components/connector/internal/revocation/cache.go
pankrator: perhaps just an error message should be enough

components/connector/internal/revocation/loader.go
DimitarPetrov: It sounds reasonable I am ok with it.

components/connector/internal/revocation/loader.go
```diff
@@ -0,0 +1,109 @@
...
+	rl.startKubeWatch(ctx)
+}
+
+func (rl *revocationListLoader) startKubeWatch(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			log.Println("context cancelled, stop revocation config map watcher")
+			return
+		default:
```
danielgospodinow, line 47: Instead of an empty default case, can we place the logic below in the default case itself? Maybe it'll look prettier. 😆

components/connector/internal/revocation/repository.go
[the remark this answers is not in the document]
pankrator: not any particular reason :D , just overlooked

components/connector/internal/revocation/loader.go
pankrator: It is standard to see empty default cases in situations like that. Also it is more readable for the happy path (scenario) not to be indented.

danielgospodinow: Yes, it makes sense.

DimitarPetrov (approved):

PR #1551, opened 2020-10-06
System broker skeleton productization

DimitarPetrov:
Tests implementation for pkg package of system-broker component.
Associated with: https://github.com/kyma-project/test-infra/pull/2826


components/system-broker/pkg/http/client.go
```diff
@@ -43,15 +46,9 @@ func NewClient(timeout time.Duration, transport http.RoundTripper) *http.Client
...
 }
 
-func NewSecuredHTTPClient(timeout time.Duration, roundTripper HTTPRoundTripper, provider TokenProvider) (*http.Client, error) {
-	transport := &SecuredTransport{
-		roundTripper:  roundTripper,
-		tokenProvider: provider,
-		lock:          sync.Mutex{},
-	}
-
+func NewSecuredHTTPClient(timeout time.Duration, securedTransport HTTPRoundTripper) (*http.Client, error) {
```
KirilKabakchiev, line 49: if we pass the transport from outside, im not sure that NewSecuredHTTPClient describes what the clients purpose is (e.g. you can pass any transport and make it not sercured http client

components/system-broker/pkg/http/correlation_id_transport.go
```diff
@@ -30,7 +31,7 @@ func NewCorrelationIDTransport(roundTripper HTTPRoundTripper, uuidService UUIDSe
 
 type CorrelationIDTransport struct {
 	roundTripper HTTPRoundTripper
-	uuidService  UUIDService
+	uuidService  uid.UUIDService
```
KirilKabakchiev, line 34: if we use package uid, maybe we can remove UUID from the name

components/system-broker/pkg/http/secured_transport.go
```diff
@@ -61,3 +70,9 @@ func (c *SecuredTransport) refreshToken(ctx context.Context) error {
 
 	return nil
 }
+
+func (c *SecuredTransport) validToken() bool {
+	c.lock.RLock()
```
KirilKabakchiev, line 75: which race condition do we solve here ? (e.g. if a token is being validated and its valid, no one will actually try to generate a new one)

components/system-broker/pkg/http/types.go
```diff
@@ -21,6 +21,9 @@ import (
 	"time"
 )
 
+const TokenExpirationBuffer = 10 * time.Second
```
KirilKabakchiev, line 24: maybe we can have this configurable ? not sure if it makes sense though

components/system-broker/pkg/oauth/token_provider_test.go
```diff
@@ -0,0 +1,157 @@
...
+	clientID      string
+	clientSecret  string
+	tokenEndpoint string
+	accessToken   string
+
+	config    *oauth.Config
+	k8sClient client.Client
+}
+
+func (suite *TokenProviderFromSecretTestSuite) SetupTest() {
```
KirilKabakchiev, line 48: some tests (test files) are one a testify suite and others are not?

components/system-broker/cmd/main.go
```diff
@@ -62,11 +61,10 @@ func main() {
 	directorGraphQLClient, err := prepareGqlClient(cfg, uuidSrv)
 	fatalOnError(err)
 
-	systemBroker := osb.NewSystemBroker(directorGraphQLClient, cfg.Server.SelfURL)
+	systemBroker := osb.NewSystemBroker(directorGraphQLClient, cfg.Server.SelfURL+cfg.Server.RootAPI)
```
KirilKabakchiev, line 64: im not sure if this will be ok for the middlewares registered in the osb/api.go file. They have to run only on broker calls and not when trying to fetch specs or hitting the health api while the filters that are registered during server creation should run on all api calls (osb+specs +health) (TODO should be verified)

DimitarPetrov: Previously SelfURL had RootAPI in it which is redundant since we have a separate configuration for the RootAPI and we can reuse it. In the end this is the same URL but crafted from two unique variables and not having double maintanance of the RootAPI one separate and one part of SelfURL.

components/system-broker/pkg/http/secured_transport.go
DimitarPetrov: Discussed separately. In short this mitigates the issue that the token will be refreshed many times one after another because in the time when it is refreshed other gouroutines will check and see that it is expired and will wait to refresh also.

components/system-broker/pkg/oauth/token_provider_test.go
DimitarPetrov: We use testify suites when there is common before/after each logic. testify suites are just a wrapper of go sub-tests and there is no need of strong convention just a convenient way to separate before/after logic. That's why it is not always needed. There are test files which has standalone top-level tests and suites aswell.

components/system-broker/pkg/http/types.go
DimitarPetrov: Probably it makes sense to be reused request timeout for this. Since this is the time that left for the token to be valid and if it is smaller than the request timeout it will expire during a request and if it is longer it does not make sense to refresh the token since it has the validity for the max request duration. What do you think?

components/system-broker/pkg/http/client.go
DimitarPetrov: Either we create the client directly and does not use misleading constructor or accept a securedTransport struct directly instead of interface. What do you think?

components/system-broker/cmd/main.go
```diff
@@ -77,30 +76,25 @@ func fatalOnError(err error) {
...
+	securedTransport := httputil.NewSecuredTransport(cfg.HttpClient.Timeout, httpTransport, oauthTokenProvider)
+	securedClient := &http.Client{
+		Transport: securedTransport,
+		Timeout:   cfg.HttpClient.Timeout,
 	}
 
 	// prepare graphql client that uses secured http client as a basis
-	gqlClient, err := graphql.NewClient(cfg.GraphQLClient, securedClient)
+	graphClient := gql.NewClient(cfg.GraphQLClient.GraphqlEndpoint, gql.WithHTTPClient(securedClient))
+	gqlClient, err := graphql.NewClient(cfg.GraphQLClient, graphClient)
```
ataleksandrov, line 97: Error check is unnecessary

components/system-broker/pkg/env/description.go
```diff
@@ -35,7 +35,7 @@ func newDescriptionTree(root string) *descriptionTree {
 	}
 }
 
-func (t *descriptionTree) AddNode(tree *descriptionTree) {
+func (t *descriptionTree) аddNode(tree *descriptionTree) {
```
ataleksandrov, line 38: "a" is written in cyrillic

components/system-broker/pkg/http/secured_transport_test.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func TestSecuredTransport_RoundTripSuccessfullyObtainsNewTokenAfterExpiration(t *testing.T) {
+	const accessToken1 = "accessToken1"
+	const accessToken2 = "accessToken2"
+
+	tokenProvider := &httpfakes.FakeTokenProvider{}
+	tokenProvider.GetAuthorizationTokenReturnsOnCall(0, httputil.Token{
+		AccessToken: accessToken1,
+		Expiration:  time.Now().Add(time.Millisecond * 100).Unix(),
```
ataleksandrov, line 54: Can we just do : ` time.Now().Unix()`?

components/system-broker/pkg/oauth/token_provider.go
```diff
@@ -48,134 +35,13 @@ const (
...
-	//request, err := c.requestProvider.Provide(ctx, input)
-	//if err != nil {
-	//	return httputils.Token{}, errors.Wrap(err, "while creating authorisation token request")
-	//}
-
-	request.SetBasicAuth(credentials.clientID, credentials.clientSecret)
-	request.Header.Set(contentTypeHeader, contentTypeApplicationURLEncoded)
-
-	response, err := c.httpClient.Do(request)
+	k8sClient, err := prepareK8sClient()
```
NickyMateev, line 42: It'sa little weird to me that a `/pkg/oauth` package with a public function with the following signiture:

```
func NewTokenProvider(config *Config, httpClient httputils.Client) (httputils.TokenProvider, error)
```

would internally have some k8s specific logic, but since this is for the sake of not having a gigantic `main.go` and I cannot give you a much better proposal for where to move this and have it as dependency, I guess we can live with it.

components/system-broker/tests/healthcheck_test/healthcheck_test.go
```diff
@@ -0,0 +1,32 @@
...
+func (suite *HealthcheckTestSuite) SetupSuite() {
+	suite.testContext = common.NewTestContextBuilder().Build(suite.T())
+}
+
+func (suite *HealthcheckTestSuite) TearDownSuite() {
+	suite.testContext.CleanUp()
+}
+
+func (suite *HealthcheckTestSuite) TestHealthcheck() {
+	suite.testContext.SystemBroker.GET("/healthz").Expect().
```
NickyMateev, line 29: Beautiful - so simple and elegant - just like old times :D

ataleksandrov (changes requested):

components/system-broker/tests/common/test_context.go
```diff
@@ -0,0 +1,167 @@
...
+	Environment env.Environment
+
+	Servers    map[string]FakeServer
+	HttpClient *http.Client
+}
+
+// NewTestContextBuilder sets up a builder with default values
+func NewTestContextBuilder() *TestContextBuilder {
+	return &TestContextBuilder{
+		Environment: TestEnv(),
```
NickyMateev, line 46: I was wondering where we get the default env values from, as far as I remember back in SM, we had a dedicated `application.yml` for default env values. But I see here that we don't really need anything more than the values that you override in the envHooks + we rely on the `DefaultConfig()` values.

NickyMateev (approved):
Overall - LGTM 👍

components/system-broker/pkg/oauth/token_provider.go
DimitarPetrov: Also it is used to unify the oauth provider with k8s secret credentials and a local provider which just returns a predefined value. This way when we rely on a single interface we can manipulate it from outside and have the 
`--oauth_provider.local` flag to easily run it locally.

components/system-broker/tests/common/test_context.go
DimitarPetrov: Yes here we have a self contained default config which is sufficient for now. If we later introduce a more sophisticated tests which need external configuration we can add it but since we started lean with the component tests framework this feature is still not needed.

ataleksandrov (approved):
Looks good to me!

KirilKabakchiev (approved):

PR #1556, opened 2020-10-09
Remove checking if request certificate is present on revocation list

koala7659:
This PR is workaround for a problem with flooding API server with requests from hydrator for endpoint ` /certificate/data/resolve` from checking certificate hash is present on revocation list. 

The revocation list was implemented with Kubernetes ConfigMap 

Please notice this is only a workaround to maintain production stability. The proper fix will be deployer later on.


KirilKabakchiev (changes requested):
This should not be merged.

PR #1559, opened 2020-10-15
Disable audit logging for graphql queries

pankrator:

- Check the type of the graphql query and if it is not a *mutation* do not audit log

### DIFF
--- a/components/gateway/pkg/proxy/transport.go
+++ b/components/gateway/pkg/proxy/transport.go
@@ -2,7 +2,9 @@ package proxy
 
 import (
 	"bytes"
+	"encoding/json"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


danielgospodinow (dismissed):

components/gateway/pkg/proxy/transport.go
```diff
@@ -62,13 +64,42 @@ func (t *Transport) RoundTrip(req *http.Request) (resp *http.Response, err error
...
+	}
+
 	err = t.auditlogSvc.Log(string(requestBody), string(responseBody), claims)
 	if err != nil {
 		return nil, errors.Wrap(err, "while sending to auditlog")
 	}
 	return resp, nil
 }
 
+func checkQueryType(requestBody []byte, typee string) (bool, error) {
```
nyordanoff, line 83: typo in `typee` ?

pankrator: it should be typee because type is a reserved keyword...

desislavaa (approved):

desislavaa (approved):

PR #1561, opened 2020-10-16
Enrich logging in Gateway

danielgospodinow:
**Description**

- Improve logging in Gateway

### DIFF
--- a/components/gateway/cmd/main.go
+++ b/components/gateway/cmd/main.go
@@ -73,7 +73,7 @@ func main() {
 
 	http.Handle("/", router)
 
-	log.Printf("Listening on %s\n", cfg.Address)
+	log.Printf("Listening on %s", cfg.Address)
 	if err := http.ListenAndServe(cfg.Address, nil); err != nil {
 		done <- true
 		panic(err)
@@ -81,7 +81,7 @@ func main() {
 }
 
 func proxyRequestsForComponent(router *mux.Router, path string, targetOrigin string, transport HTTPTransport, middleware ...mux.MiddlewareFunc) error {
-	log.Printf("Proxying requests on path `%s` to `%s`\n", path, targetOrigin)
+	log.Printf("Proxying requests on path `%s` to `%s`", path, targetOrigin)
 
 	componentProxy, err := proxy.New(targetOrigin, path, transport)
 	if err != nil {
@@ -152,7 +152,7 @@ func initAuditLogs(done chan bool) (AuditogService, error) {
 	msgChannel := make(chan auditlog.Message, cfg.MsgChannelSize)
 	initWorker(auditlogSvc, done, msgChannel)
 
-	log.Printf("Auditlog configured successfully, auth mode:%s\n", cfg.AuthMode)
+	log.Printf("Auditlog configured successfully, auth mode:%s", cfg.AuthMode)
 	return auditlog.NewSink(msgChannel, cfg.MsgChannelTimeout), nil
 }
 
--- a/components/gateway/cmd/main.go
+++ b/components/gateway/cmd/main.go
@@ -48,9 +48,11 @@ func main() {
 	done := make(chan bool)
 	var auditlogSvc AuditogService
 	if cfg.AuditlogEnabled {
+		log.Println("Auditlog is enabled")
 		auditlogSvc, err = initAuditLogs(done)
 		exitOnError(err, "Error while initializing auditlog service")
 	} else {
+		log.Println("Auditlog is disabled")
 		auditlogSvc = &auditlog.NoOpService{}
 	}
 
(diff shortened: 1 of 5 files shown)


danielgospodinow:
Should I go for [logrus](https://github.com/sirupsen/logrus) or we can just instead stick with the built-in log?

components/gateway/cmd/main.go
```diff
@@ -71,7 +73,7 @@ func main() {
 
 	http.Handle("/", router)
 
-	log.Printf("Listening on %s", cfg.Address)
+	log.Printf("Listening on %s\n", cfg.Address)
```
pankrator, line 76: The new line in the end is not needed. The log package adds new line for each print

components/gateway/internal/auditlog/service.go
```diff
@@ -35,10 +37,15 @@ func (sink *Sink) Log(request, response string, claims proxy.Claims) error {
 		Claims:   claims,
 	}
 
+	log.Printf("Attempting to register auditlog message %+v for processing to the queue (size=%d, capacity=%d)\n",
```
pankrator, line 40: I think the log after writing to the channel should be enough, so this one can be removed

pankrator (changes requested):

components/gateway/internal/auditlog/service.go
```diff
@@ -35,10 +37,15 @@ func (sink *Sink) Log(request, response string, claims proxy.Claims) error {
...
+		msg, len(sink.logsChannel), cap(sink.logsChannel))
+
 	select {
 	case sink.logsChannel <- msg:
+		log.Printf("Successfully registered auditlog message %+v for processing to the queue (size=%d, capacity=%d)\n",
+			msg, len(sink.logsChannel), cap(sink.logsChannel))
 	case <-time.After(sink.timeout):
-		return errors.New("Cannot write to the channel")
+		return errors.New(fmt.Sprintf("Failed to write message %+v to the channel", msg))
```
pankrator, line 48: The message should not be printed as it might contain sensitive data. (the same applies to auditlog anyway :D, but...)  Leave this error as it was.

danielgospodinow: Good point :D 👍🏼

pankrator: The same applies to other places where there is such msg print

components/gateway/internal/auditlog/worker.go
```diff
@@ -29,12 +29,15 @@ func (w *Worker) Start() {
...
 		select {
 		case <-w.done:
+			log.Println("Worker for auditlog message processing has finished")
 			return
 		case msg := <-w.auditlogChannel:
 			err := w.svc.Log(msg.Request, msg.Response, msg.Claims)
 			if err != nil {
-				log.Printf("error while saving auditlog: %s\n", err.Error())
+				log.Printf("error while saving auditlog message %+v with error: %s\n", msg, err.Error())
+				return
```
pankrator, line 38: This return will exit the worker goroutine so here it should be if {...} else {...}

components/gateway/cmd/main.go
```diff
@@ -139,7 +140,7 @@ func initAuditLogs(done chan bool) (AuditogService, error) {
 			msgFactory = auditlog.NewMessageFactory(oauthCfg.User, oauthCfg.Tenant, uuidSvc, timeSvc)
 		}
 	default:
-		return nil, errors.New(fmt.Sprintf("Invalid Auditlog Auth mode: %s", cfg.AuthMode))
+		return nil, errors.New(fmt.Sprintf("Invalid auditlog auth mode: %s", cfg.AuthMode))
```
DimitarPetrov, line 143: You can use `fmt.Errorf(...)` and also go errors should start with lowercase and logs with uppercase (valid for the whole change)

components/gateway/internal/auditlog/worker.go
```diff
@@ -29,12 +29,15 @@ func (w *Worker) Start() {
...
 		select {
 		case <-w.done:
+			log.Println("Worker for auditlog message processing has finished")
 			return
 		case msg := <-w.auditlogChannel:
 			err := w.svc.Log(msg.Request, msg.Response, msg.Claims)
 			if err != nil {
-				log.Printf("error while saving auditlog: %s\n", err.Error())
+				log.Printf("error while saving auditlog message %+v with error: %s", msg, err.Error())
+				return
```
DimitarPetrov: this should be continue, otherwise you will stop the worker if just one message fails.

DimitarPetrov (changes requested):

DimitarPetrov (approved):

DimitarPetrov (approved):

pankrator (approved):

PR #1566, opened 2020-10-20
Add DB logs

PetarTodorovv:
**Description**

Add DB logs

### DIFF
--- a/components/director/internal/repo/create.go
+++ b/components/director/internal/repo/create.go
@@ -47,7 +47,6 @@ func (c *universalCreator) Create(ctx context.Context, dbEntity interface{}) err
 		values = append(values, fmt.Sprintf(":%s", c))
 	}
 
-	log.Debug("Building DB query...")
 	stmt := fmt.Sprintf("INSERT INTO %s ( %s ) VALUES ( %s )", c.tableName, strings.Join(c.columns, ", "), strings.Join(values, ", "))
 
 	log.Debugf("Executing DB query: %s", stmt)
--- a/components/director/internal/repo/create.go
+++ b/components/director/internal/repo/create.go
@@ -47,9 +47,11 @@ func (c *universalCreator) Create(ctx context.Context, dbEntity interface{}) err
 		values = append(values, fmt.Sprintf(":%s", c))
 	}
 
+	log.Debug("Building DB query...")
 	stmt := fmt.Sprintf("INSERT INTO %s ( %s ) VALUES ( %s )", c.tableName, strings.Join(c.columns, ", "), strings.Join(values, ", "))
-	log.Debugf("Executing query: %s", stmt)
 
+	log.Debugf("Executing DB query: %s", stmt)
 	_, err = persist.NamedExec(stmt, dbEntity)
+
 	return persistence.MapSQLError(err, c.resourceType, "while inserting row to '%s' table", c.tableName)
 }
--- a/components/director/internal/repo/delete.go
+++ b/components/director/internal/repo/delete.go
@@ -78,16 +78,18 @@ func (g *universalDeleter) unsafeDelete(ctx context.Context, conditions Conditio
 		stmtBuilder.WriteString(" WHERE")
 	}
 
+	log.Debug("Add conditions to the query")
 	err = writeEnumeratedConditions(&stmtBuilder, conditions)
 	if err != nil {
 		return errors.Wrap(err, "while writing enumerated conditions")
 	}
 	allArgs := getAllArgs(conditions)
 
+	log.Debug("Building DB query...")
 	query := getQueryFromBuilder(stmtBuilder)
-	log.Debugf("Executing query %s", query)
+	log.Debugf("Executing DB query: %s", query)
 	res, err := persist.Exec(query, allArgs...)
-	if err = persistence.MapSQLError(err, g.resourceType, "while deleting object from database"); err != nil {
+	if err = persistence.MapSQLError(err, g.resourceType, "while deleting object from '%s' table", g.tableName); err != nil {
 		return err
 	}
 
--- a/components/director/internal/repo/delete.go
+++ b/components/director/internal/repo/delete.go
@@ -78,14 +78,12 @@ func (g *universalDeleter) unsafeDelete(ctx context.Context, conditions Conditio
 		stmtBuilder.WriteString(" WHERE")
 	}
 
-	log.Debug("Add conditions to the query")
 	err = writeEnumeratedConditions(&stmtBuilder, conditions)
 	if err != nil {
 		return errors.Wrap(err, "while writing enumerated conditions")
 	}
 	allArgs := getAllArgs(conditions)
 
-	log.Debug("Building DB query...")
 	query := getQueryFromBuilder(stmtBuilder)
 	log.Debugf("Executing DB query: %s", query)
 	res, err := persist.Exec(query, allArgs...)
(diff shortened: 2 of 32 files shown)


components/director/internal/repo/delete.go
```diff
@@ -78,16 +78,18 @@ func (g *universalDeleter) unsafeDelete(ctx context.Context, conditions Conditio
 		stmtBuilder.WriteString(" WHERE")
 	}
 
+	log.Debug("Add conditions to the query")
```
DimitarPetrov, line 81: Probably add conditions to the log otherwise it does not provide much value. Just tracing information.
Valid for other methods also.

components/director/internal/repo/get.go
```diff
@@ -67,14 +67,16 @@ func (g *universalSingleGetter) unsafeGet(ctx context.Context, conditions Condit
 		return err
 	}
 
+	log.Debug("Building DB query...")
 	query, args, err := buildSelectQuery(g.tableName, g.selectedColumns, conditions, orderByParams)
 	if err != nil {
+		log.Errorf("Error while building DB query: %s", query)
```
DimitarPetrov, line 73: Is this log needed isn't this logged in the outher scope?
If not add the error to the log also.

components/director/pkg/persistence/sql_error_mapper.go
```diff
@@ -17,18 +18,23 @@ func MapSQLError(err error, resourceType resource.Type, format string, args ...i
 	}
 
 	if err == sql.ErrNoRows {
+		log.Warnf("SQL: no rows in result set for '%s' table", resourceType)
```
DimitarPetrov, line 21: Add the error to the log also.

components/director/pkg/persistence/sql_error_mapper.go
```diff
@@ -17,18 +18,23 @@ func MapSQLError(err error, resourceType resource.Type, format string, args ...i
...
 		return apperrors.NewNotFoundErrorWithType(resourceType)
 	}
 
 	pgErr, ok := err.(*pq.Error)
 	if !ok {
+		log.Warn("Error while getting postgres error code")
```
DimitarPetrov, line 27: Is this the real error it is just a cast?

components/director/pkg/persistence/sql_error_mapper.go
```diff
@@ -17,18 +18,23 @@ func MapSQLError(err error, resourceType resource.Type, format string, args ...i
...
 		return apperrors.InternalErrorFrom(err, format, args...)
 	}
 
+	log.Debug("Checking Postgres error code...")
 	switch pgErr.Code {
 	case UniqueViolation:
+		log.Warn("Postgres unique violation error code found")
```
DimitarPetrov, line 34: You can add more context in the logs, resource type and arguments and the error itself this does not give much more value than the error returned on the next line.

components/director/internal/repo/get.go
PetarTodorovv: I've removed the log because the error is propagated to the outer scopes

components/director/internal/repo/create.go
```diff
@@ -47,9 +47,11 @@ func (c *universalCreator) Create(ctx context.Context, dbEntity interface{}) err
 		values = append(values, fmt.Sprintf(":%s", c))
 	}
 
+	log.Debug("Building DB query...")
```
DimitarPetrov, line 50: Actually the statements between Building DB Query and Executing DB Query are just one fmt.Sprintf which cannot fail. Of course it can build wrong query and this is seen in the Executing... log. It looks like this log is unnecessary.

Valid for the whole change.

components/director/internal/repo/exist.go
```diff
@@ -60,17 +62,20 @@ func (g *universalExistQuerier) unsafeExists(ctx context.Context, conditions Con
 		stmtBuilder.WriteString(" WHERE")
 	}
 
+	log.Debug("Add conditions to the query")
```
DimitarPetrov, line 65: We can add the actual conditions to the log so we can verify that the query that is executed below is right SQL representation of this conditions.

components/director/pkg/persistence/sql_error_mapper.go
```diff
@@ -17,20 +19,27 @@ func MapSQLError(err error, resourceType resource.Type, format string, args ...i
...
 	pgErr, ok := err.(*pq.Error)
 	if !ok {
-		return apperrors.InternalErrorFrom(err, format, args...)
+		log.Errorf("Error while casting to postgres error. Actual error: %s", err)
+		return apperrors.NewInternalError("Unexpected error while executing SQL query")
 	}
 
+	log.Errorf("SQL Error: %s. Caused by: %s. DETAILS: %s", fmt.Sprintf(format, args...), pgErr.Message, pgErr.Detail)
+
+	log.Debug("Checking Postgres error code...")
```
DimitarPetrov, line 34: This log does not give any information because the next statements are returns and does not represent any meaningful checkpoint in the code.

components/director/pkg/persistence/sql_error_mapper.go
```diff
@@ -17,20 +19,27 @@ func MapSQLError(err error, resourceType resource.Type, format string, args ...i
...
 	switch pgErr.Code {
```
DimitarPetrov, line 35: The two logs in the switch cases are redundant having the previous one outside of the switch giving all the necessary information

components/director/pkg/persistence/sql_error_mapper.go
```diff
@@ -17,20 +19,27 @@ func MapSQLError(err error, resourceType resource.Type, format string, args ...i
...
 	case UniqueViolation:
+		log.Errorf("Postgres unique violation error code found in '%s' table. Error message: '%s'. DETAILS: %s Postgres error code: '%s'", pgErr.Table, pgErr.Message, pgErr.Detail, pgErr.Code)
 		return apperrors.NewNotUniqueError(resourceType)
 	case ForeignKeyViolation:
+		log.Errorf("Postgres foreign key violation error found in '%s' table. Error message: '%s'. DETAILS: %s Postgres error code: '%s'", pgErr.Table, pgErr.Message, pgErr.Detail, pgErr.Code)
 		return apperrors.NewInvalidDataError("Object already exist")
 	}
 
-	return apperrors.InternalErrorFrom(err, "SQL Error: %s", fmt.Sprintf(format, args...))
+	return apperrors.NewInternalError("SQL Error occurred")
```
DimitarPetrov, line 44: `Unexpected error while executing SQL query` sounds better to me from user's perspective

DimitarPetrov (changes requested):
We still don't have unified logging infrastructure across the whole project but in director we have configured logrus entry somewhere in the outer scopes.

If you use the package directly this does not give any chance of configuration. For example by default it is on Info level and there is no way to reconfigure. Format is also predefined. And most importantly when we introduce correlationID and other tracing information which correlates all the logs from a single request this logs will be unasscoiated and therefore much less usefull.

This will be a separate effort to unify logging, but until then we should stick to the current approach to use the already configured logrus entry so we can enrich it step by step.

components/director/internal/repo/create.go
PetarTodorovv: Yes, in this case, we can remove the log but also we have cases where after the debug log we have a function and maybe it's a good idea to have such a debug log(at least for traceability)?

DimitarPetrov (approved):

PR #1568, opened 2020-10-21
Introduce timeout handler

DimitarPetrov:
A new handler is introduced which wrap other handler and take into account configured server timeout.

This change is needed as part of general timeouts configuration across the whole project.

Next PR will be from this branch: https://github.com/kyma-incubator/compass/compare/timeouts
Using this handler


nyordanoff (approved):

components/director/pkg/handler/timeout_handler.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/kyma-incubator/compass/components/director/pkg/apperrors"
+)
+
+const (
+	HeaderContentTypeKey   = "Content-Type"
+	HeaderContentTypeValue = "application/json;charset=UTF-8"
+)
+
+func WithTimeout(h http.Handler, timeout time.Duration) (http.Handler, error) {
```
KirilKabakchiev, line 16: logging the error from second writing might be useful, but we can discuss this in the 2nd PR where we have usage examples also

KirilKabakchiev (approved):

PR #1570, opened 2020-10-21
Add logs in client credentials pairing flow

nyordanoff:
**Description**
This PR adds logs in the flow where a Runtime/Application/IntegrationSystem pairs with Compass via client credentials(client_id and client_secret)

- add logs


components/director/internal/domain/oauth20/resolver.go
```diff
@@ -106,15 +112,17 @@ func (r *Resolver) generateClientCredentials(ctx context.Context, objType model.
 	}
 
 	id := clientCreds.ClientID
+	r.logger.Debugf("Creating SystemAuth model for the client credentials for %s with id %s", objType, objID)
```
DimitarPetrov, line 115: `model` seems a little confusing probably redundant. You just create a SystemAuth (valid for the logs below also)

components/director/internal/domain/systemauth/service.go
```diff
@@ -76,7 +82,7 @@ func (s *service) create(ctx context.Context, id string, objectType model.System
 
 	err = s.repo.Create(ctx, systemAuth)
 	if err != nil {
-		return "", errors.Wrapf(err, "while creating System Auth for %s", objectType)
+		return "", err
```
DimitarPetrov, line 85: Why remove the details of the error?

DimitarPetrov (approved):
Overall LGTM just some super minor remarks.

components/director/internal/domain/systemauth/service.go
nyordanoff: I moved it in the [resolver](https://github.com/kyma-incubator/compass/blob/client-creds-pairing-logs/components/director/internal/domain/oauth20/resolver.go#L123) so that I can include both the object type and the object id in the error log. 

I have to remove the `model` part there as well based on your previous comment.

DimitarPetrov (approved):

PR #1571, opened 2020-10-22
Revisit Client/Server Timeouts

DimitarPetrov:
Meaningful defaults for me:

- oathkeeper: 120s
- connectivity-adapter: 
    - 119s server 
    - 115s client
- gateway: 
    - 114s server 
    - 30s auditlog client
- connector: 
    - 100s server 
    - 95s client
- director: 
    - 110s server 
    - 105s client


components/connectivity-adapter/cmd/main.go
```diff
@@ -26,14 +31,24 @@ func main() {
...
 
-	handler, err := initAPIHandler(cfg)
+	h, err := initAPIHandler(cfg)
 	if err != nil {
 		exitOnError(err, "Failed to init External Connector handler")
 	}
 
+	handlerWithTimeout, err := handler.WithTimeout(h, cfg.ServerTimeout)
+	if err != nil {
+		exitOnError(err, "Filed configuring timeout on handler")
```
nyordanoff, line 41: `Failed`

nyordanoff (approved):

KirilKabakchiev (commented):
As discussed, lets add more robust logging on the ErrHandlerTimeout error with a new response writer wrapper

KirilKabakchiev (changes requested):
i noticed that in some clients such as https://github.com/kyma-incubator/compass/blob/bed3d78ac7328829690391f92c7a3f20f9b2ebca/components/connectivity-adapter/internal/appregistry/director/cli.go#L286 , the golang background context is passed. This is not good because when the server timeout handler cancels the request context, this won't be propagated to the outbound http call and it will keep running. I think its worth fixing this one as it kind of partically defeats the purpose of having a server timeout if we dont make sure we free up the go routine

components/connector/config/servers.go
```diff
@@ -33,13 +35,19 @@ func PrepareExternalGraphQLServer(cfg Config, certResolver api.CertificateResolv
...
-		Handler: externalRouter,
+	handlerWithTimeout, err := timeouthandler.WithTimeout(externalRouter, cfg.ServerTimeout)
+	if err != nil {
+		return nil, err
 	}
+
+	return &http.Server{
+		Addr:              cfg.ExternalAddress,
+		Handler:           handlerWithTimeout,
+		ReadHeaderTimeout: cfg.ServerTimeout,
```
KirilKabakchiev, line 46: im not sure if when the server read header timeout hits and the same time as the handlerWithTimeout, that you would be allowed to write the timeout handler message to the socket (this can be tested by making the ReadHeaderTimeout slightly less than the handlerWithTimeout timeout. If this is the case, we should have the ReadHeaderTimeout be equal to  cfg.ServerTimeout + 1 second

DimitarPetrov: ReadHeaderTimeout is hit before even ServeHTTP of the timeout handler is invoked. Therefore it is a timeout before the actual server timeout respected by the timeout handler is hit.

Therefore if the headers are read 10 seconds the time for the timeout handler will still be 110s and the whole request will be 120s long. But I think that this is not an issue at all because in our istio setup we only allow some specific headers that we know are safe and will never be needed so much time for them to be fetched.

Another approach could be to come up with a dedicated timeout for the headers and subtract it from the server timeout this way server processing will be readHeaderTimeout + ServerTimeout by timeout handler.

KirilKabakchiev (approved):

PR #1574, opened 2020-10-23
Catalog sys broker

krasish:
**Productize catalog endpoint**

- Paging director response
- Mock GQL Server


components/system-broker/tests/catalog_test/catalog_test.go
```diff
@@ -0,0 +1,351 @@
...
+		GraphqlQueryKey: common.GraphqlQueryKey{
+			Type: queryType,
+			Name: queryName,
+		},
+		Response: applicationsResponse,
+	}
+	jsonBody, err := json.Marshal(body)
+
+	assert.NoError(suite.T(), err)
+	http.DefaultClient.Post(suite.configURL, "application/json", bytes.NewReader(jsonBody))
```
KirilKabakchiev, line 350: it will be probably better to not use the default client even in tests (it has no timeout, tests can hang, etc)

components/system-broker/tests/common/fake_router.go
```diff
@@ -0,0 +1,131 @@
...
+	queryField := query.Operations[0].SelectionSet[0].(*ast.Field)
+
+	queryName := queryField.Name
+	queryType := string(query.Operations[0].Operation)
+	key := GraphqlQueryKey{
+		Type: queryType,
+		Name: queryName,
+	}
+	var response interface{}
+	func() {
```
KirilKabakchiev, line 103: is this closure needed just for the lock ? is so, why not just extract it in a separate smaller method

components/system-broker/tests/common/fake_router.go
```diff
@@ -0,0 +1,131 @@
...
+		g.m.Lock()
```
KirilKabakchiev, line 104: can this be just a read lock ?

components/system-broker/internal/director/client.go
```diff
@@ -50,47 +50,42 @@ type GraphQLizer interface {
 	PackageInstanceAuthRequestInputToGQL(in schema.PackageInstanceAuthRequestInput) (string, error)
 }
 
-func NewGraphQLClient(gqlClient Client, gqlizer GraphQLizer, gqlFieldsProvider GqlFieldsProvider) *GraphQLClient {
+func NewGraphQLClient(gqlClient Client, gqlizer GraphQLizer, gqlFieldsProvider GqlFieldsProvider, c *Config) *GraphQLClient {
 	return &GraphQLClient{
 		gcli: gqlClient,
 		//queryProvider:     queryProvider{}, - gqlizers are better
```
KirilKabakchiev, line 56: prolly you could clean up these commented lines, as this query provider seems like a bad idea anyway

components/system-broker/internal/director/client.go
```diff
@@ -50,47 +50,42 @@ type GraphQLizer interface {
...
 	//queryProvider     queryProvider
 }
 
-func (c *GraphQLClient) FetchApplications(ctx context.Context) (*ApplicationsOutput, error) {
+func (c *GraphQLClient) FetchApplications(ctx context.Context) (ApplicationsOutput, error) {
 	response := ApplicationsOutput{}
 
 	query := fmt.Sprintf(`query {
-			result: applications {
+			result: applications(first: %%d, after: %%q) {
```
KirilKabakchiev, line 75: this is a bit odd - one idea would be to use Vars() and Var() methods of the gli request (use pass default values for vars =pagesize + empty cursor and then with each request you set the correct Var values

components/system-broker/internal/director/client.go
```diff
@@ -50,47 +50,42 @@ type GraphQLizer interface {
...
 					%s
 			}
 	}`, c.outputGraphqlizer.Page(c.outputGraphqlizer.ForApplication()))
+	queryGenerator := func(pageSize int, page string) string {
```
KirilKabakchiev, line 79: this could then become something like variableResolver and accepts a map of variables to resolve with

KirilKabakchiev (changes requested):

KirilKabakchiev:
there is one TODO in catalog converter that is worth checking out. If you don't think a better approach is needed there, please remove the TODO

components/system-broker/pkg/server/server.go
```diff
@@ -99,19 +99,9 @@ func (s *Server) Start(parentCtx context.Context) {
 func (s *Server) stop(parentCtx context.Context) {
 	atomic.StoreInt32(&s.healthy, 0)
 
-	ctx, cancel := context.WithTimeout(parentCtx, s.shutdownTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), s.shutdownTimeout)
 	defer cancel()
 
-	go func(ctx context.Context) {
```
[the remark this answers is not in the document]
pankrator: The http server is started in the main goroutine, so when the server is shutdown, the main goroutine exit will kill the process

components/system-broker/internal/director/pager.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil
+	}
+
+	p.PageToken = string(response.Result.PageInfo.EndCursor)
+
+	return nil
+}
+
+func (p *Pager) HasNext() bool {
+	return p.hasNext
```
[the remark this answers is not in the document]
pankrator: Don't think that paging can be called simultaneously

components/system-broker/internal/director/pager.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (p *Pager) ListAll(ctx context.Context, output interface{}) error {
```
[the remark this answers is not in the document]
pankrator: Not sure that parallel fetching can be done as you need the previous page to fetch the next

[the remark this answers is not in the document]
pankrator: If you have any suggestions we can discuss them

components/system-broker/internal/director/pager.go
```diff
@@ -0,0 +1,98 @@
...
+	Data       interface{}       `json:"data"`
+	PageInfo   *graphql.PageInfo `json:"pageInfo"`
+	TotalCount int               `json:"totalCount"`
+}
+
+type GenericOutput struct {
+	Result *GenericPage `json:"result"`
+}
+
+type Pager struct {
```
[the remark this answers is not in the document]
pankrator: The idea is that this pager is generic by itself and should be put in the director component. At the moment I don't see the benefit to put in the director, but at the moment it will bring more complexity in my opinion

components/system-broker/tests/common/fake_router.go
krasish: response is changed on line 112 so write protection is needed.

KirilKabakchiev (approved):

KirilKabakchiev (approved):

PR #1579, opened 2020-10-27
Refactor correlation ID handling in director

desislavaa:
**Description**

* Use a `Transport` that attaches all headers that can be used for distributed tracing when the request is being executed, rather than wrapping the request creation. 
* Create an `EnvoyFilter` that takes care of transforming the correlation ID headers coming from outside components into `x-request-id` that is used across the Compass ecosystem.
More info on distributed tracing with Istio and how we can take advantage of it:
  * https://istio.io/latest/faq/distributed-tracing/#how-to-support-tracing
  * https://istio.io/latest/docs/tasks/observability/distributed-tracing/


components/director/pkg/http/types.go
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package http
+
+import (
+	"net/http"
+)
+
+//go:generate mockery -name=HTTPRoundTripper -output=automock -outpkg=automock -case=underscore
```
desislavaa, line 23: Pros of having a different repo with "util" interfaces and stuff is the fact that we can reuse the existing mocking mechanism in `system-broker` (counterfeiter) and test frame with ginkgo and gomega. If we move everything to `director` we'll have to either:
- use mockery and vanilla tests, meaning that we should rewrite all existing tests
- the mechanisms diverge until we move to counterfeiter and ginkgo everywhere

KirilKabakchiev: lets keep it here, for now. Let's see in the future when we have more utils if it will make sense to have a separate component

components/director/pkg/correlation/middleware.go
```diff
@@ -1,43 +1,84 @@
...
 
-var headerKeys = []string{
-	RequestIDHeaderKey,
-	"X-Request-ID",
-	"X-Request-Id",
-	"X-Correlation-ID",
-	"X-CorrelationID",
-	"X-ForRequest-ID"}
+// headerKeys are the expected headers that are used for distributed tracing.
+var headerKeys = []string{"x-request-id", "x-b3-traceid", "x-b3-spanid", "x-b3-parentspanid", "x-b3-sampled", "x-b3-flags", "b3"}
```
desislavaa, line 32: Should we make these configurable as well?

chart/compass/values.yaml
```diff
@@ -153,6 +153,15 @@ global:
       namespace: "compass-system"
     proxy:
       port: 15020
+    namespace: istio-system
+    ingressgateway:
+      workloadLabel: istio-ingressgateway
+      correlationHeaderRewriteFilter:
+        expectedHeaders:
+        - "x-request-id"
```
desislavaa, line 161: The headers entering the filter are all lowercased

chart/compass/templates/rewrite-correlation-headers-filter.yaml
```diff
@@ -0,0 +1,38 @@
...
+            {{ range .Values.global.istio.ingressgateway.correlationHeaderRewriteFilter.expectedHeaders }}headers['{{ . }}']=true
+            {{ end }}
+            return headers
+          end
+          function envoy_on_request(request_handle)
+            local expectedHeaders = expectedCorrelationHeaders()
+            local reqHeaders = request_handle:headers()
+            local newRequestIDHeaderValue=''
+            for key, value in pairs(reqHeaders) do
+              if (expectedHeaders[key] and newRequestIDHeaderValue=='')then
```
desislavaa, line 23: we need the `and newRequestIDHeaderValue==''` instead of breaking the loop when we find the first occurrence because we cannot break the loop :) https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter#pairs

components/director/pkg/correlation/middleware.go
```diff
@@ -1,43 +1,89 @@
...
+type Headers map[string]string
 
-type ContextEnrichMiddleware struct{}
+//AttachCorrelationIDToContext returns middleware that attaches all headers used for tracing in the current request.
+func AttachCorrelationIDToContext() func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+			if correlationHeaders := HeadersForRequest(r); len(correlationHeaders) != 0 {
+				ctx = context.WithValue(ctx, HeadersContextKey, correlationHeaders)
```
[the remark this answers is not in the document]
desislavaa: Absolutely, thank you for noticing this! 🔝

components/director/pkg/correlation/middleware.go
```diff
@@ -1,43 +1,89 @@
...
-	RequestIDHeaderKey,
-	"X-Request-ID",
-	"X-Request-Id",
-	"X-Correlation-ID",
-	"X-CorrelationID",
-	"X-ForRequest-ID"}
+// headerKeys are the expected headers that are used for distributed tracing.
+var headerKeys = []string{"x-request-id", "x-b3-traceid", "x-b3-spanid", "x-b3-parentspanid", "x-b3-sampled", "x-b3-flags", "b3"}
+
+type Headers map[string]string
```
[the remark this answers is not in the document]
desislavaa: Not really, we can use the one in the `http` packet, or you mean skip the wrapping and just use the map instead?

components/director/pkg/correlation/middleware.go
KirilKabakchiev: maybe we could. just one question - how did the tracing work properly before as we didnt propagate these ?

chart/compass/templates/rewrite-correlation-headers-filter.yaml
KirilKabakchiev: looks ok, just add this as a comment in the lua script maybe for further reference ?

components/director/pkg/correlation/middleware.go
desislavaa: It worked only for `gateway` -> `director`/`connector` requests because the gateway only proxies the request with all of its headers.

The values in the lua script are the correlation IDs that might come from outside components,
The values here are mentioned in the [Istio documentation](https://istio.io/latest/faq/distributed-tracing/#how-to-support-tracing) and are used for tracing within the cluster.
I'll add a new config entry then.

chart/compass/values.yaml
KirilKabakchiev: this might be a bit of a stretch but i remember for cf there was some vcap id or something, that was passed. Should we consider it as well ?

components/director/pkg/correlation/middleware.go
KirilKabakchiev: i guess this explains why there was no arrow in kiali between director and connector. Basically with this fix we should get this arrow, right ?
image

desislavaa: Exactly!

desislavaa: image

KirilKabakchiev (approved):

PR #1577, opened 2020-10-26
Update dependency to golang/x/text to v0.3.3

pankrator:

PR #1582, opened 2020-10-28
Increase golang/x/text dependency to v0.3.3

PetarTodorovv:
Increase golang.org/x/text dependency to version 0.3.3


desislavaa (approved):

desislavaa (changes requested):
Check the changes for `pairing-adapter` and `external-services-mock` in https://github.com/kyma-incubator/compass/pull/1577
We can also set the scan results to "false positive" for `external-services-mock` instead if updating the version, but this might be the easier approach.

desislavaa (approved):

PetarTodorovv:
> Check the changes for `pairing-adapter` and `external-services-mock` in #1577
> We can also set the scan results to "false positive" for `external-services-mock` instead if updating the version, but this might be the easier approach.

PetarTodorovv:
> Check the changes for `pairing-adapter` and `external-services-mock` in #1577
> We can also set the scan results to "false positive" for `external-services-mock` instead if updating the version, but this might be the easier approach.

Discussed offline

desislavaa (approved):

PR #1583, opened 2020-10-28
Remove the PoCs and investigations from the compass repo

ptodorovv:
**Description**
The team decided to remove the PoCs, investigations, and different researches from the compass repo to the dedicated repo.


NickyMateev (approved):

KirilKabakchiev:
What is the plan about this ? Where are we moving it ? Isn't it better to have some transparancy about the code designs there or are we going to have a separate OSS repo with those ?

DimitarPetrov:
Discussed separately and decided to leave internal investigations for now.

PR #1586, opened 2020-10-29
Ensure deletion of Runtime also deletes any associated Automatic Scenario Assignments

NickyMateev:
# Ensure deletion of Runtime also deletes any associated Automatic Scenario Assignments

## Overview

Runtimes are usually labeled with specific tenant labels denoting to which tenant they belong.

Automatic scenario assignments allow putting runtimes in scenarios by simply knowing to which tenant they belong - no need to retrieve the actual runtime and label it manually. Usually it's only necessary to specify your _scenario name_, _label key_ and _label value_ to create the scenario assignment. The _label value_ is usually some GUID which denotes the tenant ID and with which the runtime is already labeled.

When a runtime is unregistered from Compass any existing automatic scenario assignments targeting that runtime by means of the tenant label become obsolete and there's no use of such scenario assignments.

There's actually one situation in which it would be useful to keep the automatic scenario assignment and that is if the tenant decides to delete his runtime and then create a new one in the context of his tenant. In such situations it might not be the best to delete the scenario assignment as it could be reused again (with this change he'd have to create it again / place his new runtime in the respective scenario). However, from our observations it's more often the case that a runtime is deleted when the tenant is about to be deleted as well and so it doesn't make sense to keep the automatic scenario assignment when it is heavily tenant specific.

## Implementation

The implementation is based in the **unregisterRuntime** mutation which now makes sure to also:
1. Retrieve all scenarios in which the runtime is part of
2. Find any matching automatic scenario assignments based on the retrieved scenarios - these scenario assignments will be the ones who are responsible for labelling the runtime with the matching scenarios in the first place
3. Delete any found automatic scenario assignments


components/director/internal/domain/runtime/resolver_test.go
```diff
@@ -515,17 +554,453 @@ func TestResolver_DeleteRuntime(t *testing.T) {
...
+				svc.On("DeleteMultipleClientCredentials", contextParam, testAuths).Return(nil)
+
+				return svc
+			},
+			InputID:         "foo",
+			ExpectedRuntime: nil,
+			ExpectedErr:     testErr,
+		},
+		{
+			Name:            "Returns not found when listing scenarios label should succeed",
```
DimitarPetrov, line 590: Isn't this the same as the first test?

components/director/internal/domain/runtime/resolver_test.go
```diff
@@ -515,17 +554,453 @@ func TestResolver_DeleteRuntime(t *testing.T) {
...
+				svc.On("DeleteMultipleClientCredentials", contextParam, testAuths).Return(nil)
+
+				return svc
+			},
+			InputID:         "foo",
+			ExpectedRuntime: gqlRuntime,
+			ExpectedErr:     nil,
+		},
+		{
+			Name:            "Returns scenario when listing scenarios label and scenario assignment when listing scenario assignments and succeeds on deletion of scenario assignment should succeed",
```
DimitarPetrov, line 858: Isn't this duplicate with the one on line 770?

DimitarPetrov (approved):

components/director/internal/domain/runtime/resolver_test.go
NickyMateev: Good catch!

DimitarPetrov (dismissed):

DimitarPetrov (dismissed):

DimitarPetrov (approved):

PetarTodorovv (approved):

KirilKabakchiev:
Im not sure this is the correct way to go because the lifecycle of automatic scenario assignments is currently asymmentrical/inconsistent. They are created via a separate mutation during formation creation and in them they say nothing about a runtime. And now when a specific runtime is deleted this would delete any assignment that it uses. ASAs should probably be deleted when the formation from the managed UI is deleted and ofc the option that you provided in the OSS UI also makes sense. ASAs are part of the formation/scenario concept and their lifecycle should be tight to its. Provided that we keep ASAs, the correct way to go, in my opinion would be that deleting a scenario and a formation

PR #1590, opened 2020-11-01
Forbid deletion of Scenario in Compass UI if there are existing associated runtimes

NickyMateev:
# Overview

This PR fixes the bug where the **Delete** scenario button would be clickable if there are existing runtimes associated with the scenario.

The bug would allow clicking **Delete** which resulted in a failed **updateLabelDefinition** backend request ultimately failing to delete the scenario anyway (but the Compass UI makes it look for a moment as if it was successful, not showing any errors at all).

image

### DIFF
--- a/components/console/compass/src/components/Scenarios/ScenarioDetails/ScenarioDetails.js
+++ b/components/console/compass/src/components/Scenarios/ScenarioDetails/ScenarioDetails.js
@@ -13,12 +13,16 @@ ScenarioDetails.propTypes = {
 
 export default function ScenarioDetails({ scenarioName }) {
   const [applicationsCount, setApplicationsCount] = useState(0);
+  const [runtimesCount, setRuntimesCount] = useState(0);
 
   return (
     <ScenarioNameContext.Provider value={scenarioName}>
-      <ScenarioDetailsHeader applicationsCount={applicationsCount} />
+      <ScenarioDetailsHeader
+        applicationsCount={applicationsCount}
+        runtimesCount={runtimesCount}
+      />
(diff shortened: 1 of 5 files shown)


components/console/compass/src/components/Scenarios/ScenarioDetails/ScenarioDetails.js
```diff
@@ -13,12 +13,16 @@ ScenarioDetails.propTypes = {
 
 export default function ScenarioDetails({ scenarioName }) {
   const [applicationsCount, setApplicationsCount] = useState(0);
+  const [runtimesCount, setRuntimesCount] = useState(0);
```
KirilKabakchiev, line 16: looks ok in terms of fix, i was just wondering if for the long term solution, we would prefer to actually allow scenario deletion if there are apps and runtimes in it with a proper confirmation dialog and this would then result in unlabelling the apps and runtimes, updating the label definition as well as asa deletion

KirilKabakchiev (approved):

components/console/compass/src/components/Scenarios/ScenarioDetails/ScenarioDetails.js
NickyMateev: Maybe it makes sense, let's cross that bridge when we get there, though. There are a lot of considerations in what is right/wrong, which we have to discuss as a team to decide.

But I feel we're still far from such discussions as we haven't gotten too deep yet and haven't felt a need for that, hence the descision won't be that valuable.

NickyMateev:
Will be moved to https://github.com/kyma-incubator/compass-console

PR #1591, opened 2020-11-01
Use form3tech-oss/jwt-go instead of dgrijalva/jwt-go

PetarTodorovv:
**Description**
Replace github.com/dgrijalva/jwt-go v3.2.0 with the newer one github.com/form3tech-oss/jwt-go v3.2.2
The form3tech-oss is a fork of the dgrijalva library in which the security vulnerabilities are fixed.


components/director/internal/authenticator/claims.go
```diff
@@ -1,7 +1,7 @@
 package authenticator
 
 import (
-	"github.com/dgrijalva/jwt-go"
+	"github.com/form3tech-oss/jwt-go"
```
KirilKabakchiev, line 4: this is a fork with 0 stars. Im not sure how reliable it is to switch to such forks. Would it be possible to get the security fix in https://github.com/dgrijalva/jwt-go instead ?

PetarTodorovv: currently, there is no fix in the https://github.com/dgrijalva/jwt-go and we do not know when will be provided such one. So we use **github.com/form3tech-oss/jwt-go** which is a fork of the same library in which the problem is fix

KirilKabakchiev (approved):

PR #1593, opened 2020-11-01
Add E2E tests for HasScenario directive

NickyMateev:
# Overview

Some Kyma tests broke due to us not having E2E tests for the `HasScenario` directive. This PR adds such tests.

### DIFF
--- a/tests/director/api/application_api_test.go
+++ b/tests/director/api/application_api_test.go
@@ -477,33 +477,32 @@ func TestQuerySpecificApplication(t *testing.T) {
 		err = tc.RunOperation(context.Background(), queryAppReq, &actualApp)
 		saveExampleInCustomDir(t, queryAppReq.Query(), queryApplicationCategory, "query application")
 
-		//THEN
+		//THE
 		require.NoError(t, err)
 		assert.Equal(t, appID, actualApp.ID)
 	})
 
+	// update label definitions
+	ctx := context.Background()
+	tenantID := testTenants.GetDefaultTenantID()
+	scenarios := []string{defaultScenario, "test-scenario"}
+
+	jsonSchema := map[string]interface{}{
+		"type":        "array",
+		"minItems":    1,
+		"uniqueItems": true,
+		"items": map[string]interface{}{
+			"type": "string",
+			"enum": scenarios,
+		},
+	}
+	var schema interface{} = jsonSchema
+
+	updateLabelDefinitionWithinTenant(t, ctx, scenariosLabel, schema, tenantID)
+
 	t.Run("Query Application With Consumer Runtime in same scenario", func(t *testing.T) {
-		// update label definitions
-		ctx := context.Background()
-		tenantID := testTenants.GetDefaultTenantID()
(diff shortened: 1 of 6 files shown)


tests/director/api/application_api_test.go
```diff
@@ -465,17 +466,77 @@ func TestQuerySpecificApplication(t *testing.T) {
...
+	t.Run("Query Application With Consumer Runtime in same scenario", func(t *testing.T) {
+		// set application scenarios label
+		setApplicationLabel(t, ctx, appID, scenariosLabel, scenarios[1:])
+		defer setApplicationLabel(t, ctx, appID, scenariosLabel, scenarios[:1])
+
+		// set runtime scenarios label
+		setRuntimeLabel(t, ctx, runtime.ID, scenariosLabel, scenarios[1:])
+		defer setRuntimeLabel(t, ctx, runtime.ID, scenariosLabel, scenarios[:1])
+
+		// query application
```
KirilKabakchiev, line 510: query application comment here above the section of creating an empty app is abit confusing

tests/director/api/application_api_test.go
```diff
@@ -465,17 +466,77 @@ func TestQuerySpecificApplication(t *testing.T) {
...
+		err = runtimeConsumer.Run(queryAppReq, &actualApp)
+
+		//THEN
+		require.NoError(t, err)
+		assert.Equal(t, appID, actualApp.ID)
+	})
+
+	t.Run("Query Application With Consumer Runtime not in same scenario", func(t *testing.T) {
+		// set application scenarios label
+		setApplicationLabel(t, ctx, appID, scenariosLabel, scenarios[:1])
```
KirilKabakchiev, line 524: defer for cleanup like for setRuntimeLabel?

tests/director/api/application_api_test.go
```diff
@@ -465,17 +466,77 @@ func TestQuerySpecificApplication(t *testing.T) {
...
+
+		// query application
+		actualApp := graphql.Application{}
+
+		// WHEN
+		queryAppReq := fixApplicationRequest(appID)
+		err = runtimeConsumer.Run(queryAppReq, &actualApp)
+
+		//THEN
+		require.Error(t, err)
```
KirilKabakchiev, line 538: should we verify something about the error so that we can be sure it's the expected one ?

tests/director/api/application_api_test.go
NickyMateev: No need - the defer in the `setRuntimeLabel` reverts to `scenarios[:1]` which matches to `DEFAULT` scenario.
On this current line `setApplicationLabel` sets it to `DEFAULT` for the test purpose, so no need to defer/cleanup anything.

tests/director/api/application_api_test.go
NickyMateev: Sure, but I'll put an error message which might not be that specific, but will be descriptive enough, as the specific error message are not that stable.

ataleksandrov (approved):

KirilKabakchiev (approved):

PR #1599, opened 2020-11-03
Introduce common logging interface

DimitarPetrov:
A separate PR will be created for the adoption of these changes in the connector.


components/director/pkg/log/formatter.go
```diff
@@ -0,0 +1,102 @@
...
+			return nil, err
+		}
+		result[k] = field
+	}
+
+	return json.Marshal(result)
+}
+
+// Formatter is a logrus formatter that formats an entry for Kibana
+type Formatter struct {
```
KirilKabakchiev, line 64: wondering if this formatter still brings value for the integration with the cloud logging kibana? have you tried how the log entries look with and without it ?

DimitarPetrov: In Cloud Logging Kibana we don't have any predefined dashboards, so it is up to us to define ones based on our preferred format. 
However, this format is established already and I think it is better to have it in the beginning in case we want to start fast with already familiar dashboards and enrich them in the future to match our preferred format.

KirilKabakchiev (approved):

PR #1609, opened 2020-11-06
System Broker Bind API

NickyMateev:
# System Broker Bind API

This PR productizes the System Broker Bind API. This includes:

- Moving the current Provisioning logic to the Bind handlers;
- Introducing two new queries `packageInstanceAuth(id: ID!)` and `packageByInstanceAuth(id: ID!)`, and the mutation `createAutomaticScenarioAssignment` is modified to allow specifying an ID of the newly created Package Instance Auth;
- Component tests for all Bind API handlers
- E2E tests for the new queries and the modified mutation


components/system-broker/internal/osb/bind_get.go
```diff
@@ -18,17 +18,71 @@ package osb
...
+			"binding_id":  bindingID,
+		},
+	})
+
+	if err != nil && !IsNotFoundError(err) {
+		return domain.GetBindingSpec{}, errors.Wrapf(err, "while getting package instance credentials from director")
+	}
+
+	if IsNotFoundError(err) {
+		logger.Info("Package credentials for binding were not found")
```
pankrator, line 59: may be debug, as the same error is returned to the client calling the broker

components/system-broker/internal/osb/bind_get.go
```diff
@@ -18,17 +18,71 @@ package osb
...
 func (b *GetBindingEndpoint) GetBinding(ctx context.Context, instanceID, bindingID string) (domain.GetBindingSpec, error) {
 	log.C(ctx).Infof("GetBindingEndpoint instanceID: %s bindingID: %s", instanceID, bindingID)
 
-	return domain.GetBindingSpec{}, errors.New("not supported")
+	logger := log.C(ctx).WithFields(map[string]interface{}{
+		"instanceID": instanceID,
+		"bindingID":  bindingID,
+	})
+
+	logger.Info("Fetching package instance credentials")
```
pankrator, line 44: Debug

components/system-broker/internal/osb/bind_last_op.go
```diff
@@ -19,16 +19,90 @@ package osb
...
+			"instance_id": instanceID,
+			"binding_id":  bindingID,
+		},
+	})
+	if err != nil && !IsNotFoundError(err) {
+		return domain.LastOperation{}, errors.Wrapf(err, "while getting package instance credentials from director")
+	}
+
+	if IsNotFoundError(err) {
+		logger.Info("Package instance credentials not found")
```
pankrator, line 64: This is something expected to happen if the op is Unbind, but is abnormal if the operation is Bind. May be put the log on Error level is move it after the If.

pankrator (changes requested):

components/system-broker/internal/director/client.go
```diff
@@ -103,170 +104,124 @@ func (c *GraphQLClient) RequestPackageInstanceCredentialsCreation(ctx context.Co
...
-			authContextValue, found := authContext[key]
-			if !found || authContextValue != value {
-				shouldReturn = false
-			}
-		}
-
-		if shouldReturn {
-			authsResp = append(authsResp, auth)
-		}
+	if response.Package == nil || response.Package.InstanceAuth == nil || response.Package.InstanceAuth.Context == nil {
```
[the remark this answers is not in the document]
ataleksandrov: It can not be nil, see: https://github.com/kyma-incubator/compass/blob/system-broker-bind-api/components/system-broker/internal/director/client.go#L150

PR #1626, opened 2020-11-16
Enhance Auditlog logging in Gateway 

danielgospodinow:
**Description**
This PR introduces an additional Auditlog logging step before performing an operation in the Gateway.


components/gateway/internal/auditlog/service.go
```diff
@@ -146,8 +180,6 @@ func (svc *Service) parseResponse(response string) (model.GraphqlResponse, error
 func (svc *Service) createConfigChangeMsg(claims proxy.Claims, request string) model.ConfigurationChange {
 	msg := svc.msgFactory.CreateConfigurationChange()
 	msg.Object = model.Object{ID: fillID(claims, "Config Change")}
-	msg.Attributes = []model.Attribute{
```
pankrator, line 149: Better leave this as it was. Also remove the places where this attribute is added by hand

components/gateway/internal/auditlog/service.go
```diff
@@ -77,60 +64,107 @@ func NewService(client AuditlogClient, msgFactory AuditlogMessageFactory) *Servi
...
+	configChangeMsg := svc.createConfigChangeMsg(msg.Claims, msg.Request)
 	if isReadErr {
-		msg.Attributes = append(msg.Attributes, model.Attribute{
-			Name: "response",
-			Old:  "",
-			New:  "success",
-		})
-		err := svc.client.LogConfigurationChange(ctx, msg)
+		configChangeMsg.Attributes = append(configChangeMsg.Attributes,
+			model.Attribute{
```
pankrator, line 135: Remove this as it is added by the `createConfigChangeMsg `

components/gateway/internal/auditlog/service.go
```diff
@@ -77,60 +64,107 @@ func NewService(client AuditlogClient, msgFactory AuditlogMessageFactory) *Servi
...
-		msg.Attributes = append(msg.Attributes, model.Attribute{
-			Name: "response",
-			Old:  "",
-			New:  response,
-		})
-	}
-
-	err = svc.client.LogConfigurationChange(ctx, msg)
+		configChangeMsg.Attributes = append(configChangeMsg.Attributes,
+			model.Attribute{
```
pankrator, line 152: Remove this as it is added by the `createConfigChangeMsg `

components/gateway/internal/auditlog/service.go
```diff
@@ -77,60 +64,107 @@ func NewService(client AuditlogClient, msgFactory AuditlogMessageFactory) *Servi
...
+				Name: "request",
+				Old:  "",
+				New:  msg.Request,
+			}, model.Attribute{
+				Name: "response",
+				Old:  "",
+				New:  "success",
+			}, model.Attribute{
```
pankrator, line 143: Same as above correlation id. Check all places where correlation id is added as attribute and move it to the constructor

components/gateway/internal/auditlog/service.go
```diff
@@ -77,60 +64,107 @@ func NewService(client AuditlogClient, msgFactory AuditlogMessageFactory) *Servi
...
-			Old:  "",
-			New:  "success",
-		})
+	correlationID := msg.CorrelationIDHeaders[correlation.RequestIDHeaderKey]
 
-		err = svc.client.LogConfigurationChange(ctx, log)
+	if len(graphqlResponse.Errors) == 0 {
+		configChangeMsg := svc.createConfigChangeMsg(msg.Claims, msg.Request)
+		configChangeMsg.Attributes = append(configChangeMsg.Attributes,
+			model.Attribute{
```
pankrator, line 96: Remove this as it is added by the `createConfigChangeMsg`

components/gateway/internal/auditlog/service.go
```diff
@@ -77,60 +64,107 @@ func NewService(client AuditlogClient, msgFactory AuditlogMessageFactory) *Servi
...
+				Name: "request",
+				Old:  "",
+				New:  msg.Request,
+			}, model.Attribute{
+				Name: "response",
+				Old:  "",
+				New:  "success",
+			}, model.Attribute{
```
pankrator, line 104: Correlation ID seems something that will always be added to a message, so better add it to the constructor - `createConfigChangeMsg`

components/gateway/internal/auditlog/service.go
```diff
@@ -77,60 +64,107 @@ func NewService(client AuditlogClient, msgFactory AuditlogMessageFactory) *Servi
...
+
+		err = svc.client.LogConfigurationChange(ctx, configChangeMsg)
 		return errors.Wrap(err, "while sending to auditlog")
 	}
 
 	if svc.hasInsufficientScopeError(graphqlResponse.Errors) {
-		msg := svc.msgFactory.CreateSecurityEvent()
-		eventData := model.SecurityEventData{ID: fillID(claims, "Security Event"), Reason: graphqlResponse.Errors}
+		securityEventMsg := svc.msgFactory.CreateSecurityEvent()
+		eventData := model.SecurityEventData{ID: fillID(msg.Claims, "Security Event"), Reason: graphqlResponse.Errors}
```
pankrator, line 116: Add the correlation id in the security event data as it is currently missing in the security event

components/gateway/pkg/proxy/transport.go
```diff
@@ -12,30 +12,44 @@ import (
...
+}
+
+//go:generate mockery --name=PreAuditlogService --output=automock --outpkg=automock --case=underscore
+type PreAuditlogService interface {
+	AuditlogService
+	PreLog(ctx context.Context, msg AuditlogMessage) error
+}
+
+type AuditlogMessage struct {
+	CorrelationIDHeaders correlation.Headers
```
pankrator, line 37: The auditlog message only needs one correlation id, so better provide only a correlation id string

components/gateway/pkg/proxy/transport.go
```diff
@@ -44,12 +58,37 @@ func (t *Transport) RoundTrip(req *http.Request) (resp *http.Response, err error
...
 	}
-
 	req.Body = ioutil.NopCloser(bytes.NewBuffer(requestBody))
+	defer httpcommon.CloseBody(req.Body)
+
+	correlationHeaders := correlation.HeadersForRequest(req)
+
+	isMutation, err := checkQueryType(requestBody, "mutation")
+	if (err == emptyQuery) || (err == nil && isMutation) {
+		if logService, ok := t.auditlogSvc.(PreAuditlogService); ok {
```
pankrator, line 73: As discussed this can be simplified

pankrator (changes requested):

components/gateway/internal/auditlog/service.go
```diff
@@ -7,38 +7,25 @@ import (
 	"strings"
 	"time"
 
+	"github.com/kyma-incubator/compass/components/director/pkg/correlation"
 	"github.com/kyma-incubator/compass/components/gateway/pkg/auditlog/model"
 	"github.com/kyma-incubator/compass/components/gateway/pkg/proxy"
 	"github.com/pkg/errors"
 )
 
-type Message struct {
```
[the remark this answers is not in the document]
danielgospodinow: We've decided to move it up in the proxy package due to circular a dependency problem. We couldn't instantiate an auditlog.Message instance because it carries in itself a proxy.Claims instance.

components/gateway/pkg/proxy/transport.go
danielgospodinow: Yes, that's true! But we've left the message to carry all of the correlation headers because of the new child context creation in the auditlog queue workers upon calls to auditlog service. This can be seen here: https://github.com/kyma-incubator/compass/pull/1626/files#diff-252a4af9a7b98d32685344f445c56cc7ef76a7c94ca896111f515fbdec3a16afR36

components/gateway/pkg/proxy/transport.go
```diff
@@ -68,22 +111,14 @@ func (t *Transport) RoundTrip(req *http.Request) (resp *http.Response, err error
...
-		if !isMutation {
-			log.Println("Will not auditlog queries")
-			return resp, nil
-		}
-	}
-
-	correlationHeaders := correlation.HeadersForRequest(req)
-	correlationCtx := context.WithValue(context.Background(), correlation.HeadersContextKey, correlationHeaders)
-	err = t.auditlogSvc.Log(correlationCtx, string(requestBody), string(responseBody), claims)
+	err = t.auditlogSink.Log(context.TODO(), AuditlogMessage{
```
pankrator, line 114: Use `req.Context()` here, no matter that it is not used at the moment by the Sink implementation. It might be used in future when logging in enhanced

pankrator (approved):

PR #1630, opened 2020-11-25
Add grafana dashboard for gateway

pankrator:
**Description**

- collect generic go metrics for gateway
- collect auditlog channel length metric for gateway
- Create Grafana dashboard with the collected gateway metrics

image


components/gateway/internal/auditlog/worker.go
```diff
@@ -33,6 +35,7 @@ func (w *Worker) Start() {
 			return
 		case msg := <-w.auditlogChannel:
 			log.Printf("Read from auditlog channel (size=%d, cap=%d)", len(w.auditlogChannel), cap(w.auditlogChannel))
+			w.collector.SetChannelSize(len(w.auditlogChannel))
```
danielgospodinow, line 38: Can there be a race condition here? Or those Prometheus objects (prometheus.Gauge in our case) are synchronized?

chart/compass/charts/gateway/values.yaml
```diff
@@ -15,3 +15,6 @@ gateway:
   auditlog:
     enabled: false
     authMode: "basic"
+
+metrics:
+  port: 3001
```
danielgospodinow, line 20: I assume that the ports here should somehow correspond to the default values described in `main.go`'s `config` struct.

So isn't this supposed to be `port: 3003` as described in `main.go` at line 39 `MetricsAddress  string envconfig:"default=127.0.0.1:3003"` ? 

Port `3001` seems to be assigned already to the director proxy service (`DirectorOrigin  string envconfig:"default=http://127.0.0.1:3001"`).

components/gateway/internal/auditlog/worker.go
pankrator: prometheus Gauge is safe for concurrent use as it uses atomic operations

chart/compass/charts/gateway/values.yaml
pankrator: It is using port 3003 by default, as to not clash with the director local port, which is 3001. Anyway it is overwritten by the values.yaml in the chart.

danielgospodinow (dismissed):

chart/compass/charts/gateway/templates/policy.yaml
```diff
@@ -0,0 +1,12 @@
+# Required because Prometheus Operator doesn't have Istio Sidecar
+apiVersion: security.istio.io/v1beta1
+kind: PeerAuthentication
```
NickyMateev, line 3: Why is this resource needed?

components/gateway/internal/auditlog/service_test.go
```diff
@@ -285,7 +285,7 @@ func TestSink_TimeoutOnWrite(t *testing.T) {
 
 	chanMsg := make(chan proxy.AuditlogMessage)
 	defer close(chanMsg)
-	sink := auditlog.NewSink(chanMsg, time.Millisecond*100)
+	sink := auditlog.NewSink(chanMsg, time.Millisecond*100, &FakeMetricCollector{})
```
NickyMateev, line 288: Does it make sense to have some primitive unit tests for this AuditlogCollector? Like for example some assertions that when new auditlog is received some of the AuditlogCollector's methods (Describe, Collect, SetChannelSize) are invoked with certain arguments

chart/compass/charts/gateway/templates/policy.yaml
pankrator: It is required, because the prometheus server tries to call the metrics endpoint, but the prometheus server does not have a sidecar. This peerauthentication allows the prometheus to call the metrics endpoint without providing a certificate to the sidecar.

components/gateway/internal/auditlog/service_test.go
pankrator: sure, I can add one. Actually the meaningful methods to be called are only the `SetChannelSize` as the others are just to implement an interface required by prometheus package

NickyMateev (approved):

PR #1634, opened 2020-11-30
Introduce Application Name Normalization

NickyMateev:
# Overview

**Description**
 
Adds ability to add custom normalization to application names when applications are being retrieved by runtimes.
When normalization is enabled (**APP_NAME_NORMALIZATION_ENABLED = true**) the DNS validation on an application's name during registration in Compass is disabled.


components/director/pkg/graphql/app_validation.go
```diff
@@ -1,19 +1,42 @@
...
+	value, isSet := os.LookupEnv(appNormalizationEnvVar)
+	if isSet {
+		parsedValue, err := strconv.ParseBool(value)
+		if err == nil {
+			appNameNormalizationEnabled = parsedValue
+		}
+	}
+}
+
 func (i ApplicationRegisterInput) Validate() error {
```
desislavaa, line 27: Should we also add a test case in `app_validation_test.go`?

PetarTodorovv (approved):

components/director/pkg/graphql/app_validation.go
NickyMateev: Not possible due to the fact that app_validation.go has `init()` method which is called before any tests so we cannot plug the env variable with the right value.

NickyMateev:
Decommitted.

PR #1636, opened 2020-11-30
Change logging with the new logging interface in Director

StanislavStefanov:
**Description**

- Change logging in Director using the new logging interface


pankrator (changes requested):

components/director/internal/metrics/pusher.go
```diff
@@ -28,7 +30,7 @@ func NewPusher(endpoint string, timeout time.Duration) *Pusher {
 	}, []string{"method", "code", "desc"})
 
 	instanceID := uuid.New()
-	log.WithField(InstanceIDKeyName, instanceID).Infof("Initializing Metrics Pusher...")
+	log.D().WithField(InstanceIDKeyName, instanceID).Infof("Initializing Metrics Pusher...")
```
DimitarPetrov, line 33: Why do you decided to use default logger here and not to propagate a context to the pusher?

components/director/internal/tenantfetcher/client.go
```diff
@@ -92,7 +92,7 @@ func (c *Client) FetchTenantEventsPage(eventsType EventsType, additionalQueryPar
 	defer func() {
 		err := res.Body.Close()
 		if err != nil {
-			log.Warnf("Unable to close response body. Cause: %v", err)
+			log.C(context.Background()).Warnf("Unable to close response body. Cause: %v", err)
```
DimitarPetrov, line 95: `log.D().Warnf(...)`

components/director/internal/tenantmapping/staticgroup.go
```diff
@@ -19,7 +19,7 @@ type StaticGroups []StaticGroup
 
 //go:generate mockery -name=StaticGroupRepository -output=automock -outpkg=automock -case=underscore
 type StaticGroupRepository interface {
-	Get(groupnames []string) StaticGroups
+	Get(groupnames []string, logger *logrus.Entry) StaticGroups
```
DimitarPetrov, line 22: It is clearer and more consistent to receive a context instead of logger.

DimitarPetrov (changes requested):
There is a generic logic for error location in the error logs. This logic can be found in director here: https://github.com/kyma-incubator/compass/blob/master/components/director/pkg/log/error_hook.go

In short this extracts the error and it's trace to add in the log location of the error in the code. 
In order for this to work you should add the error when logging with `log.C(ctx).WithError(err).Errorf(...)`
You can try to address this by searching all the Error level logs and attach the error for better clarity.

components/director/internal/metrics/pusher.go
StanislavStefanov: Pusher is created only in TenantFetcher. The TenantFetcher uses context.Background() and ads to it log with configuration. Then this configuration is applyed to the default logger as there is no logger in this context. Nothing else is added to the context in TenantFetcher. So passing context to pusher and using the logger from it will be equivalent to using the default logger directly.

DimitarPetrov (approved):

pankrator (approved):

PR #1637, opened 2020-11-30
Add claims with consumerId to Auditlog message when preAuditloging

StanislavStefanov:
**Description**

- Add claims with consumerId to Auditlog message when preAuditloging
- Add checking for consumerId in tests


tests/director/external-services-mock-integration/auditlog_integration_test.go
```diff
@@ -56,7 +58,25 @@ func TestAuditlogIntegration(t *testing.T) {
...
+			if attr.Name == "auditlog_type" && attr.New == "pre-operation" {
+				pre = v
+			}
+			if attr.Name == "auditlog_type" && attr.New == "post-operation" {
+				post = v
+			}
+		}
+	}
+
+	assert.Equal(t, requestBody.String(), pre.Attributes[1].New)
```
NickyMateev, line 74: I don't think it's the best of ideas to access the `request` attribute by the 1-st index directly here. We know that it's the 1st, because we know the code on the other side, but for someone looking at this test for the first time and not knowing anything about our auditlog integration codebase and how we order the attributes - this will be hard to understand and derive.
So even if it would take some more lines of code it would be better practice to find the right attribute intstead of hoping we keep it at the 1st index forever and always.

NickyMateev (changes requested):

NickyMateev (approved):

PR #1638, opened 2020-12-01
Apply normalization based on Runtime label instead of env variable

NickyMateev:
# Apply normalization based on Runtime label

**Description**

Apply normalization based on Runtime label instead of env variable.


chart/compass/values.yaml
```diff
@@ -23,7 +23,7 @@ global:
       version: "PR-1618"
     director:
       dir:
-      version: "PR-1618"
+      version: "PR-1634"
```
NickyMateev, line 26: Should fix this

NickyMateev: done

desislavaa (approved):

desislavaa (approved):

PR #1641, opened 2020-12-02
Add Normalization on EventingConfiguration

NickyMateev:
**Description**

After introducing https://github.com/kyma-incubator/compass/pull/1638 it was left out that LoBs can retrieve eventing URL for given applications. These URLs include application names which should be normalized as they will be processed by Kyma Runtimes and the runtime will expect a normalized name in the corresponding eventing resources (which are based on the name provided in the corresponding Application CR).

**Related issue(s)**
https://github.com/kyma-incubator/compass/pull/1638

**Kyma Eventing Mesh Architecture**
https://kyma-project.io/docs/1.17/components/event-mesh#architecture-architecture


components/director/internal/domain/eventing/service.go
```diff
@@ -18,6 +20,7 @@ import (
 )
 
 const (
+	shouldNormalizeLabel         = "shouldNormalize"
```
KirilKabakchiev, line 23: isNormalized ?

tests/connectivity-adapter/test/api/connector_test.go
```diff
@@ -314,8 +336,8 @@ func appMgmInfoEndpointSuite(t *testing.T, directorClient director.Client, appID
 		require.Nil(t, errorResponse)
 
 		// then
-		assert.Equal(t, expectedMetadataURL, mgmInfoResponse.URLs.MetadataUrl)
-		assert.Equal(t, expectedEventsURL, mgmInfoResponse.URLs.EventsUrl)
+		assert.Equal(t, expectedMetadataURL, mgmInfoResponse.URLs.MetadataURL)
+		assert.Equal(t, expectedEventsURL, mgmInfoResponse.URLs.EventsURL)
 		assert.Equal(t, appName, mgmInfoResponse.ClientIdentity.Application)
```
KirilKabakchiev, line 341: set to app.label.name (normalized name) for backwards compatilibty

components/director/internal/domain/eventing/service.go
NickyMateev: I don't see how it fits what we're trying to convey. By **shouldNormalize** we mean that Applications (more precisely their names) _should be normalized_ for a specific runtime. We don't say that they _are normalized_ (isNormalized) or that the runtime itself _is normalized_?

KirilKabakchiev: from kyma agent PoV they will interpret shouldNormalize label as "normalization already done, skip" === isNormalized=true or atleast that was my understanding after the sync today. If the label is called shouldNormalize and has value true and the kyma agent skips normalization because we do it internally when shoulldnormalize=true, its kind of confusing (a client skips normalization when shouldNormalize is true).

NickyMateev: Renaming has been done.

KirilKabakchiev (approved):

PR #1646, opened 2020-12-05
Introduce runtime protected labels

pankrator:
**Introduce protected labels for runtimes. This will make possible to have hidden/protected metadata on runtime entity**

- Do not allow to delete protected labels
- Do not return protected labels
- Do not allow to set/update protected labels


components/director/internal/domain/runtime/service.go
```diff
@@ -337,7 +360,39 @@ func (s *service) ListLabels(ctx context.Context, runtimeID string) (map[string]
...
 	}
 
-	return labels, nil
+	return s.extractNotProtectedLabels(labels, s.protectedLabelPattern)
+}
+
+func (s *service) extractNotProtectedLabels(labels map[string]*model.Label, protectedLabelsKeyPattern string) (map[string]*model.Label, error) {
+	result := make(map[string]*model.Label)
+	for labelKey, label := range labels {
+		if protected, err := s.isProtected(labelKey, protectedLabelsKeyPattern); err != nil {
```
danielgospodinow, line 369: I got a bit confused by this inline variable initialization and the error check at the same time because the "if-else if" construction here is somewhat logically strange.

I can suggest a cleaner approach but this is just a personal preference of mine.

```suggestion
protected, err := s.isProtected(labelKey, protectedLabelsKeyPattern)
if err != nil {
    return nil, err
} 

if !protected {
    result[labelKey] = label
}
```

Note: this is also valid for `func (s *service) unsafeExtractNotProtectedLabels`

components/director/internal/domain/runtime/service.go
```diff
@@ -34,6 +36,7 @@ type LabelRepository interface {
 	ListForObject(ctx context.Context, tenant string, objectType model.LabelableObject, objectID string) (map[string]*model.Label, error)
 	Delete(ctx context.Context, tenant string, objectType model.LabelableObject, objectID string, key string) error
 	DeleteAll(ctx context.Context, tenant string, objectType model.LabelableObject, objectID string) error
+	DeleteByKeyNotPattern(ctx context.Context, tenant string, objectType model.LabelableObject, objectID string, labelKeyPattern string) error
```
NickyMateev, line 39: Wonder if we can get an easier to understand name - it took me like 5-10 seconds by just looking at the interface (without seeking the implmentation).
Maybe - `DeleteByKeyNegationPattern`?

components/director/internal/domain/label/repository.go
```diff
@@ -165,6 +167,14 @@ func (r *repository) DeleteAll(ctx context.Context, tenant string, objectType mo
 	return errors.Wrapf(err, "while deleting all Label entities from database for %s %s", objectType, objectID)
 }
 
+func (r *repository) DeleteByKeyNotPattern(ctx context.Context, tenant string, objectType model.LabelableObject, objectID string, labelKeyPattern string) error {
+	return r.deleter.DeleteMany(ctx, tenant, repo.Conditions{
+		repo.NewEqualCondition(labelableObjectField(objectType), objectID),
+		repo.NewEqualCondition(tenantColumn, tenant),
+		repo.NewNotRegexConditionString("key", labelKeyPattern),
```
NickyMateev, line 174: Extract `key` as the tenantColumn constant on the above line for consistency and better comprehension

components/director/internal/domain/runtime/service.go
```diff
@@ -182,6 +190,13 @@ func (s *service) Create(ctx context.Context, in model.RuntimeInput) (string, er
 		in.Labels[ShouldNormalize] = "true"
 	}
 
+	log.C(ctx).Debugf("Removing protected labels. Labels before: %+v", in.Labels)
+	in.Labels, err = s.unsafeExtractNotProtectedLabels(in.Labels, s.protectedLabelPattern)
+	if err != nil {
+		return "", err
+	}
+	log.C(ctx).Debugf("After removed protected labels: %+v", in.Labels)
```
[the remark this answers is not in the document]
NickyMateev: Or "Successfully stripped protected labels. Resulting labels after operation are: [...]"

components/director/internal/domain/runtime/service_test.go
```diff
@@ -245,7 +245,7 @@ func TestService_Create(t *testing.T) {
 			labelSvc := testCase.LabelUpsertServiceFn()
 			scenariosSvc := testCase.ScenariosServiceFn()
 			engineSvc := testCase.EngineServiceFn()
-			svc := runtime.NewService(repo, nil, scenariosSvc, labelSvc, idSvc, engineSvc)
+			svc := runtime.NewService(repo, nil, scenariosSvc, labelSvc, idSvc, engineSvc, ".*_defaultEventing$")
```
ataleksandrov, line 248: Maybe we could add tests for creating a runtime with provided protected labels and see that they are being ignored.

ataleksandrov (changes requested):

components/director/internal/domain/runtime/service.go
```diff
@@ -337,7 +360,39 @@ func (s *service) ListLabels(ctx context.Context, runtimeID string) (map[string]
 		return nil, errors.Wrap(err, "while getting label for Runtime")
 	}
 
-	return labels, nil
+	return s.extractNotProtectedLabels(labels, s.protectedLabelPattern)
+}
+
+func (s *service) extractNotProtectedLabels(labels map[string]*model.Label, protectedLabelsKeyPattern string) (map[string]*model.Label, error) {
```
NickyMateev, line 366: I know it's a good practice in most places, especially in the Java world to place dependant methods beneaf each other, but can we move the `extractNotProtectedLabels`, `unsafeExtractNotProtectedLabels` and `isProtected` utility functions below the `DeleteLabel` function? In most of this project I see that the pattern is to have all exported methods first and right after them are the utility/unexported methods. So just for consistency sake I was thinking along those lines - what do you think?

NickyMateev: Also maybe - `extractUnprotectedLabels` as name?

components/director/internal/domain/runtime/service.go
```diff
@@ -337,7 +360,39 @@ func (s *service) ListLabels(ctx context.Context, runtimeID string) (map[string]
...
+		if protected, err := s.isProtected(labelKey, protectedLabelsKeyPattern); err != nil {
+			return nil, err
+		} else if !protected {
+			result[labelKey] = label
+		}
+	}
+	return result, nil
+}
+
+func (s *service) unsafeExtractNotProtectedLabels(labels map[string]interface{}, protectedLabelsKeyPattern string) (map[string]interface{}, error) {
```
NickyMateev, line 378: `unsafeExtractUnprotectedLabels`?

NickyMateev (changes requested):

components/director/internal/domain/runtime/service.go
```diff
@@ -368,7 +423,14 @@ func (s *service) DeleteLabel(ctx context.Context, runtimeID string, key string)
...
 
-	if key != model.ScenariosKey {
+	protected, err := s.isProtected(key, s.protectedLabelPattern)
+	if err != nil {
+		return err
+	}
+	if protected {
+		return apperrors.NewInvalidDataError("could not delete protected label key %s", key)
+	}
+	if key != model.ScenariosKey && !protected {
```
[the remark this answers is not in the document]
pankrator: Sure, this was a left over because the above If was added later after the first check and forgot to remove it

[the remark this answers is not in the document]
NickyMateev: I agree that the check is unnecessary. But I'm also willing to say that it causes unneeded confusion as to what might be the reason for putting the additional `!protected`. Thoughts like "am I missing something, why is this here, when that's not really possible to take that value" 😄
So I'd suggest ditching the redundant check.

components/director/cmd/director/main.go
```diff
@@ -318,7 +321,7 @@ func getRuntimeMappingHandlerFunc(transact persistence.Transactioner, cachePerio
 	scenarioAssignmentRepo := scenarioassignment.NewRepository(scenarioAssignmentConv)
 	scenarioAssignmentEngine := scenarioassignment.NewEngine(labelUpsertSvc, labelRepo, scenarioAssignmentRepo)
 
-	runtimeSvc := runtime.NewService(runtimeRepo, labelRepo, scenariosSvc, labelUpsertSvc, uidSvc, scenarioAssignmentEngine)
+	runtimeSvc := runtime.NewService(runtimeRepo, labelRepo, scenariosSvc, labelUpsertSvc, uidSvc, scenarioAssignmentEngine, ".*_defaultEventing$")
```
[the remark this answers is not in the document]
NickyMateev: You never know if this instance of the `runtime service` would need the pattern. So just to be safe it's best to provide it.

components/director/internal/domain/runtime/service_test.go
```diff
@@ -269,7 +269,7 @@ func TestService_Create(t *testing.T) {
 
 	t.Run("Returns error on loading tenant", func(t *testing.T) {
 		// given
-		svc := runtime.NewService(nil, nil, nil, nil, nil, nil)
+		svc := runtime.NewService(nil, nil, nil, nil, nil, nil, ".*_defaultEventing$")
```
NickyMateev, line 272: You can provide an empty string here

components/director/cmd/director/main.go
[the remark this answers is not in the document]
pankrator: It is nothing special here, just forgot to use the configuration as it was later introduced ;)

components/director/internal/domain/runtime/service_test.go
NickyMateev: I agree - sounds like a realistic case

components/director/internal/domain/runtime/service_test.go
```diff
@@ -1672,6 +1701,37 @@ func TestService_SetLabel(t *testing.T) {
...
+		{
+			Name: "Returns an error when trying to set protected label",
+			RepositoryFn: func() *automock.RuntimeRepository {
+				repo := &automock.RuntimeRepository{}
+				repo.On("Exists", ctx, tnt, runtimeID).Return(true, nil).Once()
+				return repo
+			},
+			LabelUpsertServiceFn: func() *automock.LabelUpsertService {
+				svc := &automock.LabelUpsertService{}
+				// svc.On("UpsertLabel", ctx, tnt, &modelLabelInput).Return(nil).Once()
```
NickyMateev, line 1713: Leftover?

components/director/cmd/director/main.go
```diff
@@ -96,6 +96,8 @@ type config struct {
 	OAuth20      oauth20.Config
 
 	Features features.Config
+
+	ProtectedLabelPattern string `envconfig:"default=.*_defaultEventing"`
```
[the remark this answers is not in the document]
NickyMateev: Sure, multiple DB requests may be necessary (as you point out there is no "OR" operator in the storage layer) once we have multiple protected labels, but it will give us more flexibility going forward.

components/director/internal/domain/runtime/service.go
```diff
@@ -215,7 +230,8 @@ func (s *service) Update(ctx context.Context, id string, in model.RuntimeInput)
 		in.Labels[ShouldNormalize] = "true"
 	}
 
-	err = s.labelRepo.DeleteAll(ctx, rtmTenant, model.RuntimeLabelableObject, id)
+	// NOTE: The db layer does not support OR currently so multiple label patterns can't be implemented easily
```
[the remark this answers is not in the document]
pankrator: As discussed the provided in.Labels will be filtered out to not be created by accident

components/director/cmd/director/main.go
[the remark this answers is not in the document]
pankrator: Actually with one regex you can also match different single string values. I think this will cover the scenarios you mention.
Do you have any suggestion on how to implement the multiple protected labels without having the OR sql query builder?

components/director/internal/domain/label/repository.go
pankrator: Don't think it will bring any value, as currently the hardcoded SQL queries also use the same way "tenant" or "key" or something else. I think it will bring only the need to search for what this constant is pointing to.

components/director/internal/domain/runtime/service.go
pankrator: I agree, I didn't put them like this for any particular reason, just didn't move them below while coding in order to be able to see the code that I write. I will move them now.

components/director/cmd/director/main.go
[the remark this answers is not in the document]
NickyMateev: Oh, I think I got you. With regex we could say something like "(.*_defaultEventing)|(.someOtherPattern)" (yeah, I know my regex example and syntax probably sucks, but the idea is more important). And we could add more protected labels basically by means of regex syntax?
Anyways, I looked at the `labelRepo.DeleteByKeyNotPattern()` implementation - yeah, no sure I can suggest a clean proposal right now.. it would involve multiple DB requests for sure with the current storage layer.

ataleksandrov (approved):
It seems okay from my side.

NickyMateev (approved):

PR #1651, opened 2020-12-09
Introduce Generic Authenticators

NickyMateev:
# Introduce Generic Authenticators

**Description**

Allows plugging custom authenticators by simply providing custom configuration like:
- `gatewayHost` - defines a new host which will be secured with the generic authenticator
- `scopePrefix` (optional) - used for stripping scopes later on, if the custom authenticator prefixes them
- `attributes` - describes what's the **uniqueAttribute** by which to recognize the custom authenticator and also what are the **identityAttributes** and **tenantAttributes** which should be inspected to get identity and tenant information from the JWT token

As can be seen this generic logic currently works for JWT flows only.


chart/compass/charts/gateway/templates/oathkeeper-authenticator-rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: Rule
+metadata:
+  name: "compass-gateway-{{ $authenticatorName }}"
+spec:
+  description: Configuration of oathkeeper for secure endpoint of compass {{ $authenticatorName }} gateway
+  upstream:
+    url: "http://compass-gateway.{{ $.Release.Namespace }}.svc.cluster.local:{{ $.Values.global.gateway.port }}"
+  match:
+    methods: ["GET", "POST", "OPTIONS"]
+    url: <http|https>://compass-gateway-{{ $authenticatorName }}.{{ $.Values.global.ingress.domainName }}<(:(80|443))?>/director/graphql
```
pankrator, line 14: perhaps we should have another variable to say what would  be the URL part as the name might be something more verbose than the URL

chart/compass/charts/gateway/templates/virtualservice-authenticators.yaml
```diff
@@ -0,0 +1,61 @@
...
+  name: "compass-gateway-{{ $authenticatorName }}"
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    app: {{ template "name" . }}
+    chart: {{ $.Chart.Name }}-{{ $.Chart.Version | replace "+" "_" }}
+    heritage: {{ $.Release.Service }}
+    release: {{ $.Release.Name }}
+spec:
+  hosts:
+    - '{{ $.Values.global.gateway.tls.host }}-{{ $authenticatorName }}.{{ $.Values.global.ingress.domainName }}'
```
pankrator, line 16: the same consideration here for the authenticator name and the URL

components/director/internal/tenantmapping/mapper_for_user.go
```diff
@@ -40,26 +46,57 @@ func (m *mapperForUser) GetObjectContext(ctx context.Context, reqData oathkeeper
...
+		log.C(ctx).Info("Getting scopes from token attribute")
+		userScopes, err := reqData.GetUserScopes()
+		if err != nil {
+			return ObjectContext{}, err
+		}
+		scopes = strings.Join(userScopes, " ")
+
+		externalTenantID = gjson.Get(tknAttributes, authn.Attributes.TenantAttribute.Key).String()
+		if externalTenantID == "" {
+			return ObjectContext{}, errors.Errorf("tenant attribute %q missing from %s authenticator token", authn.Attributes.TenantAttribute.Key, authn.Name)
```
pankrator, line 96: I would extract the logic of the whole else statement in another implementation of the interface `ObjectContextProvider`. Then instanciate it as it the user mapper currently is instantiated and then provide it to the tenant mapping handler. This will allow for further extending the tenant mapping handler if it accepts map or array of different mappers.

components/director/pkg/scope/directive.go
```diff
@@ -44,6 +49,9 @@ func (d *directive) matches(actual []string, required []string) bool {
 	actMap := make(map[string]interface{})
 
 	for _, a := range actual {
+		for _, authn := range d.authenticators {
+			a = strings.TrimPrefix(a, authn.ScopePrefix)
```
pankrator, line 53: Can it happen that there are multiple authenticators with overlapping ScopePrefix? In that case it might be better to build the map keys for each scope and each authenticator

pankrator (changes requested):

components/director/internal/tenantmapping/mapper_for_user.go
NickyMateev: Sure thing - sounds like a good idea! Here's the refactoring as you proposed: https://github.com/kyma-incubator/compass/pull/1651/commits/e2b18411e7b42b4feb48f94e0bf08e2b39a5b651

chart/compass/charts/gateway/templates/oathkeeper-authenticator-rules.yaml
```diff
@@ -0,0 +1,31 @@
...
+  upstream:
+    url: "http://compass-gateway.{{ $.Release.Namespace }}.svc.cluster.local:{{ $.Values.global.gateway.port }}"
+  match:
+    methods: ["GET", "POST", "OPTIONS"]
+    url: <http|https>://{{ $config.gatewayHost }}.{{ $.Values.global.ingress.domainName }}<(:(80|443))?>/director/graphql
+  authenticators:
+  - handler: jwt
+    config:
+      trusted_issuers:
+      - {{ $config.tokenURL }}
```
NickyMateev, line 19: Not sure if this can be known beforehand

components/director/pkg/authenticator/authenticator.go
```diff
@@ -0,0 +1,84 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+)
+
+// InitFromEnv loads authenticator configurations from environment if any exist
+func InitFromEnv(envPrefix string) ([]Config, error) {
```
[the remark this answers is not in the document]
NickyMateev: I think you might have mistaken the `envconfig` library that we're using.
The functionalities that you referenced are based on: `github.com/kelseyhightower/envconfig`
We're using this much less feature-rich library: `github.com/vrischmann/envconfig`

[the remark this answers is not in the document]
KirilKabakchiev: yup, they have the exact same envconfig and default tags so I've confused them.

The vrrischmann one has some array support (but this would mean switching to an array and moving the authenticator name as a field of the authConfig struct). Depending on how much you rely on it being a map this can be tried out.
```
var conf struct {
    Shards []struct {
        Name string
        Port int
    } `envconfig:"default={foobar;localhost:2929};{barbaz;localhost:2828}"`
}
```

They also have something like the map Decoder example from kelseyhightower's version of the lib but leveraging Unmarshal 
https://github.com/vrischmann/envconfig/commit/940aedfa1b5c601fd0b27b3113cf2857662aa36f. It looks very similar just using unmarshal instead of decode.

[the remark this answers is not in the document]
NickyMateev: Actually we don't insist on the end-result type to be a map - so that's not a limitation - a slice is fine.
So I tried the slice variation of the library, but I couldn't get it to work - it never seemed to resolve anything in the config struct.
I then went on to try with a custom Unmarshaler, the concept is simple - given a string (the env variable), we should define how it would be parsed into a struct.

For example:

```go
type Authenticators []AuthnConfig

var conf struct {
    Authns Authenticators
}
```

This has the implication of the Authns env var (`APP_AUTHNS`) to be a string containing all authenticators configuration. For example:
```
APP_AUTHNS='[{"name":"uaa1", "scopePrefix":"uaa1!abc", "attributes": {"uniqueAttribute": { "key": "parent_attr.child_attr", "value": "UAA1" }, "tenant": { "key": "parent_attr.tenantid" }, "identity": { "key": "username" } }}, {"name":"uaa2", "scopePrefix":"uaa2!abc", "attributes": {"uniqueAttribute": { "key": "parent_attr.child_attr", "value": "UAA2" }, "tenant": { "key": "parent_attr.tenantid" }, "identity": { "key": "username" } }}]'
```

```go
func (a *Authenticators) Unmarshal(s string) error {
        // unmarshal `s` into `a`
	return nil
}
```

Seems simple enough, but it adds complexity to the helm charts, they'd have to change in the following manner:
```
 authenticators:
    uaa1:
      enabled: true
      gatewayHost: "compass-gateway-uaa1"
      tokenURL: "..."
      jwksURL: "..."
      config: '{"name":"uaa1", "scopePrefix":"uaa1!abc", "attributes": {"uniqueAttribute": { "key": "parent_attr.child_attr", "value": "UAA1" }, "tenant": { "key": "parent_attr.tenantid" }, "identity": { "key": "username" } }}'
    uaa2:
      enabled: true
      gatewayHost: "compass-gateway-uaa2"
      tokenURL: "..."
      jwksURL: "..."
      config: '{"name":"uaa2", "scopePrefix":"uaa2!abc", "attributes": {"uniqueAttribute": { "key": "parent_attr.child_attr", "value": "UAA2" }, "tenant": { "key": "parent_attr.tenantid" }, "identity": { "key": "username" } }}'
```

So far it might be okay, but the tricky part comes when the respective `authenticators.<authn>.config` properties need to be concatenated in the form `[{...},{...}]` so that they can be easily unmarshaled in the code. I think Helm does have some primitives for variable declaration and manipulation so in the Director deployment it would be necessary to loop all authenticators and declare some local Helm variable which is concatenated with all authenticator configs.

Another option would be to maintain a separate variable for the authenticator configs which contains all authenticators but this will be strange considering that the old structure has to continue to exist beside it (enabled, gatewayHost, tokenURL and jwksURL need to be structured and accessible individually by helm to inject in the respective resources, so we cannot put all authenticator configuration "under the same umrella".

Additionally such a change would sort introduce sort of double maintanence of the authenticator name as it is both in the structured path and also in the config json variable.

All things considered I prefer to have complexity in the code instead of on the helm chart side, so I'd rather have the env processing as is.

components/director/pkg/scope/directive.go
pankrator: Sure, I missed that logic in there. I agree that it is better this way.

components/director/internal/tenantmapping/mapper_for_user.go
pankrator: Looks nice now!

components/director/internal/authnmappinghandler/handler_test.go
```diff
@@ -0,0 +1,665 @@
...
+
+		req := httptest.NewRequest(http.MethodPost, target, strings.NewReader(""))
+		w := httptest.NewRecorder()
+
+		verifierMock := &authnMock.TokenVerifier{}
+		tokenDataMock := &authnMock.TokenData{}
+		tokenDataMock.On("Claims", &reqDataMock.Body.Extra).Return(nil).Once()
+
+		handler := authnmappinghandler.NewHandler(reqDataParserMock, httpClient, func(_ context.Context, _ authnmappinghandler.Claims) authnmappinghandler.TokenVerifier {
+			verifierMock.On("Verify", mock.Anything, signedToken).Return(tokenDataMock, nil).Once()
```
KirilKabakchiev, line 97: move the verifierMock.On after the creation of verifierMock?

components/director/internal/authnmappinghandler/handler_test.go
```diff
@@ -0,0 +1,665 @@
...
+
+		require.Contains(t, strings.TrimSpace(string(body2)), string(expectedPayload))
+
+		mock.AssertExpectationsForObjects(t, reqDataParserMock, verifierMock, tokenDataMock)
+	})
+
+	t.Run("error when new verifier fails to verify token", func(t *testing.T) {
+		wellKnownResp := fmt.Sprintf(wellKnownRespPattern, issuer, jwksURL)
+
+		httpClient := &http.Client{
```
KirilKabakchiev, line 191: if the client is the same in all tests, maybe create it before them

KirilKabakchiev (approved):

components/director/internal/authnmappinghandler/handler_test.go
```diff
@@ -574,7 +574,7 @@ func TestHandler(t *testing.T) {
 		handler.ServeHTTP(w, req)
 
 		resp := w.Result()
-		require.Equal(t, http.StatusOK, resp.StatusCode)
+		require.Equal(t, http.StatusBadRequest, resp.StatusCode)
```
KirilKabakchiev, line 577: wouldnt 401 be better ?

NickyMateev: https://github.com/ory/oathkeeper/blob/48603a1ac484b6571706021f2667f770604256b6/pipeline/mutate/mutator_hydrator.go#L218

NickyMateev: I don't mind really - I was wondering about that. I'll change it up, sure

KirilKabakchiev (approved):

desislavaa (approved):

PR #1657, opened 2020-12-16
Connector to go modules

krasish:

- Change connector to use go modules instead of dep


components/connector/Dockerfile
```diff
@@ -6,9 +6,9 @@ WORKDIR ${BASE_APP_DIR}
 COPY ./internal/ ${BASE_APP_DIR}/internal/
 COPY ./config/ ${BASE_APP_DIR}/config/
 COPY ./pkg/ ${BASE_APP_DIR}/pkg/
-COPY ./vendor/ ${BASE_APP_DIR}/vendor/
 COPY ./cmd/main.go ${BASE_APP_DIR}/
 COPY ./licenses ${BASE_APP_DIR}/licenses
+COPY go.mod go.sum ${BASE_APP_DIR}/
```
NickyMateev, line 11: Why is this necessary? I don't think I find it in our old go mod components?

components/connector/go.mod
```diff
@@ -0,0 +1,25 @@
...
+go 1.15
+
+require (
+	cloud.google.com/go v0.44.3 // indirect
+	github.com/99designs/gqlgen v0.9.3
+	github.com/evanphx/json-patch v4.5.0+incompatible // indirect
+	github.com/google/uuid v1.1.2 // indirect
+	github.com/googleapis/gnostic v0.3.1 // indirect
+	github.com/gorilla/mux v1.7.3
+	github.com/kyma-incubator/compass/components/director v0.0.0-20201215200050-802603cbf300
```
NickyMateev, line 12: Why does the Director reference have to be of a specific commit? (maybe same question for the machinebox/grapql)

tests/connector-tests/go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/kyma-incubator/compass/tests/connector-tests
+
+go 1.15
+
+require (
+	github.com/99designs/gqlgen v0.10.1 // indirect
+	github.com/kyma-incubator/compass/components/connector v0.0.0-20201217142711-105e10442111
```
NickyMateev, line 7: Same question about the commit as in the `components/connector` component

components/connector/Dockerfile
krasish: Most of the other go modules components copy the whole current directory (including go.mod and go.sum) in Docker image while in connector's Dockerfile only some of the directories were copied (e.g. /hack is not copied). They need to be in the image so that go knows which are the correct dependencies of this mod project.

components/connector/go.mod
danielgospodinow: It seems that `k8s.io/kube-openapi` is also in the same fashion. I'm also wondering.

components/connector/Dockerfile
NickyMateev: Since we're unifying the approach and if all other components copy the whole directory, but not this one - why not unify it here as well (it seems like a good opportunity while you're at it)

At the end of the day it will end up making a separate Docker stage which will copy only the binary, so I don't see much overhead in unifying this Dockerfile like most others since any additional files that you may add here will be scrapped after the final stage is created.

danielgospodinow: Can't we just add everything `COPY . ${BASE_APP_DIR}` and if there's something that needs to be excluded (e.g. `/hack`), we can add it in a `.dockerignore` file. Do you guys like the idea? It'll look a little bit cleaner.

NickyMateev: Yes, that's exactly what all our other components do `COPY . ${BASE_APP_DIR}` - and that was what @krasish was referencing and what would be needed if we unify the approaches.
Also, like I mentioned in the previous comment - the end game is to build some binaries which you can copy to the second Docker stage which is a clean slate, so it doesn't really matter if you add a /hack directory or not it will ultimately get discarded by the second stage as it only contains the binary file.
Of course if there's some logically reason that you want the first stage to ignore some files, you can always use `.dockerignore` but I'm not aware of any potential issues which may mandate something like that at this point.

components/connector/go.mod
krasish: Versions are supported only for repositories that tag their commits accordingly. We currently don't do that in compass so that is why it looks like that for director. On the other side machinebox/graphql have tagged only a couple of versions (in 2018) and have stopped tagging ever since. The commit + hash that we've got means _"the most recent versioned commit before the target commit is vX.Y.Z."_  
Reference: [Go Pseudo versions](https://golang.org/cmd/go/#hdr-Pseudo_versions); [machinebox/graphql tags](https://github.com/machinebox/graphql/tags)
On the other hand, k8s.io probably do tag their commits properly but since that is an indirect dependency I would suggest not changing it rn.

DimitarPetrov: We don't have versions in compass. If you reference a branch go mod stores the last commit reference of that branch I think.

KirilKabakchiev (approved):
looks ok to me, just remember to switch the connector in the connector tests go.mod to the proper master reference later

components/connector/Dockerfile
krasish: I agree it will give a unified look at our components and it looks cleaner. I also cannot find a specific reason to copy folders one by one instead of the whole current directory. I will change that.

components/connector/go.mod
NickyMateev: Got it - thanks for the clarification guys! 😊

P.S. @krasish I'd also suggest not changing the k8s.io dependency right now.

components/connector/Dockerfile
pankrator: There is a reason to copy only the dependencies first as it will give the ability to not rebuild all dependencies when only the code is changed. I will leave it like this and even change it in other places to first copy the go.mod and go.sum

NickyMateev: Hmm... aren't the dependencies going to be pulled(rebuild) always even when only the code is changed in the `RUN go build -v -o main .` command as the Docker container will not have the dependencies (starting from scratch always)?

pankrator: if we have `RUN go mod vendor/download` before `go build`, then go build will not try to download dependencies.

NickyMateev: Makes sense now:)

krasish: I now understand what @pankrator meant. I agree that this will be a significant improvement in our image. Thanks for the comments!

KirilKabakchiev (approved):

NickyMateev (approved):

danielgospodinow (approved):

PR #1666, opened 2020-12-31
Initial ORD Service integration

NickyMateev:
**Description**

- ORD Service related migrations
- ORD Service charts
- ORD Service E2E tests


components/schema-migrator/migrations/director/202012081537_ord-service-initial.up.sql
```diff
@@ -0,0 +1,302 @@
...
+        entries."releaseStatus" AS release_status,
+        entries.date            AS date,
+        entries.description     AS description,
+        entries.url             AS url
+ FROM event_api_definitions,
+      jsonb_to_recordset(event_api_definitions.changelog_entries) AS entries(version TEXT, "releaseStatus" TEXT,
+                                                                             date TEXT,
+                                                                             description TEXT, url TEXT));
+
+CREATE VIEW event_resource_definitions AS
```
NickyMateev, line 275: Currently each Event has only a single Event Definition, once we implement the aggregator this view should be fully represented in a separate specifications table which will have the data as well

components/schema-migrator/migrations/director/202012081537_ord-service-initial.up.sql
```diff
@@ -0,0 +1,302 @@
...
+
+CREATE VIEW package_links AS
+SELECT id                   AS package_id,
+       actions.type         AS type,
+       actions."customType" AS custom_type,
+       actions.url          AS url
+FROM packages,
+     jsonb_to_recordset(packages.package_links) AS actions(type TEXT, "customType" TEXT, url TEXT);
+
+CREATE VIEW api_resource_definitions AS
```
NickyMateev, line 213: Currently each API has only a single API Definition, once we implement the aggregator this view should be fully represented in a separate specifications table which will have the data as well

components/schema-migrator/migrations/director/202012081537_ord-service-initial.up.sql
```diff
@@ -0,0 +1,302 @@
...
+             api_res_defs."mediaType"            AS media_type
+      FROM api_definitions,
+           jsonb_to_recordset(api_definitions.api_definitions) AS api_res_defs(type TEXT, "customType" TEXT,
+                                                                               "mediaType" TEXT,
+                                                                               url TEXT)) as api_defs
+UNION ALL
+(SELECT id                                  AS api_definition_id,
+        CASE
+            WHEN spec_type::text = 'ODATA' THEN 'edmx'
+            WHEN spec_type::text = 'OPEN_API' THEN 'openapi-v3'
```
NickyMateev, line 228: in ORD there are Open API v2 and v3 but Compass supports only a single `OPEN_API` type - it's not trivial to derive reasoning about what might be the actual version, but using v3 should be good enough for legacy applications resources.

tests/ord-service/tests/api_test.go
```diff
@@ -0,0 +1,322 @@
...
+			name := gjson.Get(respBody, fmt.Sprintf("value.%d.title", i)).String()
+			require.NotEmpty(t, name)
+
+			expectedAPI := apisMap[name]
+
+			require.Equal(t, *expectedAPI.Description, gjson.Get(respBody, fmt.Sprintf("value.%d.description", i)).String())
+			require.Equal(t, expectedAPI.TargetURL, gjson.Get(respBody, fmt.Sprintf("value.%d.entryPoint", i)).String())
+			require.NotEmpty(t, gjson.Get(respBody, fmt.Sprintf("value.%d.partOfPackage", i)).String())
+			releaseStatus := gjson.Get(respBody, fmt.Sprintf("value.%d.releaseStatus", i)).String()
+			if *expectedAPI.Version.ForRemoval {
```
DimitarPetrov, line 184: These ifs and the other below can be switches.

tests/ord-service/tests/api_test.go
```diff
@@ -0,0 +1,322 @@
...
+	"github.com/kyma-incubator/compass/tests/director/pkg/idtokenprovider"
+	"github.com/kyma-incubator/compass/tests/director/pkg/ptr"
+	"github.com/kyma-incubator/compass/tests/ord-service/pkg"
+	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
+)
+
+const TestApp = "mytestapp"
+
+func TestODService(t *testing.T) {
```
DimitarPetrov, line 39: Does it makes sense to add tests for some of the most common OData queries like `expand,filter etc.`. 

I think that we can trust the library that those queries will work as expected (since we have tested our custom convertions in those top-level query tests), but we should test if our implementation (particularly the JPA representation of our DB Model) matches the DB and the relations there. 

We can just create some of those more common requests to the service and expect a 200 status code without an error message. I know that those tests seem shallow but they will test the integrity between the JPA model and our DB model which is the most essential part of the ord-service.

DimitarPetrov (changes requested):

tests/ord-service/tests/api_test.go
NickyMateev: Definitely makes sense - tests are added.

components/schema-migrator/migrations/director/202012081537_ord_service_initial.up.sql
```diff
@@ -0,0 +1,302 @@
...
+CREATE VIEW links AS
+SELECT *
+FROM (SELECT id                AS package_id,
+             NULL::uuid        AS api_definition_id,
+             NULL::uuid        AS event_definition_id,
+             links.title       AS title,
+             links.url         AS url,
+             links.description AS description
+      FROM packages,
+           jsonb_to_recordset(packages.links) AS links(title TEXT, description TEXT, url TEXT)) AS package_links
```
NickyMateev, line 104: @DimitarPetrov do you remember if the `...AS package_links` is needed for something specific here?
A similar thing can be noticed on line 139, 166, 189, 223, 261, 285.

components/schema-migrator/migrations/director/202012081537_ord_service_initial.up.sql
```diff
@@ -22,8 +22,8 @@ $$
 BEGIN
     NEW.release_status := CASE
                               WHEN NEW.release_status IS NOT NULL THEN NEW.release_status
-                              WHEN NEW.version_deprecated IS NULL AND NEW.version_for_removal THEN 'decommissioned'
                               WHEN NEW.version_deprecated THEN 'deprecated'
+                              WHEN NEW.version_for_removal THEN 'decommissioned'
```
DimitarPetrov, line 26: decommissioned should be with higher priority here because if an API is removed it is always deprecated before that and in that case, we will have both true and the realease_status will be deprecated, but the API is actually decommissioned.

DimitarPetrov (dismissed):

components/schema-migrator/migrations/director/202012081537_ord_service_initial.up.sql
NickyMateev: Discussed - issue was that without alias the migration failed with:
` error: migration failed: subquery in FROM must have an alias (column 6) in line 97: BEGIN;`

DimitarPetrov (approved):

PR #1668, opened 2021-01-05
Support authentication on fetch request

StanislavStefanov:
**Description**

- enable fetch request authentication
- add unit tests
- add end to end tests


components/director/internal/domain/fetchrequest/service.go
```diff
@@ -75,38 +83,85 @@ func (s *service) fetchAPISpec(ctx context.Context, fr *model.FetchRequest) (*st
...
+	}
+
+	if fr.Auth.Credential.Oauth != nil {
+		resp, err = s.secureClient(ctx, fr).Do(req)
+	}
+
+	return resp, err
+}
+
+func (s *service) secureClient(ctx context.Context, fr *model.FetchRequest) *http.Client {
```
KirilKabakchiev, line 139: this http client creation is not really clean code, starting from a config, then you have to pass your existing client through the context so its transport is reused. But the clients timeout is not reused and thats why you set it explicitly. And not only that but the client method is on the Config struct... I normally try to avoid this package.
Maybe if you like to keep using the package you can use only the oauth2.TokenSource and the oauth2.Transport

chart/compass/values.yaml
```diff
@@ -23,7 +23,7 @@ global:
       version: "PR-1661"
     director:
       dir:
-      version: "PR-1670"
+      version: "PR-1668"
```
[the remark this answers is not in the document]
DimitarPetrov: This will be the newest once this PR is merged since it is up to date with the master.

PetarTodorovv (dismissed):

components/director/internal/domain/fetchrequest/service.go
```diff
@@ -75,38 +83,85 @@ func (s *service) fetchAPISpec(ctx context.Context, fr *model.FetchRequest) (*st
...
+		return nil, err
+	}
+
+	var resp *http.Response
+	if fr.Auth.Credential.Basic != nil {
+		req.SetBasicAuth(fr.Auth.Credential.Basic.Username, fr.Auth.Credential.Basic.Password)
+
+		resp, err = s.client.Do(req)
+
+		if err == nil && resp.StatusCode == http.StatusOK {
```
NickyMateev, line 127: If both Basic & OAuth credentials are set and the request with basic fails it will attempt to make a request with OAuth credentials. Not sure this is expected behaviour.

A more general question is do we want to forbid FetchRequests with authentication data provided for both methods. We should discuss it with the team if it has not been already.

NickyMateev: Additionally, it may be possible for err = nil & httpstatus!=OK; this would mean that the `requestWithCredentials` function would return `nil, nil` (if on OAuth credentials provided), which would break the calling function when it inspects the response variable.

NickyMateev (changes requested):

components/director/internal/domain/api/service_test.go
```diff
@@ -371,7 +371,7 @@ func TestService_CreateToPackage(t *testing.T) {
 			},
 			FetchRequestServiceFn: func() *automock.FetchRequestService {
 				svc := &automock.FetchRequestService{}
-				svc.On("HandleAPISpec", ctx, fixModelFetchRequest(frID, frURL, timestamp)).Return(nil)
+				svc.On("HandleSpec", ctx, fixModelFetchRequest(frID, frURL, timestamp)).Return(nil)
```
NickyMateev, line 374: Since you're removing the `API` from the function definitions, remove it also from the name of test functions, ex: `TestService_RefetchAPISpec`. I also saw it in the fetch request service tests and I think also the events service tests.

components/director/internal/domain/fetchrequest/service_test.go
```diff
@@ -10,24 +10,25 @@ import (
 	"testing"
```
NickyMateev, line 10: These tests fail on my machine, any idea why?
image

components/external-services-mock/cmd/main.go
```diff
@@ -59,16 +59,26 @@ func initHTTP(cfg config) http.Handler {
 	router.HandleFunc("/v1/healtz", health.HandleFunc)
 
 	configChangeRouter := router.PathPrefix("/audit-log/v2/configuration-changes").Subrouter()
-	configChangeRouter.Use(authMiddleware)
+	configChangeRouter.Use(oauthMiddleware)
 	configurationchange.InitConfigurationChangeHandler(configChangeRouter, configChangeHandler)
 
 	router.HandleFunc("/audit-log/v2/oauth/token", oauthHandler.Generate).Methods(http.MethodPost)
+	router.HandleFunc("/oauth/token", oauthHandler.Generate).Methods(http.MethodPost)
```
NickyMateev, line 66: Since you're going to reuse the same oathHandler, fix the `APP_AUDITLOG_CLIENT_ID` and `APP_AUDITLOG_CLIENT_SECRET` environment variables to something generic, so as to not imply it's only used only for Auditlog testing.

Also externalize the Basic credentials similarly to that, because there's currently and implicit contract between the external-services-mock implementation and the corresponding e2e test that the basic credentials are "admin" & "admin".

NickyMateev (changes requested):

components/director/internal/domain/fetchrequest/service.go
StanislavStefanov: It is intentional that if the request with basic credentials fails there will be an atempt to make request with OAuth credentials if such are present. It seemed appropriate to try the second set of credentilas if the first is not good. Although I am not sure if it is possible in our scenarios the two types of credentials to be provided at the same time.

StanislavStefanov: In case of err=nil & httpstatus!=OK there are two cases. First if there are OAuth credentials provided, request with them will be made and the returned err and responce will be returned from the function. In the second case there aren't provided OAuth credentials so we are not gonig into the if statement's body so the err(in this case nil) and the responce(from the request wit basic credentials which has status code different from OK) will be returned. The resp variable is intentionaly declared before the if statement(row 121) so that the value returned from the request is not lost.

NickyMateev: Oh, yes, you're right. Seems like I mislooked that piece of code:)

chart/compass/charts/external-services-mock/templates/auditlog-secret.yaml
```diff
@@ -8,3 +8,15 @@ data:
   auditlog-client-id: {{ "client_id" | b64enc | quote }}
   auditlog-client-secret: {{ "client_secret" | b64enc | quote }}
   auditlog-oauth-url: {{ printf "http://compass-external-services-mock.%s.svc.cluster.local:%s/audit-log/v2/oauth/token" .Release.Namespace (.Values.service.port | toString) | b64enc | quote }}
+---
```
NickyMateev, line 11: Rename this file to something generic so it's unified with `chart/compass/charts/external-services-mock/templates/basic-credentials-secret.yaml`

NickyMateev (changes requested):

NickyMateev (approved):

NickyMateev (approved):

nkanchev (approved):
doc changes look good to me

PR #1671, opened 2021-01-05
Allow container in pod to be restarted on failure

dzahariev:
**Description**

Sporadically the compass-tenant-fetcher job goes in "Error" state due to issues with CloudSQL proxy pod.
This change allow the crashed container to be automatically restarted.

**Related issue(s)**
N/A


krasish (approved):

DimitarPetrov (changes requested):
This will cause crash loop problem.

Like here: https://github.com/kyma-incubator/compass/pull/1620

dzahariev:
I close this to avoid crash loop issues.

PR #1674, opened 2021-01-07
Bump alpine version

desislavaa:
**Description**

- Increased alpine version due to vulnerabilities
- Unify the alpine version used across the whole project


dzahariev (approved):
edge to 3.12 is actually a downgrade, but can be seen that no special requirement to any packages are needed from the container image.

PR #1686, opened 2021-01-13
System broker secure osb endpoints

danielgospodinow:
**Description**

- Secure OSB endpoints with client certificates
- Refactor Connector e2e tests
- Add System Broker e2e, component, and unit tests
- Enable System Broker readiness and liveness probes 

Currently, the System Broker is patched to work with two different modes - like an Integration System (the previous approach) and like a proxy that just forwards client auth.

Note:
* In this new approach, we've added an Unauthorized middleware which ensures that `Unauthorized` errors are not masked like `InternalServerError` but are instead returned as they should be.
* The System Broker is exposed via the `mtls gateway`.


chart/compass/values.yaml
```diff
@@ -1,6 +1,7 @@
 global:
   disableLegacyConnectivity: true
   defaultTenant: 3e64ebae-38b5-46a0-b1ed-9ccee153a0ae
+  fooTenant: 1eba80dd-8ff6-54ee-be4d-77944d17b10b
```
alextargov, line 4: What's that tenant for? Just curious

danielgospodinow: This tenant is used to run the System Broker tests. We considered running them in a separate tenant to ensure that we System Broker tests won't mess up Director or Connector e2e tests when they're all run concurrently.

components/system-broker/pkg/oauth/token_provider_from_secret.go
```diff
@@ -33,26 +53,68 @@ type credentials struct {
...
+	if _, err := getBearerToken(ctx); err != nil {
+		log.C(ctx).WithError(err).Warn("while obtaining bearer token")
+		return true
+	}
+
+	return false
+}
+
+func (c *TokenProviderFromSecret) GetAuthorizationToken(ctx context.Context) (httputils.Token, error) {
+	if c.validToken() {
```
danielgospodinow, line 89: Guys, we had a discussion here about some potential race conditions. We would love to leave your opinion here.

danielgospodinow: We've resolved the synchronization dilemma. You can have a look.

chart/compass/values.yaml
danielgospodinow: We've also undone this change in favor of using the default tenant for the e2e tests. We've decided to be consistent with the other component. Also, there's no chance of a race condition between all e2e tests because they're not executed concurrently at the moment. 😆

components/system-broker/pkg/http/config.go
```diff
@@ -43,6 +46,8 @@ func DefaultConfig() *Config {
 		ExpectContinueTimeout: time.Second * 2,
 		MaxIdleConns:          90,
 		SkipSSLValidation:     true,
+		ForwardHeaders:        []string{"Authorization"},
+		UnauthorizedString:    "insufficient scopes provided",
```
DimitarPetrov, line 50: These should be empty by default.

components/system-broker/pkg/server/server.go
```diff
@@ -39,15 +40,18 @@ type Server struct {
 	shutdownTimeout time.Duration
 }
 
-func New(c *Config, service log.UUIDService, routesProvider ...func(router *mux.Router)) *Server {
+func New(c *Config, service log.UUIDService, forwardHeaders []string, routesProvider ...func(router *mux.Router)) *Server {
```
DimitarPetrov, line 43: We can make New have a slice of middlewares as argument insted of providing the forwardHeaders. This way that middleware (which is specific for the system broker) can be provided in the main and other brokers can add their own middlewares.

DimitarPetrov (changes requested):

chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
```diff
@@ -95,6 +95,30 @@ spec:
 ---
 apiVersion: oathkeeper.ory.sh/v1alpha1
 kind: Rule
+metadata:
+  name: compass-broker-certs
+spec:
+  description: Configuration of oathkeeper for secure endpoint of compass system broker
+  upstream:
+    url: "http://compass-system-broker.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.global.system_broker.port }}"
```
pankrator, line 103: Do we want to Audit log operations for system broker? If this is the case then may be the requests should go through the compass-gateway component first. Not sure how this will affect authentication/authorization though.

components/system-broker/pkg/http/secured_transport.go
```diff
@@ -19,64 +19,59 @@ package http
...
+		roundTripper:   roundTripper,
+		tokenProviders: providers,
 	}
 }
 
 func (c *SecuredTransport) RoundTrip(request *http.Request) (*http.Response, error) {
-	if err := c.refreshToken(request.Context()); err != nil {
+	logger := log.C(request.Context())
+
+	token, err := c.getToken(request.Context())
```
pankrator, line 49: Why do you need to put "some configured headers" in the context in order to use them later for the token provider as you have here the request. Can't the token provider receive the whole request object and get the headers from there?
May be there is something I don't see for the need of the context, but in my opinion it would be more intuitive if you access the headers directly.

pankrator (changes requested):

chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
ataleksandrov: We decided to go with another approach. The outgoing requests to the director to go through the compass-gateway component so there are Audit logs (please see https://github.com/kyma-incubator/compass/blob/system-broker-secure-osb-endpoints/chart/compass/charts/system-broker/templates/deployment.yaml#L62-L63)

components/system-broker/pkg/http/secured_transport.go
pankrator: As discussed offline this approach is needed because of the Bind API of the broker framework because it does not have a http.Request

chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
```diff
@@ -95,6 +95,30 @@ spec:
...
+  match:
+    methods: ["GET", "POST", "PATCH", "PUT", "DELETE"]
+    url: <http|https>://{{ .Values.global.gateway.mtls.host }}.{{ .Values.global.ingress.domainName }}<(:(80|443))?>/broker<(/.*)?>
```
NickyMateev, line 106: Make use of `{{ .Values.global.system_broker.prefix }}` here

chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
```diff
@@ -95,6 +95,30 @@ spec:
 ---
 apiVersion: oathkeeper.ory.sh/v1alpha1
 kind: Rule
+metadata:
+  name: compass-broker-certs
```
NickyMateev, line 99: `compass-system-broker-certs` - we'll have other brokers in the future as well, so we need to differentiate them

chart/compass/charts/system-broker/templates/deployment.yaml
```diff
@@ -58,10 +58,11 @@ spec:
...
             - name: SERVER_SELF_URL
-              value: "https://{{ .Values.global.system_broker.tls.host }}.{{ .Values.global.ingress.domainName }}/broker"
-            - name: GRAPHQL_CLIENT_GRAPHQL_ENDPOINT
+              value: "https://{{ .Values.global.gateway.mtls.host }}.{{ .Values.global.ingress.domainName }}"
+            - name: GRAPHQL_CLIENT_GRAPHQL_ENDPOINT # directly call the compass-gateway, as we already have a crafted jwt id token
+              value: "http://{{ .Values.global.gateway.tls.host }}.{{ .Release.Namespace }}.svc.cluster.local:{{.Values.global.gateway.port }}/director/graphql"
+            - name: GRAPHQL_CLIENT_GRAPHQL_OAUTH_ENDPOINT # used when system-broker acts as an integration system, we need to convert the oauth2 opaque access token to jwt id token
               value: "https://{{ .Values.global.gateway.tls.secure.oauth.host }}.{{ .Values.global.ingress.domainName }}/director/graphql"
-
             # - name: GRAPHQL_CLIENT_ENABLE_LOGGING
```
NickyMateev, line 66: Why are all these configurations left out and commented? We should leave out dead code/configuration.

chart/compass/charts/system-broker/templates/deployment.yaml
```diff
@@ -58,10 +58,11 @@ spec:
             # - name: SERVER_SHUTDOWN_TIMEOUT
             #   value: "0.0.0.0:{{ .Values.deployment.args.containerPort }}"
             - name: SERVER_SELF_URL
-              value: "https://{{ .Values.global.system_broker.tls.host }}.{{ .Values.global.ingress.domainName }}/broker"
-            - name: GRAPHQL_CLIENT_GRAPHQL_ENDPOINT
+              value: "https://{{ .Values.global.gateway.mtls.host }}.{{ .Values.global.ingress.domainName }}"
+            - name: GRAPHQL_CLIENT_GRAPHQL_ENDPOINT # directly call the compass-gateway, as we already have a crafted jwt id token
+              value: "http://{{ .Values.global.gateway.tls.host }}.{{ .Release.Namespace }}.svc.cluster.local:{{.Values.global.gateway.port }}/director/graphql"
+            - name: GRAPHQL_CLIENT_GRAPHQL_OAUTH_ENDPOINT # used when system-broker acts as an integration system, we need to convert the oauth2 opaque access token to jwt id token
```
NickyMateev, line 64: Why do we want to keep the integration system way of authenticating?

chart/compass/charts/system-broker/templates/tests/system-broker-test.yaml
```diff
@@ -0,0 +1,59 @@
...
+            - name: DIRECTOR_HEALTHZ_URL
+              value: "http://compass-director.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.global.director.port }}/healthz"
+            - name: CONNECTOR_HEALTHZ_URL
+              value: "http://compass-connector.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.global.connector.graphql.external.port }}/healthz"
+            - name: SYSTEM_BROKER_HEALTHZ_URL
+              value: "http://compass-system-broker.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.global.system_broker.port }}/healthz"
+            - name: APP_TENANT
+              value: "{{ .Values.global.defaultTenant }}"
+            - name: APP_SYSTEM_BROKER_URL
+              value: "https://{{ .Values.global.gateway.mtls.host }}.{{ .Values.global.ingress.domainName }}/broker"
```
NickyMateev, line 44: Make use of {{ .Values.global.system_broker.prefix }} here

chart/compass/values.yaml
```diff
@@ -132,13 +135,15 @@ global:
...
 
   system_broker:
-    port: 5001
-    tls:
-      host: compass-gateway
     enabled: true
+    port: 5001
+    prefix: /broker
+    forwardHeaders: Authorization
+    unauthorizedString: "insufficient scopes provided"
```
NickyMateev, line 142: Why externalize this error message?

NickyMateev (changes requested):

components/system-broker/pkg/http/unauthorized_middleware.go
```diff
@@ -0,0 +1,66 @@
...
+	"net/http"
+	"net/http/httptest"
+	"strings"
+
+	"github.com/kyma-incubator/compass/components/system-broker/pkg/log"
+	"github.com/pkg/errors"
+	"github.com/tidwall/gjson"
+)
+
+// UnauthorizedMiddleware
```
NickyMateev, line 16: Add adequate godoc

tests/system-broker/tests/apitests/auth_test.go
```diff
@@ -0,0 +1,147 @@
+package apitests
```
NickyMateev, line 1: Ditch the `apistests` directory and place the tests directly under `tests/system-broker/tests`. There's no point for it and after refactoring the e2e tests it would be removed anyway

tests/system-broker/tests/testkit/config.go
```diff
@@ -0,0 +1,32 @@
+package testkit
```
NickyMateev, line 1: Ditch the `testkit` directory and have a top-level `tests/system-broker/pkg` directory as this would be the way to go with the new e2e test structure. `testkit` is an insidious and confusing package based on the legacy way of how the e2e tests were written.

tests/system-broker/tests/apitests/auth_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
+	connectorTestkit "github.com/kyma-incubator/compass/tests/connector-tests/test/testkit"
+	"github.com/kyma-incubator/compass/tests/connector-tests/test/testkit/connector"
+	director "github.com/kyma-incubator/compass/tests/director/gateway-integration"
+	"github.com/pivotal-cf/brokerapi/v7/domain"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTokens(t *testing.T) {
```
NickyMateev, line 23: TestSystemBrokerAuthentication

tests/system-broker/tests/apitests/auth_test.go
```diff
@@ -0,0 +1,147 @@
...
+					InsecureSkipVerify: true,
+				},
+			},
+		}
+		_, err := client.Do(req)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "tls: certificate required")
+	})
+
+	t.Run("Should fail calling bind endpoint with revoked cert", func(t *testing.T) {
```
NickyMateev, line 70: Is there something special about the `bind` endpoint compared to the `catalog` one? If not I suggest keeping things simple and consistent and request the `catalog` endpoint throughout the tests to avoid confusion.

NickyMateev (changes requested):

components/system-broker/pkg/http/forward_headers.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// HeaderForwarder stores the specified request in the context so that they can later be used and forwarded to other backends
+func HeaderForwarder(forwardHeaders []string) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+
+			for _, header := range forwardHeaders {
+				if value := r.Header.Get(header); value != "" {
```
NickyMateev, line 33: Missing unit tests for the case when none of the forward headers are found in the request

components/system-broker/pkg/http/unauthorized_middleware.go
```diff
@@ -0,0 +1,66 @@
...
+func UnauthorizedMiddleware(unauthorizedString string) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+			recorder := httptest.NewRecorder()
```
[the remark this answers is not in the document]
danielgospodinow: Currently, we don't see any alternatives. We're welcoming new suggestions but if none are present we would, unfortunately, stick with this bad practice approach.

chart/compass/charts/system-broker/templates/deployment.yaml
danielgospodinow: Yes, we should avoid dead code or dead configurations but it's okay to leave them for now just for the sake of being a template. They can be removed/uncommented when we reach the productization stage.

chart/compass/charts/system-broker/templates/deployment.yaml
danielgospodinow: Good point. I'll just comment out this env variable.

NickyMateev: My question is a more general one - why is there still code which allows for communicating with the System Broker as an integration system? There was this Token provider from secret that I saw was still in the code base. I think we decided that that would be removed all together

danielgospodinow: We've discussed this topic with a small part of the team and we've decided to leave the integration system way of communicating because there's a possibility that we'll need this in the future. As of now, we're not 100% sure that we'll never this way of communication. So we designed the System Broker to work in both ways via token provider matching.

components/system-broker/pkg/graphql/config.go
```diff
@@ -20,13 +20,16 @@ import "github.com/pkg/errors"
 
 type Config struct {
 	GraphqlEndpoint string `mapstructure:"graphql_endpoint"`
-	EnableLogging   bool   `mapstructure:"enable_logging"`
+	// used when system broker acts like an integration system
+	GraphqlOauthEndpoint string `mapstructure:"graphql_oauth_endpoint"`
```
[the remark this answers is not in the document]
danielgospodinow: Nice catch, maybe it's a leftover. This OAuth endpoint from the config is never used because the actual OAuth endpoint comes from the secret from which the System Broker read his clientID and clientSecret.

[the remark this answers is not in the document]
DimitarPetrov: Isn't that OAuth secured endpoint of director used when system broker acts as integration system? The one in the secret should be the token endpoint for issuing hydra tokens.

chart/compass/values.yaml
danielgospodinow: We externalized this error to make it configurable since the default value is an empty string as this property is something specific to different brokers.

components/system-broker/pkg/graphql/config.go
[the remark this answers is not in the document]
danielgospodinow: Yes, my mistake. @DimitarPetrov is right. So the discussed OAuth endpoint of the Director in this config file is part of the unused integration system communication. We'll remove it as well as other stuff that are related to the integration system communication scenario.

chart/compass/charts/system-broker/templates/deployment.yaml
danielgospodinow: We've cleared the dead code around the integration system flow of communicating.

components/system-broker/pkg/graphql/config.go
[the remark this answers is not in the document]
danielgospodinow: We've resolved that.

components/system-broker/pkg/graphql/config.go
```diff
@@ -20,16 +20,13 @@ import "github.com/pkg/errors"
 
 type Config struct {
 	GraphqlEndpoint string `mapstructure:"graphql_endpoint"`
-	// used when system broker acts like an integration system
-	GraphqlOauthEndpoint string `mapstructure:"graphql_oauth_endpoint"`
```
DimitarPetrov: From my PoV if we want to leave the option for system-broker to work as an integration system we want this to be achieveble only with configurations. With this commit we loose this functionality, we will need code level changes.

My proposal is to leave this config and if it is set to configure a TokenProviderFromSecret with this URL and then we can switch integration system mode with configuration only.

Additionally @StanislavStefanov should revert the job that registers integration system and guard it with some kind of prorpery in the charts for example `.Value.global.system_broker.integrationSystemAuth`

components/system-broker/pkg/graphql/config.go
```diff
@@ -20,16 +20,13 @@ import "github.com/pkg/errors"
 
 type Config struct {
```
DimitarPetrov: From my PoV if we want to leave the option for system-broker to work as an integration system we want this to be achieveble only with configurations. With this commit we loose this functionality, we will need code level changes.

My proposal is to leave this GraphqlOauthEndpoint config and if it is set to configure a TokenProviderFromSecret with this URL and then we can switch integration system mode with configuration only.

Additionally @StanislavStefanov should revert the job that registers integration system and guard it with some kind of prorpery in the charts for example `.Value.global.system_broker.integrationSystemAuth`

DimitarPetrov (changes requested):

components/system-broker/pkg/server/server.go
```diff
@@ -39,15 +39,18 @@ type Server struct {
...
+func New(c *Config, service log.UUIDService, middlewares []mux.MiddlewareFunc, routesProvider ...func(router *mux.Router)) *Server {
 	s := &Server{
 		shutdownTimeout: c.ShutdownTimeout,
 		routesProvider:  routesProvider,
 	}
 
 	router := mux.NewRouter()
+
+	router.Handle("/healthz", s.livenessHandler())
+	router.Handle("/readyz", s.readinessHandler())
```
DimitarPetrov, line 51: These should be behind the prefix in order for istio virtual service to work. You can just add the prefix to the probes.

DimitarPetrov (changes requested):

components/system-broker/pkg/server/server.go
DimitarPetrov: Discussed that all other components do not expose their health endpoints and are accessible only inside the cluster. This will be addressed everywhere if we decide to enhance our health checks.

components/system-broker/pkg/graphql/config.go
DimitarPetrov: It was decided that we want to leave the code for integration system auth, but do not enable it for now.

NickyMateev (approved):

PR #1687, opened 2021-01-14
Use tenant header when tenant isn't in token

desislavaa:
**Description**

- Fallback mechanism for when the tenant could not be found in the token

### DIFF
--- a/components/director/internal/tenantmapping/authenticator_context_provider.go
+++ b/components/director/internal/tenantmapping/authenticator_context_provider.go
@@ -18,6 +18,7 @@ package tenantmapping
 
 import (
 	"context"
+	"net/http"
 	"strings"
 
 	"github.com/tidwall/gjson"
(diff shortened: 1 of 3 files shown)


components/director/internal/tenantmapping/authenticator_context_provider.go
```diff
@@ -90,3 +91,12 @@ func (m *authenticatorContextProvider) GetObjectContext(ctx context.Context, req
 
 	return objCtx, nil
 }
+
+func tenantID(extra string, tenantKeyInExtra string, reqHeaders http.Header) string {
+	id := gjson.Get(extra, tenantKeyInExtra).String()
+	if id == "" {
+		id = reqHeaders.Get(oathkeeper.ExternalTenantKey)
```
desislavaa, line 98: Log? I'm not sure if we'll gain something, or we'll just produce more logs

KirilKabakchiev (approved):

PR #1693, opened 2021-01-20
Introduce Bundles entity

NickyMateev:
# Introduce Bundles entity

**Description**

- Rename top to bottom from package to bundle
- Packages table renamed to Bundles
- Drop the columns from the new bundle table as they belong to package table and we don't have any data in them
- Create new table package that represents the ORD package
- Add pre and post processing which modifies the GQL object on request and response - middlewarer to manipulate Query & Mutation for packages => use bundle from database => rewrite to package
- Temporary backwards compatibility is ensured using a query parameter `useBundles` specifying if the consumer wants the ORD entities
- Runtimes which provide `useBundles` in their requests shall be labeled with that same param so we can later remove this backwards compatibility logic later on
- ORD Service to expose the new Bundles and Packages


components/director/internal/packagetobundles/handler.go
```diff
@@ -74,25 +77,11 @@ func (h *Handler) Handle() func(next http.Handler) http.Handler {
...
-						Value:      "true",
-						ObjectID:   consumerInfo.ConsumerID,
-						ObjectType: model.LabelableObject(consumerInfo.ConsumerType),
-					})
+				if err := h.reconcileConsumerLabel(ctx, consumerInfo); err != nil {
+					writeAppError(ctx, w, err, http.StatusInternalServerError)
+					return
 				}
 
 				log.C(ctx).Infof("Returning request without rewriting the request body for consumer with ID %q and type %q", consumerInfo.ConsumerID, consumerInfo.ConsumerType)
```
DimitarPetrov, line 87: We can get the consumerInfo before `next.ServeHTTP(w, r)` and enrich the log on the first line to indicate which consumer has already adopted Bundles and then remove this log, since it does not give much of a value.

NickyMateev: Addressed

nkanchev (approved):
Review .md files. The only change across all files was the change of the term "package" to "bundle".

tests/director/api/application_api_test.go
```diff
@@ -223,6 +223,101 @@ func TestRegisterApplicationWithBundles(t *testing.T) {
...
+		packageID := actualApp.Packages.Data[0].ID
+
+		require.NoError(t, err)
+		require.NotEmpty(t, appID)
+		require.NotEmpty(t, packageID)
+		require.Equal(t, expectedAppName, actualApp.Name)
+
+		defer unregisterApplication(t, actualApp.ID)
+
+		t.Run("Get Application with Package when useBundles=false should succeed", func(t *testing.T) {
```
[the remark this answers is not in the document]
DimitarPetrov: We actually provide useBundles=false, since all the tests by default provide useBundles=true

components/director/internal/packagetobundles/handler.go
```diff
@@ -0,0 +1,233 @@
...
+	uidSvc := uid.NewService()
+	labelUpsertSvc := label.NewLabelUpsertService(labelRepo, labelDefRepo, uidSvc)
+
+	return &Handler{
+		transact:           transact,
+		labelUpsertService: labelUpsertSvc,
+	}
+}
+
+func (h *Handler) Handle() func(next http.Handler) http.Handler {
```
KirilKabakchiev, line 67: should be Handler() (to be consistent with other middlewares)

components/director/internal/packagetobundles/handler.go
```diff
@@ -0,0 +1,233 @@
...
+				return
+			}
+		})
+	}
+}
+
+func (h *Handler) labelRuntimeWithBundlesParam(ctx context.Context, consumerInfo consumer.Consumer) error {
+	tenantID, err := tenant.LoadFromContext(ctx)
+	if err != nil {
+		log.C(ctx).WithError(err).Error("Error determining request tenant")
```
KirilKabakchiev, line 185: double error logging (here and in caller) - relevant for next few calls as well

components/director/internal/packagetobundles/handler.go
```diff
@@ -0,0 +1,233 @@
...
+		Key:        useBundlesParam,
+		Value:      "true",
+		ObjectID:   consumerInfo.ConsumerID,
+		ObjectType: model.LabelableObject(consumerInfo.ConsumerType),
+	}); err != nil {
+		log.C(ctx).WithError(err).Errorf("An error occurred while upserting %q label", useBundlesParam)
+		return apperrors.InternalErrorFrom(err, "while executing database transaction")
+	}
+
+	err = tx.Commit()
```
KirilKabakchiev, line 209: tx.Commit().rre

components/director/internal/packagetobundles/handler.go
```diff
@@ -0,0 +1,233 @@
...
+				log.C(ctx).WithError(err).Error("Error reading request body")
+				appErr := apperrors.InternalErrorFrom(err, "while reading request body")
+				writeAppError(ctx, w, appErr, http.StatusInternalServerError)
+				return
+			}
+
+			body := string(reqBody)
+			body = strings.ReplaceAll(body, "\\n", "") // removes unnecessary complexity from the next regexes
+
+			reqPackagesJSONPattern := regexp.MustCompile(`(\s*)packages(\s*:\s*\[)`) // matches ` packages:  [`
```
KirilKabakchiev, line 110: maybe order these replaces semantically (start from query/mutation name,  then arguments, then input argument (in), then outputs)

components/director/internal/packagetobundles/handler.go
```diff
@@ -0,0 +1,233 @@
...
+			body = respPackageJSONPattern.ReplaceAllString(body, "${1}package${2}")
+
+			respPackageModeGraphQLPattern := regexp.MustCompile(`(\s*\")mode(\"\s*):(\s*\")BUNDLE(\"\s*)`) // matches ` "mode": "BUNDLE" `
+			body = respPackageModeGraphQLPattern.ReplaceAllString(body, "${1}mode${2}:${3}PACKAGE${4}")
+
+			w.WriteHeader(recorder.Code)
+			if _, err := w.Write([]byte(body)); err != nil {
+				log.C(ctx).WithError(err).Error("Error writing response body")
+				appErr := apperrors.InternalErrorFrom(err, "while writing response body")
+				writeAppError(ctx, w, appErr, http.StatusInternalServerError)
```
KirilKabakchiev, line 175: are there other graph errors that return 500 or should this ve 200 ?

KirilKabakchiev (approved):

nkanchev (approved):

PR #1712, opened 2021-01-29
Disable envoy proxy for (cron) jobs

desislavaa:
**Description**

- The `compass-migration` job and `tenant-fetcher` cron jobs do not need an envoy proxy, hence we can disable the sidecar injection. Pros: Kyma update runs smoothly, otherwise we have to manually delete all pods from these jobs because of [this script](https://github.com/kyma-project/kyma/blob/release-1.17/resources/istio/files/istio-proxy-reset.sh), that runs during the migration.

### DIFF
--- a/chart/compass/templates/migrator-job.yaml
+++ b/chart/compass/templates/migrator-job.yaml
@@ -1,84 +1,86 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-    name: compass-migration
-    labels:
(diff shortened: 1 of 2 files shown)


NickyMateev (approved):

desislavaa:
We'll open a bug on Kyma side - they should not try to restart pods that belong to jobs/cronjobs afterall

PR #1714, opened 2021-01-29
Introduce Async operations foundation

PetarTodorovv:
# Introduce Async operations foundation

- [x] Implementation
- [x] Unit tests


components/director/pkg/operation/middleware.go
```diff
@@ -0,0 +1,91 @@
...
+
+// NewMiddleware creates a new handler struct responsible for enriching the response of Async mutations with Operation URL location information
+func NewMiddleware(directorURL string) *middleware {
+	return &middleware{
+		directorURL: directorURL,
+	}
+}
+
+// ExtensionHandler enriches Async mutation responses with Operation URL location information and also empties the data property of the graphql response for such requests
+func (m *middleware) ExtensionHandler(ctx context.Context, next func(ctx context.Context) []byte) []byte {
```
KirilKabakchiev, line 44: can we not split into two but instead have all logic for operation pre and post processing in one place (in this case the directive)? I suspect it would be even simpler code there, no need for sjson deletion, just returning empty bytes

components/director/internal/domain/application/repository.go
```diff
@@ -213,3 +240,17 @@ func (r *pgRepository) multipleFromEntities(entities EntityCollection) ([]*model
...
+
+func updateOperationContext(ctx context.Context, appID string) error {
+	operations, exists := operation.FromCtx(ctx)
+	if !exists {
+		return apperrors.NewInternalError("unable to fetch operations from context")
+	}
+
+	op := (*operations)[len(*operations)-1]
+
+	op.ResourceID = appID
```
KirilKabakchiev, line 252: lets have that on a central place instead - otherwise we have to add these lines to each repo. Proposal is to put this in the directive (see the other comment on the directive). Added benefit (possible as we currently wont be needing related resoureces) is that we will have the full operation construction happening in the directive

components/schema-migrator/migrations/director/202102031458_extend_application_resources.up.sql
```diff
@@ -0,0 +1,102 @@
...
+    'SYNC'
+);
+
+ALTER TABLE webhooks
+    ADD COLUMN mode webhook_mode,
+    ADD COLUMN correlation_id_key varchar(256),
+    ADD COLUMN retry_interval int,
+    ADD COLUMN timeout int,
+    ADD COLUMN url_template jsonb,
+    ADD COLUMN input_template jsonb,
```
KirilKabakchiev, line 89: i didnt see any template definitions and validations of templates in the code. They are not arbitary, there are some rules that have to be followed and these rules should come from this component and the validation should be done during webhook creation. We can discuss ideas/proposals.

components/schema-migrator/migrations/director/202102031458_extend_application_resources.up.sql
```diff
@@ -0,0 +1,102 @@
...
+
+ALTER TABLE event_api_definitions
+    ADD CONSTRAINT event_api_id_ready_unique UNIQUE (id, ready);
+
+ALTER TABLE documents
+    ADD CONSTRAINT document_id_ready_unique UNIQUE (id, ready);
+
+ALTER TABLE bundles
+    ADD CONSTRAINT bundles_applications_ready_fk
+        FOREIGN KEY (app_id, ready) REFERENCES applications (id, ready) ON UPDATE CASCADE;
```
KirilKabakchiev, line 55: havent looked at the tests yet but do we have tests for that one ? Basically marking an as ready marks  apis, events as ready as well.

components/director/pkg/operation/scheduler.go
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package operation
+
+// Scheduler is responsible for scheduling any provided Operation entity for later processing
+//go:generate mockery -name=Scheduler -output=automock -outpkg=automock -case=underscore
+type Scheduler interface {
```
KirilKabakchiev, line 21: move that interface in the directive file

KirilKabakchiev (dismissed):

components/director/pkg/operation/api.go
```diff
@@ -0,0 +1,152 @@
...
+// ResourceFetcherFunc defines a function which fetches a particular application by tenant and app ID
+type ResourceFetcherFunc func(ctx context.Context, tenant, id string) (*model.Application, error)
+
+// TenantLoaderFunc defines a function which fetches the tenant for a particular request
+type TenantLoaderFunc func(ctx context.Context) (string, error)
+
+type handler struct {
+	transact            persistence.Transactioner
+	resourceFetcherFunc ResourceFetcherFunc
+	tenantLoaderFunc    TenantLoaderFunc
```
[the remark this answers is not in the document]
NickyMateev: It was added as generic func due to import cycle issues when referencing the `tenant` package in this one.

components/director/pkg/operation/directive.go
```diff
@@ -0,0 +1,108 @@
...
+		scheduler: scheduler,
+	}
+}
+
+// HandleOperation enriches the request with an Operation information when the requesting mutation is annotated with the Async directive
+func (d *directive) HandleOperation(ctx context.Context, _ interface{}, next gqlgen.Resolver, op graphql.OperationType) (res interface{}, err error) {
+	resCtx := gqlgen.GetResolverContext(ctx)
+	mode, ok := resCtx.Args[ModeParam].(*graphql.OperationMode)
+	if !ok {
+		return nil, apperrors.NewInternalError(fmt.Sprintf("could not get %s parameter", ModeParam))
```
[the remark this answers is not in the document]
NickyMateev: I think this should be Internal Server error, because the our gql schema has specified a default OperationMode value. So in that sense if it's missing here, something has gone wrong on our side

components/director/pkg/operation/middleware.go
NickyMateev: The initial idea was that if there are multiple gql mutations in a single request, this middleware would be called only once while the directives + resolvers are called for every mutation - this would allow us to initialize the the Operations slize in the context in a single place and also register the extensions in one shot. Maybe we can think of a way to move this in the directive, not sure if we'll be able to build on the registered extensions, but we'll try it out

components/director/internal/domain/webhook/repository_test.go
```diff
@@ -28,6 +30,8 @@ const (
 	testCaseErrorOnDBCommunication   = "got error on db communication"
 )
 
+var template = `{}`
```
desislavaa, line 33: we can move this into `fixtures_test.go` (it's also referred there)

components/schema-migrator/migrations/director/202102031458_extend_application_resources.up.sql
NickyMateev: We've only tested this manually. Maybe it won't be trivial to make a clean test for this as this would require making an async request, which we still don't have a full implementation of, we'll see if something can be done nevertheless

components/director/pkg/graphql/api.go
```diff
@@ -8,8 +8,13 @@ type APIDefinition struct {
 	Spec        *APISpec `json:"spec"`
 	TargetURL   string   `json:"targetURL"`
 	//  group allows you to find the same API but in different version
-	Group   *string  `json:"group"`
-	Version *Version `json:"version"`
+	Group     *string   `json:"group"`
+	Version   *Version  `json:"version"`
+	Ready     bool      `json:"ready"`
```
desislavaa, line 13: Same idea with the base entity can be done in the graphql package as well

components/director/pkg/operation/api.go
```diff
@@ -0,0 +1,152 @@
...
+		ResourceID   string
+		ResourceType string
+	}{
+		ResourceID:   queryParams.Get(ResourceIDParam),
+		ResourceType: queryParams.Get(ResourceTypeParam),
+	}
+
+	log.C(ctx).Infof("Executing Operation API with resourceType: %s and resourceID: %s", inputParams.ResourceType, inputParams.ResourceID)
+
+	if err := validation.ValidateStruct(&inputParams,
```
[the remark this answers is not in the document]
NickyMateev: The op struct is internally created processed, I think we should expect it to be properly populated. This here is for the purpose to tell the client that his provided query params are invalid and he should correct them.

components/director/pkg/operation/middleware_test.go
```diff
@@ -0,0 +1,268 @@
...
+			gqlResults: gqlResults,
+		}
+
+		middleware := operation.NewMiddleware(directorURL)
+		resp := middleware.ExtensionHandler(context.Background(), dummyResolver.SuccessResolve)
+
+		require.Equal(t, gqlResultResponse(gqlResults[0].resultName), resp)
+	})
+
+	t.Run("when an async operation is found in the context, location extension would be attached and data would dropped", func(t *testing.T) {
```
desislavaa, line 56: "...and data would _be_ dropped" here and everythere below

desislavaa (commented):
A few minor ones from me

components/director/pkg/operation/api.go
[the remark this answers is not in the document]
KirilKabakchiev: Operation is still passed to scheduler and scheduler will has public Schedule method. Ideally a public method that has a contract with the outside world via such a struct should do a .Validate on it. So the best option would be to have such validate and call it in scheduler so that it can make sure it is given a valid op before sending to api server

components/director/pkg/operation/api.go
```diff
@@ -0,0 +1,152 @@
...
+		validation.Field(&inputParams.ResourceID, is.UUID),
+		validation.Field(&inputParams.ResourceType, validation.Required, validation.In(strings.ToLower(resource.Application.ToLower())))); err != nil {
+		apperrors.WriteAppError(ctx, writer, apperrors.NewInvalidDataError("Unexpected resource type and/or GUID"), http.StatusBadRequest)
+		return
+	}
+
+	tx, err := h.transact.Begin()
```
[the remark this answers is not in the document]
NickyMateev: Let's look this in a separate task.

components/schema-migrator/migrations/director/202102031458_extend_application_resources.up.sql
NickyMateev: These tests will be part of the operations controller e2e tests.

components/director/pkg/operation/api.go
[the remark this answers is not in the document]
KirilKabakchiev: @dpanayotov can we add one for this one ?

components/director/pkg/operation/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+	operations := make([]*Operation, 0)
+	ctx = SaveToContext(ctx, &operations)
+
+	resp := next(ctx)
+
+	locations := make([]string, 0)
+	for _, operation := range operations {
+		operationURL := fmt.Sprintf("%s/operations?resourceID=%s&resourceType=%s", m.directorURL, operation.ResourceID, operation.ResourceType)
```
desislavaa, line 52: you can use the constants from api.go instead of hardcoded `resourceID` and `resourceType` (they should be "snake-cased")

components/director/internal/domain/api/entity.go
```diff
@@ -23,3 +29,31 @@ type EntitySpec struct {
 	SpecFormat sql.NullString `db:"spec_format"`
 	SpecType   sql.NullString `db:"spec_type"`
 }
+
+func (e *Entity) SetReady(ready bool) {
+	e.Ready = ready
+}
+
+func (e *Entity) GetCreatedAt() time.Time {
```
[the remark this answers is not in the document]
NickyMateev: This would cause conflict with the variable names `CreatedAt`, `UpdatedAt` and `DeletedAt`

components/director/pkg/operation/directive.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	ctx = SaveToContext(ctx, &[]*Operation{operation})
+
+	resp, err := next(ctx)
+	if err != nil {
+		log.C(ctx).WithError(err).Errorf("An error occurred while processing operation: %s", err.Error())
+		return nil, apperrors.NewInternalError("Unable to process operation")
+	}
+
+	operationID, err := d.scheduler.Schedule(*operation)
```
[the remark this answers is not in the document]
NickyMateev: Will be added in the Scheduler PR.

components/director/cmd/director/main.go
```diff
@@ -225,6 +236,13 @@ func main() {
 
 	mainRouter.HandleFunc(cfg.AuthenticationMappingEndpoint, authnMappingHandlerFunc.ServeHTTP)
 
+	appRepo := defaultApplicationRepo()
+	operationHandler := operation.NewHandler(transact, appRepo.GetByID, tenant.LoadFromContext)
+
+	operationsAPIRouter := mainRouter.PathPrefix(cfg.OperationEndpoint).Subrouter()
```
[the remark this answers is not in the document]
NickyMateev: In such a scenario the client receives EOF, same is the case with the other Director APIs. I suggest adding this as a general fix-it.

components/director/pkg/operation/api.go
```diff
@@ -0,0 +1,152 @@
...
+		return
+	}
+
+	if err := tx.Commit(); err != nil {
+		log.C(ctx).WithError(err).Errorf("An error occurred while closing database transaction: %s", err.Error())
+		apperrors.WriteAppError(ctx, writer, apperrors.NewInternalError("Unable to finalize database operation"), http.StatusInternalServerError)
+		return
+	}
+
+	opResponse := &OperationResponse{
```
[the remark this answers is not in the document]
NickyMateev: I really like the proposal with the method on `model.Model` - I faced a lot of import cycle issues between the `model` and `operation` packages trying to achieve this unfortunately. For now I'd like to extract this as a separate function (your first proposal) instead and try to improve it with following PRs as I don't want to rush it now, due to the fact that two other PRs are waiting on this one.

components/director/pkg/operation/directive.go
```diff
@@ -0,0 +1,108 @@
...
+
+// HandleOperation enriches the request with an Operation information when the requesting mutation is annotated with the Async directive
+func (d *directive) HandleOperation(ctx context.Context, _ interface{}, next gqlgen.Resolver, op graphql.OperationType) (res interface{}, err error) {
+	resCtx := gqlgen.GetResolverContext(ctx)
+	mode, ok := resCtx.Args[ModeParam].(*graphql.OperationMode)
+	if !ok {
+		return nil, apperrors.NewInternalError(fmt.Sprintf("could not get %s parameter", ModeParam))
+	}
+
+	ctx = SaveModeToContext(ctx, *mode)
```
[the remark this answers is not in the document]
NickyMateev: We rely on knowing what mode the current operation is in two places in the repo layer (we do mainly checks of the sort `if operation.ModeFromCtx(ctx) == graphql.OperationModeAsync {`).

If we remove the mode from the context we could  achieve the same check logic by looking for an operation object in the context, but this breaks when there are multiple requests (one is async and another is sync). We need to know for every single request what operation mode we're in and we can't infer that from the operations in the context.

In this sense we need to keep the operation mode.

components/schema-migrator/migrations/director/202102031458_extend_application_resources.up.sql
NickyMateev: Webhook validation added with: https://github.com/kyma-incubator/compass/pull/1714/commits/e8c3613bd623371576d93844ba5f4326f3c640c2

components/director/cmd/director/main.go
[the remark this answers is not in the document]
KirilKabakchiev: for graphql i think it returns proper error. If you've verified that the whole process/director does not crash on panic,  its ok to have it as a separate bli

components/director/pkg/operation/directive.go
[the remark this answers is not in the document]
KirilKabakchiev: i didnt understand why it breaks with sync requests/still dont see the purpose of mode. There is no operation on the context for them at all == sync. 


If you take a step back though, Ideally since you dont have related resouces with this implementation it does not make sense to pass the op in the context at all. You can just set the ready field here (keep the timestamps in the storage)

desislavaa (approved):

components/director/internal/domain/webhook/repository.go
```diff
@@ -44,7 +46,7 @@ func NewRepository(conv EntityConverter) *repository {
 	return &repository{
 		singleGetter: repo.NewSingleGetter(resource.Webhook, tableName, tenantColumn, webhookColumns),
 		creator:      repo.NewCreator(resource.Webhook, tableName, webhookColumns),
-		updater:      repo.NewUpdater(resource.Webhook, tableName, []string{"type", "url", "auth"}, tenantColumn, []string{"id", "app_id"}),
+		updater:      repo.NewUpdater(resource.Webhook, tableName, updatableColumns, tenantColumn, []string{"id", "app_id"}),
```
pankrator, line 49: As there are now more than just `id` and `app_id`, but also `runtime_id` and `integration_system` is this `idColumns` enough?.
We have to keep in mind that if in the future we want to have for example `runtime_id` and not `app_id` the update SQL will now work as it will add where clause with `app_id`

pankrator: not sure if this is enough descriptive, but you can ask offline to further discuss. I am not seeing it as a problem as of now, but perhaps in the future

pankrator: As a second thought, each of them can be empty right? If this is true, then there will only be where clause with empty `app_id` for example. Then you can just add the rest of the ID fields in this array for completeness or leave only the `id` of the webhook. Not sure why the `app_id` is needed at all though.

components/director/internal/model/base_entity.go
```diff
@@ -0,0 +1,101 @@
...
+	SetDeletedAt(t time.Time)
+
+	GetError() *string
+	SetError(err string)
+}
+
+type BaseEntity struct {
+	ID        string
+	Ready     bool
+	CreatedAt time.Time
```
pankrator, line 49: shouldn't it have `json` tags at least for createdAt, UpdatedAt and DeletedAt?

pankrator: nvm....

components/director/pkg/operation/api.go
```diff
@@ -0,0 +1,148 @@
...
+
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	res, err := h.resourceFetcherFunc(ctx, tenantID, op.ResourceID)
+	if err != nil {
+		log.C(ctx).WithError(err).Errorf("An error occurred while fetching resource from database: %s", err.Error())
+
+		if apperrors.IsNotFoundError(err) {
+			apperrors.WriteAppError(ctx, writer, apperrors.NewNotFoundErrorWithMessage(resource.Application, op.ResourceID,
+				fmt.Sprintf("Operation for application with id %s not found", op.ResourceID)), http.StatusNotFound)
```
dpanayotov, line 100: some places it's `resource`, others - `application`. I guess because it's hardcoded for now, but we should have it in mind when introducing other types

components/director/pkg/operation/api.go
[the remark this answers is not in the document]
dpanayotov: I don't understand why there's a transaction when nothing is done after a select? To discuss offline and decide on a BLI

pankrator (approved):

PR #1719, opened 2021-02-04
Support multiple API/Event Specifications

DimitarPetrov:
**Description**
Currently, each API/Event has an embedded spec_data, spec_type, and spec_format fields which represent the specification of the API/Event. In ORD each API/Event can provide the spec in different formats, therefore more than one specification per API/Event should be supported.

- Introducing new Entity - Specification
- Drop spec_data, spec_type, and spec_format columns of API/Event in favor of the new Specifications Entity
- Adapt Fetch Requests to write to the Specification table rather than to the API/Event directly.
- Remove Fetch Request manipulation from APIs/Events since it now belongs to the Specification
- Adapt all the tests and add new ones
- Adapts ORD Service In: https://github.com/kyma-incubator/ord-service/pull/13

**Pull Request status**

- [x] Implementation
- [x] Unit tests


NickyMateev (commented):
I think my main comment is that it seems like the way specs have been removed from the APIs and Events it seems like is not backwards compatible for GraphQL clients.
Fortunately we don't have such already, other than the Connectivity Adapter, which I don't see adapted

NickyMateev (approved):

PR #1721, opened 2021-02-08
Downgrade kube builder's dependencies in the system broker

PetarTodorovv:
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


NickyMateev (changes requested):
Bump values.yaml and validate operations-controller works properly when the director/system broker dependency is added now after the downgrade

NickyMateev (approved):

NickyMateev (approved):

PR #1723, opened 2021-02-08
Allow e2e test execution using Docker driver

ataleksandrov:
**Description**

- Patch test definitions when using docker driver

**Related issue(s)**
- The Minikube IP using Docker driver is 127.0.0.1. The IP is used in most of the e2e TestDefinitions in hostAliases field, see: https://github.com/kyma-incubator/compass/blob/master/chart/compass/charts/connector/templates/tests/test-connector.yaml#L48-L52


installation/scripts/testing.sh
```diff
@@ -8,6 +8,20 @@ readonly TMP_DIR=$(mktemp -d)
...
+USED_DRIVER=$(minikube profile list -o json | jq -r ".valid[0].Config.Driver")
+if [[ $USED_DRIVER == "docker" ]]; then
+  echo "----------------------------"
+  echo "- Docker driver is used, patching Test definitions..."
+  echo "----------------------------"
+
+  MINIKUBE_IP=$(minikube ssh egrep "minikube$" /etc/hosts | cut -f1)
+  NAMESPACES=( "kyma-system" "compass-system" )
+  for namespace in "${NAMESPACES[@]}"; do
+   kubectl get testdefinitions.testing.kyma-project.io -o name -n ${namespace} | sed -e 's/.*\///g' | xargs -I {} kubectl patch testdefinitions.testing.kyma-project.io {} -n ${namespace} --type=json -p="[{\"op\": \"replace\", \"path\": \"/spec/template/spec/hostAliases/0/ip\", \"value\": \"${MINIKUBE_IP}\"}]" 2>/dev/null
```
KirilKabakchiev, line 21: this should be done as part of the charts installation imo. Otherwise now if i setup a minikube with compass and docker driver, the test definitions will be broken until running testing.sh (and one might decide to simply use `kyma  test run` and `kyma test status` and not testing.sh).

One proposal - why don't you extend run.sh - when `--docker-driver` is set, attach another override (configmap) in the installer to set `global.minikubeIP` to the correct value  ?

ataleksandrov: Actually I tried this approach at first. My concern was about the test definition that come from the Kyma (dex-connection for instance). I could not think of another way to patch them unless we contribute this Minikube IP override in their code base as well. I was not sure if they would accept more branching(just for one driver) in their code, also is the Kyma cli the only way of installing Kyma?

KirilKabakchiev: I think in some test definitions they have the hostAliases and in some they don't, it won't be any new branching just making all test definitions consistent. I will look a bit deeper.

On the sidenote, I think it will also make sense that we actually test that everything works with docker driver if we intend to use it. There is a compass-integration job that already does minikube testing but we need to pass an additional parameter to the script it uses - `deploy-and-test.sh` located in compass repo so that it uses docker driver as well

DimitarPetrov:
Addressed in: https://github.com/kyma-incubator/compass/pull/1746

PR #1731, opened 2021-02-15
Scheduler for operations and update operation handler

pankrator:
**Scheduler to create operations CRDs in the k8s cluster which will be handled by certain controller. Update operation handler which will be called upon certain operation completeness either FAID or SUCCEED**

- Scheduler for operations
- Update operation handler


components/director/pkg/operation/update_operation_handler.go
```diff
@@ -0,0 +1,154 @@
...
+func NewUpdateOperationHandler(transact persistence.Transactioner, resourceUpdaterFuncs map[resource.Type]ResourceUpdaterFunc, resourceDeleterFuncs map[resource.Type]ResourceDeleterFunc) *updateOperationHandler {
+	return &updateOperationHandler{
+		transact:             transact,
+		resourceUpdaterFuncs: resourceUpdaterFuncs,
+		resourceDeleterFuncs: resourceDeleterFuncs,
+	}
+}
+
+// ServeHTTP handles the Operations API requests
+func (h *updateOperationHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
```
NickyMateev, line 70: Am I missing something or does the writer never get used to set and return appropriate status code (+ response body? empty). Probably 200 OK is the default but maybe it would be good to be explicit and set it manually

components/director/pkg/operation/update_operation_handler.go
```diff
@@ -0,0 +1,154 @@
...
+		log.C(ctx).WithError(err).Errorf("An error occurred while opening db transaction: %s", err.Error())
+		apperrors.WriteAppError(ctx, writer, apperrors.NewInternalError("Unable to establish connection with database"), http.StatusInternalServerError)
+		return
+	}
+	defer h.transact.RollbackUnlessCommitted(ctx, tx)
+
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	resourceUpdaterFunc := h.resourceUpdaterFuncs[operation.ResourceType]
+	switch operation.OperationType {
```
NickyMateev, line 115: I see that this API was build with the undestanding that it will be invoked in the scenarios when there's nothing more to be done regarding an operation and the `application.ready` property should be flipped to `true` (whether there's an error or not).
After some discussions around the operations controller reconciliation loop I think we'll have to do a little bit more "intermediate" invocations of this API from the controller side - essentially every time the reconciliation loop ends we'll be calling this API. This is so that we can periodically reconcile the state in Director DB (the current state) with what's specified by the Operation CR (desired state). This also means that we shouldn't be flipping the `ready` property as we currently do here. I suggest extending the `OperationRequest` struct with `Ready` bool field so that the operations controller can itself dictate and specify what's the actually value of this property at a particular point in time.

components/director/pkg/operation/update_operation_handler.go
pankrator: you are right, it is missing

components/director/pkg/operation/update_operation_handler.go
```diff
@@ -0,0 +1,154 @@
...
+	case graphql.OperationTypeUpdate:
+		if err := resourceUpdaterFunc(ctx, operation.ResourceID, true, getError(operation.Error)); err != nil {
+			log.C(ctx).WithError(err).Errorf("While updating resource %s with id %s", operation.ResourceType, operation.ResourceID)
+			apperrors.WriteAppError(ctx, writer, apperrors.NewInternalError("Unable to update resource %s with id %s", operation.ResourceType, operation.ResourceID), http.StatusInternalServerError)
+			return
+		}
+	case graphql.OperationTypeDelete:
+		resourceDeleterFunc := h.resourceDeleterFuncs[operation.ResourceType]
+		if operation.Error != "" {
+			if err := resourceUpdaterFunc(ctx, operation.ResourceID, true, getError(operation.Error)); err != nil {
```
ataleksandrov, line 127: Maybe `ready=false`? see: https://github.com/kyma-incubator/compass/blob/master/components/director/pkg/operation/api.go#L139-L142

components/director/pkg/operation/update_operation_handler.go
```diff
@@ -0,0 +1,154 @@
...
+		resourceUpdaterFuncs: resourceUpdaterFuncs,
+		resourceDeleterFuncs: resourceDeleterFuncs,
+	}
+}
+
+// ServeHTTP handles the Operations API requests
+func (h *updateOperationHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	ctx := request.Context()
+
+	if request.Method != http.MethodPost {
```
ataleksandrov, line 73: `PUT` instead of `POST`?

components/director/cmd/director/main.go
```diff
@@ -468,3 +503,18 @@ func webhookService() webhook.WebhookService {
...
 	return webhook.NewService(webhookRepo, uidSvc)
 }
+
+func buildScheduler(namespace string) (operation.Scheduler, error) {
+	cfg, err := rest.InClusterConfig()
+	if err != nil {
+		log.D().WithError(err).Warn("Could not start K8S Scheduler. Will use the default one")
+		return &operation.DefaultScheduler{}, nil
+	}
+	k8sClient, err := client.NewForConfig(cfg)
```
ataleksandrov, line 513: We could add timeout like: https://github.com/kyma-incubator/compass/blob/master/components/connector/cmd/main.go#L141

components/director/pkg/operation/update_operation_handler.go
desislavaa: Ah, we'll need to change that, nice catch! The resource is `ready=false` until an operation is running. At first it was with `ready=false` when the operation failed or was still running, but the concept was changed.

components/director/pkg/operation/k8s/k8s_scheduler.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		return "", err
+	}
+	return string(updatedOperation.UID), err
+}
+
+func isOpInProgress(op *v1alpha1.Operation) bool {
+	for _, cond := range op.Status.Conditions {
+		if cond.Status == v1.ConditionTrue {
+			return false
```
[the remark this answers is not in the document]
desislavaa: We have two possible conditions - Ready and Error.
When an operation is running, whey would both be `false` and when it finishes only one will become `true`.

There is a check for concurrent operations before scheduling one, so basically the op should never be in progress.

components/director/cmd/director/main.go
```diff
@@ -468,3 +503,19 @@ func webhookService() webhook.WebhookService {
 
 	return webhook.NewService(webhookRepo, uidSvc)
 }
+
+func buildScheduler(namespace string, timeout time.Duration) (operation.Scheduler, error) {
+	cfg, err := rest.InClusterConfig()
```
NickyMateev, line 508: btw while testing the combined scheduler + controller flow I had to change this to the following in order to run the  director on localhost:

```
	var kubeconfig *string
	if home := homedir.HomeDir(); home != "" {
		kubeconfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
	} else {
		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
	}
	flag.Parse()

	// use the current context in kubeconfig
	cfg, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
	if err != nil {
		panic(err.Error())
	}
```

Maybe it makes to introduce an env var switch which will be enabled when using `run.sh` so we can develop with local clusters as well.

components/director/cmd/director/main.go
```diff
@@ -468,3 +503,19 @@ func webhookService() webhook.WebhookService {
...
+	if err != nil {
+		log.D().WithError(err).Warn("Could not start K8S Scheduler. Will use the default one")
+		return &operation.DefaultScheduler{}, nil
```
NickyMateev, line 511: Not sure we want to keep this default scheduler - it doesn't make sense to allow async operations with it. It was left out because we had to have a stub implementation before the actual one was implemented.

I think this should either throw an error here, or the directives responsible for determining what kind of operation is made (ASYNC or SYNC) it should throw an http client error if an ASYNC operation is requested but only the default scheduler is enabled.

Essentially my concern is that the default scheduler is not really any implementation at all.

NickyMateev (changes requested):

components/director/cmd/director/main.go
NickyMateev: Reference links to running a k8s client in out-of-cluster mode:
1. https://github.com/kubernetes/client-go#how-to-use-it
2. https://github.com/kubernetes/client-go/tree/master/examples/out-of-cluster-client-configuration

components/director/pkg/operation/k8s/k8s_scheduler.go
```diff
@@ -0,0 +1,94 @@
...
+			Name: operationName,
+		},
+	}
+	return updateOperationSpec(op, result)
+}
+
+func updateOperationSpec(op *operation.Operation, k8sOp *v1alpha1.Operation) *v1alpha1.Operation {
+	k8sOp.Spec = v1alpha1.OperationSpec{
+		OperationCategory: op.OperationCategory,
+		OperationType:     v1alpha1.OperationType(title(string(op.OperationType))),
```
NickyMateev, line 82: This part here - `title(string(op.OperationType))`, will cause the new Operations update handler to fail validating the request as it expects all upper case letters for the operation type:

```
if err := validation.ValidateStruct(&operation,
	validation.Field(&operation.ResourceID, is.UUID),
	validation.Field(&operation.OperationType, validation.Required, validation.In(graphql.OperationTypeCreate, graphql.OperationTypeUpdate, graphql.OperationTypeDelete)),
	validation.Field(&operation.ResourceType, validation.Required, validation.In(resource.Application))); err != nil {
	apperrors.WriteAppError(ctx, writer, apperrors.NewInvalidDataError("Invalid operation properties: %s", err), http.StatusBadRequest)
	return
}
```

Reference link: https://github.com/kyma-incubator/compass/blob/5debfe3ffcc4e31a3487995b17f25e53435a9eff/components/director/pkg/operation/update_operation_handler.go#L98

NickyMateev (changes requested):

components/director/pkg/operation/update_operation_handler.go
```diff
@@ -0,0 +1,156 @@
...
+	case graphql.OperationTypeUpdate:
+		if err := resourceUpdaterFunc(ctx, operation.ResourceID, true, getError(operation.Error)); err != nil {
+			log.C(ctx).WithError(err).Errorf("While updating resource %s with id %s", operation.ResourceType, operation.ResourceID)
+			apperrors.WriteAppError(ctx, writer, apperrors.NewInternalError("Unable to update resource %s with id %s", operation.ResourceType, operation.ResourceID), http.StatusInternalServerError)
+			return
+		}
+	case graphql.OperationTypeDelete:
+		resourceDeleterFunc := h.resourceDeleterFuncs[operation.ResourceType]
+		if operation.Error != "" {
+			if err := resourceUpdaterFunc(ctx, operation.ResourceID, true, getError(operation.Error)); err != nil {
```
NickyMateev, line 127: The API seems to fail deep in the storage layer whenever there's an error to persist - the raw SQL error returns something of the kind: `Token "reconciliation is invalid` (the original error message to persist was `reconciliation timeout reached`

NickyMateev (changes requested):

components/director/cmd/director/main.go
```diff
@@ -86,8 +91,9 @@ import (
 const envPrefix = "APP"
 
 type config struct {
-	Address string `envconfig:"default=127.0.0.1:3000"`
-	AppURL  string `envconfig:"APP_URL"`
+	Address         string `envconfig:"default=127.0.0.1:3000"`
+	InternalAddress string `envconfig:"default=127.0.0.1:3002"`
```
NickyMateev, line 95: The `InternalAddress`, just like the `Address` should be defined somewhere centrally in our charts and configurable through environment variables. This way the host+port will be defined in one place and all other components which needed, like the `operations-controller` and charts, like the network policy, can reference it.

components/director/pkg/operation/k8s/k8s_scheduler.go
desislavaa: We'll need to tweak the validation then, good catch! The "title"-thingy was needed because of the OperationType in the CRD: https://github.com/kyma-incubator/compass/blob/22003412f0640de76ce17b07fcbc424964e6e6c6/components/operations-controller/api/v1alpha1/operation_types.go#L22

desislavaa: It started to look really workaround-ish and I've just changed the types in the `graphql` package to match the ones in the CRD

components/director/cmd/director/main.go
```diff
@@ -468,3 +506,51 @@ func webhookService() webhook.WebhookService {
...
+
+	var kubeconfig *string
+	home := homedir.HomeDir()
+	if home == "" {
+		return nil, errors.New("failed to get home dir in order to build k8s client for scheduler")
+	}
+
+	log.C(ctx).Info("Using local configuration for k8s client")
+	kubeconfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
+	flag.Parse()
```
KirilKabakchiev, line 554: this flag is already defined by controller-runtime, also the outsideCluster variable seems unneeded. If config lookup is done in the correct order, it should seemlessly look for inclusterconfig if it is not found else where

I'd propose to drop this method and instead use `func GetConfig() (*rest.Config, error)`

from the godoc 
```
// Config precedence
//
// * --kubeconfig flag pointing at a file
//
// * KUBECONFIG environment variable pointing at a file
//
// * In-cluster config if running in cluster
//
// * $HOME/.kube/config if exists
```

components/director/pkg/operation/directive.go
```diff
@@ -148,10 +151,53 @@ func (d *directive) HandleOperation(ctx context.Context, _ interface{}, next gql
...
+	app, err := d.resourceFetcherFunc(ctx, tenant, resourceID)
+	if err != nil {
+		if apperrors.IsNotFoundError(err) {
+			return err
+		}
+
+		return apperrors.NewInternalError("failed to fetch resource with id %s", resourceID)
+	}
+
+	if app.GetDeletedAt().IsZero() && app.GetUpdatedAt().IsZero() && !app.GetReady() && (app.GetError() == nil || *app.GetError() == "") { // CREATING
```
dzahariev, line 187: (app.GetError() == nil || *app.GetError() == "") can be extracted as helper function as I can see this check on several places

components/director/pkg/operation/update_operation_handler.go
```diff
@@ -0,0 +1,184 @@
...
+		return
+	}
+
+	writer.WriteHeader(http.StatusOK)
+}
+
+func operationRequestFromBody(ctx context.Context, request *http.Request) (*OperationRequest, *errResponse) {
+	bytes, err := ioutil.ReadAll(request.Body)
+	if err != nil {
+		return nil, &errResponse{apperrors.NewInternalError("Unable to read request body"), http.StatusInternalServerError}
```
dzahariev, line 154: Probably http.StatusBadRequest is better in this case.

components/director/pkg/str/strings.go
```diff
@@ -49,3 +50,7 @@ func PrefixStrings(in []string, prefix string) []string {
 
 	return prefixedFieldNames
 }
+
+func Title(s string) string {
```
dzahariev, line 54: Reuse this function on all places.

components/director/run.sh
```diff
@@ -13,6 +13,7 @@ ROOT_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 SKIP_DB_CLEANUP=false
 REUSE_DB=false
+DISABLE_ASYNC_MODE=true
```
dzahariev, line 16: It is a personal opinion, but I do not like reversed logic variables. 
I prefer EnabledX vs DisabledX, as boolean values are easily guessed later in code:
TRUE=ENABLED
FALSE=DISABLED

components/operations-controller/client/client.go
```diff
@@ -0,0 +1,119 @@
...
+func (c *OperationsClient) Operations(namespace string) OperationsInterface {
+	c.namespace = namespace
+	return c
+}
+
+func (c *OperationsClient) List(ctx context.Context, opts metav1.ListOptions) (*v1alpha1.OperationList, error) {
+	result := v1alpha1.OperationList{}
+	err := c.restClient.Get().
+		Namespace(c.namespace).
+		Resource("operations").
```
dzahariev, line 52: "operations" can be extracted as constant, probably in operation.go and reused.

components/director/run.sh
desislavaa: The default behavior is with async enabled, so explicitly disabling it seemed more reasonable - it will be harder to misconfigure it

components/operations-controller/Makefile
```diff
@@ -1,6 +1,6 @@
-APP_NAME = compass-operations-controller
+APP_NAME = compass-operations-controller-manager
 APP_PATH = components/operations-controller
-BUILDPACK = eu.gcr.io/kyma-project/test-infra/buildpack-golang-toolbox:v20210125-6234473e
+BUILDPACK = eu.gcr.io/kyma-project/test-infra/pr/buildpack-golang-kubebuilder2:PR-3299
```
NickyMateev, line 3: https://github.com/kyma-project/test-infra/pull/3299 has been merged. Use the newly built image as a result of it, instead of the one tagged with the PR.

NickyMateev (approved):

components/operations-controller/Makefile
NickyMateev, line 3: https://github.com/kyma-project/test-infra/pull/3299 has been merged. Use the newly built image as a result of it, instead of the one tagged with the PR.

NickyMateev (dismissed):

components/director/pkg/operation/update_operation_handler.go
pankrator: In this case we try to read the request body which might not me a user (request) mistake

components/director/internal/domain/application/repository.go
```diff
@@ -80,6 +85,26 @@ func (r *pgRepository) Delete(ctx context.Context, tenant, id string) error {
...
+	opMode := operation.ModeFromCtx(ctx)
+	if opMode == graphql.OperationModeAsync {
+		app, err := r.GetGlobalByID(ctx, id)
+		if err != nil {
+			return err
+		}
+
+		app.Ready = false
+		app.SetError("")
+		if app.DeletedAt.IsZero() { // Needed for the tests but might be useful for the production also
```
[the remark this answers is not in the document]
pankrator: It is needed because when delete is started, director has to set the deleted at property

PetarTodorovv (approved):

PR #1732, opened 2021-02-16
Use bundles without query param

ataleksandrov:
**Description**

- Remove useBundles query param

### DIFF
--- a/components/director/internal/packagetobundles/handler.go
+++ b/components/director/internal/packagetobundles/handler.go
@@ -75,7 +75,7 @@ func (h *Handler) Handler() func(next http.Handler) http.Handler {
 			}
 
 			body := string(reqBody)
-			//body = strings.ReplaceAll(body, "\\n", "") // removes unnecessary complexity from the next regexes
+			body = strings.ReplaceAll(body, "\\n", " ") // removes unnecessary complexity from the next regexes
 
 			unmodifiedBody := body
 
--- a/components/director/internal/packagetobundles/handler.go
+++ b/components/director/internal/packagetobundles/handler.go
@@ -42,7 +42,7 @@ import (
 	"github.com/vektah/gqlparser/gqlerror"
 )
 
-const usesBundlesLabel = "usesBundles"
+const usesBundlesLabel = "useBundles"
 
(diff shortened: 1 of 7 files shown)


components/director/internal/packagetobundles/handler.go
```diff
@@ -148,6 +122,31 @@ func (h *Handler) Handler() func(next http.Handler) http.Handler {
...
+					writeAppError(ctx, w, appErr, http.StatusInternalServerError)
+					return
+				}
+
+				log.C(ctx).Infof("Will proceed without rewriting the request body. Bundles are adopted for consumer with ID %q and type %q", consumerInfo.ConsumerID, consumerInfo.ConsumerType)
+
+				next.ServeHTTP(w, r)
+
+				if consumerInfo.ConsumerType == consumer.Runtime {
+					if err := h.labelRuntimeWithBundlesParam(ctx, consumerInfo); err != nil {
```
NickyMateev, line 140: Now that we removed the explicit query parameter the following problem has appeared - whenever `unmodifiedBody == body` and `consumerInfo.ConsumerType == consumer.Runtime`, we'll label the runtime with the appropriate Bundles label.
But `unmodifiedBody == body` could be true even if a Runtime has not adopted Bundles yet. This is because this directive is invoked for all possible queries/mutations (not only those related to packages/bundles). So essentially every request query/mutation which does not contain packages will be treated as `unmodifiedBody == body` and the Runtime will be labeled as if it has already migrated without this being the case.
This is not just a theoretical possibility, but it will actually happen as soon as this change is merged (if it is merged in the current way) - this is because the Runtime Agent periodically (every couple minutes) does the following:
1. `applicationsForRuntime` query
2. `setRuntimeLabel` mutation

Until migration the query will contain packages and so we'll replace them and not label the runtime, but immediately after that query, the runtime agent does the `setRuntimeLabel` mutation which does not contain any packages and so the runtime will be labeled as if it has migrated already.

This is why the labelling logic here needs to be reworked, maybe we should discuss this as a team.

NickyMateev (changes requested):

NickyMateev (approved):

ataleksandrov:
fixes #1734

PR #1733, opened 2021-02-17
Introduce Operations Controller

NickyMateev:
# Introduce Operations Controller

**Description**

- Initial implementation of the Operation Controller component
  - Operation retrieval
  - Webhook request preparation according to predefined templates and request objects
  - Webhook execution
  - Webhook polling
  - Maintaining API Server + Director state
- Async + Sync webhooks supported
- 1 webhook per Operation (for now, this covers our use-cases)
- Improved Templating logic
- Adapted the Director component CorrelationID library so that the Correlation ID transport decorator can be reused dynamically
- Evolved the `TokenProvider` abstraction in System Broker component into the more general `AuthorizationProvider` to fit the Operations Controller component more general use-cases for a more flexible secured transport decorator 
- Unit & Integration tests for Operations Controller

**Related issue(s)**

- https://github.com/kyma-incubator/compass/pull/1714
- https://github.com/kyma-incubator/compass/pull/1731
- https://github.com/kyma-incubator/compass/pull/1717

**Pull Request status**

- [x] Implementation
- [x] Unit tests


components/operations-controller/internal/director/client.go
```diff
@@ -0,0 +1,83 @@
...
+
+// defaultClient is the default implementation of the Client interface
+type defaultClient struct {
+	types.ApplicationLister
+	HTTPClient  http.Client
+	DirectorURL string
+}
+
+// TODO: Remove this struct and resuse the one from Director once the Scheduler logic is merged
+type Request struct {
```
NickyMateev, line 45: Reuse type defined in https://github.com/kyma-incubator/compass/pull/1731 after it is merged:
https://github.com/kyma-incubator/compass/blob/b5fa7f582b2ec7f4000948f70959293669e306d6/components/director/pkg/operation/update_operation_handler.go#L41

components/director/pkg/webhook/types.go
```diff
@@ -0,0 +1,259 @@
...
+type Mode string
+
+const (
+	ModeSync  Mode = "SYNC"
+	ModeAsync Mode = "ASYNC"
+)
+
+// Resource is used to identify entities which can be part of a webhook's request data
+type Resource interface {
+	Sentinel()
```
pankrator, line 38: Try removing this method. I think it is not needed as every struct will implicitly implement it. As the method is not used it should be safe to remove.

components/operations-controller/internal/director/client.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
+	"github.com/kyma-incubator/compass/components/director/pkg/resource"
+	"github.com/kyma-incubator/compass/components/system-broker/pkg/types"
+	"net/http"
+)
+
+// Client defines a Director client which is capable of fetching an application
+// and notifying Director for state changes of said application
+type Client interface {
+	types.ApplicationLister
```
pankrator, line 33: Perhaps for another change, but may be it is a good idea to move the dependencies in director

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,423 @@ package controllers
...
+		return 0, nil
+	}
+
+	lastPollTimestamp, err := time.Parse(timeLayout, webhookStatus.LastPollTimestamp)
+	if err != nil {
+		return 0, err
+	}
+
+	nextPollTime := lastPollTimestamp.Add(time.Duration(*webhook.RetryInterval) * time.Minute)
+	return nextPollTime.Sub(time.Now()), nil
```
pankrator, line 297: You can use `time.Until` instead of `t.Sub(time.Now())`

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,423 @@ package controllers
...
+func retrieveWebhooks(appWebhooks []graphql.Webhook, opWebhookIDs []string) (map[string]graphql.Webhook, error) {
+	if len(opWebhookIDs) == 0 {
+		return nil, errors.New("no webhooks found for operation")
+	}
+
+	if len(opWebhookIDs) > 1 {
+		return nil, errors.New("multiple webhooks per operation are not supported")
+	}
+
+	webhooks := make(map[string]graphql.Webhook, 0)
```
pankrator, line 392: No need for 0 when making a map

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,423 @@ package controllers
...
+		case graphql.WebhookModeSync:
+			return r.updateStatus(ctx, logger, operation)
+		default:
+			return ctrl.Result{}, errors.New("unsupported webhook mode")
+		}
+	}
+
+	requeueAfter, err := getNextPollTime(&webhookEntity, *webhookStatus, r.webhookConfig.TimeLayout)
+	if err != nil {
+		return ctrl.Result{}, err
```
pankrator, line 162: In case when parsing the requeue time shouldn't we set the operation status to Failed, as it will not be able to self-heal? At least the spec of the operation should be edited to fix that

pankrator: This will give more information that something bad has happened and we can further troubleshoot. Perhaps there is no need to send back information to director, but at least the Operation CRD can save that error, so we can later inspect it

pankrator (changes requested):

components/operations-controller/main.go
```diff
@@ -52,33 +65,54 @@ func init() {
...
 		os.Exit(1)
 	}
 
-	if err = (&controllers.OperationReconciler{
-		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("Operation"),
-		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
+	httpClient := http.Client{
+		Transport: httputil.NewHTTPTransport(cfg.HttpClient),
```
desislavaa, line 99: The transport can be wrapped as in the gql client:
```
httputil.NewCorrelationIDTransport(httputil.NewErrorHandlerTransport(httputil.NewHTTPTransport(httpCfg)))
```

components/operations-controller/internal/director/client.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return err
+	}
+
+	resp, err := dc.HTTPClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
```
desislavaa, line 78: We can log the body here as well, but if the correlation ID is attached to the logger, it may not be that helpful, we'll have to look at the director logs anyhow, it's your choice

components/operations-controller/internal/webhook/client.go
```diff
@@ -0,0 +1,211 @@
...
+	return &web_hook.ResponseData{
+		Headers: headers,
+		Body:    body,
+	}, nil
+}
+
+func checkForErr(resp *http.Response, successStatusCode *int, error *string) error {
+	var errMsg string
+	if *successStatusCode != resp.StatusCode {
+		errMsg += fmt.Sprintf("response success status code was not met - expected %q, got %q; ", *successStatusCode, resp.StatusCode)
```
desislavaa, line 199: Log the body? It may be helpful for troubleshooting

components/operations-controller/internal/webhook/client.go
```diff
@@ -0,0 +1,211 @@
...
+	mode := web_hook.ModeSync
+	if webhook.Mode != nil {
+		mode = web_hook.Mode(*webhook.Mode)
+	}
+	response, err := web_hook.ParseOutputTemplate(webhook.InputTemplate, webhook.OutputTemplate, *responseData)
+	if err != nil {
+		return nil, err
+	}
+
+	if *response.Location == "" && mode == web_hook.ModeAsync {
```
desislavaa, line 115: if `response` is `nil` the controller crashes

components/director/pkg/webhook/types.go
NickyMateev: I know we discussed it and I couldn't remember why we wanted to have some dummy method, but now I remember - removing the dummy method would practically make any type implicitly implement this interface. This would be just an indirect way of using `interface{}`. So the idea behind the dummy method was to at least restrict to some extent the types which implement this so that we can limit any potential problems later on.

components/operations-controller/controllers/operation_controller.go
NickyMateev: Yeah, I guess we should discuss that, as there are other similar situations like when the webhooks are not 1, or if the requestData fails to parse. All these are problems which should not happen and are sort of internal server errors which cannot self-heal on their own.

So I agree that here we should not return `err` as this would try to requeue. We've added an error log so that we can diagnose the issue, not sure if we should update the operation and/or notify director in such situations - like I said we have a bunch of those, so we should come to a unified agreement what to do in such situations.

components/operations-controller/main.go
NickyMateev: I agree with the `ErrorHandlerTransport`, but not sure we want to wrap the default http client with the `CorrelationIDTransport` decorator. Currently the default http client is used for webhook requests and the correlation ID which was used with the original async mutation is reused (if provided, in our current use case with it will be). Maybe it would be confusing to send two correlation ids to them.

The default http client is also used for making the Update Operation status requests to the Director from Controller side. There again I'm not sure if we should generate random correlation ids rather than reusing the one from the original mutation.

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,428 @@ package controllers
...
+	}
+
+	if operation.Status.Conditions == nil || overrideStatus {
+		operation.Status.Conditions = make([]v1alpha1.Condition, 0)
+	}
+
+	hasReady := false
+	hasError := false
+	for _, condition := range operation.Status.Conditions {
+		if condition.Type == v1alpha1.ConditionTypeReady {
```
KirilKabakchiev, line 328: its a common pattern to define some utils for working with conditions, otherwise its too verbose and hard to follow.. Getting a condition by type, Initializing conditions to default values, Setting a condition by type and so on.. 

In general a well written loop likes like a "sentence" - basically most if checks are outsourced to smaller methods so it can be a readable thing in the end, without caring too much about the details. Example https://github.com/kyma-project/kyma/blob/132d719c59388fc01af7e8b382f3e65dbf346dee/components/kyma-operator/pkg/k8s/controller.go. There are probably other good examples, havent looked too much yet, I can do if you need further details

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,428 @@ package controllers
...
+	operation, err := r.k8sClient.Get(ctx, req.NamespacedName)
+	if err != nil {
+		logger.Error(err, fmt.Sprintf("Unable to retrieve %s from API server", req.NamespacedName))
+		return ctrl.Result{}, nil
+	}
+
+	operation = setDefaultStatus(*operation)
+
+	webhookCount := len(operation.Spec.WebhookIDs)
+	if webhookCount == 0 || webhookCount > 1 {
```
KirilKabakchiev, line 84: I think that executing an operation with 0 hooks should also work end to end. Imagine that for app X there are delete hooks and we delete it async. We also delete app Y async, there are no webhooks for it, but its delete app api that is async in the end so it should still work

KirilKabakchiev (changes requested):

components/operations-controller/controllers/operation_controller.go
KirilKabakchiev: option 1 - we return the error and let controller manager do backoff requeues. We add some alerts about op being requeued more than X times. This is how we detect such bugs. Problem is that very often after we fix the bug, we wont be updating the op if the bug was related to how it was constructed or updated so the user will be patiently waiting for its reconcliation timeout.

option 2 - we just fail those directly, we still generate an alert - atleast this way the client won't be waiting for x hours for something that is doomed to fail :)

components/director/pkg/webhook/types.go
```diff
@@ -0,0 +1,259 @@
...
+
+	return headers, nil
+}
+
+func ParseOutputTemplate(inputTmpl, outputTmpl *string, respData ResponseData) (*Response, error) {
+	if outputTmpl == nil && inputTmpl != nil {
+		return nil, errors.New("missing webhook output template")
+	}
+
+	if outputTmpl == nil {
```
KirilKabakchiev, line 216: such behaviour should be avoided, this is hidden/unexpected and will often result in bugs - e.g. whoever calls Parseoutput template should rather get an error that he provided an invalid param and not call it rather than get a nil,nil

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,420 @@ package controllers
...
+
+		return ctrl.Result{RequeueAfter: r.webhookConfig.RequeueInterval}, nil
+	}
+
+	operation = setDefaultStatus(*operation)
+
+	if app.Result.Ready {
+		operation = setConditionStatus(*operation, v1alpha1.ConditionTypeReady, corev1.ConditionTrue, "")
+		operation.Status.Phase = v1alpha1.StateSuccess
+		if app.Result.Error != nil && *app.Result.Error != "" {
```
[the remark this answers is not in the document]
NickyMateev: `app.Result.Error` is a pointer type, which `len` doesn't accept, nevertheless a utility function has been added to improve the readability of such code segments

components/operations-controller/controllers/operation_controller.go
NickyMateev: I like Option 2 - which is related to what @pankrator  said in his second comment. Let's go with that one, as adequate alerts will take some time to prepare. With Option 2 the client will get feedback as soon as possible that something has gone wrong and not to wait for nothing.
Also, I think that we should do this same thing in the aforementioned scenarios as well, which are sort of internal server errors - request Data parsing fails, or webhooks > 1, etc.

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,420 @@ package controllers
...
+		return ctrl.Result{}, nil
+	}
+
+	requestData, err := parseRequestData(operation)
+	if err != nil {
+		logger.Error(err, operationError("Unable to parse request data", operation))
+		return ctrl.Result{}, nil
+	}
+
+	ctx = tenant.SaveToContext(ctx, requestData.TenantID)
```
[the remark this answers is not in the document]
NickyMateev: The tenant is dynamic and we have to plug it into the token transport decorator somehow. Using the context seems to be the most streamlined solution.
Otherwise we'd have to propagate the tenantID through numerous functions, even some functions which doesn't make much sense. Also the end goal is to generate a self-signed token with which to communicate with the Director, this sounds like it makes most sense to be near the transport where the header is attached.

https://github.com/kyma-incubator/compass/blob/8d38369c87bee7ad5c693b0e3310b87e0fba54d3/components/operations-controller/internal/director/token_provider.go#L71

Also propagating the tenant throughout the director client functions would look strange in terms of common interfaces as none of the gql client functions from the system broker do it like that.

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,420 @@ package controllers
...
+		case graphql.WebhookModeSync:
+			return r.updateStatus(ctx, logger, operation)
+		default:
+			return ctrl.Result{}, errors.New("unsupported webhook mode")
+		}
+	}
+
+	requeueAfter, err := getNextPollTime(&webhookEntity, *webhookStatus, r.webhookConfig.TimeLayout)
+	if err != nil {
+		logger.Error(err, operationError("Unable to calculate next poll time", operation))
```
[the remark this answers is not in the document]
NickyMateev: The only reason for the `getNextPollTime` to fail is if the `r.webhookConfig.TimeLayout` configuration is somehow invalid - this I think we can refer to as an internal server error.
Without manual intervention the operations controller would always break here. This is why I think the discussion from above applies here and I think we should actually update the operation status in k8s and update the operation in Director, so that the client can have immediate feedback and we can get notified immediately (as it would take some time for us to have adequate alerts). Here's the discussion from above, which comments on this exact situation: https://github.com/kyma-incubator/compass/pull/1733#discussion_r582827243

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,420 @@ package controllers
...
+			operation.Status.Phase = v1alpha1.StateFailed
+		}
+		if err := r.k8sClient.UpdateStatus(ctx, operation); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if app.Result.Error != nil && *app.Result.Error != "" {
+		operation = setErrorStatus(*operation, *app.Result.Error)
```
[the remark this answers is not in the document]
NickyMateev: - `setErrorStatus` ensures that only the Error condition is active (True) and flips off the Ready condition (False)
- `setReadyStatus` does the opposite of `setErrorStatus`
- `setConditionStatus` allows specifying the value of a particular condition without flipping the other one

components/operations-controller/internal/director/client.go
NickyMateev: We don't have access to the logger at this point, we'll have to plug it in through the DirectorClient interface, which I'm not sure we want to do here..

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,420 @@ package controllers
...
+		logger.Error(err, operationError("Unable to retrieve webhooks", operation))
+		return ctrl.Result{}, nil
+	}
+
+	if isReconciliationTimeoutReached(operation, calculateReconciliationTimeout(webhooks, r.webhookConfig.WebhookTimeout)) {
+		opErr := errors.New("reconciliation timeout reached")
+		return r.updateStatusWithError(ctx, logger, operation, opErr)
+	}
+
+	webhookEntity := webhooks[operation.Spec.WebhookIDs[0]]
```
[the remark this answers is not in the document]
NickyMateev: Initially the idea was to try to make the controller generic and lay the foundation for multiple webhook execution in the future. Eventually however webhook validation logic was added to ensure that only 1 webhook is expected to be executed and also the 0-indexing came along. This obviously causes conflict with the original premise to have the controller generic.
At this point in time I think it might be wiser to limit the controller to a single webhook and eliminate the confusion with multiple webhooks for the sake of more robust controller execution (ensuring always that the input operation will try to execute only a single webhook). Even though the Director component ensures that a single webhook will be placed in the `WebhookIDs` slice, I think it's important for the Operations Controller to also validate this as someone might have accidentally (or not) updated the Operation CR at some point and put the resource in an invalid state which might break the controller execution eventually.

components/operations-controller/controllers/operation_controller.go
NickyMateev: We also have this restriction on the Director side and it will not allow to schedule the operation and fail the mutation if there are no webhooks for an async mutation.
It sounds strange to me to want to delete an async application with no webhooks registered for such a request. If that's intentional, then the delete should be sync, and if it is not intentional and the webhook registration was actually forgotten, then that should be fixed by registering the necessary webhooks before attempting the async request.

components/director/pkg/webhook/types.go
```diff
@@ -0,0 +1,259 @@
...
+type Response struct {
+	Location          *string `json:"location"`
+	SuccessStatusCode *int    `json:"success_status_code"`
+	Error             *string `json:"error"`
+}
+
+// ResponseStatus defines the schema for Webhook status templates when dealing with async webhooks
+type ResponseStatus struct {
+	Status                     *string `json:"status"`
+	SuccessStatusCode          *int    `json:"success_status_code"`
```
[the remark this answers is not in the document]
NickyMateev: It allows us to catch if some of these fields have not been provided in the templates as they will be `nil` and we can return the client an error upon webhook registration.

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,420 @@ package controllers
...
+		return ctrl.Result{}, nil
+	}
+
+	ctx = tenant.SaveToContext(ctx, requestData.TenantID)
+	app, err := r.directorClient.FetchApplication(ctx, operation.Spec.ResourceID)
+	if err != nil {
+		logger.Error(err, operationError("Unable to fetch application", operation))
+
+		if isReconciliationTimeoutReached(operation, time.Duration(r.webhookConfig.TimeoutFactor)*r.webhookConfig.ReconciliationTimeout) {
+			if err := r.k8sClient.Delete(ctx, operation); err != nil {
```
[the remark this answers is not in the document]
NickyMateev: Firstly, the Director does not return an appropriate error message when app not found - it simply returns 200 OK and:

```
{
    "data": {
        "result": null
    }
}
```

Not sure it was intentionally left like so, because there were some old PRs which introduce more client-friendly errors in similar situations, but were not applied to the app not found case. Anyway, we'd have to do some changes in the Director to fix that, which is of course not something to sweat about.

 But it got me thinking in general, do we even need to interpret the error message? Why should we make the condition for deleting an Operation CR a combination of ROT expired + App Not Found? I think that the ROT expiration should dominate this decision and if it has passed, then there's nothing to try anymore with this operation and waste reconcile loop recourses.

Just image if there was some consistent non-app-not-found error which kept getting constantly returned - we'd never delete this Operation CR, because we'll never get the necessary App Not Found error.

What do you think?

components/operations-controller/internal/webhook/client.go
```diff
@@ -0,0 +1,211 @@
...
+	client := &d.HTTPClient
+	if webhook.Auth != nil {
+		basicCreds, isBasicAuth := webhook.Auth.Credential.(*graphql.BasicCredentialData)
+		if isBasicAuth {
+			req.SetBasicAuth(basicCreds.Username, basicCreds.Password)
+		}
+
+		oauthCreds, isOAuth := webhook.Auth.Credential.(*graphql.OAuthCredentialData)
+		if isOAuth {
+			client = d.OAuthClientProviderFunc(ctx, *client, oauthCreds)
```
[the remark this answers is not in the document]
NickyMateev: I agree, it was kinda of messy with the multiple tranports/clients. I decided to do some general refactoring of the `TokenProvider` abstraction in the System Broker and turned it into a more general `AuthorizationProvider` interface. This is now leveraged in the bootstrapping of the operations controller and only a single transport is initialized with a secured transport decorator initialized with `BasicAuthProvider`, `TokenProvider` and `UnsignedTokenProvider`.
I think it is much cleaner now.

components/operations-controller/main.go
```diff
@@ -52,33 +65,54 @@ func init() {
...
+	unsignedTokenProvider, err := director.NewUnsignedTokenProvider(cfg.GraphQLClient.GraphqlEndpoint)
+	if err != nil {
+		fatalOnError(err)
+	}
+
+	directorGraphQLClient, err := graphql.PrepareGqlClient(cfg.GraphQLClient, cfg.HttpClient, unsignedTokenProvider)
+	fatalOnError(err)
+
+	controller := controllers.NewOperationReconciler(client.New(mgr.GetClient()), cfg.Webhook,
+		director.NewClient(cfg.Director.InternalAddress, httpClient, directorGraphQLClient), &webhook.DefaultClient{HTTPClient: httpClient, OAuthClientProviderFunc: defaultOAuthClientProviderFunc},
```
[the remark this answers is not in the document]
NickyMateev: As mentioned in the previous comment - the transport abstraction has been reworked and also the correlation IDs are attached through the established correlation id transport decorator.

components/operations-controller/main.go
NickyMateev: Correlation IDs attaching has been added. I wasn't able to find anything around the `controller-runtime` regarding extracting additional tracing info. Maybe we can re-visit this in a separate tast.

components/operations-controller/controllers/operation_controller.go
NickyMateev: Thanks for the feedback - I introduced a `StatusManager` abstraction to hide all the details regarding maintaining an Operation CR's status. Also the `setDefaultStatus` method was removed.

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,428 @@ package controllers
...
+
+	if !hasError {
+		operation.Status.Conditions = append(operation.Status.Conditions, v1alpha1.Condition{
+			Type:   v1alpha1.ConditionTypeError,
+			Status: corev1.ConditionFalse,
+		})
+	}
+
+	if operation.Status.Webhooks == nil || overrideStatus {
+		operation.Status.Webhooks = make([]v1alpha1.Webhook, 0)
```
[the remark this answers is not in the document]
NickyMateev: This `setDefaultStatus` function has been removed in favour of the new `StatusManager` abstraction.

components/operations-controller/controllers/operation_controller.go
```diff
@@ -16,36 +16,428 @@ package controllers
...
+		return 0, err
+	}
+
+	nextPollTime := lastPollTimestamp.Add(time.Duration(*webhook.RetryInterval) * time.Second)
+	return nextPollTime.Sub(time.Now()), nil
+}
+
+func setDefaultStatus(operation v1alpha1.Operation) *v1alpha1.Operation {
+	overrideStatus := false
+	if operation.Generation != operation.Status.ObservedGeneration {
```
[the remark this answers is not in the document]
NickyMateev: This `setDefaultStatus` function has been removed in favour of the new `StatusManager` abstraction.

components/operations-controller/controllers/operation_controller.go
NickyMateev: Also in general I tried streamlining the main loop logic to leverage the different abstractions that we have so that the whole loop reads easier. The main loop function has reduced more than twice in terms of code lines as a result, but of course there are limits to how much we can shorten it as the reconcile loop has some corner case branches which need addressing and which cannot be covered cleanly by abstractions (ex. invalid requestObject, deriving the correct webhook entity from the retrieved application, calculating next poll time.. to name a few).

components/operations-controller/controllers/operation_controller_test.go
```diff
@@ -0,0 +1,1776 @@
...
+	}
+)
+
+func TestReconcile_FailureToGetOperationCR_ShouldResultNoRequeueNoError(t *testing.T) {
+	logger := &mockedLogger{}
+
+	k8sClient := &clientfakes.FakeClient{}
+	k8sClient.GetReturns(nil, mockedErr)
+
+	controller := NewOperationReconciler(k8sClient, nil, nil, nil, logger)
```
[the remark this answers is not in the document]
NickyMateev: I added a separate suite of semi-integration tests as you proposed while also keeping these tests here which took some time and effort to write and I think both test suites would be valuable.

components/operations-controller/main.go
```diff
@@ -132,16 +123,27 @@ func fatalOnError(err error) {
 	}
 }
 
-func defaultOAuthClientProviderFunc(ctx context.Context, client http.Client, oauthCreds *director_graphql.OAuthCredentialData) *http.Client {
-	conf := &clientcredentials.Config{
-		ClientID:     oauthCreds.ClientID,
-		ClientSecret: oauthCreds.ClientSecret,
-		TokenURL:     oauthCreds.URL,
-		AuthStyle:    oauth2.AuthStyleInParams,
+func prepareHttpClient(cfg *httputil.Config, graphqlCfg *graphql.Config) (*http.Client, error) {
```
KirilKabakchiev, line 129: passing the graphql target url here is a bit strange, also the unsigned provider uses this method, others return nil, so this probably means that the abstraction of target url does not fit well here. Basically now the secure transport has to understand that not all providers provide a target url and needs to provide its own value then.

Ideally the directorURL passed to the director client should be reused and targeturl() removed from interface - you should pass it to PrepareGqlClient and provide it in 	graphClient := graphql.NewClient("", graphql.WithHTTPClient(httpClient)) where currently an empty string is passed as url

components/operations-controller/main.go
```diff
@@ -95,23 +93,16 @@ func main() {
...
 
-	directorGraphQLClient, err := graphql.PrepareGqlClient(cfg.GraphQLClient, cfg.HttpClient, unsignedTokenProvider)
+	graphqlClient, err := graphql.PrepareGqlClientWithHttpClient(cfg.GraphQLClient, httpClient)
 	fatalOnError(err)
 
 	controller := controllers.NewOperationReconciler(cfg.Webhook, ctrl.Log.WithName("controllers").WithName("Operation"),
 		k8s.NewClient(mgr.GetClient()),
-		director.NewClient(cfg.Director.InternalAddress, httpClient, directorGraphQLClient),
-		webhook.NewClient(httpClient, defaultOAuthClientProviderFunc))
+		director.NewClient(cfg.Director.InternalAddress, httpClient, graphqlClient),
```
KirilKabakchiev, line 107: you could create the graphql client within the director client constructor maybe ? Pass the graphql client config as part of the director client config

components/operations-controller/controllers/operation_controller_integration_test.go
```diff
@@ -0,0 +1,692 @@
...
+ in-memory Kubernetes API Server is deployed in order to test out some of the critical flows
+ of the Operation Controller reconciliation loop.
+
+ Unlike the tests in operation_controller_test.go the tests here do not aim to be as exhaustive as possible
+ but rather strive to cover main flows and scenarios. Also - similarly to the operation_controller_test.go
+ tests, here the Director and Webhook clients have been mocked.
+
+*/
+
+func TestController_Scenarios(t *testing.T) {
```
KirilKabakchiev, line 81: maybe add a test that verifies that overidding of operation for a resource works properly (e.g. a create op was just executed, now executing a delete - status is then currently reflecting the delete op, webhooks of delete were called, etc)

KirilKabakchiev, line 81: you could verify that deleting an op does not trigger the controller similarly to how you did for status update (separate test or one of the existing tests?)

components/operations-controller/internal/server/config.go
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package server
+
+import (
+	"github.com/kyma-incubator/compass/components/system-broker/pkg/server"
+)
+
+// Settings type to be loaded from the environment
+type Config struct {
```
krasish, line 24: Validate method is inherited from *server.Config and will get called as part of the Validation of config.Config. Thus MetricAddress and EnableLeaderElection fields are not validated though. If we want to validate those fileds a local Validate method should be introduced.

components/operations-controller/Makefile
```diff
@@ -1,4 +1,4 @@
-APP_NAME = compass-operations-controller-manager
+APP_NAME = compass-operations-controller
 APP_PATH = components/operations-controller
 BUILDPACK = eu.gcr.io/kyma-project/test-infra/buildpack-golang-kubebuilder2:v20210216-1.14.0-rc1-535-gb45951c9
```
krasish: There was some vulnerability in versions of go <=1.15.6 if I recall correctly. This buildpack has [a new version](https://github.com/kyma-project/test-infra/pull/3341) from 4 days ago. As far as I can see it only bumps version of go. Consider using that one instead.

components/operations-controller/internal/k8s/status/manager.go
```diff
@@ -0,0 +1,147 @@
...
+// are scheduled on the same Operation CR and the old status should be wiped off.
+func (m *manager) Initialize(ctx context.Context, name types.NamespacedName) error {
+	return m.updateStatusFunc(ctx, name, func(operation v1alpha1.Operation, status *v1alpha1.OperationStatus) {
+		if status.ObservedGeneration != nil && operation.ObjectMeta.Generation == *status.ObservedGeneration {
+			return
+		}
+
+		status.ObservedGeneration = &operation.Generation
+
+		status.Phase = v1alpha1.StateInProgress
```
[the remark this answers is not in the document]
NickyMateev: I actually liked the clarity of seeing the status being initialized verbosely and directly in front of you. The properties are not that many and I don't think it's an issue in this case.
By the way, an initial version of the "status manager" (there wasn't such an abstraction yet) tried to optimize code duplication with utility functions, and while it was shorter it wasn't as clear as it it is now.

components/operations-controller/internal/config/config.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/kyma-incubator/compass/components/system-broker/pkg/env"
+	"github.com/kyma-incubator/compass/components/system-broker/pkg/graphql"
+	"github.com/kyma-incubator/compass/components/system-broker/pkg/http"
+	"github.com/pkg/errors"
+	"github.com/spf13/pflag"
+)
+
+// Validatable defines entities capable of being validated
+type Validatable interface {
```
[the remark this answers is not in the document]
NickyMateev: `Validatable` is a pretty common interface and so you might see that there are other dependencies/libs which we use, which provide the same interface. While we could technically import their own version of this interface I think it's cleaner to redefine the interface on our own instead of importing and coupling our package with some other arbitrary one which happened to declare the same interface (ex.`import inputvalidation.Validatable`).
Also generally in Go I think it's accepted to redefine small constructs/types/interfaces instead of always living with the idea of reusing everything as much as possible. It makes the codebase simpler and easier to understand.

components/operations-controller/internal/server/config.go
NickyMateev: `MetricAddress` and `EnableLeaderElection` are not mandatory properties and are pretty wide open in terms of possible values they could have assigned. If we have a need for explicit validation of them and have a clear understanding what an invalid value for each of them means, then yes, we could introduce a `Validate` method for this struct, have it call the embedded `*server.Config`'s Validate and then do explicit validation of the two properties.

components/operations-controller/internal/webhook/client.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	responseObject, err := parseResponseObject(resp)
+	if err != nil {
+		return nil, err
+	}
+
+	log.C(ctx).Info(fmt.Sprintf("Webhook response object: %v", *responseObject))
+
+	response, err := responseObject.ParseOutputTemplate(webhook.OutputTemplate)
```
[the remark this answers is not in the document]
NickyMateev: We cannot enforce clients to always register these templates as that would be backward incompatible, so theoretically, you're right, it is possible for it it be nil here. Since we cannot have validation upon webhook registration, we'll add additional checks here.

KirilKabakchiev (approved):

PR #1738, opened 2021-02-18
Unify System Broker and ORD specification APIs

krasish:
**Description**

- Expose 'id' fields of Specs in ApiDefinitions and EventDefinitions
- ORD Service can now be called using a certificate
- SystemBroker returns URLs to ORD service
- Drop SystemBroker's specs API 

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] E2E tests


components/system-broker/pkg/ord/ord_test.go
```diff
@@ -0,0 +1,51 @@
+package ord_test
+
```
PetarTodorovv, line 2: Maybe _config_test.go_ is better name for this file?

PetarTodorovv (approved):

components/system-broker/pkg/ord/ord_test.go
krasish: Addressed it! 🚀

nkanchev (approved):

tests/system-broker/tests/auth_test.go
```diff
@@ -95,6 +115,66 @@ func TestSystemBrokerAuthentication(t *testing.T) {
 	})
 }
 
+func TestCallingORDServiceWithCert(t *testing.T) {
+	ORDContext, err := pkg.NewTestContext(config)
```
NickyMateev, line 119: Instead of `ORDContext` this can simply be `testContext`, or `testCtx`, as it is not entirely ORD specific.

tests/system-broker/tests/auth_test.go
```diff
@@ -95,6 +115,66 @@ func TestSystemBrokerAuthentication(t *testing.T) {
...
+
+	logrus.Infof("registering runtime with name: %s, within tenant: %s", runtimeInput.Name, ORDContext.Tenant)
+	runtime := director.RegisterRuntimeFromInputWithinTenant(t, ORDContext.Context, ORDContext.DexGraphqlClient, ORDContext.Tenant, runtimeInput)
+	defer director.UnregisterRuntimeWithinTenant(t, ORDContext.Context, ORDContext.DexGraphqlClient, ORDContext.Tenant, runtime.ID)
+
+	app, err := director.RegisterApplicationWithinTenant(t, ORDContext.Context, ORDContext.DexGraphqlClient, ORDContext.Tenant, applicationInput)
+	require.NoError(t, err)
+	defer director.UnregisterApplication(t, ORDContext.Context, ORDContext.DexGraphqlClient, ORDContext.Tenant, app.ID)
+
+	securedClient, configuration, certChain := getSecuredClientByContext(t, testCtx, runtime.ID)
```
NickyMateev, line 130: Probably reorder and move this secured client construction line below the register bundle + api (but before the test) so that the bundle + api are naturally after the application registration.

tests/system-broker/tests/auth_test.go
```diff
@@ -95,6 +115,66 @@ func TestSystemBrokerAuthentication(t *testing.T) {
...
+
+		resp, err := securedClient.Do(req)
+
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+		require.Equal(t, http.StatusBadRequest, resp.StatusCode) //oathkeeper cannot be configured to return 401 the way we use it
+
+		body, err := ioutil.ReadAll(resp.Body)
+		require.NoError(t, err)
+		require.Contains(t, string(body), "invalid tenantID")
```
NickyMateev, line 173: invalid tenantID?

tests/system-broker/tests/auth_test.go
```diff
@@ -95,6 +115,66 @@ func TestSystemBrokerAuthentication(t *testing.T) {
...
+
+		reqORD := createORDServiceRequest(t, ORDContext, api.ID, api.Spec.ID)
+		respORD, err := securedClient.Do(reqORD)
+
+		require.NoError(t, err)
+		require.NotNil(t, respORD)
+		require.Equal(t, http.StatusOK, respORD.StatusCode)
+	})
+
+	t.Run("Should fail calling ORD service with revoked cert", func(t *testing.T) {
```
NickyMateev, line 155: Is it possible that this test runs before the previous one (the happy path)? If it is possible then this would break the happy path test and we should probably create a dedicated secured client for each test and defer the revocation.

tests/system-broker/tests/auth_test.go
krasish: When the certificate is not resolved successfully (bc it had been revoked) the call to ORD Service will contain an (invalid) token which does not have the tenantId claim so this message is [returned](https://github.com/kyma-incubator/ord-service/blob/722f1717632f358a052cd4da0c759804d35af190/components/ord-service/src/main/java/com/sap/cloud/cmp/ord/service/controller/SpecificationsController.java#L28).

tests/system-broker/tests/auth_test.go
krasish: It looked to me that subtests are run sequentially because they are simply method invocations (and will get executed in the same order in which they appear in code). [This answer](https://stackoverflow.com/a/39734200/9986012) seems to support that. Also, I have run all tests multiple times now and had never noticed a different execution order.

NickyMateev (approved):

nkanchev (approved):

PR #1743, opened 2021-02-24
Ord Aggregator implementation

nyordanoff:
**Description**
This PR introduces the `ORD Aggregator` component. It will aggregate the information served by all `ORD Providers` according to the ORD aggregation rules and re-expose it through the `ORD service`.

**Changes proposed in this pull request:**
- implementation of the aggregator
- unit tests

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/open_resource_discovery/service.go
```diff
@@ -0,0 +1,529 @@
...
+	}); found {
+		packageID = &packagesFromDB[i].ID
+	}
+
+	specs := make([]*model.SpecInput, 0, len(api.ResourceDefinitions))
+	for _, resourceDef := range api.ResourceDefinitions {
+		specs = append(specs, resourceDef.ToSpec())
+	}
+
+	if !found {
```
pankrator, line 415: Perhaps rename the variable `found` to something else like `isAPIFound` because currently it is a bit puzzling what `found` is referring to as it is referring to found on line 391

pankrator (changes requested):
Review is done only on the Go style and without the tests. I will do another one after I understand better the spec.

components/director/cmd/ordaggregator/main.go
```diff
@@ -0,0 +1,154 @@
...
+	productSvc := product.NewService(productRepo)
+	vendorSvc := ordvendor.NewService(vendorRepo)
+	tombstoneSvc := tombstone.NewService(tombstoneRepo)
+
+	ordClient := open_resource_discovery.NewClient(httpClient)
+
+	return open_resource_discovery.NewAggregatorService(transact, appSvc, webhookSvc, bundleSvc, apiSvc, eventAPISvc, specSvc, packageSvc, productSvc, vendorSvc, tombstoneSvc, ordClient)
+}
+
+func createAndRunConfigProvider(ctx context.Context, cfg config) *configprovider.Provider {
```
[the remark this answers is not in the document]
nyordanoff: It is indeed the same, but if exported in the director, the function can not be imported in the aggregator as the package in the director is `main`. So, a new package should be created where the function is moved and then imported to the two places but is it worth the effort for a single function?

components/director/internal/open_resource_discovery/client.go
```diff
@@ -0,0 +1,100 @@
...
+			return nil, errors.Wrapf(err, "error fetching ORD document from: %s", url+config.URL)
+		}
+		doc.SystemInstanceAware = config.SystemInstanceAware
+		docs = append(docs, doc)
+	}
+
+	return docs, nil
+}
+
+func (c *client) fetchOpenDiscoveryDocumentWithAccessStrategy(ctx context.Context, documentURL string, accessStrategy AccessStrategyType) (*Document, error) {
```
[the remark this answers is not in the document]
nyordanoff: Currently we only support `open` accessStrategy which means no auth. The whole `accessStrategy` flow will be implemented in the future so we decided to leave it there as a reminder.

There's no issue removing it but it will be once again added when the flow is being implemented. Do you think it's worth it?

components/director/internal/domain/api/service.go
```diff
@@ -100,21 +110,28 @@ func (s *service) GetForBundle(ctx context.Context, id string, bundleID string)
 	return apiDefinition, nil
 }
 
-func (s *service) CreateInBundle(ctx context.Context, bundleID string, in model.APIDefinitionInput, spec *model.SpecInput) (string, error) {
+func (s *service) CreateInBundle(ctx context.Context, appId, bundleID string, in model.APIDefinitionInput, spec *model.SpecInput) (string, error) {
+	return s.Create(ctx, appId, &bundleID, nil, in, []*model.SpecInput{spec})
```
pankrator, line 114: In which cases can the packageID be empty?

DimitarPetrov: When the API is created through the GraphQL api

components/director/internal/model/api.go
```diff
@@ -43,32 +48,87 @@ func (_ *APIDefinition) GetType() resource.Type {
...
-	LineOfBusiness      json.RawMessage
-	Industry            json.RawMessage
-	Version             *VersionInput
+	OrdBundleID         *string         `json:"partOfConsumptionBundle"`
+	OrdPackageID        *string         `json:"partOfPackage"`
+	Tenant              string          `json:",omitempty"`
+	Name                string          `json:"title"`
+	Description         *string         `json:"description"`
+	TargetURL           string          `json:"entryPoint"`
+	Group               *string         `json:",omitempty"` //  group allows you to find the same API but in different version
```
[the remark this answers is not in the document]
DimitarPetrov: These properties are relevant for the GraphQL API only and since there is a GraphQL layer that has relevant JSON tags these JSON tags on model entities are only relevant for the ORD scenario.

components/director/internal/open_resource_discovery/client.go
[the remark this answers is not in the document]
pankrator: It can stay here as a reminder

components/director/cmd/ordaggregator/main.go
```diff
@@ -0,0 +1,154 @@
...
+	provider := configprovider.NewProvider(cfg.ConfigurationFile)
+	err := provider.Load()
+	exitOnError(err, "Error on loading configuration file")
+	executor.NewPeriodic(cfg.ConfigurationFileReload, func(ctx context.Context) {
+		if err := provider.Load(); err != nil {
+			exitOnError(err, "Error from Reloader watch")
```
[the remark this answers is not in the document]
nyordanoff: resolved offline

components/director/cmd/ordaggregator/main.go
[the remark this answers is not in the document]
pankrator: sure, it's ok this way

components/director/internal/domain/bundle/service_test.go
```diff
@@ -851,3 +851,81 @@ func TestService_ListByApplicationID(t *testing.T) {
...
+		fixBundleModel(name, desc),
+		fixBundleModel(name, desc),
+	}
+
+	ctx := context.TODO()
+	ctx = tenant.SaveToContext(ctx, tenantID, externalTenantID)
+
+	testCases := []struct {
+		Name               string
+		PageSize           int
```
pankrator, line 873: Probably `PageSize` is a left over as I don't see it used

components/director/internal/domain/eventdef/service_test.go
```diff
@@ -282,35 +284,126 @@ func TestService_ListForBundle(t *testing.T) {
...
+		fixEventDefinitionModel(id, bndlID, name),
+		fixEventDefinitionModel(id, bndlID, name),
+	}
+
+	ctx := context.TODO()
+	ctx = tenant.SaveToContext(ctx, tenantID, externalTenantID)
+
+	testCases := []struct {
+		Name               string
+		PageSize           int
```
pankrator, line 306: Same here for the `PageSize`

components/director/internal/open_resource_discovery/service_test.go
```diff
@@ -0,0 +1,1026 @@
...
+			clientFn: func() *automock.Client {
+				client := &automock.Client{}
+				doc := fixORDDocument()
+				doc.Tombstones[0].OrdID = vendorORDID
+				client.On("FetchOpenResourceDiscoveryDocuments", txtest.CtxWithDBMatcher(), baseURL).Return(open_resource_discovery.Documents{doc}, nil)
+				return client
+			},
+		},
+		{
+			Name:            "Does not resync resources if product resource deletion due to tombstone fails",
```
pankrator, line 891: The naming of the test cases is a bit confusing. If I understand it correctly it should be something like: 
`Does not resync resources if product resource deletion fails due to tombstone`

nyordanoff: The context is that the product resource should be deleted because there is a tombstone resource for it (that's the `due to tombstone` part of the name). And the test case is that the product deletion fails - everything with the tombstone is ok.

You can think of it as something along the lines of: `Does not resync resources if there's a tombstone for a product resource and the product deletion fails`

components/director/internal/domain/bundle/resolver.go
```diff
@@ -235,6 +237,16 @@ func (r *Resolver) DeleteBundle(ctx context.Context, id string) (*graphql.Bundle
 		return nil, err
 	}
 
+	err = r.apiSvc.DeleteAllByBundleID(ctx, id)
```
pankrator, line 240: Just an idea - shouldn't this be part of the bundle service and not the resolver? My idea behind this is that the business logic (service) should contain the deletion of relative resources.
Not sure if this was already discussed.

pankrator: As discussed this is accepted by design

pankrator (approved):

PR #1741, opened 2021-02-22
Move OTT to Director

ataleksandrov:
**Description**

Changes proposed in this pull request:
- Introduce internal graphQL API to issue tokens for CSR in Director
- Move OTT validation hydrator to Director
- Save OTT in `system_auths`

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests
- [x] `chart/compass/values.yaml` is updated

PR #1744, opened 2021-03-02
Fix operations policy

desislavaa:
**Problem Description**
The current `NetworkPolicy` configuration denies access to all ports available on director, besides the internal one. The internal one is accessible only from the controller (which is the only intended behaviour in the mentioned case).

**Changes proposed in this pull request**
Use AuthorizationPolicy instead of NetworkPolicy for the internal communication between the operations controller and director.
Pros:
- Easier to implement:
   > Another difference worth mentioning is that NetworkPolicies work in an additive, whitelist model. When a NetworkPolicy selects a specific pod, that pod will reject any connections, except those that are explicitly allowed.
- We'll have some consistency with https://github.com/kyma-incubator/compass/pull/1741
- Since it's managed by Istio, and we have Istio in the minimal installation used by the e2e tests, it will be easier to catch issues caused by the policy. The `NetworkPolicy` is managed by Calico, which is not part of the minimal installation.

References regarding AuthorizationPolicy vs NetworkPolicy:
- https://istio.io/latest/blog/2017/0.1-using-network-policy/ (quite old)
- https://banzaicloud.com/blog/istio-authorization-policies/


chart/compass/values.yaml
```diff
@@ -216,6 +217,9 @@ global:
         - "x-vcap-request-id"
         - "x-broker-api-request-identity"
 
+  ingress:
```
desislavaa, line 220: fixes `helm template` execution

ataleksandrov (dismissed):

PetarTodorovv (dismissed):

ataleksandrov (approved):

PR #1746, opened 2021-03-02
Fix minikube IP for docker driver

DimitarPetrov:
**Description**

`minikube ip` returns 127.0.0.1 in the case of docker driver (by design) which is not valid inside the cluster as it represents localhost for the container. Therefore test-definitions fail on minikube with docker driver.


KirilKabakchiev (approved):
the deploy-and-test.sh script is used by the prow pr job that starts the tests with minikube (https://github.com/kyma-project/test-infra/blob/14fab4c9de15bd8a3a3de76e384700e8b1af7354/prow/scripts/provision-vm-compass.sh#L130). You could add the docker driver arg in the script (so that we also have the tests on pr running with minikube with docker driver)

PR #1748, opened 2021-03-04
Async deletion test

dzahariev:
**Description**

- e2e test for Async operation

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


dzahariev:
Validated

nkanchev (dismissed):

components/external-services-mock/pkg/webhook/webhook.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+const (
+	OperationPath                     = "webhook/delete/operation"
+	DeletePath                        = "webhook/delete"
+	OperationResponseStatusOK         = "SUCCEEDED"
+	OperationResponseStatusINProgress = "IN_PROGRESS"
+)
+
+var isOk bool
```
NickyMateev, line 23: `isOk` is too general for me, I can't get any semantics for its meaning until I've looked around in the handlers. Maybe rename it to something like `deletionInvoked` or `hasDeletionStarted`, or something which has more meaning.

components/external-services-mock/pkg/webhook/webhook.go
```diff
@@ -0,0 +1,83 @@
+package webhook
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+type OKRequestData struct {
```
NickyMateev, line 8: Move the struct definitions below the `const` and `var` definitions

components/external-services-mock/pkg/webhook/webhook.go
```diff
@@ -0,0 +1,83 @@
...
+			rw.WriteHeader(http.StatusOK)
+		} else {
+			rw.WriteHeader(http.StatusLocked)
+		}
+	}
+}
+
+func NewWebHookOperationHTTPHandler() func(rw http.ResponseWriter, r *http.Request) {
+	return func(rw http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodGet {
```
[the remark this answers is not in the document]
NickyMateev: Also, if we're always going to just flip the switch to `isOK: false` then it probably doesn't make sense to have the dedicated `OKRequestData` struct which parses some payload in order to determine what value to set to the boolean flag.

components/external-services-mock/pkg/webhook/webhook_test.go
```diff
@@ -0,0 +1,87 @@
...
+		handler := http.HandlerFunc(webhook.NewWebHookOperationHTTPHandler())
+		handler.ServeHTTP(rr, req)
+		require.Equal(t, http.StatusOK, rr.Code)
+		require.Contains(t, rr.Body.String(), webhook.OperationResponseStatusINProgress)
+
+		// Verify is locked
+		req, err = http.NewRequest(http.MethodGet, "/webhook/delete", nil)
+		require.NoError(t, err)
+		rr = httptest.NewRecorder()
+		handler = http.HandlerFunc(webhook.NewDeleteHTTPHandler())
```
NickyMateev, line 51: Unnecessary cast to `http.HandlerFunc`

components/external-services-mock/pkg/webhook/webhook_test.go
```diff
@@ -0,0 +1,87 @@
...
+		handler := http.HandlerFunc(webhook.NewWebHookOperationHTTPHandler())
+		handler.ServeHTTP(rr, req)
+		require.Equal(t, http.StatusOK, rr.Code)
+		require.Contains(t, rr.Body.String(), webhook.OperationResponseStatusOK)
+
+		// Verify is unlocked
+		req, err = http.NewRequest(http.MethodGet, "/webhook/delete", nil)
+		require.NoError(t, err)
+		rr = httptest.NewRecorder()
+		handler = http.HandlerFunc(webhook.NewDeleteHTTPHandler())
```
NickyMateev, line 83: Also unnecessary cast

tests/director/external-services-mock-integration/async_api_delete_test.go
```diff
@@ -0,0 +1,171 @@
+package external_services_mock_integration
```
NickyMateev, line 1: Not sure these tests should be in the `tests/director/external-service-mock-intergration` directory. We created the [`tests/operations-controller`](https://github.com/kyma-incubator/compass/tree/master/tests/operations-controller) directory for precisely these tests.
From one point of view we do test scenarios relating to external systems in general in the `tests/director/external-service-mock-intergration` directory (ex. auditlog & fetch requests) and in this sense these tests do make sense to be here.
But from another point of view these tests cover much bigger scenarios than the auditlog & fetch requests scenarios. Also eventually we might want to add more operations-controller e2e tests which would potentially make the `tests/director/external-service-mock-intergration` directory very operations-controller -centric instead of having more general purpose tests in it which don't have a better place to be in. For these exact reasons I think we wanted to have the `tests/operations-controller` directory.

Maybe we can discuss it in the next daily standup? 🙂

tests/director/external-services-mock-integration/async_api_delete_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+func TestAsyncAPIDeleteApplication(t *testing.T) {
+	var (
+		operationFullPath = fmt.Sprintf("%s%s", testConfig.ExternalServicesMockBaseURL, webhook.OperationPath)
+		deleteFullPath    = fmt.Sprintf("%s%s", testConfig.ExternalServicesMockBaseURL, webhook.DeletePath)
+	)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	appName := fmt.Sprintf("app-async-del-%s", time.Now().Format("060102150405"))
```
NickyMateev, line 38: You can use ``time.Now().Format(time.RFC3339Nano)` here instead of hardcoding.
Example: https://github.com/kyma-incubator/compass/blob/252d77cd7a5028dc39e6d6b8f782b7e58ad07bc6/components/operations-controller/internal/k8s/status/manager_test.go#L252

tests/director/external-services-mock-integration/async_api_delete_test.go
```diff
@@ -0,0 +1,171 @@
...
+	operationName := fmt.Sprintf("application-%s", app.ID)
+	t.Log(fmt.Sprintf("Check operation CR with name %s is created", operationName))
+	operation, err := operationsK8sClient.Get(ctx, operationName, metav1.GetOptions{})
+	require.NoError(t, err)
+	require.NotEmpty(t, operation)
+
+	t.Log(fmt.Sprintf("Verify operation CR with name %s is in progress", operationName))
+	counter := 0
+	isInDesiredState, webhookResponse = isWebhookOperationInDesiredState(t, operationFullPath, webhook.OperationResponseStatusINProgress)
+	for (!isInDesiredState) && counter < 100 {
```
NickyMateev, line 92: Instead of maintaining a counter and looping with `for` you can leverage `require.Eventually`, something we did in SM as well. Here's an example:

```
require.Eventually(t, func() bool {
		// logic here -> function will run every 5 seconds for 2minutes until the function returns `true`
	}, 2*time.Minute, 5*time.Second)
```

https://github.com/kyma-incubator/compass/blob/59756f9b7ca08a29d21f3de68ff2c9584f3284ea/components/operations-controller/controllers/operation_controller_integration_test.go#L157

tests/director/external-services-mock-integration/async_api_delete_test.go
```diff
@@ -0,0 +1,171 @@
...
+		time.Sleep(time.Second * 5)
+		isInDesiredState, webhookResponse = isWebhookOperationInDesiredState(t, operationFullPath, webhook.OperationResponseStatusINProgress)
+		counter++
+	}
+	isInDesiredState, webhookResponse = isWebhookOperationInDesiredState(t, operationFullPath, webhook.OperationResponseStatusINProgress)
+	require.True(t, isInDesiredState, fmt.Sprintf("Webhook responded with: %s. Expected state: %s", webhookResponse, webhook.OperationResponseStatusINProgress))
+
+	t.Log("Verify the application status in director is 'ready:false'")
+	application := getApplication(t, ctx, dexGraphQLClient, testConfig.DefaultTenant, app.ID)
+	require.False(t, application.Ready, "Application is not in status ready:false")
```
NickyMateev, line 103: Tweak the fixtures and also assert the `application.Error` field that it is empty. Additionally assert the timestamps, most notably the deletion timestamp that it's not null (and it's around the current time? because you cannot assert an exact time)

NickyMateev (changes requested):

chart/compass/templates/gateway-external-services-integration.yaml
```diff
@@ -85,4 +85,40 @@ spec:
...
+kind: ServiceAccount
+metadata:
+  name: {{ .Chart.Name }}-operations-tests-service-account
+  namespace: kyma-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Chart.Name }}-operations-tests-operations-role
+  namespace: compass-system
```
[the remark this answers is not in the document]
dzahariev: Yes. The reason is that the test is running in kyma-system namespace, and the operations are created in compass-systm namespace. So we need to have this binding to allow test to read operations.

installation/scripts/testing-suite.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
dzahariev: Sure, we can remove it.

tests/director/run-compass-gateway-external-services-integration.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
dzahariev: Will be removed too

tests/director/external-services-mock-integration/async_api_delete_test.go
dzahariev: time.RFC3339Nano  contains special symbols, spaces and is too long for this field.

tests/director/external-services-mock-integration/async_api_delete_test.go
dzahariev: Sure, let discuss it.

components/external-services-mock/pkg/webhook/webhook.go
```diff
@@ -0,0 +1,83 @@
...
+
+var isOk bool
+
+func NewDeleteHTTPHandler() func(rw http.ResponseWriter, r *http.Request) {
+	return func(rw http.ResponseWriter, r *http.Request) {
+		if isOk {
+			isOk = false
+			rw.WriteHeader(http.StatusOK)
+		} else {
+			rw.WriteHeader(http.StatusLocked)
```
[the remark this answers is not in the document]
dzahariev: For this test the "lock" represent the real state. As the operation is now locked for deletion.

components/external-services-mock/pkg/webhook/webhook.go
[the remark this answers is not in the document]
dzahariev: We make refactoring and now - GET is used to read state, POST to change it and DELETE to initiate deletion of system.

nkanchev (approved):

NickyMateev (approved):

PR #1761, opened 2021-03-10
Add webhooks to application templates 

krasish:
**Description**

- Extend GraphQL API for Application Templates to be able to receive webhooks on top-level
- Add app_template_id FK to  application table & webhook table
- addWebhook to take applicationTemplateId. Check that only one is provided (applicationId or applicationTemplateId).
- Limit the webhooks of the same type on the resource to 1 - SQL CONSTRAINTS
- Fetching an application with webhooks should return the webhooks calculated by priority and united with the ones for its template (QUERY)
- Implement webhook priority where the same webhook type on application is with high priority than the same webhook on application template (DIRECTIVE)
- SQL to populate `app_template_id` in the applications table
- E2E tests

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/apptemplate/converter.go
```diff
@@ -214,3 +227,19 @@ func (c *converter) placeholdersToGraphql(in []model.ApplicationTemplatePlacehol
...
 }
+
+func (c *converter) webhooksToGraphql(in []model.Webhook) ([]graphql.Webhook, error) {
+	webhookPtrs := make([]*model.Webhook, len(in))
+	for i := 0; i < len(in); i++ {
+		webhookPtrs[i] = &in[i]
+	}
+	gqlWebhookPtrs, err := c.webhookConverter.MultipleToGraphQL(webhookPtrs)
+	if err != nil {
+		return nil, fmt.Errorf("while converting webhooks to graphql: %w", err)
```
dzahariev, line 238: Why do not use errors.Wrapf here too?

components/director/internal/domain/apptemplate/repository_test.go
```diff
@@ -20,7 +20,8 @@ import (
 func TestRepository_Create(t *testing.T) {
 	t.Run("Success", func(t *testing.T) {
 		// GIVEN
-		appTemplateModel := fixModelAppTemplate(testID, testName)
+		webhooksModel := fixModelApplicationWebhooks(testWebhookID, testID)
```
dzahariev, line 23: We should come with a convention "App" or "Application" in fixture function names.
The difference I have in mind is "fixModelAppTemplate" and "fixModelApplicationWebhooks".

dzahariev (changes requested):

components/director/internal/domain/application/service.go
```diff
@@ -450,6 +402,82 @@ func (s *service) createRelatedResources(ctx context.Context, in model.Applicati
...
 }
+
+func (s *service) genericCreate(ctx context.Context, in model.ApplicationRegisterInput, repoCreatorFunc repoCreatorFunc) (string, error) {
+	appTenant, err := tenant.LoadFromContext(ctx)
+	if err != nil {
+		return "", err
+	}
+	log.C(ctx).Debugf("Loaded Application Tenant %s from context", appTenant)
+
+	applications, err := s.appRepo.ListAll(ctx, appTenant)
```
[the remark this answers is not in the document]
NickyMateev: The normalization logic used to compare the new name with any existing that follows is not trivial to be moved to an sql-style solution. Also this logic will be here only for a few more sprints as the Kyma colleagues will start taking care of normalization and then we won't need to have this here.

components/schema-migrator/migrations/director/20210302174853_add_app_template_to_webhooks.down.sql
```diff
@@ -0,0 +1,21 @@
...
+    ALTER COLUMN tenant_id SET NOT NULL,
+    DROP CONSTRAINT IF EXISTS webhooks_app_template_id_fkey,
+    DROP COLUMN IF EXISTS app_template_id;
+
+UPDATE applications
+SET app_template_id = NULL
+WHERE id = id;
+
+ALTER TABLE applications
+    DROP CONSTRAINT IF EXISTS applications_app_template_id_fkey,
```
dzahariev, line 18: `applications_app_template_id_fkey` is not defined at UP script, and I cannot find it in other SQLs too. Can you please check this.
Probably you should name the constraint in UP script:
```
ADD CONSTRAINT applications_app_template_id_fkey 
   FOREIGN KEY (app_template_id) 
   REFERENCES app_templates(id) 
   ON DELETE SET NULL;
```

components/director/internal/domain/apptemplate/converter.go
```diff
@@ -3,6 +3,7 @@ package apptemplate
 import (
 	"database/sql"
 	"encoding/json"
+	"fmt"
```
dzahariev, line 6: Not used import - please remove.

components/schema-migrator/migrations/director/20210302174853_add_app_template_to_webhooks.down.sql
NickyMateev: Good catch - `applications_app_template_id_fkey` was the default name generated for the FK which is added in the UP migration. But it will be more correct to add the name explicitly like you proposed so that we can be sure that the constraint will be named this way rather than expect Postgres to always name it like that.

dzahariev (approved):

PR #1762, opened 2021-03-10
Address operation controller review feedback

PetarTodorovv:
**Description**

- The _**ready**_ field is no longer return in the GraphQL requests
- Rename the operation API and use path params instead of query params
- Introduce a new _**creation_time**_ field that reflects the timestamp of the latest operation(CREATE, UPDATE, DELETE)
- Add new application statuses and maintain them

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/cmd/director/main.go
```diff
@@ -103,7 +103,7 @@ type config struct {
 	TenantMappingEndpoint         string `envconfig:"default=/tenant-mapping"`
 	RuntimeMappingEndpoint        string `envconfig:"default=/runtime-mapping"`
 	AuthenticationMappingEndpoint string `envconfig:"default=/authn-mapping"`
-	OperationEndpoint             string `envconfig:"default=/operations"`
+	OperationEndpoint             string `envconfig:"default=/last_operation"`
```
NickyMateev, line 106: The PUT API URL Path of the Operations API should not be changed - it should remain `/operation`.
Introduce `LastOperationEndpoint string` for the last operation GET API

components/director/pkg/graphql/schema.graphql
```diff
@@ -658,7 +666,6 @@ type Application {
 	bundle(id: ID!): Bundle
 	auths: [SystemAuth!]
 	eventingConfiguration: ApplicationEventingConfiguration
-	ready: Boolean!
```
NickyMateev, line 661: I see that our `pkg/graphql` entities still have a BaseEntity which contains a `ready` field? Not sure we want to keep it there, but maybe it's fine to stay as long as the end result from the GraphQL query does not contain the field. Have you tried it?

components/director/pkg/graphql/schema.graphql
```diff
@@ -51,6 +51,15 @@ enum ApplicationStatusCondition {
 	INITIAL
 	CONNECTED
 	FAILED
+	CREATING
```
NickyMateev, line 54: You need to add these types to the model + db layers as well. This includes migrations to the enum field.
Also I noticed that after you do that, you need to adapt the converters in places like this one: https://github.com/kyma-incubator/compass/blob/16d72a407f83a06550044ca11ad07fd6050cfe25/components/director/internal/domain/application/converter.go#L211

components/director/pkg/operation/api.go
```diff
@@ -128,21 +132,39 @@ func prepareLastOperation(resource model.Entity) *OperationResponse {
 		Error: resource.GetError(),
 	}
 
+	prepareOperationType(resource, opResponse)
```
NickyMateev, line 135: Maybe rename the `prepareLastOperation` function into `buildLastOperation`. Also maybe move the `prepareOperationType`, `prepareOperationStatus` and `prepareCreationTime` functions as methods to the `OperationResponse` and rename them to something like `initializeOperationType`, `initializeOperationStatus`, `initializeCreationTime` or if you have a better word than **prepare** or **initialize**.

components/director/pkg/operation/directive.go
```diff
@@ -292,3 +299,16 @@ func executeSyncOperation(ctx context.Context, next gqlgen.Resolver, tx persiste
...
+func determineApplicationInProgressStatus(opType graphql.OperationType) model.ApplicationStatusCondition {
+	switch opType {
+	case graphql.OperationTypeCreate:
+		return model.ApplicationStatusConditionCreating
+	case graphql.OperationTypeUpdate:
+		return model.ApplicationStatusConditionUpdating
+	case graphql.OperationTypeDelete:
+		return model.ApplicationStatusConditionDeleting
+	default:
+		return model.ApplicationStatusConditionInitial
```
NickyMateev, line 312: Setting `Initial` as will lead to confusion and inconsitent state. We should rather return an error of some kind if opType is not among the supported ones.

components/director/pkg/operation/directive.go
```diff
@@ -119,6 +121,11 @@ func (d *directive) HandleOperation(ctx context.Context, _ interface{}, next gql
 		return nil, apperrors.NewInternalError("Failed to process operation")
 	}
 
+	if err := d.resourceUpdaterFunc(ctx, entity.GetID(), false, nil, determineApplicationInProgressStatus(operationType)); err != nil {
+		log.C(ctx).WithError(err).Errorf("While updating resource %s with id %s", entity.GetType(), entity.GetID())
```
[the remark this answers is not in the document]
NickyMateev: Also the test for this case is missing

components/director/pkg/operation/update_operation_handler.go
```diff
@@ -182,3 +186,33 @@ func stringifiedJsonError(errorMsg string) (*string, error) {
...
+		appConditionStatus = model.ApplicationStatusConditionDeleteSucceeded
+		if opError != nil || ptrStrToStr(opError) != "" {
+			appConditionStatus = model.ApplicationStatusConditionDeleteFailed
+		}
+	}
+
+	return appConditionStatus
+}
+
+func ptrStrToStr(s *string) string {
```
NickyMateev, line 213: Move this function variation to this file: https://github.com/kyma-incubator/compass/blob/6e6973361cc9940351bbe3f7ba99ec9869e778e3/components/director/pkg/str/strings.go#L33

NickyMateev (changes requested):
Tweak the Operations Controller E2E test to assert that the Application status = `DELETING`

components/operations-controller/internal/director/client.go
```diff
@@ -64,7 +64,7 @@ func (c *client) UpdateOperation(ctx context.Context, request *Request) error {
 		return err
 	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPut, c.directorURL+"/operations", bytes.NewBuffer(body))
+	req, err := http.NewRequestWithContext(ctx, http.MethodPut, c.directorURL+"/last_operation", bytes.NewBuffer(body))
```
alextargov, line 67: Probably we could use an env variable here?

components/director/pkg/operation/update_operation_handler.go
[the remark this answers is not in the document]
PetarTodorovv: Maybe in the system-broker component, but in the director, there is no such function

components/director/go.mod
```diff
@@ -6,59 +6,41 @@ require (
 	github.com/99designs/gqlgen v0.9.3
 	github.com/DATA-DOG/go-sqlmock v1.5.0
 	github.com/Masterminds/goutils v1.1.1 // indirect
-	github.com/Masterminds/semver v1.5.0 // indirect
```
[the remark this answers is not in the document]
PetarTodorovv: I think it's not from tidy, but when synchronized my dependencies. I'll double-check the go.mod and go.sum files

components/director/pkg/graphql/schema.graphql
PetarTodorovv: Yes, I've tested it and when the query request has the ready field it fails, regardless a base entity has a ready field

NickyMateev (dismissed):

NickyMateev (dismissed):

NickyMateev (approved):

NickyMateev (dismissed):

NickyMateev (dismissed):

NickyMateev (approved):

PR #1768, opened 2021-03-16
Strip sensitive data

desislavaa:
**Description**

- Add a new directive that can remove sensitive fields
- Introduce new scopes for each field
- Add the needed scopes for each consumer 
- Add a new job that synchronizes the scopes in the current Compass configuration and in Hydra

**How to test the scopes synchronizer locally**
1. Start compass
2. Create a new integration system and request oauth credentials for it
3. Exec into compass-director and update the client created on step 2
```
curl http://ory-hydra-admin.kyma-system.svc.cluster.local:4445/clients/<client_id> -X PUT -d '{"scope": "application:read"}'
```
4. "Restart" the `compass-director-clients-scopes-synchronization` job (get it, delete it, and create it again)
5. Exec into compass-director again, execute `curl http://ory-hydra-admin.kyma-system.svc.cluster.local:4445/clients` and validate that the scopes are back

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


chart/compass/charts/director/config.yaml
```diff
@@ -101,4 +130,6 @@ clientCredentialsRegistrationScopes:
     - "label_definition:read"
     - "label_definition:write"
     - "automatic_scenario_assignment:read"
-    - "automatic_scenario_assignment:write"
\ No newline at end of file
+    - "automatic_scenario_assignment:write"
+    - "integration_system.auths:read"
```
KirilKabakchiev, line 134: one of the recommendations in the Treat modeling workshop is that we don't return credentials that were provided to us to the entity that provided them. So maybe we could remove line 134.

tests/Makefile
```diff
@@ -21,7 +21,7 @@ e2e-test:
 	@$(INSTALLATION_SCRIPTS_DIR)/testing.sh
 
 e2e-test-clean:
-	@kubectl delete clustertestsuites.testing.kyma-project.io compass-e2e-tests || true
+	@kubectl delete clustertestsuites.testing.kyma-project.io compass-tests || true
```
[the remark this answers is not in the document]
desislavaa: Correct, this is a leftover

tests/director/tests/fetch_request_test.go
```diff
@@ -15,7 +15,11 @@ import (
 	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
 )
 
-const OpenAPISpec = "https://raw.githubusercontent.com/kyma-incubator/github-slack-connectors/beb8e5b6d8f3a644b8380e667a9376bc353e54dd/github-connector/internal/registration/configs/githubopenAPI.json"
+const (
+	OpenAPISpec   = "https://raw.githubusercontent.com/kyma-incubator/github-slack-connectors/beb8e5b6d8f3a644b8380e667a9376bc353e54dd/github-connector/internal/registration/configs/githubopenAPI.json"
+	AsyncAPISpec  = "https://raw.githubusercontent.com/kyma-incubator/github-slack-connectors/2fcc4728f63cdf23656417d9f85679a02367df97/slack-connector/internal/registration/configs/slackasyncapi.json"
+	MDDocumentURL = "https://raw.githubusercontent.com/kyma-incubator/compass/v1.4.2/README.md"
+)
```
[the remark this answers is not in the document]
desislavaa: Indeed, but the `OpenAPISpec` is used in both files, and it seemed reasonable to add the other specs there as well - they're still used in the context of a fetch request

StanislavStefanov (approved):

components/director/pkg/sanitize/directive.go
```diff
@@ -0,0 +1,45 @@
...
+	scopesGetter ScopesGetter
+}
+
+func NewDirective(getter ScopesGetter) *directive {
+	return &directive{
+		scopesGetter: getter,
+	}
+}
+
+func (d *directive) Sanitize(ctx context.Context, obj interface{}, next graphql.Resolver, scopesDefinition string) (interface{}, error) {
```
KirilKabakchiev, line 28: just an idea to remove some duplication 

Currently hasScopes and sanitize directives are exactly the same apart from what they return on scope missmatch (error forbidden vs nil error). You can have a single directive struct that accepts scopesGetter and "scopesMissmatchErrorProvider" or something with better name.  Then you can then reuse the same struct for the directive in the code for both instantiations passing different arguments.

components/director/pkg/sanitize/directive.go
```diff
@@ -0,0 +1,45 @@
+package sanitize
```
KirilKabakchiev, line 1: this is actually hasScopes, just with different "error outcome" - in this case nil. Maybe move all these things in has scopes (this comment builds ontop of the one about using the same directive struct). Probably in regards to the tests you would have to add one or two extra tests to the existing scope directive tests and avoid some copying there as well

tests/director/tests/sensitive_data_strip_test.go
```diff
@@ -0,0 +1,257 @@
...
+				require.Equal(t, application.Bundles.Data[0].APIDefinitions.Data[0].Spec.FetchRequest != nil, test.fieldExpectations.apiSpecFetchRequest)
+				require.Equal(t, application.Bundles.Data[0].EventDefinitions.Data[0].Spec.FetchRequest != nil, test.fieldExpectations.eventSpecFetchRequest)
+				require.Equal(t, application.Bundles.Data[0].Documents.Data[0].FetchRequest != nil, test.fieldExpectations.documentFetchRequest)
+				require.Equal(t, application.Bundles.Data[0].InstanceAuths != nil, test.fieldExpectations.bundleInstanceAuths)
+			})
+		}
+	})
+
+	t.Run("Application template access", func(t *testing.T) {
+		t.Run("from runtime", func(t *testing.T) {
```
KirilKabakchiev, line 178: what about accessing app templates from int systems ?

tests/director/tests/sensitive_data_strip_test.go
```diff
@@ -0,0 +1,257 @@
...
+			apiSpecFetchRequest       bool
+		}
+		type testCase struct {
+			name              string
+			consumer          *gcli.Client
+			fieldExpectations accessRequired
+		}
+		testCases := []testCase{
+			{
+				name:     "Runtime consumer",
```
KirilKabakchiev, line 121: we also have int system consumer

chart/compass/charts/director/templates/oauth-clients-scopes-sync-job.yaml
```diff
@@ -0,0 +1,96 @@
+apiVersion: batch/v1
+kind: Job
```
KirilKabakchiev, line 2: This job will probably rarely do any actual syncs (probably most of the times it would be just running without doing much). One idea for optimization would be that we annotate the job with the hash of the configmap its going to run with. And the next time we are about to run it we can check if the configmap hash has chaged and if not we skip running it.


Edit: I looked for some resources how this can be achieved, something similar for deployments is discussed in https://github.com/helm/helm/issues/2639 

One idea can be whats done above with the deployment in the above link. 

Another idea/variation of the one above could be that we look for the configmap, calculate the hash, look for the job and get the hashed annotation. Then we guard the job.yaml with an if template to only be created if the hash of the configmap is different from the hash of the job. lookups in charts we already do here https://github.com/kyma-incubator/compass/pull/1552/files

tests/director/tests/sensitive_data_strip_test.go
desislavaa: line 130 :) no fields should be accessible

components/director/internal/domain/oauth20/service.go
```diff
@@ -72,6 +90,22 @@ func (s *service) CreateClientCredentials(ctx context.Context, objectType model.
 	return credentialData, nil
 }
 
+func (s *service) UpdateClientScopes(ctx context.Context, clientID string, objectType model.SystemAuthReferenceObjectType) error {
+	scopes, err := s.GetClientCredentialScopes(objectType)
+	if err != nil {
+		if !model.IsIntegrationSystemNoTenantFlow(err, objectType) {
```
ataleksandrov, line 96: There are few `GetClientCredentialScopes` calls and the returned error is handled differently. Here and in `CreateClientCredentials` we are looking for `IsIntegrationSystemNoTenantFlow`.(in sanitize directive we do not check for this kind of error) I could not get why this check is needed at all.

components/director/internal/domain/oauth20/service.go
```diff
@@ -195,6 +248,54 @@ func (s *service) doRequest(ctx context.Context, method string, endpoint string,
 	return resp, closeBodyFn, nil
 }
 
+func (s *service) clientsFromHydra(ctx context.Context, endpoint string) ([]Client, error) {
```
ataleksandrov, line 251: Could we use hydra sdk [client](https://github.com/ory/hydra/blob/fe8fdc5f6d1ca1ebe16a5241cd06a32d249a4362/docs/versioned_docs/version-v1.8/sdk/go.md), instead of custom code?

chart/compass/charts/director/config.yaml
```diff
@@ -81,15 +81,44 @@ graphql:
...
     deleteAutomaticScenarioAssignmentsForSelector: ["automatic_scenario_assignment:write"]
 
+  field:
+    application:
+      auths: ["application.auths:read"]
+      webhooks: ["application.webhooks:read"]
+    application_template:
+      webhooks: ["application_template.webhooks:read"]
+    bundle:
+      instance_auth: ["bundle.instance_auths:read"]
```
ataleksandrov, line 91: Should we add this scope to the integration systems as for now they have "runtime:write" which is needed to request bundleInstanceAuth creation?

components/director/hack/plugins/descriptionsdecorator/plugin.go
```diff
@@ -11,7 +11,6 @@ import (
 	"github.com/99designs/gqlgen/codegen/config"
 	"github.com/99designs/gqlgen/plugin"
 	"github.com/kyma-incubator/compass/components/director/hack/plugins"
-	"github.com/kyma-incubator/compass/components/director/pkg/log"
```
ataleksandrov, line 14: Is this intentional?

components/director/internal/scopes_sync/service.go
```diff
@@ -0,0 +1,123 @@
...
+	UpdateClientScopes(ctx context.Context, clientID string, objectType model.SystemAuthReferenceObjectType) error
+}
+
+type service struct {
+	oAuth20Svc OAuthService
+	transact   persistence.Transactioner
+	repo       SystemAuthRepo
+}
+
+func NewService(oAuth20Svc OAuthService, transact persistence.Transactioner, repo SystemAuthRepo) *service {
```
ataleksandrov, line 40: return `SyncService` interface XOR delete the interface

components/director/hack/plugins/descriptionsdecorator/plugin.go
StefanCholakov: Log package was used only to print a debug statement, for which now we use "fmt". There was no actual benefit from it.

chart/compass/charts/director/config.yaml
desislavaa: Currently, there are no use cases where integration systems create instance auths, so we can skip them in the response here and forbid the creation in another PR.

chart/compass/charts/director/templates/oauth-clients-scopes-sync-job.yaml
desislavaa: We went with the second approach, thanks for the suggestion! The first one was not sufficient, because of the helm hooks we use in order to run after the migration job

tests/director/tests/sensitive_data_strip_test.go
desislavaa: the client was the int systems one, just forgot to change the message in th test case

KirilKabakchiev (approved):

nkanchev (approved):

PR #1769, opened 2021-03-17
Forbid system deletion when in formation

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/application/service.go
```diff
@@ -301,6 +303,21 @@ func (s *service) Delete(ctx context.Context, id string) error {
 		return errors.Wrapf(err, "while loading tenant from context")
 	}
 
+	scenarios, runtimes, err := s.isInScenarioAndRuntime(ctx, appTenant, id)
```
danielgospodinow, line 306: Is it necessary to check for the runtimes?  I personally think that we can skip this because the runtimes in a given formation can be easily seen from the UI.

components/director/internal/domain/application/service.go
```diff
@@ -301,6 +303,21 @@ func (s *service) Delete(ctx context.Context, id string) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "while checking if application is in formation and runtime")
+	}
+
+	if len(scenarios) > 0 {
+		msg := fmt.Sprintf("System deletion failed: the system is part of a formation - %s", strings.Join(scenarios, ", "))
+
+		if len(runtimes) > 0 {
```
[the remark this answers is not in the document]
alextargov: The jira ticket requires the runtimes in the message. :(

components/director/internal/domain/application/service.go
alextargov: Same as https://github.com/kyma-incubator/compass/pull/1769#discussion_r599410457

components/director/internal/domain/application/service.go
```diff
@@ -30,7 +31,7 @@ const (
 
 type repoCreatorFunc func(ctx context.Context, application *model.Application) error
 
-//go:generate mockery -name=ApplicationRepository -output=automock -outpkg=automock -case=underscore
+//go:generate mockery --name=ApplicationRepository --output=automock --outpkg=automock --case=underscore
```
pankrator, line 34: Either change all generated mocks to use the newer version of mockery or adjust your version to the one used before. Currently is it impossible to compile all mocks with one go command

tests/director/api/labeling_api_test.go
```diff
@@ -595,11 +602,13 @@ func TestSearchApplicationsByLabels(t *testing.T) {
 
 	firstApp := registerApplication(t, ctx, "first")
 	require.NotEmpty(t, firstApp.ID)
+	//defer updateApplicationScenariosToDefaultState(t, ctx, firstApp.ID)
```
pankrator, line 605: Delete this if not needed. Check for other commented lines like this.

tests/pkg/fixtures/application_queries.go
```diff
@@ -146,3 +146,32 @@ func GenerateOneTimeTokenForApplication(t *testing.T, ctx context.Context, gqlCl
...
+
+	scenarios := []string{defaultValue}
+	var labelValue interface{} = scenarios
+
+	t.Log("Updating Application scenario to a default state")
+	setApplicationLabel(t, ctx, gqlClient, applicationID, tenantID, labelKey, labelValue)
+}
+
+func setApplicationLabel(t *testing.T, ctx context.Context, gqlClient *gcli.Client, applicationID string, tenantID string, labelKey string, labelValue interface{}) graphql.Label {
+	request := FixGetApplicationsRequestWithPagination()
```
ataleksandrov, line 162: Not sure, what is the purpose of this request?

components/director/internal/domain/application/service.go
```diff
@@ -514,3 +531,69 @@ func (s *service) ensureIntSysExists(ctx context.Context, id *string) (bool, err
...
+
+	var runtimesNames []string
+	for _, r := range runtimes {
+		runtimesNames = append(runtimesNames, r.Name)
+	}
+
+	return validScenarios, runtimesNames, nil
+}
+
+func removeDefaultScenario(scenarios []string) []string {
```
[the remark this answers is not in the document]
alextargov: Right. So after a communication with Dimitar Patanayotov, we decided that applications within `DEFAULT` formations should be able to be delated. The Cockpit does not introduce the concept of a `DEFAULT` formation and that's why we want to be able to delete such applications.

components/director/internal/domain/application/service.go
```diff
@@ -514,3 +533,55 @@ func (s *service) ensureIntSysExists(ctx context.Context, id *string) (bool, err
 	log.C(ctx).Infof("Integration System with id %s exists", *id)
 	return true, nil
 }
+
+func (s *service) getScenariosAndRuntimes(ctx context.Context, tenant, applicationID string) ([]string, []string, error) {
+	log.C(ctx).Infof("Getting formations for application with id %s", applicationID)
```
KirilKabakchiev, line 538: formation is not a word that is used in the code (i just saw a couple of tests that use it) - every where else its called scenario. So for consistency use scenarios

components/director/internal/domain/application/service.go
```diff
@@ -301,6 +305,21 @@ func (s *service) Delete(ctx context.Context, id string) error {
 		return errors.Wrapf(err, "while loading tenant from context")
 	}
 
+	scenarios, runtimes, err := s.getScenariosAndRuntimes(ctx, appTenant, id)
```
KirilKabakchiev, line 308: usually methods that contain  "and" are handling multiple responsibilities which is not advicable from clean code pov.  It's also making the code confusing. For example, right now just reading through this method im left under the impression that you are first fetching both scenarios and runtimes and if there are no scenarios we return. And this is not the case - i have to go into the getScenariosAndRuntimes method and see that even though we are saying we fetch both, we only fetch rts if scenarios exist. 

So split those into two methods

components/director/internal/domain/application/service.go
```diff
@@ -301,6 +305,21 @@ func (s *service) Delete(ctx context.Context, id string) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "while checking if application is in formation and runtime")
```
KirilKabakchiev, line 310: "is in **scenario** **with** runtime"

components/director/internal/domain/application/service.go
```diff
@@ -301,6 +305,21 @@ func (s *service) Delete(ctx context.Context, id string) error {
...
+	}
+
+	if len(scenarios) > 0 {
+		msg := fmt.Sprintf("System deletion failed: the system is part of a formation - %s", strings.Join(scenarios, ", "))
+
+		if len(runtimes) > 0 {
+			msg = fmt.Sprintf("%s and of runtime - %s", msg, strings.Join(runtimes, ", "))
+		}
+
+		return apperrors.NewInvalidOperationError(msg)
```
KirilKabakchiev, line 320: i dont think you should be returning an error if the scenario contains no runtimes. With the current impl of scenarios - no runtimes = no one is using the apis of the app

components/director/internal/domain/application/service.go
```diff
@@ -301,6 +305,21 @@ func (s *service) Delete(ctx context.Context, id string) error {
 		return errors.Wrapf(err, "while loading tenant from context")
 	}
 
+	scenarios, runtimes, err := s.getScenariosAndRuntimes(ctx, appTenant, id)
+	if err != nil {
+		return errors.Wrapf(err, "while checking if application is in formation and runtime")
+	}
+
+	if len(scenarios) > 0 {
```
KirilKabakchiev, line 313: this is abit messy in terms of error handling but when you split the methods it should become more straight forward out of the box

components/director/internal/domain/application/service.go
```diff
@@ -514,3 +533,55 @@ func (s *service) ensureIntSysExists(ctx context.Context, id *string) (bool, err
...
+	validScenarios := removeDefaultScenario(scenarios)
+
+	if len(validScenarios) == 0 {
+		return nil, nil, nil
+	}
+
+	scenariosQuery := eventing.BuildQueryForScenarios(validScenarios)
+	runtimeScenariosFilter := []*labelfilter.LabelFilter{labelfilter.NewForKeyWithQuery(model.ScenariosKey, scenariosQuery)}
+
+	log.C(ctx).Infof("Listing runtimes matching the query %s", scenariosQuery)
```
KirilKabakchiev, line 559: this should be probably on debug

tests/director/tests/application_api_test.go
```diff
@@ -492,34 +494,133 @@ func TestCreateApplicationWithDuplicatedNamesWithinTenant(t *testing.T) {
...
+
+		err = testctx.Tc.RunOperationWithCustomTenant(ctx, dexGraphQLClient, tenantID, req, nil)
+
+		//THEN
+		require.EqualError(t, err, expectedErrorMsg)
+	})
+
+	t.Run("Error when application is in formation and runtime", func(t *testing.T) {
+		//GIVEN
+		expectedErrorMsg := "graphql: The operation is not allowed [reason=System deletion failed: the system is part of a formation - test-scenario and of runtime - one-runtime]"
```
KirilKabakchiev, line 576: use the word scenario instead (rename everywhere)

tests/director/tests/application_api_test.go
```diff
@@ -492,34 +494,133 @@ func TestCreateApplicationWithDuplicatedNamesWithinTenant(t *testing.T) {
...
+		(*runtimeInput.Labels)[ScenariosLabel] = scenarios
+		runtimeInputWithNormalizationGQL, err := testctx.Tc.Graphqlizer.RuntimeInputToGQL(runtimeInput)
+		require.NoError(t, err)
+		registerRuntimeRequest := fixtures.FixRegisterRuntimeRequest(runtimeInputWithNormalizationGQL)
+
+		runtime := graphql.Runtime{}
+		err = testctx.Tc.RunOperationWithCustomTenant(ctx, dexGraphQLClient, tenantID, registerRuntimeRequest, &runtime)
+		require.NoError(t, err)
+		require.NotEmpty(t, runtime.ID)
+		defer fixtures.UnregisterRuntime(t, ctx, dexGraphQLClient, tenantID, runtime.ID)
```
KirilKabakchiev, line 592: usually you defer first, then do whatever you would be doing. This makes sure defer would be called even if what you had to do failed half way

tests/director/tests/application_api_test.go
```diff
@@ -492,34 +494,133 @@ func TestCreateApplicationWithDuplicatedNamesWithinTenant(t *testing.T) {
...
+		require.NoError(t, err)
+
+		createApplicationReq := fixtures.FixRegisterApplicationRequest(appInputGQL)
+		application := graphql.Application{}
+
+		err = testctx.Tc.RunOperationWithCustomTenant(ctx, dexGraphQLClient, tenantID, createApplicationReq, &application)
+		require.NoError(t, err)
+		require.NotEmpty(t, application.ID)
+
+		defer func() {
```
KirilKabakchiev, line 606: same as above

tests/director/tests/application_api_test.go
```diff
@@ -492,34 +494,133 @@ func TestCreateApplicationWithDuplicatedNamesWithinTenant(t *testing.T) {
...
+			fixtures.UpdateApplicationScenariosToDefaultState(t, ctx, dexGraphQLClient, tenantID, application.ID)
```
KirilKabakchiev, line 561: could this be just UnassignApplicationFromScenarios?

tests/pkg/fixtures/application_queries.go
```diff
@@ -146,3 +146,24 @@ func GenerateOneTimeTokenForApplication(t *testing.T, ctx context.Context, gqlCl
...
+	defaultValue := "DEFAULT"
+
+	scenarios := []string{defaultValue}
+	var labelValue interface{} = scenarios
+
+	t.Log("Updating Application scenario to a default state")
+	setApplicationLabel(t, ctx, gqlClient, applicationID, tenantID, labelKey, labelValue)
+}
+
+func setApplicationLabel(t *testing.T, ctx context.Context, gqlClient *gcli.Client, applicationID string, tenantID string, labelKey string, labelValue interface{}) graphql.Label {
```
KirilKabakchiev, line 161: this is used only the method above, prolly no need go be a separate method

components/director/internal/domain/application/service.go
alextargov: This is true, but it is contraversive with what the ticket says:

> Deleting a system which is assigned to a formation must fail

tests/director/tests/application_api_test.go
alextargov: I think the error message shall stay `formation` as it is by requirements. However, I'll change the method, variable, etc namings to `scenario`

dpanayotov: Can remain scenario. Extracting scenario label as formation entity will include adjusting all errors as well. This error will not be propagated to users anyway atm.

components/director/internal/domain/application/service.go
dpanayotov: UX would be better if when there is no runtime in the scenario, deleting the system should not be stopped.

PR #1792, opened 2021-03-26
Aggregator chart and e2e test

nyordanoff:
**Description**

This PR introduces a chart and e2e test for the `ord-aggregator` component. The setup of the test is as follows: 
- it loops for a specified time that is equal to the aggregator schedule plus a small buffer;
- an application with webhook of type `OPEN_RESOURCE_DISCOVERY` is registered so that when the regular `ord-aggregator` job passes, the aggregator itself will be able to aggregate info from the document that this app serves;
- all ord specific responses are pre-defined and handled by the `external-services-mock`;
- during this whole loop, the `ord-service` regularly checks if the `ord-aggregator` successfully populated the new data and if so checks its validity;
- if the aggregator did not finish for the specified time (see point 1), the e2e test timeouts and fails

- newly introduced chart for the `ord-aggregator` of type `CronJob`
- e2e test for the `ord-aggregator` component

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [x] Integration tests


tests/ord-aggregator/Makefile
```diff
@@ -0,0 +1,34 @@
...
+	@$(INSTALLATION_SCRIPTS_DIR)/testing.sh ord-aggregator
+
+e2e-test-clean:
+	@kubectl delete clustertestsuites.testing.kyma-project.io compass-e2e-tests || true
+
+sandbox-test:
+	@../sandbox.sh ord-aggregator
+
+run:
+	@kubectl exec -n kyma-system oct-tp-compass-e2e-tests-compass-e2e-ord-aggregator-local-0 -c ord-aggregator-tests -- ./ord-aggregator.test test -run=$(testName)
```
StanislavStefanov, line 24: replace with @kubectl exec -n kyma-system oct-tp-compass-e2e-tests-compass-e2e-ord-aggregator-local-0 -c ord-aggregator-tests -- ./ord-aggregator.test -test.run $(testName)

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -0,0 +1,254 @@
...
+			respBody = makeRequestWithHeaders(t, httpClient, testConfig.ORDServiceURL+"/systemInstances?$format=json", map[string][]string{tenantHeader: {testConfig.DefaultTenant}})
+			if len(gjson.Get(respBody, "value").Array()) == 0 {
+				t.Log("Missing System Instances...will try again")
+				return false
+			}
+
+			require.Equal(t, expectedNumberOfSystemInstances, len(gjson.Get(respBody, "value").Array()))
+			require.Equal(t, applicationName, gjson.Get(respBody, "value.0.title").String())
+			require.Equal(t, applicationDescription, gjson.Get(respBody, "value.0.description").String())
+
```
StanislavStefanov, line 123: can we extract these lines into function and place it into pkg/assertions

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -0,0 +1,254 @@
...
+		Webhooks: []*directorSchema.WebhookInput{
+			{
+				Type: directorSchema.WebhookTypeOpenResourceDiscovery,
+				URL:  &testConfig.ExternalServicesMockBaseURL,
+			},
+		},
+	}
+}
+
+func makeRequestWithHeaders(t *testing.T, httpClient *http.Client, url string, headers map[string][]string) string {
```
StanislavStefanov, line 252: maybe move to pkg/requests

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+		select {
+		case <-done:
+			return errors.New("timeout waiting for entities to be present in DB")
+		case <-ticker:
+		}
+	}
+}
+
+func createApp() directorSchema.ApplicationRegisterInput {
```
StanislavStefanov, line 239: maybe move to pkg/fixtures

tests/ord-service/tests/api_test.go
```diff
@@ -539,65 +539,27 @@ func TestORDService(t *testing.T) {
...
 
 	t.Run("Errors generate user-friendly message", func(t *testing.T) {
-		respBody := makeRequestWithHeadersAndStatusExpect(t, httpClient, testConfig.ORDServiceURL+"/test?$format=json", map[string][]string{tenantHeader: {testConfig.DefaultTenant}}, http.StatusNotFound)
+		respBody := request.MakeRequestWithHeadersAndStatusExpect(t, httpClient, testConfig.ORDServiceURL+"/test?$format=json", map[string][]string{tenantHeader: {testConfig.DefaultTenant}}, http.StatusNotFound, testConfig.ORDServiceDefaultResponseType)
 
 		require.Contains(t, gjson.Get(respBody, "error.message").String(), "Use odata-debug query parameter with value one of the following formats: json,html,download for more information")
 	})
 }
 
 func makeRequest(t *testing.T, httpClient *http.Client, url string) string {
```
StanislavStefanov, line 553: can we move the makeRequest* functions to pkg/requests

nyordanoff: I was considering it but these `makeRequest*` functions are just wrappers for the extracted one. If we do extract them all, we will have to pass an additional argument `ORDServiceDefaultResponseType string` to them which is ord-specific and somehow nullifies the wrapping.

Moreover, we will have to add this specific `ord` argument to each function execution which, again, kind of breaks the wrapping logic. I decided that it will be cleaner to leave these functions and pass only once the `ord` argument.

tests/ord-aggregator/tests/handler_test.go
nyordanoff: Same as below

tests/ord-aggregator/tests/handler_test.go
nyordanoff: Makes sense, I guess it got a bit cleaner with this approach.

StanislavStefanov (approved):

tests/ord-aggregator/tests/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return 0, errors.New("error while parsing cron time")
+	}
+
+	// This is the starting time that will be subtracted from the next activation cron time below. This way the cron time duration can be estimated.
+	year, month, day := time.Now().Date()
+	startingTime := time.Date(year, month, day, 0, 0, 0, 0, time.Now().Location())
+
+	nextTime := scheduleTime.Next(startingTime)
+	cronTimeDuration := nextTime.Sub(startingTime)
```
DimitarPetrov, line 63: can't this be simplified like this:

```go
now := time.Now()
next := scheduleTime.Next(now)
next.Sub(now)
```

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -0,0 +1,213 @@
...
+func TestORDAggregator(t *testing.T) {
+	appInput := fixtures.FixSampleApplicationRegisterInputWithORDWebhooks(expectedSystemInstanceName, expectedSystemInstanceDescription, testConfig.ExternalServicesMockBaseURL)
+
+	eventsMap := make(map[string]string, 0)
+	eventsMap[firstEventTitle] = firstEventDescription
+	eventsMap[secondEventTitle] = secondEventDescription
+
+	ctx := context.Background()
+
+	dexToken, err := idtokenprovider.GetDexToken()
```
DimitarPetrov, line 68: I think that @StanislavStefanov  moved dex token generation to a central place as a test optimization maybe you can look into it.

tests/ord-aggregator/tests/main_test.go
nyordanoff: That was my initial approach but it is incorrect based on how `scheduleTime.Next()` works. In the current setup, the cron time that we pass has value of 5mins (execute every 5mins). By design, that means execution in: xx:05, xx:10, xx:15, xx:20 etc.

`scheduleTime.Next(<time>)` gives you the next scheduled execution after the given `<time>`. So, if we pass `time.Now()`  and then subtract it, the result will be how much time is left to the next scheduled execution (e.g: 
time.Now() is 10:12, `next := scheduleTime.Next(now); next.Sub(now)` will result in 3 mins: "10:15 - 10:12"). So, it is impossible to calculate how much is the cron time.

With the implemented approach, we set a “fixed” initial time (00:00), so that when we subtract it from the next scheduled cron time (00:05 in our case), the result will be the actual duration of the cron time. You can think of it as subtraction of zero.

StanislavStefanov (approved):

StanislavStefanov (approved):

PR #1798, opened 2021-03-28
Add correlation id to timeout handler

karaimin:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/pkg/handler/timeout_handler.go
```diff
@@ -47,15 +46,18 @@ type timeoutLoggingHandler struct {
 }
 
 func (h *timeoutLoggingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := log.ContextWithLogger(r.Context(), log.LoggerWithCorrelationId(r))
```
DimitarPetrov, line 49: In `log.LoggerWithCorrelationId(r)` you use `correlation.CorrelationIDForRequest(r)` which generates a new correlation_id if one is not found in headers/context.

Currently, if this handler is before the request middleware and there is no correlation_id in the headers you generate a new one and add it to the logger and the logger to the context.

However, you do not add this new context to the request. This way in the middleware a new correlation_id will be generated and used for the other logs and the whole flow won't be possible to follow.

karaimin: Nice catch! However, I think this is working correctly despite the missing part you mentioned. I checked the logs in both timeout_handler and some component handlers (in connectivity-adapter) and the logged **x-request_id** 

I debugged it and found [this side effect](https://github.com/kyma-incubator/compass/blob/master/components/director/pkg/correlation/middleware.go#L90):

Well, I am still able to attach the current context to the request, but maybe we have a bigger problem. 
What do you think ?

DimitarPetrov: I think it is fine since it is also described in the go doc comment above the method then it is intended to work like this.

However the name of the method sounds like a getter, that's why I misinterpreted it. But it is somehow reasonable to work like this since we want to be sure that when we generate the correlation_id it will be available for every handler down the chain.

karaimin: I attached the context to request just in case something in the implementation changes

danielgospodinow (approved):

chart/compass/values.yaml
```diff
@@ -42,25 +42,25 @@ global:
       path: eu.gcr.io/kyma-project/incubator
     connector:
       dir:
-      version: "PR-1767"
+      version: "PR-1798"
```
danielgospodinow, line 45: Since there are changes to the e2e tests, there should also be a PR number bump in the `e2e_tests` field below.

danielgospodinow (changes requested):
Since there are changes to the e2e tests, there should also be a PR number bump in the `e2e_tests` field in the `values.yaml` file.

danielgospodinow (dismissed):

danielgospodinow (approved):

PR #1811, opened 2021-03-31
Fix last operation handling

pankrator:


tests/director/tests/main_test.go
```diff
@@ -40,6 +42,7 @@ func TestMain(m *testing.M) {
 		log.Fatal(errors.Wrap(err, "while getting dex token"))
 	}
 	dexGraphQLClient = gql.NewAuthorizedGraphQLClient(dexToken)
+	directorHTTPClient = gql.NewAuthorizedHTTPClient(dexToken)
```
danielgospodinow, line 45: Should we also comment this line? Since this Director HTTP client is only used in the test which is currently commented out.

danielgospodinow (dismissed):
Looks good 👌🏼

tests/director/tests/main_test.go
pankrator: I left it uncommented, as it might be used outside the current test if someone needs it before it

danielgospodinow (approved):

PR #1814, opened 2021-04-01
Fix asynchronous system deletion failure

danielgospodinow:
**Description**
Asynchronous system deletion fails due to unmarshal error.

- Fix unmarshalling issue in cases when webhook call fails
- Fix unclosed HTTP response bodies


components/operations-controller/internal/webhook/client.go
```diff
@@ -124,7 +124,12 @@ func (c *client) Do(ctx context.Context, request *Request) (*web_hook.Response,
 	isAsyncWebhook := webhook.Mode != nil && *webhook.Mode == graphql.WebhookModeAsync
 
 	if isLocationEmpty && isAsyncWebhook {
-		return nil, errors.New("missing location url after executing async webhook")
+		respBody, err := ioutil.ReadAll(resp.Body)
```
ataleksandrov, line 127: The response body is already read in the `parseResponseObject`. We should close it in order to avoid resource leak. Closing the body there won't allow such second read.

components/operations-controller/internal/webhook/client.go
```diff
@@ -207,7 +221,13 @@ func parseResponseObject(resp *http.Response) (*web_hook.ResponseObject, error)
...
 	var errMsg string
 	if *successStatusCode != resp.StatusCode {
-		errMsg += fmt.Sprintf("response success status code was not met - expected %q, got %q; ", *successStatusCode, resp.StatusCode)
+		respBody, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("failed to parse webhook HTTP response with body %s", respBody))
+		}
+		resp.Body = ioutil.NopCloser(bytes.NewBuffer(respBody))
+
+		errMsg += fmt.Sprintf("response success status code was not met - expected %d, got %d and body %s; ", *successStatusCode, resp.StatusCode, respBody)
```
KirilKabakchiev, line 230: have you tried out how such an error would look like ? To me it looks like if an error is present in the template, and you also add the body, you will get the same error twice in the message..

components/operations-controller/internal/webhook/client.go
```diff
@@ -207,7 +221,13 @@ func parseResponseObject(resp *http.Response) (*web_hook.ResponseObject, error)
 func checkForErr(resp *http.Response, successStatusCode *int, error *string) error {
 	var errMsg string
 	if *successStatusCode != resp.StatusCode {
-		errMsg += fmt.Sprintf("response success status code was not met - expected %q, got %q; ", *successStatusCode, resp.StatusCode)
+		respBody, err := ioutil.ReadAll(resp.Body)
```
KirilKabakchiev, line 224: reading the resp body twice in the same method Do/Poll should really be avoided. Instead of adding workarounds such as resp.Body = ioutil.NopCloser(bytes.NewBuffer(respBody)) simply refactor your methods so you dont read the body twice in the same method

KirilKabakchiev (changes requested):

components/operations-controller/internal/webhook/client.go
danielgospodinow: Yes, you're absolutely correct. It's my fault here because I had setup a mock server which didn't return an error JSON in the response body, thus the `error` field in the output template was empty. That's why I didn't see any good error messages in the first place and that's why I though that it would be a good idea to inject the response body in there. I'll correct this.

KirilKabakchiev (approved):

PR #1817, opened 2021-04-02
make docker driver default for minikube

StanislavStefanov:
- make docker driver default for minikube
- introduce `--hyperkit` flag in cese we want to use hyperkit driver


installation/cmd/run.sh
```diff
@@ -81,10 +81,10 @@ fi
 
 if [[ ! ${SKIP_MINIKUBE_START} ]]; then
   echo "Provisioning Minikube cluster..."
-  if [[ ! ${DOCKER_DRIVER} ]]; then
-    kyma provision minikube --cpus ${MINIKUBE_CPUS} --memory ${MINIKUBE_MEMORY} --timeout ${MINIKUBE_TIMEOUT}
-  else
+  if [[ ! ${HYPERKIT} ]]; then
```
danielgospodinow, line 84: I would personally prefer the straight case `if the user has passed the $HYPERKIT arg, then use hyperkit driver. Otherwise, if he didn't do so, just use docker-driver`. This is to avoid the confusing negation logics.

It's up to you to decide if this fits better for you also.

```suggestion
  if [[ ${HYPERKIT} ]]; then
    kyma provision minikube --cpus ${MINIKUBE_CPUS} --memory ${MINIKUBE_MEMORY} --timeout ${MINIKUBE_TIMEOUT}
  else
    kyma provision minikube --cpus ${MINIKUBE_CPUS} --memory ${MINIKUBE_MEMORY} --timeout ${MINIKUBE_TIMEOUT} --vm-driver docker --docker-ports 443:443 --docker-ports 80:80
  fi
```

StanislavStefanov: As discussed offline, the initial approach will be used.

PR #1820, opened 2021-04-06
Fix jwt signing keys

alextargov:
**Description**
There was a missing check when we were parsing the JWT keys and because of that GA creation/deletion failed. If we have more than one JWT key there is no guarantee that the correct one will be used.

- Add the necessary check and ensure the correct JWK will be used

- [x] Implementation
- [x] Unit tests


desislavaa (approved):
Validated 💯

PR #1825, opened 2021-04-08
Pass grant type and client id in update body

desislavaa:
**Description**
Yet another fix for https://github.com/kyma-incubator/compass/pull/1768

- Fix Hydra client update - pass required grant types as well, otherwise they're set to `[]`
- The required grant types are now part of the same configuration as the scopes, instead of a hardcoded value
- [**Important/TBD**] The Kyma runtime requires access to the system auths' IDs of applications - they're exctracted in the [compass-runtime-agent](https://github.com/kyma-project/kyma/blob/main/components/compass-runtime-agent/internal/compass/director/converter.go#L31) and used for certificate subject validation in [application-connectivity-validator](https://github.com/kyma-project/kyma/blob/main/components/application-connectivity-validator/internal/validationproxy/handler.go#L198) 

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


desislavaa:
✅ Validatinon procedure:
1. Create a new integration system and request client credentials
2. Manually update the client in Hydra via:
```
curl http://ory-hydra-admin.kyma-system.svc.cluster.local:4445/clients/9e974835-f577-4b23-9e8a-73bc00f71d75 -X PUT -d '{"scope": "application:read"}'
```
3. Run the synchronization job
4. Check that the client is updated in Hydra - both `scope` and `grant_types` fields are present
5. Request token for the client credentials - it should not fail. Previously it did with `The OAuth 2.0 Client is not allowed to use authorization grant "client_credentials"` error message.
6. Fetch any resource with the integration system client credentais

pankrator (approved):

PR #1826, opened 2021-04-09
Fix async system deletion stuck on failed error field parse

danielgospodinow:
**Description**

Fix a bug that caused failure of async system deletion. 

Skipping keys with `null` value should prevent this issue from happening.


components/operations-controller/internal/webhook/client.go
```diff
@@ -206,6 +206,9 @@ func parseResponseObject(resp *http.Response) (*web_hook.ResponseObject, error)
 		}
 
 		for k, v := range tmpBody {
+			if v == nil {
```
KirilKabakchiev, line 209: extend the unit tests in client.go and the e2e tests in external service async delete (the e2e test is already there, it just returns not the real response body) to cover this case

KirilKabakchiev (changes requested):

KirilKabakchiev (approved):
Looks great, thanks very much for taking the time to add the tests

PR #1827, opened 2021-04-12
Sanitize directive

StanislavStefanov:
**Description**

- introduce separate SystemAuth types for Application, Runtime and Integration System
- strip sensitive data from SystemAuth but leave SysytemAuth.ID visible by default

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


tests/director/tests/sensitive_data_strip_test.go
```diff
@@ -26,17 +26,13 @@ func TestSensitiveDataStrip(t *testing.T) {
 
 	t.Log("Creating application template")
 	appTmpInput := fixtures.FixApplicationTemplateWithWebhook("app-template-test")
-	var appTemplateID string
-	defer fixtures.DeleteApplicationTemplate(t, ctx, dexGraphQLClient, tenantId, appTemplateID)
 	appTemplate := fixtures.CreateApplicationTemplateFromInput(t, ctx, dexGraphQLClient, tenantId, appTmpInput)
-	appTemplateID = appTemplate.ID
+	defer fixtures.DeleteApplicationTemplate(t, ctx, dexGraphQLClient, tenantId, appTemplate.ID)
```
desislavaa, line 30: deferring before we try to create the resource was suggested in the previous PR - it makes sence, since the creation could go wrong, and in case of an error somewhere in the function that creates the resource, we'll never reach the defer statement

desislavaa (approved):
✅ Validated locally by:
Running full Kyma installation and checking if the compass-runtime-agent can connect successfuly. Previously, it could not.

PR #1831, opened 2021-04-13
Handle Moved Runtime events

karaimin:
**Description**

- Tenant Fetcher Cron Job now can handle new event (Moved labeled runtime to different tenant). The processing is placed between create and delete tenant operations
- update runtimes with preconfigured label key and received label value, with new tenant id
- upsert label definition for the new tenant id
- add cascade update (runtime, tenant_id) foreign keys (to update related entities as well)

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/labeldef/repository.go
```diff
@@ -61,6 +66,15 @@ func (r *repository) Create(ctx context.Context, def model.LabelDefinition) erro
 	return nil
 }
 
+func (r *repository) Upsert(ctx context.Context, label model.LabelDefinition) error {
```
NickyMateev, line 69: Unit testing?

components/director/internal/domain/labeldef/service.go
```diff
@@ -128,6 +131,18 @@ func (s *service) Update(ctx context.Context, def model.LabelDefinition) error {
 	return nil
 }
 
+func (s service) Upsert(ctx context.Context, def model.LabelDefinition) error {
```
NickyMateev, line 134: Unit testing?

components/director/internal/domain/runtime/repository.go
```diff
@@ -202,6 +202,16 @@ func (r *pgRepository) Update(ctx context.Context, item *model.Runtime) error {
 	return r.updater.UpdateSingle(ctx, runtimeEnt)
 }
 
+func (r *pgRepository) UpdateTenantID(ctx context.Context, runtimeID, newTenantID string) error {
```
NickyMateev, line 205: Unit testing?

components/director/internal/domain/runtime/service.go
```diff
@@ -140,6 +141,10 @@ func (s *service) GetByTokenIssuer(ctx context.Context, issuer string) (*model.R
 	return runtime, nil
 }
 
+func (s *service) GetByFiltersGlobal(ctx context.Context, filter []*labelfilter.LabelFilter) (*model.Runtime, error) {
```
NickyMateev, line 144: Unit testing?

components/director/internal/domain/runtime/service.go
```diff
@@ -370,6 +375,13 @@ func (s *service) ListLabels(ctx context.Context, runtimeID string) (map[string]
 	return extractUnProtectedLabels(labels, s.protectedLabelPattern)
 }
 
+func (s *service) UpdateTenantID(ctx context.Context, runtimeID, newTenantID string) error {
```
NickyMateev, line 378: Unit testing?

chart/compass/values.yaml
```diff
@@ -354,6 +354,7 @@ global:
         tenantCreated: "127.0.0.1/events?type=created"
         tenantDeleted: "127.0.0.1/events?type=deleted"
         tenantUpdated: "127.0.0.1/events?type=updated"
+        movedRuntimeByLabel: "127.0.0.1/events?type=moved"
```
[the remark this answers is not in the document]
karaimin: Because it is semantically not true. The old tenant will still exist.

components/schema-migrator/migrations/director/20210413140518_update-runtime-tenantId-update-cascade.down.sql
```diff
@@ -0,0 +1,23 @@
+BEGIN;
+
+ALTER TABLE labels DROP CONSTRAINT labels_tenant_id_fkey1;
+ALTER TABLE labels
+    ADD CONSTRAINT labels_tenant_id_fkey1
+        FOREIGN KEY (tenant_id, runtime_id) REFERENCES runtimes (tenant_id, id)
+            ON UPDATE NO ACTION
```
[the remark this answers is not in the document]
karaimin: No, The FK is recreated. So, there is no problem to delete the whole `ON UPDATE`  cascade. But, it looks slightly  better ... like code documentation what is reverted.

components/director/internal/tenantfetcher/service.go
```diff
@@ -88,84 +154,103 @@ func (s Service) SyncTenants() error {
...
-		return err
+	if err = s.tenantStorageService.CreateManyIfNotExists(ctx, tenantsToCreate); err != nil {
+		return errors.Wrap(err, "while storing new tenants")
 	}
 
-	err = s.kubeClient.UpdateTenantFetcherConfigMapData(ctx, convertTimeToUnixNanoString(startTime))
+	return nil
+}
+
+func (s Service) deleteTenants(ctx context.Context, ts *currentTenantsSupplier, eventsTenants []model.BusinessTenantMappingInput) error {
```
NickyMateev, line 236: Could you place the `moveRuntimesByLabel` method below this `deleteTenants` so as to sort them in a logical way as this file is already pretty overcrowded and there are numerous functions/methods everywhere?

components/director/internal/tenantfetcher/service.go
```diff
@@ -88,84 +154,103 @@ func (s Service) SyncTenants() error {
...
 	}
 	defer s.transact.RollbackUnlessCommitted(ctx, tx)
-
 	ctx = persistence.SaveToContext(ctx, tx)
 
-	currentTenants, err := s.tenantStorageService.List(ctx)
-	if err != nil {
-		return errors.Wrap(err, "while listing tenants")
-	}
+	existingTenantsSup := s.newCurrentTenantsSupplier()
```
NickyMateev, line 187: Is there a reason for this additional abstraction `currentTenantsSupplier`? As far as I see it's only used so that `createTenants` and `deleteTenants` can call it's `Get(ctx)` method. Why not simply make the `Get(ctx)` method a standalone function which returns existing tenants here and then provide them to the `createTenants` and `deleteTenants`?

components/director/internal/tenantfetcher/service.go
```diff
@@ -38,44 +49,99 @@ type QueryConfig struct {
...
+	for _, ct := range currentTenants {
+		supplier.currentTenantsMap[ct.ExternalTenant] = true
 	}
+
+	return supplier.currentTenantsMap, nil
+}
+
+func (a syncTenantsAction) Execute() error {
+	if err := a.fn(); err != nil {
+		return errors.Wrap(err, "while "+a.name)
```
NickyMateev, line 141: Either remove "while" from here or from the actions as it currently puts 2x "while"

components/schema-migrator/migrations/director/20210413140518_update-runtime-tenantId-update-cascade.down.sql
```diff
@@ -0,0 +1,23 @@
+BEGIN;
```
NickyMateev, line 1: These tables also have a relation to the `runtimes` table:
- `api_runtime_auths`
- `webhooks`

components/director/internal/tenantfetcher/service.go
karaimin: There were a meaning, but after all the refactoring maybe we don't need it anymore :D

components/director/internal/domain/runtime/service.go
krasish: Is it really necessary to write such a test? What I think it would do is simply test that error from repo layer is forwarded which is obvious. Is there any other reason for writing such a test, that I am missing?

components/director/internal/tenantfetcher/client.go
```diff
@@ -29,9 +29,10 @@ type OAuth2Config struct {
 }
 
 type APIConfig struct {
-	EndpointTenantCreated string `envconfig:"APP_ENDPOINT_TENANT_CREATED"`
-	EndpointTenantDeleted string `envconfig:"APP_ENDPOINT_TENANT_DELETED"`
-	EndpointTenantUpdated string `envconfig:"APP_ENDPOINT_TENANT_UPDATED"`
+	EndpointTenantCreated       string `envconfig:"APP_ENDPOINT_TENANT_CREATED"`
+	EndpointTenantDeleted       string `envconfig:"APP_ENDPOINT_TENANT_DELETED"`
+	EndpointTenantUpdated       string `envconfig:"APP_ENDPOINT_TENANT_UPDATED"`
+	EndpointRuntimeMovedByLabel string `envconfig:"optional,APP_ENDPOINT_RUNTIME_MOVED_BY_LABEL"`
```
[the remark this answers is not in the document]
krasish: Because we would not be able to start the program if we don't provide the APP_ENDPOINT_RUNTIME_MOVED_BY_LABEL environment variable if optional is omitted here. In that case, we want to start the program and skip getting events for moved runtimes due to the check that you pointed.

components/director/internal/domain/labeldef/service.go
```diff
@@ -128,6 +131,18 @@ func (s *service) Update(ctx context.Context, def model.LabelDefinition) error {
...
+	def.ID = s.uidService.Generate()
```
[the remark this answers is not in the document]
krasish: That ID will be used in case there is not an already existing record. We could use the Create function but we would need to make an Exist call prior to calling Create. That way we would need to execute 2 database calls while using the Upsert we only make 1.

components/director/internal/tenantfetcher/service.go
```diff
@@ -88,84 +154,103 @@ func (s Service) SyncTenants() error {
...
 		return err
 	}
 
-	totalNewEvents := len(tenantsToCreate) + len(tenantsToDelete)
+	runtimesToMove, err := s.getRuntimesToMoveByLabel(lastConsumedTenantTimestamp)
+	if err != nil {
+		return err
+	}
+
+	totalNewEvents := len(tenantsToCreate) + len(tenantsToDelete) + len(runtimesToMove)
```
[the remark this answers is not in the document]
krasish: Changed that to the previous behavior.

components/director/internal/tenantfetcher/service.go
```diff
@@ -88,84 +154,103 @@ func (s Service) SyncTenants() error {
...
-	if err != nil {
-		return errors.Wrap(err, "while listing tenants")
-	}
+	existingTenantsSup := s.newCurrentTenantsSupplier()
 
-	currentTenantsMap := make(map[string]bool)
-	for _, ct := range currentTenants {
-		currentTenantsMap[ct.ExternalTenant] = true
+	actions := make([]*syncTenantsAction, 0)
+	if len(tenantsToCreate) == 0 && len(tenantsToDelete) == 0 {
```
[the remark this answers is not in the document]
krasish: In order to make this, we would need to append the delete actions and prepend the create actions to the actions slice which would be something like te following: 
``` go
        actions := make([]*syncTenantsAction, 0)
	actions = append(actions, &syncTenantsAction{"while moving runtimes by labels", func() error { return s.moveRuntimesByLabel(ctx, runtimesToMove) }})
	if len(tenantsToCreate) != 0 || len(tenantsToDelete) == 0 {
		actions = append([]*syncTenantsAction{&syncTenantsAction{"while storing tenants", func() error { return s.createTenants(ctx, existingTenantsSup, tenantsToCreate) }}}, actions...)
		actions = append(actions, &syncTenantsAction{"while deleting tenants", func() error { return s.deleteTenants(ctx, existingTenantsSup, tenantsToDelete) }})
	}
```
To me, it looks a bit confusing and it does not emphasize elements order which is sufficient in this case.

components/director/internal/domain/labeldef/service.go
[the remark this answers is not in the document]
NickyMateev: `s.uidService.Generate()` will always generate and put a non-existing GUID into `def.ID`. In which scenario will this `Upsert` function act as an update rather than a create?

NickyMateev (approved):

components/director/internal/tenantfetcher/service.go
```diff
@@ -88,82 +131,135 @@ func (s Service) SyncTenants() error {
...
-	currentTenants, err := s.tenantStorageService.List(ctx)
-	if err != nil {
-		return errors.Wrap(err, "while listing tenants")
-	}
+	actions := make([]*syncTenantsAction, 0)
 
-	currentTenantsMap := make(map[string]bool)
-	for _, ct := range currentTenants {
-		currentTenantsMap[ct.ExternalTenant] = true
+	if len(tenantsToCreate) == 0 && len(tenantsToDelete) == 0 {
```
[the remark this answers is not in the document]
karaimin: Yeah, they handle it, but it is not visible enough. You have to look a lot deeper in the call hierarchy in order to understand this. IMO, ensuring that you pass only valid/meaningful data on top level method seems to be more appropriate way to go. Also, ... 
- this prevent passing nil values to the action methods. 
- one db call can be avoided _(getting the existing tenants)_

It is also possible to place an explicit check (the same check) at the beginning of each action method. This will make entrypoint method `SyncTenants` more compact, but it will lead to:
 - add extra check for empty slice at the beginning of each action method
 - bring back the previously introduced `currentTenantsSupplier` struct (get tenants if not did it yet) in order to ensure one db call (or none -  for missing create/delete events) is executed to fetch the current tenants. (`createTenants / deleteTenants` need it)

Both approaches have their advantages. If you insist on cleaner `SyncTenants` method we can go with the second approach. ;)

NickyMateev (approved):

PR #1845, opened 2021-04-26
Adaptation of ORD RC2

nyordanoff:
**Description**

Currently, the relation between Bundles and Apis/Events is `One-to-Many` (a single Bundle can have many Apis/Events but a single Api/Event can be part of only one Bundle).

This PR introduces `Many-to-Many` relation between Bundles and Apis/Events (a single Bundle can have many Apis/Events and a single Api/Event can be part of many Bundles) due to ORD requirements. In addition, one API can have a different `targetURL` per Bundle.

The whole change is backwards compatible from GraphQL layer point of view, meaning that GraphQL requests still support only the `One-to-Many` relation.

- to achieve the Many-to-Many relation, a new table `bundle_references` is introduced (JOIN Table in SQL terms) which holds information about `bundle_id/api_id/event_id` and `targetURL` of the given API in the given Bundle;
- a new Entity `BundleReference` is introduced which propagates the new logic across all layers;
- API/Event service/repo layer is adapted to handle both `One-to-Many` requests from the GraphQL as well as `Many-to-Many` ones from the aggregator;
- adaptation in the aggregator so that it can properly handle the new `Many-to-Many` relation;
- Changes and adaptations to ORD specific fields;

**Related issue(s)**
- https://github.com/kyma-incubator/ord-service/pull/21

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests
- [x] `chart/compass/values.yaml`


components/director/internal/domain/bundlereferences/repository.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (r *repository) GetBundleIDsForObject(ctx context.Context, tenantID string, objectType model.BundleReferenceObjectType, objectID *string) (ids []string, err error) {
+	fieldName, err := r.referenceObjectFieldName(objectType)
+	if err != nil {
+		return nil, err
+	}
+
+	var objectBundleIDs IDs
+
+	r.lister.SetSelectedColumns([]string{bundleIDColumn})
```
pankrator, line 102: as the lister is one for the repository and the repository is shared between resolver requests it might happen that multiple requests use this method and try to change a property of the object which will lead to concurrency problem.
Perhaps instantiating a new instance of the lister on change will have to be done in order not to use mutex which will be slower IMHO.

components/director/internal/repo/list.go
```diff
@@ -67,6 +68,10 @@ func (l *universalLister) List(ctx context.Context, tenant string, dest Collecti
 	return l.unsafeList(ctx, dest, additionalConditions...)
 }
 
+func (l *universalLister) SetSelectedColumns(selectedColumns []string) {
+	l.selectedColumns = strings.Join(selectedColumns, ", ")
```
pankrator, line 72: Perhaps in order to avoid concurrency problems, a copy of the lister can be created, modified and returned

pankrator (changes requested):

pankrator (dismissed):

pankrator (approved):

PR #1850, opened 2021-05-13
NoOpService now implements PreLog

dzahariev:
**Description**
- NoOpService now implements PreLog

**Pull Request status**
- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


ataleksandrov (commented):
We didn't catch that because there is a bug here: https://github.com/kyma-incubator/compass/blob/master/components/gateway/cmd/main.go#L65. This service should be PreAuditlogService. Moreover here: https://github.com/kyma-incubator/compass/blob/master/components/gateway/pkg/proxy/transport.go#L47 this should also be PreAuditlogService and then there is no need of this cast: https://github.com/kyma-incubator/compass/blob/master/components/gateway/pkg/proxy/transport.go#L82

ataleksandrov (approved):

ataleksandrov:
LGTM

ataleksandrov (approved):

PR #1851, opened 2021-05-13
Adjust livenessProbe settings

dzahariev:
**Description**
- Adjust livenessProbe settings

**Pull Request status**
- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


ataleksandrov (commented):
I would suggest to add a separate livenessProbe configuration to the ORD service in particular. All another components have no problem starting under 30 seconds. This change to global livenessProbe would significantly slow down the whole installation.

ataleksandrov (approved):

PR #1858, opened 2021-05-17
Ord rc3 adaptation

alextargov:
**Description**

 - Validation when the ORD docs have APIResources of policyLevel='sap' or 'sap-partner' and with apiProtocol='soap-inbound' or 'soap-outbound', then it is mandatory to provide either WSDL V2 or WSDL V1 definitions.
 - Validation when the ORD docs have APIResources of policyLevel='sap' or 'sap-partner' and with apiProtocol='odata-v2' or 'odata-v4', then it is mandatory to not only provide edmx definitions, but also OpenAPI definitions.
- Replaced Vendor.sapPartner property with a more generic Vendor.partners property
- Renamed PackageLink.type enum value licence to license

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/internal/open_resource_discovery/validation.go
+++ b/components/director/internal/open_resource_discovery/validation.go
@@ -35,6 +35,32 @@ const (
 	VendorPartnersRegex               = "^([a-zA-Z0-9._\\-]+):(vendor):([a-zA-Z0-9._\\-]+):()$"
 )
 
+const (
+	PolicyLevelSap        string = "sap"
+	PolicyLevelSapPartner string = "sap-partner"
+	PolicyLevelCustom     string = "custom"
+
+	ReleaseStatusBeta       string = "beta"
+	ReleaseStatusActive     string = "active"
+	ReleaseStatusDeprecated string = "deprecated"
+
+	ApiProtocolODataV2      string = "odata-v2"
+	ApiProtocolODataV4      string = "odata-v4"
+	ApiProtocolSoapInbound  string = "soap-inbound"
+	ApiProtocolSoapOutbound string = "soap-outbound"
+	ApiProtocolRest         string = "rest"
+	ApiProtocolSapRrc       string = "sap-rfc"
+
+	ApiVisibilityPublic   string = "public"
+	ApiVisibilityPrivate  string = "private"
+	ApiVisibilityInternal string = "internal"
+
+	ApiImplementationStandardDocumentApi   string = "sap:ord-document-api:v1"
+	ApiImplementationStandardServiceBroker string = "cff:open-service-broker:v2"
+	ApiImplementationStandardCsnExposure   string = "sap:csn-exposure:v1"
+	ApiImplementationStandardCustom        string = "custom"
+)
+
 var shortDescriptionRules = []validation.Rule{
 	validation.Required, validation.Length(1, 255), validation.NewStringRule(noNewLines, "short description should not contain line breaks"),
 }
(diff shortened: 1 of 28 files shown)


dzahariev (approved):

dzahariev (approved):

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -132,21 +161,55 @@ func validateAPIInput(api *model.APIDefinitionInput) error {
 		validation.Field(&api.Industry, validation.By(func(value interface{}) error {
 			return validateJSONArrayOfStrings(value, regexp.MustCompile(StringArrayElementRegex))
 		})),
-		validation.Field(&api.ResourceDefinitions, validation.Required),
+		validation.Field(&api.ResourceDefinitions, validation.Required, validation.By(func(value interface{}) error {
+			if value == nil {
+				return nil
+			}
+
+			jsonArr := api.ResourceDefinitions
```
nyordanoff, line 169: `jsonArr` is a bit misleading because the ResourceDefinitions are not represented as `json.RawMessage`.

`resourceDefinitions` is better from my point of view

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -132,21 +161,55 @@ func validateAPIInput(api *model.APIDefinitionInput) error {
...
+
+			if len(jsonArr) == 0 {
+				return nil
+			}
+
+			ordID := str.PtrStrToStr(api.OrdPackageID)
```
nyordanoff, line 175: minor: maybe rename it to `pkgOrdID` for further clarification

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -455,6 +522,39 @@ func validateJSONArrayOfObjects(arr interface{}, elementFieldRules map[string][]
 	return nil
 }
 
+func validateJSONObjects(obj interface{}, isOptional bool, elementFieldRules map[string][]validation.Rule, crossFieldRules ...func(gjson.Result) error) error {
+	jsonObj, ok := obj.(json.RawMessage)
```
nyordanoff, line 526: Why the check `if obj == nil {return nil}` is skipped here?

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -455,6 +522,39 @@ func validateJSONArrayOfObjects(arr interface{}, elementFieldRules map[string][]
...
+	if !ok {
+		return errors.New("should be json")
+	}
+
+	if len(jsonObj) == 0 && isOptional {
+		return nil
+	}
+
+	if !gjson.ValidBytes(jsonObj) && !isOptional {
```
nyordanoff, line 535: I don't quite get the logic behind the `isOptional`

For example: here if invalid JSON is passed but isOptional == true, then error won't be thrown and we continue with invalid json. Is this a desired behaviour?

I think we can reuse the validation part of `validateJSONArrayOfObjects` function

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -3005,6 +3237,51 @@ func TestDocuments_ValidateEvent(t *testing.T) {
 				return []*open_resource_discovery.Document{doc}
 			},
 		},
+		{
+			Name: "Missing `supported` field in the `extensible` object for API",
```
nyordanoff, line 3241: `for Event`

the same applies to all test case names below

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -3304,6 +3581,31 @@ func TestDocuments_ValidateVendor(t *testing.T) {
 				doc := fixORDDocument()
 				doc.Vendors[0].Labels = json.RawMessage(invalidLabelsWhenKeyIsWrong)
 
+				return []*open_resource_discovery.Document{doc}
+			},
+		},
+		{
```
nyordanoff, line 3587: New test case must be added where a value of the `Partners` array doesn't match the required regex

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -3304,6 +3581,31 @@ func TestDocuments_ValidateVendor(t *testing.T) {
...
+			Name: "Invalid JSON object `Partners` field for Vendor",
+			DocumentProvider: func() []*open_resource_discovery.Document {
+				doc := fixORDDocument()
+				doc.Vendors[0].Partners = json.RawMessage(`[]`)
```
nyordanoff, line 3591: The test case name suggests testing with invalid JSON, so instead of `[]` we can pass  the `invalidJson` variable which is declared in the tests

In addition, a new test case must be created where `Invalid `Partners` field when the JSON array is empty` where `[]` will be passed. 

This distinction is needed because these are two separate checks in the code.

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -3304,6 +3581,31 @@ func TestDocuments_ValidateVendor(t *testing.T) {
...
+
+				return []*open_resource_discovery.Document{doc}
+			},
+		}, {
+			Name: "`Partners` values are not array for Vendor",
+			DocumentProvider: func() []*open_resource_discovery.Document {
+				doc := fixORDDocument()
+				doc.Vendors[0].Labels = json.RawMessage(invalidPartnersWhenValueIsNotArray)
```
nyordanoff, line 3599: `doc.Vendors[0].Partners`

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -3304,6 +3581,31 @@ func TestDocuments_ValidateVendor(t *testing.T) {
...
+
+				return []*open_resource_discovery.Document{doc}
+			},
+		}, {
+			Name: "`Partners` values are not array of strings for Vendor",
+			DocumentProvider: func() []*open_resource_discovery.Document {
+				doc := fixORDDocument()
+				doc.Vendors[0].Labels = json.RawMessage(invalidPartnersWhenValuesAreNotArrayOfStrings)
```
nyordanoff, line 3607: Two things here:

1. `doc.Vendors[0].Partners`
2. the `invalidPartnersWhenValuesAreNotArrayOfStrings` JSON is not correct for the use-case of that test. The json should be `["<validPartner>", 31231]` i.e a validPartner and a number

PR #1860, opened 2021-05-19
Allow Director to use DB dump

NickyMateev:
**Description**

When running the Director `run.sh` script or the global Compass `installation/cmd/run.sh` script, simply pass `--dump-db` flag. A DB dump will be pulled from trusted GCP remote source if a dump file is not present already.

If you want to make a fresh dump directly from a DB instance, the `components/schema-migrator/dump_db.sh` script can be used. (**Prerequisite:** *kubectl should point to existing dev CMP cluster*)

- [x] Allow for local Director setup to reuse DB dump [**Scenario:** Quick check whether new migrations are valid & also play around with new Director functionality]
- [x] Allow for local Minikube setup to reuse DB dump [**Scenario:** Quick check whether new migrations are valid & also play around with new functionality in the context of a complete Compass cluster]
- [x] Allow for local Minikube E2E tests to reuse DB dump ** (some tests are breaking, need to re-check)
- [x] Allow for remote E2E tests to reuse DB dump
- [x] Allow for local `validate.sh` Schema Migrator which executes migrations to reuse DB dump [**Scenario:** Quick check whether new migrations are valid with data]
- [x] Allow for remote `validate.sh` Schema Migrator on PR job which executes migrations to reuse DB dump

> For the `validate.sh` Schema Migrator script the DOWN migrations don't make sense as there is data already in the database, so the DOWN migrations would error.


components/director/run.sh
```diff
@@ -121,12 +131,83 @@ else
 
     echo -e "${GREEN}Populate DB${NC}"
 
-    cat ${ROOT_PATH}/../schema-migrator/migrations/director/*.up.sql | \
-        docker exec -i ${POSTGRES_CONTAINER} psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
-    cat ${ROOT_PATH}/../schema-migrator/seeds/director/*.sql | \
-        docker exec -i ${POSTGRES_CONTAINER} psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
+    if [[ ${DUMP_DB} = false ]]; then
+        cat ${ROOT_PATH}/../schema-migrator/migrations/director/*.up.sql | \
```
KirilKabakchiev, line 135: why not simply use `migrate` here ? or even better just call run.sh of the schema migrator

components/director/run.sh
```diff
@@ -121,12 +131,83 @@ else
...
+            echo -e "${GREEN}Database dumped!${NC}"
+
+            pkill kubectl
+        fi
+
+        cat ${ROOT_PATH}/../schema-migrator/seeds/dump.sql | \
+            docker exec -i ${POSTGRES_CONTAINER} psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
+
+
+       LATEST_MIGRATION_VERSION=$(docker exec -i ${POSTGRES_CONTAINER} psql -qtAX -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}" -c "SELECT version FROM schema_migrations")
```
KirilKabakchiev, line 175: why is this with latest migration version done ? (mb you want to make sure new migrations are applied even without restarting the local director? cant you simply use `migrate` / run.sh here aswell ?

components/director/run.sh
```diff
@@ -121,12 +131,83 @@ else
...
+            trap cleanup_temp_migrations EXIT
+
+            PGPASSWORD=$REMOTE_DB_PWD pg_dump --dbname=director --file=${ROOT_PATH}/../schema-migrator/seeds/dump.sql --host=localhost --port=5555 --username=director --column-inserts --no-owner --no-privileges
+
+            echo -e "${GREEN}Database dumped!${NC}"
+
+            pkill kubectl
+        fi
+
+        cat ${ROOT_PATH}/../schema-migrator/seeds/dump.sql | \
```
KirilKabakchiev, line 171: if the db on dev is on different version than the local db, it would be good to atleast get some warning that  the data  import might fail to save someone troubleshooting time. Alternatively you might fix it so that if the 2 db versions differ, you might "roll back"(apply only up to) the local db to match the remote db before the import, do the data import and than roll up the extra migrations ontop - this might even be used to kind of quickly verify that the new migrations can be properly applied with existing data while starting up the local dev (a useful side effect?). Then the minikube PR prow job (if we configure it to do dump importing) will actually always verify that the new migrations are OK. 
Anyway, I think we don't need to spin up gke clusters for this so prolly using minikube should be fine/faster esp as its just migration stuff.

components/director/run.sh
```diff
@@ -121,12 +131,83 @@ else
...
+           echo -e "${GREEN}New migrations will be applied...${NC}"
+           cat ${ROOT_PATH}/../schema-migrator/migrations/new_migrations.sql | \
+               docker exec -i ${POSTGRES_CONTAINER} psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
+       else
+           echo -e "${GREEN}No new migrations will be applied.${NC}"
+       fi
+    fi
 fi
 
+INTERNAL_TENANT_ID=$(docker exec -i ${POSTGRES_CONTAINER} psql -qtAX -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}" -c "SELECT id FROM business_tenant_mappings WHERE external_tenant = '3e64ebae-38b5-46a0-b1ed-9ccee153a0ae'")
```
KirilKabakchiev, line 203: mb extract this in another script  so it can be usable standalone and call it here? at some point one might need just a token without wanting to rerun run.sh to get one

KirilKabakchiev (changes requested):

components/director/cmd/jwtgenerator/main.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/form3tech-oss/jwt-go"
+)
+
+type Claims struct {
+	Scopes string `json:"scopes"`
+	Tenant string `json:"tenant"`
+	jwt.StandardClaims
+}
+
+func main() {
```
KirilKabakchiev, line 16: do you really need a whole go main file for that ? cant this be just a shell script ?

components/director/run.sh
NickyMateev: It's definitely smarter to use `migrate` here 👍

components/director/run.sh
NickyMateev: It will not be uncommon for local db to mismatch dev db during development. If a mismatch exists it will be pretty much always in favor of the local db which has newer migrations (based on latest master + current feature that's being developed).
Having said that I can't think of a scenario where the newer migrations when applied on a db based on the migrations to that point + corresponding data (talking about the db dump from dev here) might fail unless the migrations themselves are corrupt, which should error anyway?

KirilKabakchiev: if a migration introduces a constraint which is not compatible with the data, the migration will fail. The point was that one option to detect this would be in run.sh

KirilKabakchiev: I guess I went with too few words - there were two parts of the proposal (not sure if it makes sense, you tell me :) )
1. If you do the adjustments in `this` run.sh when one boots up locally his director with dump-data parameter he will get out of the box validation for his new migrations.
2.  if you do the same in the `installation/cmd` run.sh the minikube prow job will also do this same validation on PR. (Here some ip whitelisting in gcp might be needed)

NickyMateev: The validation behaviour that you described above is the exact same way that it is currently implemented. The DB dump file contains not only `INSERT` statements which import the data, but it also includes DDL statements which create all tables, constraints etc.
So applying the dump file to a local DB would essentially mean that we apply all migrations up to the point of the dump along with it's data, and once that's done, we leverage the migrate tool to apply all new migrations and thus validating that any introduced constraints properly work with the existing data we have on dev.

NickyMateev: I will however add a sort of warning to notify when the two versions mismatch. While testing this, I had the unfortunate time wasting situation where the local migrations (which presumably should be ahead of dev migrations) was actually behind dev's migrations - this happens when you work on a branch for some period without merging the latest master changes..

The `migrate` tool was returning an error with a somewhat unintuitive message and it didn't make sense until I realized what was happening. This was the error message: `error: file does not exist`

components/schema-migrator/run.sh
```diff
@@ -36,11 +36,32 @@ if [ "${dbReady}" != true ] ; then
...
+
+if [[ -f seeds/dump.sql ]]; then
+    echo "Will reuse existing dump in seeds/dump.sql"
+    cat seeds/dump.sql | \
+        PGPASSWORD="${DB_PASSWORD}" psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}" --set="${SSL_OPTION}"
+
+    REMOTE_MIGRATION_VERSION=$(PGPASSWORD="${DB_PASSWORD}" psql -qtAX -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}" -c "SELECT version FROM schema_migrations")
+    LOCAL_MIGRATION_VERSION=$(echo $(ls migrations/director | tail -n 1) | grep -o -E '[0-9]+' | head -1 | sed -e 's/^0\+//')
+
+    if [[ ${REMOTE_MIGRATION_VERSION} = ${LOCAL_MIGRATION_VERSION} ]]; then
```
krasish, line 55: Maybe add coloring to messages displayed in this if's body just like you've done in director/run.sh. Looks quite good to me :D

components/schema-migrator/validate.sh
```diff
@@ -92,7 +123,10 @@ function migrationProcess() {
 
     echo -e "${GREEN}Migrations for \"${db}\" database and \"${path}\" path${NC}"
     migrationUP "${path}" "${db}"
-    migrationDOWN "${path}" "${db}"
+
+    if [[ ! -f seeds/dump.sql ]]; then
```
krasish, line 127: Why don't we execute down migrations if we've used a db dump? Is that ok?

installation/cmd/run.sh
```diff
@@ -79,6 +83,16 @@ if [ -z "$KYMA_INSTALLATION" ]; then
   KYMA_INSTALLATION="minimal"
 fi
 
+if [[ ${DUMP_DB} ]]; then
+    sed -i - "s/image\:.*compass-schema-migrator.*/image\: compass-schema-migrator\:latest/" ${ROOT_PATH}/chart/compass/templates/migrator-job.yaml
```
krasish, line 87: As a result of these few commands after the installation is done, we are left with a modified migrator-job.yaml file. If we only wanted to install compass on minikube with a db-dump once, in order to test something with "realish" data we would probably want the migrator-job's image to be left unchanged afterwards. The image should be different only untill the installation takes place. Probably we can save the image's name in a variable and change it back to the original one, once the installation is done.

krasish:
I think you should update values.yaml.

components/schema-migrator/dump_db.sh
```diff
@@ -0,0 +1,36 @@
...
+    REMOTE_DB_PWD=$(base64 -d <<< $(kubectl get secret -n compass-system compass-postgresql -o=jsonpath="{.data['postgresql-director-password']}"))
+    DIRECTOR_POD_NAME=$(kubectl get pods -n compass-system | grep "director" | head -1 | cut -c -33)
+    kubectl port-forward --namespace compass-system $DIRECTOR_POD_NAME 5555:5432 &
+    sleep 5 # necessary for the port-forward to open in time for the next command
+
+    echo -e "${GREEN}Dumping database. This will take about 2-3 minutes...${NC}"
+
+    PGPASSWORD=$REMOTE_DB_PWD pg_dump --dbname=director --file=${ROOT_PATH}/seeds/dump.sql --host=localhost --port=5555 --username=director --column-inserts --no-owner --no-privileges
+
+    echo -e "${GREEN}Database dumped!${NC}"
```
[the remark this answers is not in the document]
NickyMateev: Good catch - in such cases we can simply guard the script with `set -e` and `set +e`.

components/director/run.sh
```diff
@@ -121,12 +127,39 @@ else
...
+        if [[ ! -f ${ROOT_PATH}/../schema-migrator/seeds/dump.sql ]]; then
+            echo -e "${GREEN}Will pull DB dump from GCR bucket${NC}"
+            gsutil cp gs://sap-cp-cmp-dev-db-dump/dump.sql ${ROOT_PATH}/../schema-migrator/seeds/dump.sql
+        fi
+
+        cat ${ROOT_PATH}/../schema-migrator/seeds/dump.sql | \
+            docker exec -i ${POSTGRES_CONTAINER} psql -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}"
+
+        REMOTE_MIGRATION_VERSION=$(docker exec -i ${POSTGRES_CONTAINER} psql -qtAX -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}" -c "SELECT version FROM schema_migrations")
+        LOCAL_MIGRATION_VERSION=$(echo $(ls ${ROOT_PATH}/../schema-migrator/migrations/director | tail -n 1) | grep -o -E '[0-9]+' | head -1 | sed -e 's/^0\+//')
```
PetarTodorovv, line 146: what is the purpose of the | head -1 | sed -e 's/^0\+//' at the end?

components/schema-migrator/dump_db.sh
```diff
@@ -0,0 +1,42 @@
...
+    if [[ $CURRENT_KUBE_CONTEXT != *dev* ]]; then
+        echo -e "${RED}Error: Current kubectl context does not point to an existing Compass development cluster${NC}"
+        exit 1
+    fi
+
+    echo -e "${GREEN}Will dump and use database data from connected kubernetes Compass development cluster${NC}"
+
+
+    REMOTE_DB_PWD=$(base64 -d <<< $(kubectl get secret -n compass-system compass-postgresql -o=jsonpath="{.data['postgresql-director-password']}"))
+    DIRECTOR_POD_NAME=$(kubectl get pods -n compass-system | grep "director" | head -1 | cut -c -33)
```
PetarTodorovv, line 28: cut command at the end ensure we've got the correct director pod? if for some reason we get the wrong pod containing director in the name e.g.compass-director-clients-scopes-synchronization-llrk7?

components/director/run.sh
NickyMateev: Just a fancy way to extract the number of the latest migration

components/schema-migrator/dump_db.sh
NickyMateev: Cut -33 ensures that we only get the name of the pod, otherwise there's some unnecessary junk till the end of the line:
image

components/schema-migrator/validate.sh
NickyMateev: There are some conceptual things to be agreed upon for this to work. Let's discuss it.

installation/cmd/run.sh
NickyMateev: fixed

components/schema-migrator/validate.sh
NickyMateev: Discussed with team.

PetarTodorovv:
Looks good to me ✅

PetarTodorovv (approved):

PR #1875, opened 2021-05-28
Set docker as default driver

StanislavStefanov:

make docker driver default for minikube
introduce --hyperkit flag in cese we want to use hyperkit driver

### DIFF
--- a/installation/cmd/run.sh
+++ b/installation/cmd/run.sh
@@ -15,7 +15,7 @@ MIGRATOR_FILE=$(cat $ROOT_PATH/chart/compass/templates/migrator-job.yaml)
 
 MINIKUBE_MEMORY=8192
 MINIKUBE_TIMEOUT=25m
-MINIKUBE_CPUS=5
+MINIKUBE_CPUS=4
 
 POSITIONAL=()
 while [[ $# -gt 0 ]]
@@ -42,8 +42,8 @@ do
             SKIP_KYMA_START=true
             shift # past argument
         ;;
-        --docker-driver)
-            DOCKER_DRIVER=true
+        --hyperkit)
+            HYPERKIT=true
             shift # past argument
         ;;
         --dump-db)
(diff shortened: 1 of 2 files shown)


installation/cmd/run.sh
```diff
@@ -117,10 +117,18 @@ fi
...
   echo "Provisioning Minikube cluster..."
-  if [[ ! ${DOCKER_DRIVER} ]]; then
-    kyma provision minikube --cpus ${MINIKUBE_CPUS} --memory ${MINIKUBE_MEMORY} --timeout ${MINIKUBE_TIMEOUT}
-  else
+  if [[ ! ${HYPERKIT} ]]; then
+    DOCKER_CPUS=$(docker info --format '{{json .}}' | jq -r '.NCPU')
+    DOCKER_MEMORY=$(($(docker info --format '{{json .}}' | jq -r '.MemTotal') / 1000000))
+    echo "Docker CPUS: " $DOCKER_CPUS
+    echo "Docker Memory: " $DOCKER_MEMORY
+    if [[ ("$DOCKER_CPUS" -lt "$MINIKUBE_CPUS") || ("$DOCKER_MEMORY" -lt "$MINIKUBE_MEMORY")]]; then
```
DimitarPetrov, line 125: This check is not stable because MINIKUBE_CPUS and MINIKUBE_MEMORY can be provided from outside via --minikube-cpus and --minikube-memory flags.

installation/scripts/installer.sh
```diff
@@ -88,11 +88,13 @@ if [ ${ADMIN_PASSWORD} ]; then
 fi
 
 USED_DRIVER=$(minikube profile list -o json | jq -r ".valid[0].Config.Driver")
+
 if [[ $USED_DRIVER == "docker" ]]; then
-  MINIKUBE_IP=$(minikube ssh egrep "minikube$" /etc/hosts | cut -f1)
+  MINIKUBE_IP=$(minikube ssh cat /etc/hosts | grep "minikube" | head -n 1 | cut -f1)
```
DimitarPetrov, line 93: Why you need to cat the file then pipe to grep, when you can do jus `grep "minikube" /etc/hosts | head -n 1`

Also, keep in mind that this same command is used in the kyma-cli. If you think that it is wrong, consider changing it in the kyma-cli as well.

Personally i don't see any difference between those commands (before and after).

DimitarPetrov (changes requested):

DimitarPetrov:
We are removing minikube as local env with kyma 2.0 adoption.

PR #1878, opened 2021-05-28
Adapt ORD v1

nyordanoff:
**Description**

This PR adopts the ORD 1.0 changes.

- Dropped `NOT NULL` constraint for `Api.TargetUrls` column
- Ensuring that API without `targetURLs` can not be part of a Bundle

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/open_resource_discovery/validation.go
```diff
@@ -685,6 +693,16 @@ func isValidDate(date *string) validation.RuleFunc {
 	}
 }
 
+func isAPIPartOfBundles(partOfConsumptionBundles []*model.ConsumptionBundleReference) validation.RuleFunc {
+	return func(value interface{}) error {
+		var err error
+		if partOfConsumptionBundles != nil && len(partOfConsumptionBundles) > 0 {
+			err = errors.New("api without entry points can not be part of consumption bundle")
```
DimitarPetrov, line 700: Minor: You can return the error here and nil below. 

It will be more readable from my PoV. 

If you think this is better I am ok with it also.

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -685,6 +693,16 @@ func isValidDate(date *string) validation.RuleFunc {
 	}
 }
 
+func isAPIPartOfBundles(partOfConsumptionBundles []*model.ConsumptionBundleReference) validation.RuleFunc {
```
DimitarPetrov, line 696: The name `isAPIPartOfBundles` does not imply which is true and which is false.

This is fine when used as a standalone function, however it is used in validation like:

```
validation.When(api.TargetURLs == nil, validation.By(isAPIPartOfBundles(api.PartOfConsumptionBundles))
```

When you read this line you do not know that it is valid when it is part of consumption bundle or when it is not.

I would suggest something like `notPartOfConsumptionBundles`

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -3256,6 +3264,14 @@ func TestDocuments_ValidateEvent(t *testing.T) {
 				doc := fixORDDocument()
 				doc.EventResources[0].PartOfConsumptionBundles[0].DefaultTargetURL = "https://exmaple.com/test/v3"
 
+				return []*open_resource_discovery.Document{doc}
+			},
+		}, {
+			Name: "Empty `PartOfConsumptionBundle` field for Event",
```
DimitarPetrov, line 3270: Missing test for: `an API without targetURLs and not part of consumption bundle should be valid`

components/director/internal/open_resource_discovery/validation.go
nyordanoff: I prefer your approach as well.

DimitarPetrov (dismissed):

DimitarPetrov (approved):

PR #1879, opened 2021-05-31
Ensure Schema Migrator cleans up dump before release

NickyMateev:
# Ensure Schema Migrator cleans up dump before release

**Related issue(s)**
https://github.com/kyma-incubator/compass/pull/1860

### DIFF
--- a/components/schema-migrator/run.sh
+++ b/components/schema-migrator/run.sh
@@ -49,10 +49,12 @@ if [ -n "${DB_SSL}" ] ; then
 fi
 
 
+set -e
+
 if [[ -f seeds/dump.sql ]] && [[ "${DIRECTION}" == "up" ]]; then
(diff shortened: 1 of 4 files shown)


components/schema-migrator/run.sh
```diff
@@ -75,3 +77,5 @@ if [[ "${NON_INTERACTIVE}" == "true" ]]; then
 else
     $CMD
 fi
+
+set +e
```
KirilKabakchiev, line 81: usually this is put on the top of the script. Or is there any particular reason  to put it in the bottom ?

NickyMateev: I wasn't sure if once setting `set -e` the effect would stay for parent scripts calling this one and so I added this just in case. If you think it's unnecessary, I can drop it or change it somehow.

KirilKabakchiev (approved):

PR #1881, opened 2021-06-01
E2E tests execution on real environment

pankrator:
**E2E tests execution on real environment**

- Prepare tests to accept token for tests execution
- Rewrite some tests to pass on real environment


docs/compass/09-01-e2e-test-design.md
```diff
@@ -0,0 +1,47 @@
...
+### Admin test user
+
+* When e2e tests are started they should create the test tenants as they are not created by default in real environments. Perhaps use the tenants that are in the values.yaml. Possibly new ones can be created on the fly, but tests should be adapted to use the newly created tenants
+* Dex static users can be added and used for test purposes. These users (one user?) can be associated with only test tenants, so it doesn't have any permissions on other tenants. However the addition of static users will slightly change the login UI and present the user the option to choose between login with static user or with SAP ID (current).
+
+### Test tenants
+* Test tenants can be loaded by using the tenant loader.
+* The same tenants should be configured in director with the corresponding test admin user.
+
+### Branches with POC
```
NickyMateev, line 39: Do we need this `Branches with POC` section here?

docs/compass/09-01-e2e-test-design.md
```diff
@@ -0,0 +1,47 @@
...
+* After tests are executed the configmap with overrides can be deleted
+* Run installer again which will remove the external services mock
+* `global.externalServicesMock.auditlog: "false"` this is needed because we want to disable the auditlog configurations around the external services mock as it will not be tested in real environment
+* some resources like secrets from external services mock chart should be guarded and not created in case the auditlog is disabled when installing external services mock
+
+
+### Admin test user
+
+* When e2e tests are started they should create the test tenants as they are not created by default in real environments. Perhaps use the tenants that are in the values.yaml. Possibly new ones can be created on the fly, but tests should be adapted to use the newly created tenants
+* Dex static users can be added and used for test purposes. These users (one user?) can be associated with only test tenants, so it doesn't have any permissions on other tenants. However the addition of static users will slightly change the login UI and present the user the option to choose between login with static user or with SAP ID (current).
```
NickyMateev, line 33: Since this is OSS maybe not imply that SAP ID is the current chosen. smth like `between. login with static user or with the currently configured dex connector (ex. SAP ID)`

installation/resources/installer-config-local.yaml.tpl
```diff
@@ -39,6 +39,7 @@ data:
   global.minikubeIP: ""
   global.ingress.domainName: "kyma.local"
   global.externalServicesMock.enabled: "true"
+  global.externalServicesMock.auditlog: "true"
```
NickyMateev, line 42: It's probably not a problem, but I just noticed this:
1. We set auditlog: true here
2. We also set it in two places in the test-infra scripts:
- https://github.com/kyma-project/test-infra/blob/a606c71b5a88cd86e1d05263427488cd51a8f826/prow/scripts/cluster-integration/compass-gke-integration.sh#L236
- https://github.com/kyma-project/test-infra/blob/e3fb15e8805853a02037df4bb57b78bf0f0c31d4/prow/scripts/cluster-integration/compass-gke-benchmark.sh#L237

It's probably not an issue that it's set twice, but I just wanted to point it out, because you're more in the context so you might know something more if it's an issue. (if it's not an issue, we might even go ahead and remove it from test-infra?)

chart/compass/templates/managed-postgresql-secret.yaml
```diff
@@ -14,6 +14,32 @@ metadata:
   name: compass-postgresql
   namespace: {{ .Release.Namespace }}
 type: Opaque
+data:
```
NickyMateev, line 17: Leftover duplication of secrets?

chart/compass/templates/tests/director/director-test.yaml
```diff
@@ -29,7 +29,7 @@ spec:
       containers:
         - name: director-tests
           image: {{ .Values.global.images.containerRegistry.path }}/{{ .Values.global.images.e2e_tests.dir }}compass-tests:{{ .Values.global.images.e2e_tests.version }}
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: Always
```
NickyMateev, line 32: Why is this necessary?

chart/compass/templates/tests/director/director-test.yaml
```diff
@@ -93,20 +93,62 @@ spec:
               value: "https://{{ .Values.global.gateway.tls.secure.oauth.host }}.{{ .Values.global.ingress.domainName }}/director/graphql"
             - name: APP_WEBHOOK_URL
               value: "https://kyma-project.io"
+            - name: APP_DEFAULT_SCENARIO_ENABLED
```
NickyMateev, line 96: Why did we want to enable/disable the Default scenario?

tests/connector/tests/main_test.go
```diff
@@ -40,20 +43,40 @@ var (
...
+	var dexToken string
+
+	tokenConfig := server.Config{}
+	err := envconfig.InitWithPrefix(&tokenConfig, "APP")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Info("Get Dex id_token")
+	if tokenConfig.IsWithToken {
```
NickyMateev, line 60: I see this block on multiple places:
```
if tokenConfig.IsWithToken {
		tokenConfig.Log = log.Infof
		ts := server.New(&tokenConfig)
		dexToken = server.WaitForToken(ts)
	} else {
		dexToken, err = idtokenprovider.GetDexToken()
		if err != nil {
			log.Fatal(errors.Wrap(err, "while getting dex token"))
		}
	}
```

I was wondering if we could hide it behind some sort of abstraction. One idea maybe is to have `clients.NewStaticUserClient` & `clients.NewStaticUserClientWithToken` and the logic hidden behind those two. But not sure we want to hide the server opening inside the `clients` package abstraction.

docs/compass/09-01-e2e-test-design.md
pankrator: I will remove it

installation/resources/installer-config-local.yaml.tpl
pankrator: The configuration from test infra is for the tests deployments and it is needed in order to start the auditlog component of the external-services-mock. This configuration here is for local installations

chart/compass/templates/managed-postgresql-secret.yaml
pankrator: Actually the secret duplication is needed, because tests that are running in kyma-system namespace will need access to those secrets in order for the pods to access the database.

chart/compass/templates/tests/director/director-test.yaml
pankrator: This is a leftover

chart/compass/templates/tests/director/director-test.yaml
pankrator: As discussed, this is needed because e2e tests assume that this is enabled by default. But on real envs this is disabled

chart/compass/templates/managed-postgresql-secret.yaml
NickyMateev: ah yes, I didn't see that the namespace was different. Now it makes sense

NickyMateev (approved):

NickyMateev:
lgtm

NickyMateev (dismissed):

nkanchev (approved):

NickyMateev (approved):

PR #1886, opened 2021-06-07
Set oathkeeper timeouts on Istio level

DimitarPetrov:
**Description**

- Oathkeeper timeouts
- Local overrides refactoring and simplification

Currently, oathkeeper does not have effective timeouts.

The only timeout that is for proxying to the director is the server ones. See [here](https://github.com/ory/oathkeeper/blob/d942c043aa2370b87e0dd822440ad39d809755f9/cmd/server/server.go#L60)

However, server timeouts are enforced once the server tries to return a response. In this case, the server is proxying to the director which can respond much slower than that and this will be identified only when the director responds.

Furthermore oathkeeper reverse proxy uses the default transport. See [here](https://github.com/ory/oathkeeper/blob/d942c043aa2370b87e0dd822440ad39d809755f9/proxy/proxy.go#L91)

This makes it impossible to enforce any effective timeout on the proxying of the oathkeeper.

In order to mitigate this, we add an EnvoyFilter on each oathkeeper sidecar and attach a timeout header which envoy respects on each request.


dzahariev (dismissed):

dzahariev (approved):

chart/compass/charts/connectivity-adapter/templates/virtualservice-mtls.yaml
```diff
@@ -25,6 +25,7 @@ spec:
             {{- range .Values.global.gateway.headers.request.remove }}
             - {{ . }}
             {{- end }}
+      timeout: {{ .Values.global.oathkeeper.timeout }}
```
KirilKabakchiev, line 28: how about adding a destination rule for oathkeeper so that the timeout can be centrally defined on it ?

nkanchev (approved):

installation/cmd/run.sh
```diff
@@ -126,7 +126,7 @@ fi
 
 echo "Label Minikube node for benchmark execution..."
 NODE=$(kubectl get nodes | tail -n 1 | cut -d ' ' -f 1)
-kubectl label node "$NODE" benchmark=true
+kubectl label node "$NODE" benchmark=true || true
```
NickyMateev, line 129: Minor: Why is this necessary?

NickyMateev (approved):

KirilKabakchiev (approved):

installation/cmd/run.sh
DimitarPetrov: If you run installation with --skip-minikube-start the node is already labelled.

PR #1888, opened 2021-06-08
Tenant fetcher deployment customer tenant storage and tenant table remodelling

danielgospodinow:
**Description**

Adapt the Tenant Fetcher deployment to store information about the customer tenant.

- Introduce two fields `parent` and `type` to BusinessTenant model in the persistence layer
- Handle the two new fields in the Tenant Fetcher component
- Introduce migrations which add those fields to the `business_tenant_mappings` table
- Adapt tests

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/schema-migrator/migrations/director/20210608113500_adpot_tenant_customer_id_subdomain.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE business_tenant_mappings ADD COLUMN customerId VARCHAR(256);
```
dzahariev, line 3: customer_id

components/schema-migrator/migrations/director/20210608113500_adpot_tenant_customer_id_subdomain.down.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE business_tenant_mappings DROP COLUMN customerId;
```
dzahariev, line 3: customer_id

components/tenant-fetcher/internal/tenant/service.go
```diff
@@ -62,10 +62,14 @@ func (s *service) Create(writer http.ResponseWriter, request *http.Request) {
 		http.Error(writer, fmt.Sprintf("Property %q not found in body or it is not of String type", s.config.TenantProviderTenantIdProperty), http.StatusInternalServerError)
 		return
 	}
+	customerId := gjson.GetBytes(body, s.config.TenantProviderCustomerIdProperty)
```
dzahariev, line 65: Customer ID is mandatory and is always present - it is better to return error in case it is missing.

components/tenant-fetcher/internal/tenant/service.go
```diff
@@ -62,10 +62,14 @@ func (s *service) Create(writer http.ResponseWriter, request *http.Request) {
...
+	subdomain := gjson.GetBytes(body, s.config.TenantProviderSubdomainProperty)
```
dzahariev, line 66: Subdomain is missing is CIS1 case and you may log here the for corresponding tenant the subdomain was not found.

dzahariev (dismissed):

components/director/internal/domain/tenant/converter.go
```diff
@@ -21,8 +23,13 @@ func (c *converter) ToEntity(in *model.BusinessTenantMapping) *tenant.Entity {
 		ID:             in.ID,
 		Name:           in.Name,
 		ExternalTenant: in.ExternalTenant,
-		ProviderName:   in.Provider,
-		Status:         tenant.TenantStatus(in.Status),
+		Parent: sql.NullString{
+			String: in.Parent,
+			Valid:  true,
```
dzahariev, line 28: should be set to true only if in.Parent in not empty string or nil

Probably the best would be to set defaults for BusinessTenantMapping.Parent to be empty string - "", and later here just set this to len(in.Parent)!=0

components/director/internal/domain/tenant/converter.go
```diff
@@ -21,8 +23,13 @@ func (c *converter) ToEntity(in *model.BusinessTenantMapping) *tenant.Entity {
...
+		},
+		Type:         tenant.Type(in.Type),
```
dzahariev, line 30: Also can be nil for now and we may consider using sql.NullString here too.

components/director/pkg/tenant/entity.go
```diff
@@ -1,19 +1,30 @@
...
-	Name           string       `db:"external_name"`
-	ExternalTenant string       `db:"external_tenant"`
-	ProviderName   string       `db:"provider_name"`
-	Initialized    *bool        `db:"initialized"` // computed value
-	Status         TenantStatus `db:"status"`
+	ID             string         `db:"id"`
+	Name           string         `db:"external_name"`
+	ExternalTenant string         `db:"external_tenant"`
+	Parent         sql.NullString `db:"parent"`
+	Type           Type           `db:"type"`
```
dzahariev, line 10: Type probably also should be sql.NullString

components/director/internal/domain/tenant/converter.go
```diff
@@ -34,6 +41,8 @@ func (c *converter) FromEntity(in *tenant.Entity) *model.BusinessTenantMapping {
 		ID:             in.ID,
 		Name:           in.Name,
 		ExternalTenant: in.ExternalTenant,
+		Parent:         in.Parent.String,
```
dzahariev, line 44: Should check if it is valid and assign "" or in.Parent.String depend on validity.

components/director/internal/domain/tenant/converter.go
```diff
@@ -34,6 +41,8 @@ func (c *converter) FromEntity(in *tenant.Entity) *model.BusinessTenantMapping {
...
+		Type:           string(in.Type),
```
dzahariev, line 45: Also check should be done here, as currently this field can be empty.

components/tenant-fetcher/internal/model/tenant.go
```diff
@@ -4,7 +4,11 @@ import "github.com/kyma-incubator/compass/components/director/pkg/tenant"
 
 type TenantModel struct {
 	ID             string
+	Name           string
 	TenantId       string
-	TenantProvider string
-	Status         tenant.TenantStatus
+	ParentInternal string
```
dzahariev, line 9: Probably ParentID is more meaningful name?

components/tenant-fetcher/internal/model/tenant.go
```diff
@@ -4,7 +4,11 @@ import "github.com/kyma-incubator/compass/components/director/pkg/tenant"
...
+	ParentExternal string
```
dzahariev, line 10: Probably better naming is: ParentExternalID

components/tenant-fetcher/internal/tenant/converter.go
```diff
@@ -14,21 +16,33 @@ func NewConverter() *converter {
...
-		return nil
-	}
-	return &model.TenantModel{
+func (c *converter) FromEntity(in tenant.Entity) model.TenantModel {
+	return model.TenantModel{
 		ID:             in.ID,
+		Name:           in.Name,
 		TenantId:       in.ExternalTenant,
-		TenantProvider: in.ProviderName,
+		ParentInternal: in.Parent.String,
```
dzahariev, line 33: The same check for sql.NullString validity and assign either the value or the "".

components/tenant-fetcher/internal/tenant/converter.go
```diff
@@ -14,21 +16,33 @@ func NewConverter() *converter {
...
+		Type:           in.Type,
```
dzahariev, line 34: Ant the same for the Type (as it may be nil now)

components/tenant-fetcher/internal/tenant/converter.go
```diff
@@ -14,21 +16,33 @@ func NewConverter() *converter {
...
+		Provider:       in.ProviderName,
 		Status:         in.Status,
 	}
 }
+
+func NewNullString(s string) sql.NullString {
+	if len(s) == 0 {
+		return sql.NullString{}
```
dzahariev, line 42: Make it explicit:
sql.NullString{
Valid:false,
}

components/tenant-fetcher/internal/model/tenant.go
danielgospodinow: Yes, it's more meaningful to have the ID suffix. 👍🏼 But is it necessary to remove the `Internal` part since it's a bit more descriptive this way?

components/director/internal/domain/tenant/converter.go
danielgospodinow: Good catch! But I was planning to use [this function](https://github.com/kyma-incubator/compass/pull/1888/files#diff-e4867418df28596ed6cf910ad54f304cc1b560eb25e35a1c327297f6ed90f657R40) instead since it resolves exactly this issues. 😄

components/director/internal/domain/tenant/converter.go
danielgospodinow: I think that `in.Parent.String` is by default `""` if it's a `NULL` in the DB and it's a value different than `NULL` if there's a valid value for this column in the DB. Can there be a case in which `nullstring.String != ""` and `nullstring.Valid == false`?

components/tenant-fetcher/internal/tenant/converter.go
danielgospodinow: I'm not sure whether Type should be a nullable field. How does it sound if we add an additional Type type `empty`/`none` to indicate an empty tenant, so that we enforce this column to be always non-null? For this to work, it would be also necessary to modify the enum in the SQL migration.

components/tenant-fetcher/internal/tenant/converter.go
danielgospodinow: It think it would be `""` as default if it's `NULL` in the DB because the `in.Parent.String` is a `string` object, not a `*string`.

desislavaa (approved):

components/director/internal/domain/tenant/converter.go
```diff
@@ -66,3 +72,16 @@ func (c *converter) MultipleToGraphQL(in []*model.BusinessTenantMapping) []*grap
 
 	return tenants
 }
+
+func newNullString(s string) sql.NullString {
+	if len(s) == 0 {
```
dzahariev, line 77: What about:
```
func newNullString(s string) sql.NullString {
    return sql.NullString{
 			String: s,
 			Valid:  len(s)!=0,
 		}
}
```

components/tenant-fetcher/internal/tenant/converter.go
```diff
@@ -14,21 +16,36 @@ func NewConverter() *converter {
...
+		Status:           in.Status,
 	}
-	return &model.TenantModel{
-		ID:             in.ID,
-		TenantId:       in.ExternalTenant,
-		TenantProvider: in.ProviderName,
-		Status:         in.Status,
+}
+
+func newNullString(s string) sql.NullString {
```
dzahariev, line 40: Code duplication - this function was already available in components/director/internal/domain/tenant/converter.go.

Probably we may consider extracting it and reusing it in common place for both packages like: github.com/kyma-incubator/compass/components/director/pkg/tenant package?

dzahariev (approved):

components/director/internal/domain/tenant/converter.go
dzahariev: Will be addressed in next change

components/tenant-fetcher/internal/tenant/converter.go
dzahariev: Will be addressed in next change

PR #1891, opened 2021-06-09
Make WebhookInput's outputTemplate field mandatory for certain webhook types

karaimin:
**Description**

- Currently it is possible to register/unregister resource in `ASYNC` mode with REGISTER_APPLICATION / UNREGISTER_APPLICATION webhook type. The WebhookInput field `outputTemplate` should be mandatory. `operations-controller` expect it to be available.

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/pkg/graphql/webhook_validation.go
+++ b/components/director/pkg/graphql/webhook_validation.go
@@ -84,11 +84,11 @@ func (i WebhookInput) Validate() error {
 }
 
 func isOutTemplateMandatory(webhookType WebhookType) bool {
-	outTmplMandatoryTypes := map[WebhookType]struct{}{
-		WebhookTypeRegisterApplication:   {},
-		WebhookTypeUnregisterApplication: {},
+	switch webhookType {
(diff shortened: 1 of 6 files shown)


components/director/pkg/graphql/webhook_validation.go
```diff
@@ -78,3 +82,13 @@ func (i WebhookInput) Validate() error {
 		validation.Field(&i.Auth),
 	)
 }
+
+func isOutTemplateMandatory(webhookType WebhookType) bool {
```
desislavaa, line 86: [minor suggestion] what if this check uses `switch`
```suggestion
switch webhookType {
	case WebhookTypeRegisterApplication, WebhookTypeUnregisterApplication:
		return true
	default:
		return false
	}
```

desislavaa (approved):

desislavaa (approved):

PR #1893, opened 2021-06-10
Remove runtime name restriction

desislavaa:
**Description**

- Remove runtime name restriction

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/pkg/inputvalidation/validators_string.go
```diff
@@ -34,6 +36,20 @@ func (v *dnsNameRule) Validate(value interface{}) error {
 	return nil
 }
 
+func (v *runtimeNameRule) Validate(value interface{}) error {
```
DimitarPetrov, line 39: Should we leave the DNS compliance since right now we allow spaces in the name?

components/director/pkg/graphql/runtime_validation_test.go
```diff
@@ -26,9 +26,9 @@ func TestRuntimeInput_Validate_Name(t *testing.T) {
 			ExpectedValid: false,
 		},
 		{
-			Name:          "Invalid - Invalid Name",
+			Name:          "Expected valid with digit",
 			Value:         inputvalidationtest.InvalidName,
-			ExpectedValid: false,
+			ExpectedValid: true,
```
DimitarPetrov, line 31: Change the `inputvalidationtest.InvalidName` to be invalid instead of removing the invalid test.

components/director/pkg/inputvalidation/validators_string.go
desislavaa: and special symbols, and so on... I'm trying to find the place where the runtime name is "randomly generated", so we can tweak aour validation accordingly

desislavaa: I wasn't able to find where the generation is happening, but I saw that in the cockpit, the name can contain underscore, which is not DNS compliant

DimitarPetrov (approved):

components/director/pkg/inputvalidation/validators_string.go
krasish: We should probably still have some validation for the special characters though?

DimitarPetrov: This whole function can be reduced to `validation.Field(&i.Name, validation.Required, validation.RuneLength(1, appNameLengthLimit) validation.Match(runtimeNameRgx))`

desislavaa: Absolutely! Thanks @DimitarPetrov 🍻

DimitarPetrov (approved):

DimitarPetrov (approved):

PR #1894, opened 2021-06-10
Introduce automatic vendor assignment

nyordanoff:
**Description**

Currently it isn't mandatory Vendor resources to be declared in the Documents because there is a concept of a central registry from where Vendors will be fetched once. However, until this registry concept is production ready, we should make sure that if no SAP Vendor resource is explicitly declared across all Documents of a System Instance to assign it once for it so that all resources referencing that SAP Vendor will not fail.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/internal/open_resource_discovery/service_test.go
+++ b/components/director/internal/open_resource_discovery/service_test.go
@@ -22,7 +22,6 @@ func TestService_SyncORDDocuments(t *testing.T) {
 	sanitizedDoc := fixSanitizedORDDocument()
 	var nilSpecInput *model.SpecInput
 	var nilBundleID *string
-	sapVendorTitle := "SAP SE"
 
 	secondTransactionNotCommited := func() (*persistenceautomock.PersistenceTx, *persistenceautomock.Transactioner) {
 		persistTx := &persistenceautomock.PersistenceTx{}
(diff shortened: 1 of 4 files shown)


components/director/internal/open_resource_discovery/service_test.go
```diff
@@ -1058,6 +1101,81 @@ func TestService_SyncORDDocuments(t *testing.T) {
 				return client
 			},
 		},
+		// TODO: Delete the two tests below after the concept of central registry for Vendors fetching is productive
+		{
+			Name: "Success when resources are not in db and no Vendor is declared in Documents should Create them",
```
DimitarPetrov, line 1106: "Success when resources are not in db and no `SAP` Vendor is declared in Documents should Create them"

DimitarPetrov (changes requested):

components/director/internal/open_resource_discovery/service_test.go
```diff
@@ -1058,6 +1101,81 @@ func TestService_SyncORDDocuments(t *testing.T) {
...
+			clientFn: func() *automock.Client {
+				client := &automock.Client{}
+				doc := fixORDDocument()
+				doc.Vendors = nil
+				client.On("FetchOpenResourceDiscoveryDocuments", txtest.CtxWithDBMatcher(), baseURL).Return(open_resource_discovery.Documents{doc}, nil)
+				return client
+			},
+		},
+		{
+			Name: "Success when resources are already in db and no Vendor is declared in Documents should Update them",
```
DimitarPetrov, line 1142: "Success when resources are already in db and no `SAP` Vendor is declared in Documents should Update them"

DimitarPetrov (approved):

DimitarPetrov (approved):

PR #1898, opened 2021-06-14
Validation rule for resource def

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/open_resource_discovery/validation.go
```diff
@@ -937,3 +1060,24 @@ func validateExtensibleInnerFields(el gjson.Result) error {
...
+	return hash, nil
+}
+
+func checkHashEquality(rdFromDBVersion, rdFromDocVersion, hashFromDB, hashFromDoc string) error {
+	rdFromDBVersion = fmt.Sprintf("v%s", rdFromDBVersion)
+	rdFromDocVersion = fmt.Sprintf("v%s", rdFromDocVersion)
+
+	areEventVersionsEqual := semver.Compare(rdFromDocVersion, rdFromDBVersion)
+	if areHashesEqual := cmp.Equal(hashFromDB, hashFromDoc); !areHashesEqual && areEventVersionsEqual <= 0 {
+		return errors.New("there is a change in the resource definitions; version value should be incremented")
```
nyordanoff, line 1079: Probably rephrase as: `there is a change in the resource; ....` because `resource definitions` is a particular field in Apis/Events and the error is a bit misleading as if the change is only in the resource defs.

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -599,7 +600,26 @@ func TestDocuments_ValidatePackage(t *testing.T) {
...
 				return []*open_resource_discovery.Document{doc}
 			},
-		}, {
+		},
+		{
+			Name: "Not incremented `version` field when Package",
+			DocumentProvider: func() []*open_resource_discovery.Document {
+				doc := fixORDDocument()
+				doc.Packages[0].Industry = json.RawMessage(`["Mining"]`)
+				doc.Packages[0].Version = fixPackagesWithLowerVersion()[0].Version
```
nyordanoff, line 609: No need to lower the version value here and in all similar test cases (for Apis/Events).

For further explanation check the comment in the `open_resource_discovery/fixtures_test.go`

components/director/internal/open_resource_discovery/fixtures_test.go
```diff
@@ -608,6 +641,44 @@ func fixBundleCreateInput() []*model.BundleCreateInput {
 	}
 }
 
+func fixAPIsWithLowerVersion() []*model.APIDefinition {
+	apis := fixAPIs()
+
+	for _, api := range apis {
+		api.Version.Value = "1.1.0"
```
nyordanoff, line 648: From my POV this is not necessary because:
    - if we do not lower the version here, the hash below will be with the default version that is defined in the fixture itself; that will be our `*fromDb resource`;
    - in the `resourceHashes` map which represents the `input` resources, the hash will again include the version value defined in the fixture;
    - so, hashes for both `input` and `*fromDB` resources will have the same version value -> then in the test cases where some field will be changed without version bump, the new hash will be re-evaluated and it will see that there are changes but the versions are still equal -> validation error

To, sum it up, we will simplify things by not having to explicitly lower the version field in the test cases -> `doc.<Resource>[0].Version = fixPackagesWithLowerVersion()[0].Version` will be obsolete.

components/director/internal/open_resource_discovery/fixtures_test.go
```diff
@@ -608,6 +641,44 @@ func fixBundleCreateInput() []*model.BundleCreateInput {
 	}
 }
 
+func fixAPIsWithLowerVersion() []*model.APIDefinition {
```
nyordanoff, line 644: Based on the comment below and the removal of the version lowering, these functions can be renamed to `fix<Resource>WithHash()`

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -1493,6 +1519,86 @@ func TestDocuments_ValidateAPI(t *testing.T) {
...
+			AfterTest: func() {
+				resourceHashes = fixResourceHashes()
+			},
+		}, {
+			Name: "Valid incremented `version` field when resource definition has changed for API",
+			DocumentProvider: func() []*open_resource_discovery.Document {
+				doc := fixORDDocument()
+				doc.APIResources[0].ResourceDefinitions[0].Type = model.APISpecTypeCustom
+				doc.APIResources[0].ResourceDefinitions[0].CustomType = "sap:custom-definition-format:v1"
+				doc.APIResources[0].VersionInput.Value = "2.1.3"
```
nyordanoff, line 1597: In all positive test cases for the version bump, it is pointless to just set a bumped version to the given Resource without re-evaluating the hash.

As mentioned in the `fixtures_test.go` comment, the hash comparison is made between the two fixed objects. If you don't override the hash value for the given changed resource in the `resourceHashes` map (and then reset it in the AfterTest func) then the comparison will never "find out" that there was a change.

To sum up, after a change is made and a correct version bump is set, the hash needs to be re-evaluated and the hash map to be overridden. By doing so, you will simulate actual change in the input and the comparison will be made with the changed input object.

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -1493,6 +1519,86 @@ func TestDocuments_ValidateAPI(t *testing.T) {
 
 				return []*open_resource_discovery.Document{doc}
 			},
+		}, {
+			Name: "Not incremented `version` field when resource definition's URL has changed for API",
```
nyordanoff, line 1523: All API/Event tests change the fields in the `resourceDefinitions` field. 

Probably, for completeness sake, add one more test case where some other field is changed and the version is not bumped.

nyordanoff (changes requested):

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -1813,6 +1919,7 @@ func TestDocuments_ValidateAPI(t *testing.T) {
 				doc := fixORDDocument()
 				doc.APIResources[0].ResourceDefinitions = nil
 				doc.APIResources[0].Visibility = str.Ptr(open_resource_discovery.ApiVisibilityPrivate)
+				doc.APIResources[0].VersionInput.Value = "2.1.3"
```
nyordanoff, line 1922: These version bumps are obsolete.

As this and the remaining similar test cases do not test the version bump logic, I think these bumps can be deleted as they do not change anything if the hash is not re-evaluated

components/director/internal/open_resource_discovery/interfaces.go
```diff
@@ -49,12 +49,13 @@ type EventService interface {
 type SpecService interface {
 	CreateByReferenceObjectID(ctx context.Context, in model.SpecInput, objectType model.SpecReferenceObjectType, objectID string) (string, error)
 	DeleteByReferenceObjectID(ctx context.Context, objectType model.SpecReferenceObjectType, objectID string) error
+	ListByReferenceObjectID(ctx context.Context, objectType model.SpecReferenceObjectType, objectID string) ([]*model.Spec, error)
```
nyordanoff, line 52: This is obsolete.

components/director/internal/open_resource_discovery/service.go
```diff
@@ -135,7 +137,68 @@ func (s *Service) processDocuments(ctx context.Context, appID string, baseURL st
...
+		return err
+	}
+
+	resourceHashes := make(map[string]uint64)
+
+	for _, doc := range documents {
+		for _, apiInput := range doc.APIResources {
+			normalizedAPIDef, err := normalizeAPIDefinition(apiInput)
+			if err != nil {
+				return nil
```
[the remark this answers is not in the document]
alextargov: It was a leftover logic. Err should have been returned

components/director/internal/open_resource_discovery/service.go
```diff
@@ -135,7 +137,68 @@ func (s *Service) processDocuments(ctx context.Context, appID string, baseURL st
...
+	packageDataFromDB, err := s.fetchPackagesFromDB(ctx, appID)
+	if err != nil {
+		return err
+	}
+
+	resourceHashes := make(map[string]uint64)
+
+	for _, doc := range documents {
+		for _, apiInput := range doc.APIResources {
+			normalizedAPIDef, err := normalizeAPIDefinition(apiInput)
```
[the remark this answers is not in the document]
alextargov: Yes. So when doing marshal/unmarshal we get consistent array spacings, formatting, etc. Ex: `["val1",  "val2" ]` would be transformed to `["val1","va2"]`

components/director/internal/model/api.go
```diff
@@ -13,9 +16,9 @@ import (
 )
 
 type APIDefinition struct {
-	ApplicationID                           string
-	PackageID                               *string
-	Tenant                                  string
+	ApplicationID                           string  `hash:"ignore"`
```
nyordanoff, line 19: Why are these tags put here as you never hash the "model" struct but only the "input" one? Same for Events/Packages.

From my POV only the `Version` value in the `Input` struct should be ignored.

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -937,3 +1060,24 @@ func validateExtensibleInnerFields(el gjson.Result) error {
...
+
+	return hash, nil
+}
+
+func checkHashEquality(rdFromDBVersion, rdFromDocVersion, hashFromDB, hashFromDoc string) error {
+	rdFromDBVersion = fmt.Sprintf("v%s", rdFromDBVersion)
+	rdFromDocVersion = fmt.Sprintf("v%s", rdFromDocVersion)
+
+	areVersionsEqual := semver.Compare(rdFromDocVersion, rdFromDBVersion)
+	if areHashesEqual := cmp.Equal(hashFromDB, hashFromDoc); !areHashesEqual && areVersionsEqual <= 0 {
```
nyordanoff, line 1078: Here a validation case is missed - if there are no changes to the resource but the version is changed (incremented, decremented). This should be illegal and an error should be thrown.

Example:

	areHashesEqual := cmp.Equal(hashFromDB, hashFromDoc)

	if areHashesEqual && areVersionsEqual != 0 {
		return errors.New("changed version without any resource changes")
	}

	if !areHashesEqual && areVersionsEqual <= 0 {
		return errors.New("there is a change in the resource; version value should be incremented")
	}

components/director/internal/open_resource_discovery/fixtures_test.go
```diff
@@ -615,6 +648,41 @@ func fixBundleCreateInput() []*model.BundleCreateInput {
 	}
 }
 
+func fixAPIsWithHash() []*model.APIDefinition {
+	apis := fixAPIs()
```
nyordanoff, line 652: The `Version` value in the returned Apis from `fixAPIs` should be the same as the `Version` value in the `fixORDDocument`

Same for Events/Packages.

components/director/internal/model/api.go
alextargov: These should have been leftovers as previously we hashed the db model as well for comparison, do you remember

components/director/internal/open_resource_discovery/fixtures_test.go
alextargov: This breaks a lot of tests on the service layer if you meant to change the version in fixAPIs().
If we put the same version only in the scope of this function, then it is okay

nyordanoff (approved):

PR #1905, opened 2021-06-21
Fetch known systems from Systems service

pankrator:
**Fetch known systems from Systems service**

- Parse systems by configuration
- Parallel fetching of systems


components/director/internal/systemfetcher/service.go
```diff
@@ -0,0 +1,195 @@
...
+
+			log.C(ctx).Info(fmt.Sprintf("Successfully synced systems for tenant %s", tenantSystems.tenant.ExternalTenant))
+		}
+	}()
+
+	wg := sync.WaitGroup{}
+	for number, t := range tenants {
+		wg.Add(1)
+		s.workers <- struct{}{}
+		go func(t *model.BusinessTenantMapping, number int) {
```
NickyMateev, line 91: This `number` var seems obsolete

components/director/internal/systemfetcher/service.go
```diff
@@ -0,0 +1,195 @@
...
+	systems []System
+}
+
+func (s *SystemFetcher) SyncSystems(ctx context.Context) error {
+	tenants, err := s.listTenants(ctx)
+	if err != nil {
+		return errors.Wrap(err, "failed to list tenants")
+	}
+
+	systemsQueue := make(chan TenantSystems, 100)
```
NickyMateev, line 69: Maybe 100 should be a constant or config?

components/director/internal/systemfetcher/service.go
```diff
@@ -0,0 +1,195 @@
...
+	wgDB := sync.WaitGroup{}
+	wgDB.Add(1)
+	go func() {
+		defer func() {
+			wgDB.Done()
+		}()
+		for tenantSystems := range systemsQueue {
+			err = s.saveSystemsForTenant(ctx, tenantSystems.tenant, tenantSystems.systems)
```
NickyMateev, line 77: The `saveSystemsForTenant` opens a transaction inside. Is it thread-safe to use `s.transaction.Begin()` multiple times by different goroutines? If yes, then does every one of these transactions open up a new db connection, because if it is the case, this could exhaust our connections

NickyMateev (changes requested):

components/director/internal/systemfetcher/client.go
```diff
@@ -0,0 +1,125 @@
...
+
+type Client struct {
+	apiConfig         APIConfig
+	oAuth2Config      OAuth2Config
+	clientCreatorFunc ClientCreatorFunc
+}
+
+type ClientCreatorFunc func(ctx context.Context, oauth2Config OAuth2Config, scopes []string, tenant string) *http.Client
+
+func DefaultClientCreator(ctx context.Context, oAuth2Config OAuth2Config, scopes []string, tenant string) *http.Client {
```
NickyMateev, line 43: Unused `tenant` arg

NickyMateev: also no need to export func

components/director/internal/systemfetcher/client.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+
+	"github.com/kyma-incubator/compass/components/director/pkg/log"
+	"github.com/pkg/errors"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+var (
+	scopes = []string{"uaa.resource"}
```
NickyMateev, line 19: Seems like this should be a config

components/director/internal/systemfetcher/client.go
```diff
@@ -0,0 +1,125 @@
...
+			tenant: tenant,
+			base:   http.DefaultTransport,
+		},
+	}
+	// The client provided here is used only to get token, not for
+	// the actual request to the system service
+	ctx = context.WithValue(ctx, oauth2.HTTPClient, httpTokenClient)
+
+	httpClient := c.clientCreatorFunc(ctx, c.oAuth2Config, scopes, tenant)
+	httpClient.Timeout = c.apiConfig.Timeout
```
NickyMateev, line 76: Why set the timeout again? It's already configured at the start of the function

components/director/internal/systemfetcher/client.go
```diff
@@ -0,0 +1,125 @@
...
+
+	url := c.apiConfig.Endpoint + "?$filter=" + urlpkg.QueryEscape(c.apiConfig.FilterCriteria)
```
NickyMateev, line 78: Can we have some happy-path logs here for both remote requests?

components/director/internal/systemfetcher/client.go
```diff
@@ -0,0 +1,125 @@
...
+
+func NewClient(apiConfig APIConfig, oAuth2Config OAuth2Config, clientCreatorFunc ClientCreatorFunc) *Client {
+	return &Client{
+		apiConfig:         apiConfig,
+		oAuth2Config:      oAuth2Config,
+		clientCreatorFunc: clientCreatorFunc,
+	}
+}
+
+func (c *Client) FetchSystemsForTenant(ctx context.Context, tenant string) ([]System, error) {
```
NickyMateev, line 63: Can you add a brief comment on this function about what it does and particularly why it seemingly makes the same HTTP remote call twice (but with different query params)?

components/director/internal/systemfetcher/client.go
NickyMateev: On second thought, is it because of the comment above that states the `httpTokenClient` is not used in the actual request to the system service? If so, maybe we can extract the client creation in a subfunction so as to minimize confusion and remove these oauth2 package specifics of bootstrapping and creating an oauth client from the main `FetchSystemsForTenant` function.

NickyMateev (changes requested):

components/director/internal/domain/application/service.go
```diff
@@ -514,6 +538,43 @@ func (s *service) genericCreate(ctx context.Context, in model.ApplicationRegiste
...
+		alreadyExits := false
+
+		for _, a := range allApps {
+			if ai.Name == a.Name && ((ai.SystemNumber == nil && a.SystemNumber == nil) || (ai.SystemNumber != nil && a.SystemNumber != nil && *(ai.SystemNumber) == *(a.SystemNumber))) {
+				alreadyExits = true
+				break
+			}
+		}
+
+		for _, a := range uniqueNonExistingApps {
```
NickyMateev, line 563: If I understand correctly this second loop here is because it might happen so that the external system service returns duplicate system records.

If that's the case I think the code would be easier to read if `uniqueNonExistingApps` was a map[[]model.ApplicationRegInput]boolean (essentially a Set data struct). Then after the loop:

```
if !alreadyExits {
			uniqueNonExistingApps[ai] = true
		}
```

And afterwards just convert it back to an array.
I think it's better that way and it's an implementation detail that we might get duplicate systems, this should not overcomplicate our generic code.

components/director/internal/domain/application/service.go
```diff
@@ -260,6 +260,30 @@ func (s *service) CreateFromTemplate(ctx context.Context, in model.ApplicationRe
...
 }
 
+func (s *service) CreateManyIfNotExistsWithEventualTemplate(ctx context.Context, applicationInputs []model.ApplicationRegisterInput, templateTypes []string) error {
+	appsToAdd, err := s.filterUniqueNonExistingApplications(ctx, applicationInputs)
+	if err != nil {
+		return errors.Wrap(err, "while filtering unique and non-existing applications")
+	}
+
+	for index, a := range appsToAdd {
+		template := templateTypes[index]
```
NickyMateev, line 270: I'm not a fan of this hidden strong coupling between `applicationInputs` and `templateTypes` and then relying on the common `index`.

I suggest grouping them in a map all the way back before passing them to the `CreateManyIfNotExistsWithEventualTemplate` function instead of as two separate arguments to the function.

Furthermore, this coupling will immediately brake if there are duplicate systems as the index will mismatch after the `filterUniqueNonExistingApplications` functions runs.

NickyMateev (changes requested):

components/director/internal/systemfetcher/model.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/tidwall/gjson"
+)
+
+var (
+	// Mappings global static configuration which is set after reading the configuration during startup, should only be used for the unmarshaling of system data
+	Mappings []TempMapping
+)
+
+type TempMapping struct {
```
NickyMateev, line 14: Can you rename this to `TemplateMapping`, because it otherwise sounds like a temporary struct for holding things?

components/director/internal/systemfetcher/model.go
```diff
@@ -0,0 +1,58 @@
...
+	Name        string
+	SourceKey   []string
+	SourceValue []string
+}
+
+type AdditionalUrls map[string]string
+
+type AdditionalAttributes map[string]string
+type SystemBase struct {
```
NickyMateev, line 23: My OCD spidey sense tells me there should be an empty new line between `AdditionalAttributes` and `SystemBase` 😄

components/director/internal/systemfetcher/model.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/tidwall/gjson"
+)
+
+var (
+	// Mappings global static configuration which is set after reading the configuration during startup, should only be used for the unmarshaling of system data
+	Mappings []TempMapping
```
NickyMateev, line 11: `TemplateMappings` ?

NickyMateev: Maybe also add to the godoc what its purpose, because it is a little controversial how we bootstrap it with the global variable 😄

NickyMateev (changes requested):

components/director/cmd/systemfetcher/main.go
```diff
@@ -0,0 +1,166 @@
+package main
```
NickyMateev, line 1: I may have missed it and it could have been discussed already - but what about e2e tests?

chart/compass/templates/system-fetcher-job.yaml
```diff
@@ -0,0 +1,165 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ $.Chart.Name }}
+    release: {{ $.Release.Name }}
+    helm.sh/chart: {{ $.Chart.Name }}-{{ $.Chart.Version | replace "+" "_" }}
+    app.kubernetes.io/name: {{ template "name" . }}
+    app.kubernetes.io/managed-by: {{ $.Release.Service }}
+    app.kubernetes.io/instance: {{ $.Release.Name }}
+data:
+  system-information-api-endpoint: "{{ .Values.global.systemFetcher.systemsAPIEndpoint }}"
```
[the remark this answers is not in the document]
pankrator: No particular reason for this, just left it as it was from the previous branch. But you are right that it adds a level of indirection which is not necessary and the information is not shared with anyone else

chart/compass/values.yaml
```diff
@@ -335,6 +335,25 @@ global:
       maxOpenConnections: 2
       maxIdleConnections: 2
 
+  systemFetcher:
+    enabled: true
```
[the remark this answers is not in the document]
pankrator: leftover

chart/compass/values.yaml
```diff
@@ -335,6 +335,25 @@ global:
...
+    name: "system-fetcher"
+    schedule: "0 0 * * *"
```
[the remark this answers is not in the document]
pankrator: I think this should be fine, as it is once a day at midnight - https://crontab.guru/#0_0_*_*_*
The job is very slow at the moment, about 30-40 minutes.

components/director/cmd/systemfetcher/main.go
```diff
@@ -0,0 +1,166 @@
...
+func main() {
+	cfg := config{}
+	err := envconfig.InitWithPrefix(&cfg, "APP")
+	if err != nil {
+		log.D().Fatal(errors.Wrap(err, "failed to load config"))
+	}
+	if err = json.Unmarshal([]byte(cfg.SystemToTemplateMappingsString), &cfg.SystemToTemplateMappings); err != nil {
+		log.D().Fatal(errors.Wrap(err, "failed to read system template mappings"))
+	}
+	systemfetcher.Mappings = cfg.SystemToTemplateMappings
```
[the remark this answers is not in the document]
pankrator: Unfortunately, the global configuration could not be avoided :(

components/director/internal/domain/application/service.go
```diff
@@ -514,6 +538,43 @@ func (s *service) genericCreate(ctx context.Context, in model.ApplicationRegiste
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "while listing all applications for tenant %s", appTenant)
+	}
+
+	var uniqueNonExistingApps []model.ApplicationRegisterInput
+	for _, ai := range applicationInputs {
+		alreadyExits := false
+
+		for _, a := range allApps {
+			if ai.Name == a.Name && ((ai.SystemNumber == nil && a.SystemNumber == nil) || (ai.SystemNumber != nil && a.SystemNumber != nil && *(ai.SystemNumber) == *(a.SystemNumber))) {
```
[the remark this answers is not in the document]
pankrator: On the second comment, there shouldn't be any systems without a system number, but the check is for completeness.

components/director/internal/systemfetcher/auth.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "net/http"
+
+type HeaderTransport struct {
+	base   http.RoundTripper
+	tenant string
+}
+
+func (ht *HeaderTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("X-zid", ht.tenant)
```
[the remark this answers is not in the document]
pankrator: Perhaps would be better yes.

components/director/internal/systemfetcher/client.go
pankrator: It needs to be exported, as it is used from main pkg. I think the tenant argument is used in tests to set different tenants

components/director/internal/systemfetcher/client.go
```diff
@@ -0,0 +1,125 @@
...
+
+	url := c.apiConfig.Endpoint + "?$filter=" + urlpkg.QueryEscape(c.apiConfig.FilterCriteria)
+	systems, err := fetchSystemsForTenant(ctx, httpClient, url)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to fetch systems from %s", url)
+	}
+
+	tenantFilter := fmt.Sprintf(c.apiConfig.FilterTenantCriteriaPattern, tenant)
+	url = c.apiConfig.Endpoint + "?$filter=" + urlpkg.QueryEscape(tenantFilter)
+	systemsByTenantFilter, err := fetchSystemsForTenant(ctx, httpClient, url)
```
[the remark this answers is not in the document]
pankrator: I agree that this is not the best solution at the moment

components/director/internal/systemfetcher/service.go
pankrator: You are correct to some extent. It does start a new db connection and can possibly exhaust them, but it actually does not run multiple times in parallel. So to say it should be safe to use as long as you don't run it in parallel.

The new goroutine is only running one `saveSystemsForTenant` at a time, but it does start a new connection each time after it has finished the previous processing loop and closed(commited/rolled back) the previous connection.

components/director/internal/systemfetcher/service.go
pankrator: leftover master🥇

components/director/internal/systemfetcher/service.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	defer s.transaction.RollbackUnlessCommitted(ctx, tx)
+
+	ctx = tenant.SaveToContext(ctx, tenantMapping.ID, tenantMapping.ExternalTenant)
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	for _, system := range systems {
+		appInput, templateType := s.convertSystemToAppRegisterInput(system)
+		template, err := s.appTemplateService.GetByName(ctx, templateType)
+		if err != nil && !apperrors.IsNotFoundError(err) {
```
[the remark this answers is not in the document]
pankrator: Perhaps an error will tell us that we have a non-existing template mapping. As a second thought this can be done at the startup and convert once the template mapping to IDs and there will no need to get them one by one here...

components/schema-migrator/migrations/director/20220609403210_system_number.up.sql
```diff
@@ -0,0 +1,43 @@
+BEGIN;
+
+ALTER TABLE applications
+    ADD COLUMN system_number VARCHAR(256) NULL;
+
+ALTER TABLE applications DROP CONSTRAINT application_tenant_id_name_unique;
+
+ALTER TABLE applications
+    ADD CONSTRAINT application_tenant_id_name_unique UNIQUE (tenant_id, name, system_number);
```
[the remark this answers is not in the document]
pankrator: At least from what we have as information until now, yes

components/schema-migrator/migrations/director/20220609403210_system_number.up.sql
```diff
@@ -0,0 +1,43 @@
...
+    'CREATING',
+    'CREATE_FAILED',
+    'CREATE_SUCCEEDED',
+    'UPDATING',
+    'UPDATE_FAILED',
+    'UPDATE_SUCCEEDED',
+    'DELETING',
+    'DELETE_FAILED',
+    'DELETE_SUCCEEDED'
+    'MANAGED'
```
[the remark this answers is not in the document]
pankrator: This seems like something that needs to be explored a bit more. But to me it seems like that`MANAGED` status is abused.

components/schema-migrator/migrations/director/20220609403210_system_number.up.sql
```diff
@@ -0,0 +1,43 @@
...
+    ALTER COLUMN status_condition TYPE application_status_condition
+        USING status_condition::text::application_status_condition;
+
+ALTER TABLE applications
+    ALTER COLUMN status_condition
+        SET DEFAULT 'INITIAL' ::application_status_condition;
+
+DROP TYPE application_status_condition_old;
+
+ALTER TABLE applications ALTER column "name" TYPE varchar(256);
```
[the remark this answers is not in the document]
pankrator: I will further check this, good to point it out.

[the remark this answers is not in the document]
pankrator: `NOT NULL` still stays after the migration, so it should be ok without specifying it here.

components/director/internal/systemfetcher/client.go
pankrator: I saw what you mean :D. Made the function type private.

components/director/internal/domain/application/service.go
pankrator: cannot use `model.ApplicationRegInput` as key, but perhaps a simpler key containing only the searched properties can be used. Only `systemNumber`, `Name` and `tenant`

components/schema-migrator/migrations/director/20220609403210_system_number.up.sql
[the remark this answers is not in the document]
pankrator: As discussed label will be used to distinguish managed systems

components/director/internal/systemfetcher/client.go
```diff
@@ -0,0 +1,133 @@
...
+	return &Client{
+		apiConfig:     apiConfig,
+		oAuth2Config:  oAuth2Config,
+		clientCreator: clientCreator,
+	}
+}
+
+// FetchSystemsForTenant fetches systems from the service by making 2 HTTP calls with different filter criteria
+func (c *Client) FetchSystemsForTenant(ctx context.Context, tenant string) ([]System, error) {
+	ctx, httpClient := c.clientForTenant(ctx, tenant)
```
danielgospodinow, line 63: We can think of a way to avoid creating a new HTTP client for each tenant to avoid using too much unnecessary memory. Maybe we can use a single HTTP client here in the SystemFetcher `client.go` and we can use `clientcredentials.Config{...}.Token(...)` to fetch only tokens for each tenant and just update the `Authorization` header of the SystemFetcher's HTTP client.

nkanchev (dismissed):

NickyMateev (approved):

nkanchev (approved):

PR #1906, opened 2021-06-22
Fix oathkeeper token cache

DimitarPetrov:
### Problem

id_token mutator from oathkeeper maintains a token cache: https://github.com/ory/oathkeeper/blob/d942c043aa2370b87e0dd822440ad39d809755f9/pipeline/mutate/mutator_id_token.go

With the previous release, we updated the TTL of the token to 1h in order to mitigate interference with all the retry chain bugs identified. However, it looks like that the oathkeeper uses the same TTL for the token cache.

In addition, kyma have a job - https://github.com/kyma-project/kyma/blob/main/resources/ory/charts/oathkeeper/templates/job.yaml which rotates all the JWKs keys on helm upgrade.

So what happens is:
Me as a customer (or a runtime agent) perform a call to CMP:
1. Oathkeeper performs tenant-mapping and auth-mapping and then crafts an id_token
2. Caches this id_token for ttl duration - 1h
3. CMP CD pipepline installs new compass version
4. JWKs are rotated by the kyma post-upgrade job
5. All the tokes cached for 1h with the same set of claims are invalid. Therefore all the calls for a given tenant of a given client are failing.

This is especially bad for the runtimes, since they periodically do some kind of resyncs. This way they always will have a token cached and will fail for up to 1h after upgrade.

### Solution

Add a nonce value in the claims. For example we can add the time.Now().UTC() in the claims in order to guarantee that each token will have unique claims effectively disabling the cache.

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -53,7 +53,7 @@ global:
       version: "PR-1884"
     director:
       dir:
-      version: "PR-1889"
+      version: "PR-1906"
     gateway:
       dir:
       version: "PR-1884"
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -269,7 +269,8 @@ global:
     port: 4455
     timeout_ms: 120000
     idTokenConfig:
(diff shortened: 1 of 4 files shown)


KirilKabakchiev:
Please hold off on this one. Apart from the extra memory, there seems to be a performance degradation due to the signing methods used when disabling the cache. https://github.com/ory/oathkeeper/issues/364

DimitarPetrov:
Another approach was taken.

PR #1908, opened 2021-06-22
Introduce panic handler

DimitarPetrov:
## Description

We need a generic panic handler middleware in order to guarantee that panics return proper response if unrecovered.


NickyMateev (approved):

krasish (approved):

components/director/pkg/panic_recovery/handler.go
```diff
@@ -0,0 +1,35 @@
...
+	"encoding/json"
+	"net/http"
+	"runtime/debug"
+
+	"github.com/kyma-incubator/compass/components/director/pkg/apperrors"
+	"github.com/kyma-incubator/compass/components/director/pkg/log"
+)
+
+// NewPanicRecoveryMiddleware returns a standard mux middleware that provides panic recovery
+func NewPanicRecoveryMiddleware() func(http.Handler) http.Handler {
```
KirilKabakchiev, line 13: Could you remove the one from system broker as well ?

DimitarPetrov: It uses a specific broker error and I thought it is better to be consistent with the specific domain errors. Broker errors in the broker and apperrors in the director.

PR #1911, opened 2021-06-23
Use suggested token when pairing applications

desislavaa:
**Description**

- Replace the returned token for `requestOneTimeTokenForApplication` mutation with the one that should be used, depending on the application type:
  - legacy applications (with "legacy" label): token = legacyConnectorURL
  - applications, managed by integration system: token = token (no change)
  - basic applications: token = rawEncoded

_Approach:_
In order to provide backwards compatability, the new behaviour is observed only when a "suggest_token" header is present.
This functionality could not be moved in a "label" resolver,  because the app ID is already gone.

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/onetimetoken/service.go
```diff
@@ -218,3 +244,44 @@ func (s *service) getTokenFromAdapter(ctx context.Context, adapterURL string, ap
...
 	}, nil
 }
+
+func (s *service) getSuggestedTokenForApp(ctx context.Context, app *model.Application, oneTimeToken *model.OneTimeToken) string {
+	if !tokenSuggestionEnabled(ctx, s.suggestTokenHeaderKey) {
+		return oneTimeToken.Token
+	}
+
+	if app.IntegrationSystemID != nil {
+		log.C(ctx).Infof("Application belongs to an integration system, will use the actual token")
```
desislavaa, line 254: All of these -> Debug

NickyMateev: Whatever we choose, the log should include the app guid at least

components/director/internal/domain/onetimetoken/token.go
```diff
@@ -0,0 +1,86 @@
...
+	reqHeaders, ok := ctx.Value(header.ContextKey).(http.Header)
+	if !ok || reqHeaders.Get(suggestTokenHeaderKey) != "true" {
+		return false
+	}
+
+	log.C(ctx).Infof("Token suggestion is required by client")
+	return true
+}
+
+func extractToken(currentToken string) string {
```
NickyMateev, line 68: Can you add godoc to this func, because otherwise we have to read it to understand what it's purpose is with this generic name

NickyMateev (commented):
Only thing I still can't figure out is why we call it "suggested" 😄

NickyMateev (approved):

NickyMateev (approved):

NickyMateev (approved):

PR #1913, opened 2021-06-24
Do not allow Adapter to make two services having the same display names

franpog859:
**Description**

Connectivity Adapter works among others like the old Application Registry API. Some time ago there was [a change to the Application Registry API regarding the validation of service display names](https://github.com/kyma-project/kyma/pull/11295/files#diff-43ee20121ee1c6cbaaf8a4cec3e4e11992be7224c73b5520b51e211a113f7569). This was done due to the introduction of the Central Application Gateway which operates on the display names. Therefore, the Adapter should work in a similar way. In the Compass scenario, the service display name is the [API Package name](https://github.com/kyma-project/kyma/blob/main/components/compass-runtime-agent/internal/kyma/applications/converter.go#L85). [API Package is actually an API Bundle](https://github.com/kyma-incubator/compass/commit/db75220d5d942e03bf55b7c24569031c87b1c538#diff-ea6eb741146d2eaf8065b51a83d212dc8aeb5bcbd05c5bde61dd235e0e338c36), so here we need to check the normalized API Bundle names

- Do not allow the Connectivity Adapter to register two services with the same names after the normalization. Do not allow also such an update

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/10361

**Pull Request status**

- [ ] Implementation and unit tests
- [ ] Chart changes
- [ ] Automatic or manual integration test
- [ ] End to end test with the real integration systems


components/connectivity-adapter/internal/appregistry/service/handler.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/kyma-incubator/compass/components/connectivity-adapter/pkg/res"
 	"github.com/kyma-incubator/compass/components/director/pkg/graphql"
 	"github.com/kyma-incubator/compass/components/director/pkg/log"
+	"github.com/kyma-project/kyma/components/application-operator/pkg/normalization"
```
everesio, line 14: I am not sure if code owner will be happy that code dependencies between kyma and compass is added.

franpog859: True. Do you suggest duplicating this package and pasting it also into this repository? What do you think about it, @kyma-incubator/cmp_team ?

franpog859:
It looks like we should not merge this change. I'm closing this PR. If you are interested in this topic, go to [the follow-up issue](https://github.com/kyma-project/kyma/issues/11609)

PR #1914, opened 2021-06-28
Async health checks

ataleksandrov:
1. `/livez` always return 200 OK and is used during the livenessProbe
2. Execute health checks as background jobs and return the last calculated state
     * All checks are fatal
     * Only db check is added for now
     * ` /healthz` returns OK/DOWN, but logs have rich information needed for troubleshooting
     * Indicator config respects - interval, timeout, initialDelay and threshold


components/director/internal/healthz/indicator.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	if i.timeout <= 0 {
+		return errors.New("indicator timeout could not be <= 0")
+	}
+	if i.initialDelay < 0 {
+		return errors.New("indicator initial delay could not be < 0")
+	}
+	if i.threshold < 0 {
+		return errors.New("indicator threshold could not be < 0")
+	}
```
desislavaa, line 97: Can this snippet be part of the configure method?

ataleksandrov: I moved it to the `New` health function. This way we could pipe multiple `RegisterIndicator` calls.

desislavaa (approved):

PR #1923, opened 2021-06-30
Enable periodic full resync on tenant-fetcher jobs

DimitarPetrov:
## Description

Currently, tenant-fetcher caches the last run timestamp and fetches only newer events on the next run. This way if for whatever reason we miss an event we will never try to fetch it again until it is gone.

A new periodic full resync was introduced in order to mitigate such situations.


KirilKabakchiev (approved):
Discussed other approaches, lets stay with this one for now as its already implemented

KirilKabakchiev (approved):

PR #1926, opened 2021-07-01
Expose healthz API

ataleksandrov:
**Description**

- Expose director healthz API
- Enhance indicator logging
- Clean-up e2e test definitions

- [X] Implementation
- [X] Integration tests


chart/compass/charts/director/templates/virtual-service.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: networking.istio.io/v1alpha3
```
danielgospodinow, line 1: Can we instead expose the Director health APIs through the Gateway? Since currently all calls towards the Director go through the Gateway, we can just stay consistent. And something more important, exposing a Director API directly, we increase our potential attack surface.

ataleksandrov: Going through `compass-gateway.<domain>/director/healthz` could be a bit misleading as you would expect the call to go first - compass gateway component. IMO the call should be directly to the Director, this way we bypass some sporadic network error, that may occur. Moreover there is a `connector.<domain>/healthz` that already exists.

danielgospodinow (approved):

PR #1927, opened 2021-07-01
Refetch failed specs in ORD Aggregator

DimitarPetrov:
**Description**

Currently, If fetch request fails to pull the spec data for a particular API/Event when the ORD aggregator is fetching all the document information, then the API/Event spec fetching will not be retried, even after the aggregator restarts and tries to refetch the ORD documents again.

**Fix**

Refetch all specs which have fetch request in a failed state.


NickyMateev (commented):
LGTM 🔨
This does seem to me like a good candidate for an e2e test, though

NickyMateev (dismissed):

NickyMateev (approved):
🔨

NickyMateev (approved):

PR #1933, opened 2021-07-01
Add temporary managed label

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/domain/application/resolver.go
```diff
@@ -264,6 +264,9 @@ func (r *Resolver) RegisterApplication(ctx context.Context, in graphql.Applicati
 	log.C(ctx).Infof("Registering Application with name %s", in.Name)
 
 	convertedIn, err := r.appConverter.CreateInputFromGraphQL(ctx, in)
+
+	convertedIn.Labels["managed"] = false
```
krasish, line 268: Maybe do this after check for err? I personally got confused while reading this part and thought that we swallow an error on line 266.

tests/pkg/assertions/assertions.go
```diff
@@ -34,6 +34,9 @@ func AssertApplication(t *testing.T, in graphql.ApplicationRegisterInput, actual
 //TODO: After fixing the 'Labels' scalar turn this back into regular assertion
 func AssertLabels(t *testing.T, in graphql.Labels, actual graphql.Labels, app graphql.ApplicationExt) {
 	appNameNormalizier := normalizer.DefaultNormalizator{}
+
+	in["managed"] = false
```
desislavaa, line 38: This will mess up Nicky Stoykov's test, let's add a check if the label does not exist, and only then add the "false" value

desislavaa (approved):

desislavaa (approved):

PR #1939, opened 2021-07-06
Use kibana formatter

StefanCholakov:
- Add Kibana formatter for logging


components/director/pkg/log/formatter.go
```diff
@@ -0,0 +1,79 @@
...
+	return json.Marshal(result)
+}
+
+// KibanaFormatter is a logrus formatter that formats an entry for Kibana
+type KibanaFormatter struct {
+}
+
+// Format formats a logrus entry for Kibana logging
+func (f *KibanaFormatter) Format(e *logrus.Entry) ([]byte, error) {
+	// todo remove
```
NickyMateev, line 52: todo?

NickyMateev (commented):
Overall lgtm, but we need to understand what causes Kibana not to provide the json properties as queryable fields sometimes 🤔

chart/compass/charts/connectivity-adapter/templates/deployment.yaml
```diff
@@ -51,6 +51,8 @@ spec:
               value: "https://{{ .Values.global.connectivity_adapter.tls.host }}.{{ .Values.global.ingress.domainName }}"
             - name: APP_CONNECTOR_ADAPTER_MTLS_BASE_URL
               value: "https://{{ .Values.global.connectivity_adapter.mtls.host }}.{{ .Values.global.ingress.domainName }}"
+            - name: APP_LOG_FORMAT
+              value: "kibana"
```
NickyMateev, line 55: Maybe it would be better to introduce in values.yaml a property called `log.format` which will contain the default log settings (only format for now) and reference this here and everywhere else where `kibana` is hardcoded.

This way if we want to switch the default log format we'll only have to do it in one place.

chart/compass/values.yaml
```diff
@@ -138,14 +138,17 @@ global:
     suggestTokenHeaderKey: suggest_token
 
     tests:
-      scopes: "runtime:write application:write label_definition:write integration_system:write application:read runtime:read label_definition:read integration_system:read health_checks:read application_template:read application_template:write eventing:manage tenant:read automatic_scenario_assignment:read automatic_scenario_assignment:write"
+      scopes: "runtime:write application:write label_definition:write integration_system:write application:read runtime:read label_definition:read integration_system:read healthhealth_checks:read application_template:read application_template:write eventing:manage tenant:read automatic_scenario_assignment:read automatic_scenario_assignment:write"
```
NickyMateev, line 141: `healthhealth_checks:read` typo mistake?

NickyMateev (dismissed):

components/director/pkg/log/formatter_test.go
```diff
@@ -0,0 +1,80 @@
+package log
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
DimitarPetrov, line 10: Do we want to introduce a new testing framework for just a few tests? 

I think it is better to rewrite the tests with the standard go testing framework for consistency.

desislavaa (approved):

PR #1940, opened 2021-07-06
Tenant fetcher server no-op delete

danielgospodinow:
**Description**

Make the tenant fetcher deployment's delete API no-op. The tenant fetcher cronjob will manage the deletion of tenants.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/tenant-fetcher/internal/tenant/service.go
```diff
@@ -22,12 +21,11 @@ import (
 )
 
 const (
-	genericDatabaseError               = "Something went wrong with our database"
-	parentTenantConfigureDatabaseError = "Something went wrong while configuring parent tenant"
-	tenantCreateDatabaseError          = "Something went wrong while creating tenant"
-	tenantGetDatabaseError             = "Something went wrong while getting tenant"
-	tenantUpdateDatabaseError          = "Something went wrong while updating tenant"
-	tenantDeleteDatabaseError          = "Something went wrong while deleting tenant"
+	GenericDatabaseError               = "Something went wrong with our database"
```
danielgospodinow, line 24: Move those error constants to `components/tenant-fetcher/internal/consts/errors.go`? Or nah?

krasish: Using the standard Go library as an example this seems to be a practice in the Go language since built-in packages like [io](https://golang.org/pkg/io/#pkg-variables) and [bufio](https://golang.org/pkg/bufio/#pkg-variables) declare their errors similarly. It is a little strange that we declare them as Strings (instead of errors) but I think it is ok since we only use them as string constants and not errors. Moreover, constants are usually also declared in the same fashion anyways. I would suggest leaving them as they are.

krasish (approved):

PR #1946, opened 2021-07-08
Introduce hierarchical tenancy based on CRM ID

DimitarPetrov:
**Description**

- Hierarchical tenant Isolation based on CRM ID and its child "account" tenants implemented as recursive SQL query in the generic repo layer
- GraphQL middleware that intercepts mutation fields in order to find the right tenant to use for the mutation (the owner tenant of the requested resource).
- DB view acting as index containing all the entity ids with their owning tenant. It is dynamically updated with the data from each table in the public schema containing 'id' and 'tenant_id' columns.
- Refactor various entities (labels and bundles) using raw SQL queries to use the generic repo layer instead.
- Improve tenant-loader to insert the tenants provided in the right order based on parent relationship. (insert parent before child)

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/repo/exist.go
```diff
@@ -41,7 +41,7 @@ func (g *universalExistQuerier) Exists(ctx context.Context, tenant string, condi
 	if tenant == "" {
 		return false, apperrors.NewTenantRequiredError()
 	}
-	conditions = append(Conditions{NewEqualCondition(*g.tenantColumn, tenant)}, conditions...)
+	conditions = append(Conditions{NewTenantIsolationCondition(*g.tenantColumn, tenant)}, conditions...)
```
ataleksandrov, line 44: Baking the tenant isolation logic in the repo layer seems reasonable, but I have concerns.. Most of our queries and mutations do more than one db call. Take for example `appsForRuntime`, we make 3 db calls (runtime exists, getScenarios and list by scenario), this will result in executing the recursive sql 3 times on `appsForRuntime`.  I would suggest creating a separate middleware that executes the recursive query once and stores the tenant ids in the context, so that we could later add `tenant_id in (?,?,...,?)` conditions in the repo layer.

components/director/internal/domain/document/fixtures_test.go
```diff
@@ -154,3 +156,23 @@ func timeToTimestampPtr(time time.Time) *graphql.Timestamp {
...
+
+func fixUpdateTenantIsolationSubquery() string {
+	return regexp.QuoteMeta(`tenant_id IN ( with recursive children AS (SELECT t1.id, t1.parent FROM business_tenant_mappings t1 WHERE id = ? UNION ALL SELECT t2.id, t2.parent FROM business_tenant_mappings t2 INNER JOIN children t on t.id = t2.parent) SELECT id from children )`)
+}
+
+func fixTenantIsolationSubquery() string {
+	return fixTenantIsolationSubqueryWithArg(1)
+}
+
+func fixUnescapedTenantIsolationSubquery() string {
```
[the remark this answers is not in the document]
DimitarPetrov: I thought about removing them but decided that those functions are small enough (and hard enough to figure out without context) to leave them as is. Furthermore, this tenant isolation query will not change with adding more and more tenant types, therefore it is probably useful to have those utility fixtures when writing new tests in any repository and don't think about it.

If you think it is better to remove them and leave only the used ones I will do so, however, I think it will be easier for further repository tests which potentially can have those requirements.

components/director/internal/repo/exist.go
DimitarPetrov: Discussed offline and decided that there are optimizations coming (dataloaders) that will make the performance implications of this change irrelevant. However, it is much better from design PoV to leave it in the repo layer as if we move it to a central place the repository layer is not self-contained and cannot be used without the said middleware.

ataleksandrov (approved):

PR #1948, opened 2021-07-12
Down migrations

desislavaa:
**Description**

If an error occurs while a new version of compass is being deployed, the Compass installer initiates Helm rollback of the release. However, if there were any new incompatible migrations, the roll-back won't be possible.

- Add a new pre-rollback K8s job for executing down migrations.
Unfortunately, we cannot rely on “pre-rollback” hooks, because the version of the migration job won’t have the old migrations - the “pre-rollback” hook uses the target release (the one we’re rolling back to) - [Pre-rollback hooks execution  · Issue #5825 · helm/helm · GitHub](https://github.com/helm/helm/issues/5825)
- Not all cloud providers support `ReadWriteMany` access mode for PVs, hence we'll have to stick with `ReadWriteOnce` - that means that only one node can be attached to the volume, and we have to ensure that all migration pods are running on the same node by adding `nodeSelector`, and labeling the node we want to use.

**Related PRs(s)**
- https://github.com/kyma-project/test-infra/pull/3948

**Local Validation**
1. Install Compass from this branch
2. Add a new migration that will break the new version, e.g:
```sql
ALTER TABLE api_definitions
    DROP COLUMN id;
```
Build an image of `schema-migrator`, or use this one - `docker.io/dessislava/schema-migrator:final-v2-fail`, and trigger the installation again:
```bash
./installation/cmd/run.sh --skip-minikube-start --skip-kyma-start --docker-driver
```
3. Wait for the rollback, and see that the version will be rollbacked successfully, and Compass is working as expected - that might take a while, because Helm retries immediately after the rollback (5 retries in total, and the `compass-migration` job has default `backOffLimit` of `6`). My suggestion - break the installation, and come back to check the state after ~1h.

Here's the console output of such roll-backed release:
image

Compass is working as expected:
image

**Notes**
- We need the `"helm.sh/resource-policy": keep` annotation of the PVC because if the installation of this version, that introduces the PVC for the first time fails, Helm will try to delete the PVC, and won't be able to do so, because it's still used by the `compass-migration` job. If we manually delete the job, the PVC deletion will be successful, however, it is required by the pre-rollback `compass-migration-down`, and we'll get ourselves into a dead-lock situation.
- There are two very important checks:
  - Migration actually exists - if that check is missing, we'll drop the whole DB
  - If the migration is for a clean installation - when the `compass-migration` job runs for the first time, there's no valid version that can be saved in the config map, that holds the last successful migration. In that case we save a "clean" version in the CM - if the version is "clean", then the DB can be safely dropped.
- The `compass-migration` job is not executed during the rollback (it does not have a hook for that).

**Pull Request status**

- [x] Implementation


desislavaa:
We need to merge the PR in test-infra first

desislavaa:
Validation once again:
image

pankrator (dismissed):

NickyMateev (approved):

nkanchev (approved):

chart/compass/templates/migrator-down-job.yaml
```diff
@@ -0,0 +1,99 @@
...
+            - name: DB_SSL
+              valueFrom:
+                secretKeyRef:
+                  name: compass-postgresql
+                  key: postgresql-sslMode
+            - name: MIGRATION_PATH
+              value: "director"
+            - name: DIRECTION
+              value: "down"
+            - name: CM_NAME
```
KirilKabakchiev, line 77: do we need these or are they left over ? CM_NAME + CM_NAMESPACE

chart/compass/templates/migrator-down-job.yaml
```diff
@@ -0,0 +1,99 @@
...
+                  name: compass-postgresql
+                  key: postgresql-director-db-name
+            - name: DB_SSL
+              valueFrom:
+                secretKeyRef:
+                  name: compass-postgresql
+                  key: postgresql-sslMode
+            - name: MIGRATION_PATH
+              value: "director"
+            - name: DIRECTION
```
KirilKabakchiev, line 75: does it make sense that this direction is actually an arg to the script ? it would be easier to use (run.sh --up, run.sh --down), also the "migration_path" - run.sh --up -component director

KirilKabakchiev (commented):
minor comments, you may ignore them as well

chart/compass/templates/migrator-down-job.yaml
desislavaa: Leftovers, thanks!

desislavaa:
Validated once again:
image

nkanchev (approved):

KirilKabakchiev (approved):

PR #1950, opened 2021-07-13
Delete systems in lifecycle status deleted

pankrator:
**Delete systems in lifecycle status deleted**

- ...

- [x] Implementation
- [ ] Unit tests
- [x] Integration tests


chart/compass/values.yaml
```diff
@@ -374,6 +374,8 @@ global:
     queueSize: 100
     # fetchTimeout - shows the timeout to wait for oauth token and for fetching systems (in one request) separately
     fetchTimeout: "5s"
+    # directorRequestTimeout - graphql requests timeout to director
+    directorRequestTimeout: "40s"
```
NickyMateev, line 378: Minor: Is `40s` something specific or just random? We usually set something that's a multiple of `30s`

Also maybe rename `fetchTimeout` to `fetchRequestTimeout` so that the configuration naming is consistent.

components/director/pkg/http/error_transport_test.go
```diff
@@ -7,10 +7,9 @@ import (
 	"net/http"
 	"testing"
 
-	"github.com/kyma-incubator/compass/components/system-broker/pkg/http/httpfakes"
```
NickyMateev, line 10: Why are a bunch of system-broker files moved to director component?

components/director/internal/systemfetcher/director.go
```diff
@@ -0,0 +1,60 @@
...
+			    }
+			}`, id))
+	token, err := d.authenticator.GetAuthorization(ctx, tenant)
+	if err != nil {
+		return err
+	}
+
+	gqlRequest.Header.Set("Authorization", "Bearer "+token)
+
+	if err := d.Run(ctx, gqlRequest, nil); err != nil {
```
NickyMateev, line 37: We ought to handle gracefully cases where a delete async request is already in progress (i.e. concurrency check code in director middleware will return an error to this client here)

components/director/internal/systemfetcher/director.go
```diff
@@ -0,0 +1,60 @@
...
+type claims struct {
+	Scopes string `json:"scopes"`
+	Tenant string `json:"tenant"`
+	jwt.StandardClaims
+}
+
+type authProvider struct {
+}
+
+func (a *authProvider) GetAuthorization(ctx context.Context, tenant string) (string, error) {
```
NickyMateev, line 52: Maybe you could use the `unsignedTokenAuthorizationProvider` and have it in place of the `authenticator Authenticator` field of the director client?
https://github.com/kyma-incubator/compass/blob/5deba32726254d9997880bbb21723ef8b22a24c6/components/operations-controller/internal/auth/unsigned_token_provider.go#L60

components/director/internal/systemfetcher/service.go
```diff
@@ -149,6 +180,22 @@ func (s *SystemFetcher) saveSystemsForTenant(ctx context.Context, tenantMapping
...
 	for _, system := range systems {
+		if system.AdditionalAttributes[LifecycleAttributeName] == LifecycleDeleted {
+			log.C(ctx).Infof("Getting system by name %s and system number %s", system.DisplayName, system.SystemNumber)
+			app, err := s.systemsService.GetByNameAndSystemNumber(ctx, system.DisplayName, system.SystemNumber)
+			if err != nil {
+				log.C(ctx).WithError(err).Errorf("Could not get system with name %s and system number %s", system.DisplayName, system.SystemNumber)
+				continue
+			}
+
+			if err := s.directorClient.DeleteSystemAsync(ctx, app.ID, tenantMapping.ID); err != nil {
```
NickyMateev, line 191: Not only can the `s.directorClient.DeleteSystemAsync` func try to gracefully handle concurrent operation errors, but we could even add an additional check here where if app.ready = false & deletedAt != nil => then don't make the `s.directorClient.DeleteSystemAsync` request as the system is already in deletion mode.

Also a log for a concurrent operation would be good

NickyMateev: An e2e test for this would be nice too

components/system-broker/pkg/http/correlation_id_transport.go
```diff
@@ -1,44 +0,0 @@
-/*
```
NickyMateev, line 1: Why does this go down the trash

NickyMateev (changes requested):

chart/compass/templates/tests/system-fetcher/role-binding.yaml
```diff
@@ -14,7 +14,7 @@ metadata:
 rules:
 - apiGroups: ["*"]
   resources: ["jobs"]
-  resourceNames: ["system-fetcher-test"]
+  resourceNames: ["system-fetcher-test", "system-fetcher-test2"]
```
NickyMateev, line 17: кофти

components/director/tools/tools.go
```diff
@@ -0,0 +1,10 @@
+// +build tools
+
+package tools
+
+import (
+	_ "github.com/maxbrunsfeld/counterfeiter/v6"
+)
+
+// This file imports packages that are used when running go generate, or used
```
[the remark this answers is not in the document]
KirilKabakchiev: This is a pattern for locking some external tools that are usually used for code generation to a particular version within a module. This way everyone can use the same counterfeither version. However since this is excluded from the build process via the tools comment, whats missing right now is a makefile step that we can run to set the version of the tool for the module.
Something like 
```
install-tools: download
  @echo Installing tools from tools.go
  @cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
```

components/director/go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/lestrrat-go/jwx v1.2.0
 	github.com/lib/pq v1.10.2
 	github.com/machinebox/graphql v0.2.3-0.20181106130121-3a9253180225
+	github.com/maxbrunsfeld/counterfeiter/v6 v6.4.1
```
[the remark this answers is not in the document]
pankrator: This was needed because I moved a common package from system-broker. System broker uses counterfeiter for mocks, so I decided to introduce it in director.
We discussed in one of the Retrospectives to start using counterfeiter for new mocks and lazily move older mocks to counterfeiter when needed, because mockery is very fragile.

chart/compass/values.yaml
pankrator: `40s` is just a random number that will be sufficient for director requests - I think `30s` will be fine also.

components/director/internal/systemfetcher/director.go
pankrator: At the moment it does not fail. Just prints the error returned and continues with the rest of the systems. Even if a check is added whether the operation is in progress, not sure what else we can do aside from log it.

components/director/pkg/http/error_transport_test.go
pankrator: They are used both in director and in system broker and director is the component to be used for shared packages.

components/director/internal/systemfetcher/director.go
pankrator: Agree, but it should be moved to director first and then extended to support different scopes.

components/system-broker/pkg/http/correlation_id_transport.go
pankrator: There is the same implementation in director that can be used

NickyMateev (dismissed):

PR #1955, opened 2021-07-19
Fix logging setting

StefanCholakov:
- Fix logging setting in pairing adapter and tenant-fetcher job

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -72,7 +72,7 @@ global:
       version: "PR-1939"
     pairing_adapter:
       dir:
-      version: "PR-1939"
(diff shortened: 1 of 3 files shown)


chart/compass/values.yaml
```diff
@@ -72,7 +72,7 @@ global:
       version: "PR-1939"
     pairing_adapter:
       dir:
-      version: "PR-1939"
+      version: "PR-1955"
```
desislavaa, line 75: Changes in the charts don't require version update. The versions should be updated only when there are changes in the code of a component - a new image is built in that case, and we refer to that new image here.

desislavaa (approved):

PR #1960, opened 2021-07-21
Delete eventing label

StefanCholakov:
-Delete default eventing label, when removing from formation

- [x] Implementation
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/components/director/internal/domain/eventing/service_test.go
+++ b/components/director/internal/domain/eventing/service_test.go
@@ -3,6 +3,7 @@ package eventing
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"testing"
 
 	"github.com/kyma-incubator/compass/components/director/pkg/normalizer"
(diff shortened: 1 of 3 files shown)


components/director/internal/domain/eventing/service_test.go
```diff
@@ -1040,25 +1017,31 @@ func Test_GetForApplication(t *testing.T) {
...
 
 		// WHEN
-		_, err := svc.GetForApplication(ctx, app)
+		conf, err := svc.GetForApplication(ctx, app)
 
 		// THEN
-		require.Error(t, err)
-		require.EqualError(t, err, expectedError)
+		require.NoError(t, err)
+		require.Equal(t, emptyConfiguration, conf)
```
DimitarPetrov, line 1044: This test relies on that the GetByKey returns NotFound and then on the second time on 187 line no runtime is found and therefore emptyConfig.

However, I think it makes sense to add a test that on the second execution on GetByKey returns a real runtime and therefore validate the full flow.

1. Get defatul eventing
2. See that there is no scenario anymore
3. Delete the default eventing
4. Get new runtime in scenario with the APP
5. Add default eventing to it
6. Return real config

DimitarPetrov (changes requested):

NickyMateev (approved):

PR #1961, opened 2021-07-22
Introduce dataloaders

nyordanoff:
**Description**

This PR integrates the usage of dataloaders. The concept is explained [here](https://gqlgen.com/reference/dataloaders/) and we are using [this library](https://github.com/vektah/dataloaden).  The dataloaders are used in several of the resolvers so that the number of queries being sent to the database is reduced.

The current approach is:
- HTTP call with graphql body -> Resolvers are hit; DB call for each resolver/subresolver

The new dataloader approach is:
- HTTP call with graphql body -> Middleware is hit and all dataloaders are initialized -> Resolvers are hit; if one has to be called multiple times, it is blocked so that the info (ids) is collected -> FetchFunc of the dataloader is executed with the gathered info; that's only one DB call

- implemented dataloader for `Bundles` resolver
- implemented dataloader for `APIDefinitions` resolver
- implemented dataloader for `EventDefinitions` resolver
- implemented dataloader for `Documents` resolver
- implemented dataloader for `FetchRequest` resolver
- adapt all layers for the mentioned resolvers above 
- introduced `UnionLister` in repo layer allowing `UNION` queries to be built
- remove all unused `List`/`Get` functions across the layers used in the old approach
- enhance the `TestRegisterApplicationWithBundles` e2e tests so that Bundle sub-resources (APIs, Events, Documents, FetchRequests) have unique field names and potential swaps while listing them are caught
- in some resolvers the transaction was committed before graphQL conversion which can throw an error -> adapted to commit the transaction after the conversion

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


pankrator (changes requested):
Check places where array initialisation can be done with capacity. I mentioned some, but may have missed others.

pankrator (approved):

PR #1963, opened 2021-07-23
Configure logging in  operation controller

StefanCholakov:
Before, only a part from the operations-controller's logs was in JSON format. Example:
`operations-controller <timestamp>    INFO   Operation executed successfully    {"operation": "compass-system/application-<application_id>"} `. After the change, the whole log is in JSON format. Example:
`{"level":"info","written_at":<timestamp>,"msg":"Successfully updated operation status to succeeded","operation":"compass-system/application-<application_id>"}`

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


StanislavStefanov (approved):

desislavaa:
we need to agree on whether we'll introduce new dependency, or we'll upgrade the kubebuilder version

desislavaa (approved):

PR #1967, opened 2021-07-30
Enable formation filtering in ORD service

ataleksandrov:
Enable filtering based on formation in the ORD service:
- mapping views are added
- tenant_mapping handler is modified to accept external tenants that are not present in tenant_mapping table.
- e2e tests

Also see: https://github.com/kyma-incubator/ord-service/pull/33

PR #1981, opened 2021-08-13
Secure ORD Service with external certificate issuer

DimitarPetrov:
**Description**

We need a way to trust an external certificate issuer for ORD service communication.

- New istio gateway and vs trusting the external cert issuer's root CA
- Improve certificate mapping hydrator to genericly match a subject and extract the client from it.
- Add a new flow in tenant-mapping hydrator for CertService

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/tenantmapping/cert_service_context_provider.go
```diff
@@ -0,0 +1,55 @@
...
+	//
+	//		log.C(ctx).Infof("Returning tenant context with empty internal tenant ID and external ID %s", externalTenantID)
+	//		return NewObjectContext(NewTenantContext(externalTenantID, ""), "", authDetails.AuthID, consumer.Runtime), nil
+	//	}
+	//	return ObjectContext{}, errors.Wrapf(err, "while getting external tenant mapping [ExternalTenantId=%s]", externalTenantID)
+	//}
+	//
+	//objCtx := NewObjectContext(NewTenantContext(externalTenantID, tenantMapping.ID), "", authDetails.AuthID, consumer.Runtime)
+
+	objCtx := NewObjectContext(NewTenantContext(externalTenantID, externalTenantID), "", authDetails.AuthID, consumer.Runtime)
```
NickyMateev, line 50: So until we start storing subaccs, this will result in no records being returned by the Director as there will be no actual internal tenant with the given ID and sql queries `WHERE tenant_id=<exernalTenantID` will return empty result set - correct?

tests/ord-service/tests/main_test.go
```diff
@@ -34,11 +39,18 @@ var (
...
-	ORDServiceURL                 string
-	ORDServiceStaticURL           string
-	ORDServiceDefaultResponseType string
-	DefaultScenarioEnabled        bool `envconfig:"default=true"`
+	DirectorURL                      string
+	ORDServiceURL                    string
+	ORDExternalCertSecuredServiceURL string
+	ORDServiceStaticPrefix           string
+	ORDServiceDefaultResponseType    string
+	CACertificate                    []byte `envconfig:"-"`
```
NickyMateev, line 47: Maybe it makes sense now to bunch the cert config into a separate struct and embed it here

NickyMateev (changes requested):

components/director/internal/tenantmapping/cert_service_context_provider.go
DimitarPetrov: The ORD services works with views introduced with that has subaccounts as tenants: https://github.com/kyma-incubator/compass/pull/1967

NickyMateev (approved):

NickyMateev (approved):

PR #1985, opened 2021-08-17
Add paging to system fetcher

krasish:
The system fetcher job currently does not do paging when it calls APIs for systems. Instead, it only calls once. This leads to missing systems for tenants with more than the default page size for APIs which are pageable. This PR introduces a generic PageIterator and refactors system fetcher to use it. 

**Description**

- Introduce PageIterator
- Change System Fetcher to use the PageIterator
- Add unit tests

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [ ] Integration tests


components/director/internal/systemfetcher/client_test.go
```diff
@@ -144,10 +206,10 @@ func fixHttpClient(t *testing.T) (*mockData, string) {
 	mock := mockData{
 		callNumber: 1,
 	}
-
+	// 0 1 2 0 1 2
```
alextargov, line 209: Guess that's unnecessary anymore?

components/director/pkg/paging/page_iterator.go
```diff
@@ -0,0 +1,89 @@
...
+		skip:                  0,
+		pageSize:              pageSize,
+		do:                    do,
+	}
+}
+
+//Next fetches the next page of the PageIterator. It returns true if there possibly are more pages that can eb fetched.
+//In order to get all the results that the PageIterator can fetch Next should be called until it returns false and no error.
+//Once next returns false with no error it should not be called anymore. If it does get called the behaviour is undefined.
+func (p *PageIterator) Next() (bool, error) {
```
pankrator, line 46: Store in the `PageIterator` that the final page is encountered in a variable. Then if `Next` is called even after there are no more pages, do nothing.

components/director/pkg/paging/page_iterator.go
```diff
@@ -0,0 +1,89 @@
...
+		return false, nil
+	}
+	p.skip += p.pageSize
+	return true, nil
+}
+
+//FetchAll fetches all the pages (calls Next method) that the PageIterator can fetch.
+func (p *PageIterator) FetchAll() (err error) {
+	var again bool
+	for again, err = p.Next(); again; again, err = p.Next() {
```
pankrator, line 63: instead of `again` -> `hasNext`?

components/director/pkg/paging/page_iterator.go
```diff
@@ -0,0 +1,89 @@
...
+//In order to get all the results that the PageIterator can fetch Next should be called until it returns false and no error.
+//Once next returns false with no error it should not be called anymore. If it does get called the behaviour is undefined.
+func (p *PageIterator) Next() (bool, error) {
+	p.buildNextURL()
+	count, err := p.do(p.nextURL)
+	if err != nil {
+		return false, errors.Wrapf(err, "while fetching next page: ")
+	}
+
+	if count < p.pageSize {
```
pankrator, line 53: What will happen if there are exactly `pageSize` elements returned and no more. The code using page iterator will call `Next` and it will fail to fetch any more elements. 
`p.do` function should return whether there is next page or not depending on the response.

components/director/internal/systemfetcher/client_test.go
krasish: It definitely is :D

components/director/pkg/paging/page_iterator.go
pankrator: As discussed, the backend should respond with empty result set if there are no more elements.

components/director/pkg/paging/page_iterator.go
```diff
@@ -0,0 +1,89 @@
...
+func (p *PageIterator) FetchAll() (err error) {
+	var again bool
+	for again, err = p.Next(); again; again, err = p.Next() {
+	}
+	return err
+}
+
+func (p *PageIterator) buildNextURL() {
+	p.resetURL()
+	p.setQueryParam(fmt.Sprintf(p.pageSkipFormat, p.skip))
```
pankrator, line 70: see `url.Values` type and `url.Parse`

components/director/pkg/paging/page_iterator.go
pankrator: If there is an error in the first page, the for loop will continue until again is true. Perhaps move the login inside for loop body, it will be easier to read and maintain

pankrator (changes requested):

components/director/pkg/paging/page_iterator.go
```diff
@@ -0,0 +1,89 @@
...
+		pageSkipFormat:        skipFormat,
+		pageSizeFormat:        sizeFormat,
+		additionalQueryParams: additionalQueryParams,
+		skip:                  0,
+		pageSize:              pageSize,
+		do:                    do,
+	}
+}
+
+//Next fetches the next page of the PageIterator. It returns true if there possibly are more pages that can eb fetched.
```
alextargov, line 43: Typo here `eb`

components/director/pkg/paging/page_iterator.go
krasish: As discussed, we need our implementation of the URL query formatting.

components/external-services-mock/internal/systemfetcher/handler.go
```diff
@@ -15,13 +15,14 @@ import (
 type SystemFetcherHandler struct {
 	mutex         sync.Mutex
 	defaulTenant  string
-	mockedSystems []byte
+	mockedSystems [][]byte
```
danielgospodinow, line 18: What is the purpose of those `mockedSystems`?

components/external-services-mock/internal/systemfetcher/handler.go
```diff
@@ -60,10 +61,24 @@ func (s *SystemFetcherHandler) HandleFunc(rw http.ResponseWriter, req *http.Requ
...
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
-	_, err := rw.Write(s.mockedSystems)
+	_, err := rw.Write(resp)
 	if err != nil {
 		httphelpers.WriteError(rw, errors.Wrap(err, "error while writing response"), http.StatusInternalServerError)
 	}
 }
+
+func (s *SystemFetcherHandler) HandleReset(rw http.ResponseWriter, _ *http.Request) {
```
danielgospodinow, line 78: I'm also wondering for the purpose of this reset handler.

components/external-services-mock/internal/systemfetcher/handler.go
krasish: **mockedSystems** stores the systems which the external-services-mock will return on calls to `/systemfetcher/fetch`. It serves as a queue - on each call the first system that has been mocked through the `/systemfetcher/configure` endpoint will get returned. In order to be able to test paging functionality, we need to configure more than one response (at least 2 pages). That is why it is a slice of slices now.

components/external-services-mock/internal/systemfetcher/handler.go
krasish: When an e2e test is done we want to make sure that no leftover mock responses are left which can potentially break the next one. That is why all system-fetcher tests [have a deferred cleanup](https://github.com/kyma-incubator/compass/pull/1985/files#diff-8ab34488d3ef47b9a7de83c838c9fab56baaa87ef7b2d022de34fc46318e61d3R230) for the systems that they have mocked to the external-services-mock. This handler simply clears the queue and gets called at the end of each test.

pankrator (approved):

components/external-services-mock/internal/systemfetcher/handler.go
danielgospodinow: Oh, sorry, I've misread the location of this file. I didn't see it was in the `ext-svc-mock` service. I though it was part of the system fetcher service itself which then would've been a concern. It's all good! 👍🏼

PR #1991, opened 2021-08-19
Remove description from ory rules

StefanCholakov:
Remove description from ory rules as there is no such property since oathkeeper-maester v0.1.4

### DIFF
--- a/chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
+++ b/chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
@@ -3,7 +3,7 @@ kind: Rule
 metadata:
   name: compass-gateway-oauth
 spec:
-  description: Configuration of oathkeeper for secure endpoint of compass gateway
+  # Configuration of oathkeeper for secure endpoint of compass gateway
   upstream:
     url: "http://compass-gateway.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.global.gateway.port }}"
   match:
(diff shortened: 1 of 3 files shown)


NickyMateev (approved):

chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
```diff
@@ -163,7 +163,7 @@ metadata:
   name: compass-connector-one-time-tokens
   namespace: {{ .Release.Namespace }}
 spec:
-  description: Configuration of oathkeeper for secure endpoint of compass gateway
+  # Configuration of oathkeeper for secure endpoint of compass gateway
```
dzahariev, line 166: Having "description" seem to be still valid syntax.
https://github.com/ory/oathkeeper-maester/blob/master/config/samples/oathkeeper_v1alpha1_rule.yaml#L13

dzahariev (changes requested):

chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
StefanCholakov: "description" seems to miss here:
https://github.com/ory/oathkeeper-maester/blob/master/api/v1alpha1/rule_types.go#L59
Is it possible this example to be outdated?

dzahariev:
The error seem to be here:
https://github.com/kyma-project/test-infra/blob/a4f8d51b794b7ad962b23b8178c98b71ada1d0eb/prow/scripts/cluster-integration/compass-gke-benchmark.sh#L281

The new kyma is installed (from PR) with previous compass (from master). This I the reason new Kyma cannot work with old syntax.

PR #1993, opened 2021-08-24
Add request monitoring per client identificator for all authentication flows

danielgospodinow:
**Description**

Introduce Prometheus counters that keep track of the number of requests from all clients. The idea is to have information about the clients that are calling us.

- Prometheus counters for client IDs of calling clients

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

It should look something like this:
Screenshot 2021-09-10 at 13 47 44


components/director/internal/tracking/request.go
```diff
@@ -0,0 +1,85 @@
+package tracking
```
pankrator, line 1: In general this file seems like doing the same as the different flows for authenticating JWT, Cert and One-time-token. 

Perhaps it would be better to split the *trackingService* in 3 different calls which are accordingly to each responsible component. For certificate it would be connector, for JWT and OTT it would be director. This will also remove the dependency of director to connector. 

The tracking service can put the client identification in the header/body and then it can be inspected in director (as the last component in the chain) and used in prometheus metric.

pankrator (changes requested):

components/director/internal/tracking/handler.go
```diff
@@ -0,0 +1,54 @@
+package tracking
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/kyma-incubator/compass/components/director/internal/metrics"
+	doathkeeper "github.com/kyma-incubator/compass/components/director/internal/oathkeeper"
```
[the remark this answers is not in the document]
danielgospodinow: For `Director Oathkeeper` but it's useless in this case. 😆

components/director/internal/tracking/request.go
danielgospodinow: We discussed offline that it would be better to instrument client calls directly into the Tenant Mapping mutator right before we many any DB calls.

components/director/internal/tenantmapping/handler_test.go
```diff
@@ -227,7 +233,10 @@ func TestHandler(t *testing.T) {
 			tenantmapping.AuthenticatorObjectContextProvider: userMockContextProvider,
 		}
 
-		handler := tenantmapping.NewHandler(authn, reqDataParserMock, transact, objectContextProviders)
+		clientInstrumenter := &automock.ClientInstrumenter{}
+		clientInstrumenter.On("InstrumentClient", identityUsername, "JWT", mock.Anything)
```
pankrator, line 237: instead "JWT", oathkeeper.JWTAuthFlow

components/director/internal/tenantmapping/handler_test.go
```diff
@@ -340,7 +352,10 @@ func TestHandler(t *testing.T) {
 			tenantmapping.SystemAuthObjectContextProvider: systemAuthMockContextProvider,
 		}
 
-		handler := tenantmapping.NewHandler(nil, reqDataParserMock, transact, objectContextProviders)
+		clientInstrumenter := &automock.ClientInstrumenter{}
+		clientInstrumenter.On("InstrumentClient", systemAuthID.String(), "OAuth2", mock.Anything)
```
pankrator, line 356: instead "JWT",  `oathkeeper.OAuth2Flow`

pankrator (changes requested):

components/director/internal/tenantmapping/handler.go
```diff
@@ -52,25 +49,34 @@ type TenantRepository interface {
 	GetByExternalTenant(ctx context.Context, externalTenant string) (*model.BusinessTenantMapping, error)
 }
 
+// ClientInstrumenter describes instrumentation tools that support client instrumentation.
```
pankrator, line 52: better description would be:
`collects metrics for different client and auth flows

pankrator (changes requested):

pankrator (approved):

PR #1996, opened 2021-08-30
Upgrade go to version 1.17

StefanCholakov:

- Upgrade go to version 1.17
- Release note: https://golang.org/doc/go1.17
- Note: Since the current  `// +build` lines for build tag selection are transformed to new `//go:build` lines, running `go fix` will remove all `// +build` lines from the source tree, leaving behind equivalent, easier-to-read `//go:build` lines


desislavaa (approved):
Just add the link to the release notes in the description, and note that we have to keep both versions of the "build" command until Go 1.18, then we should execute `go fix` to get rid of the old syntax.
- https://golang.org/doc/go1.17
- https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

PR #1998, opened 2021-08-31
Introduce regional subaccount tenants

desislavaa:
**Description**

- This PR introduces а new tenant type called "subaccount", which will be a child to the "account" tenant. The tenant hierarchy becomes: "customer" -> "account" -> "subaccount". The new tenant must have a parent tenant.
We figure the type of the tenant we create by the presence of a specific property in the incoming create request - if the body contains only the property for account tenant ID (e.g.`accountID`), then we should create an account tenant, but if it also contains a property for a subaccount tenant ID (e.g. `subaccountID`), then we should create a new tenant of type "subaccount".
The name "subaccount" for the new type looks suitable, because of the parent-child relationship between the new tenant type, and the "Account" tenants we currently have. Another option was "regional", but the "Account" tenants can also be labeled with a region, hence "regional" was no longer an option. 
- A new endpoint which can be called to get dependent services when onboarding a tenant - returns an empty json for now
- Small improvement in the E2E tests - use the configured json properties, instead of hardcoded ones, when sending tenant creation request.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/tenantfetchersvc/handler.go
```diff
@@ -109,16 +114,65 @@ func (h *handler) DeleteByExternalID(writer http.ResponseWriter, req *http.Reque
...
+
+	vars := mux.Vars(request)
+	region, ok := vars[h.config.RegionPathParam]
+	if !ok {
+		log.C(ctx).Errorf("Region path parameter is missing from request")
+		http.Error(writer, "Region path parameter is missing from request", http.StatusBadRequest)
+		return
+	}
+
+	body, err := readBody(request)
```
DimitarPetrov, line 128: In the readBody, request body is closed, which is not necessary:

```
The Server will close the request body. The ServeHTTP Handler does not need to.
```

components/director/internal/tenantfetchersvc/provisioner.go
```diff
@@ -58,6 +94,10 @@ func (p *provisioner) ensureParentExists(ctx context.Context, parentTenantID, ch
 		return id, nil
 	}
 
+	if !automaticParentCreationEnabled(childTenant) {
```
DimitarPetrov, line 97: Why automatic GA creation as part of SA event is disabled? What will happen if we miss a subscription event for a subaccount because the GA is still not fetched?

tests/tenant-fetcher/tests/handler_test.go
```diff
@@ -163,27 +189,221 @@ func TestDecommissioningHandler(t *testing.T) {
...
+		require.NoError(t, err)
+	}
+	if len(tenant.CustomerID) > 0 {
+		body, err = sjson.Set(body, config.CustomerIDProperty, tenant.CustomerID)
+		require.NoError(t, err)
+	}
+	if len(tenant.Subdomain) > 0 {
+		body, err = sjson.Set(body, config.SubdomainProperty, tenant.Subdomain)
+		require.NoError(t, err)
+	}
```
DimitarPetrov, line 398: why not just json marshall the tenant?

DimitarPetrov (changes requested):

tests/tenant-fetcher/tests/handler_test.go
desislavaa: That was actually the behaviour before, and while testing locally with the props from prod, the tests started failing because they sent for example "subdomain", but our prod configuration requires "subscribedSubdomain" in the tenant-fetcher deployment.

components/director/internal/tenantfetchersvc/handler.go
```diff
@@ -109,16 +115,73 @@ func (h *handler) DeleteByExternalID(writer http.ResponseWriter, req *http.Reque
...
+		log.C(ctx).WithError(err).Errorf("Failed to provision tenant with ID %s: %v", regionalTenant.ExternalTenant, err)
+		http.Error(writer, fmt.Sprintf(tenantCreationFailureMsgFmt, regionalTenant.ExternalTenant), http.StatusInternalServerError)
+		return
+	}
+
+	writer.Header().Set("Content-Type", "text/plain")
+	writer.WriteHeader(http.StatusOK)
+	if _, err := writer.Write([]byte(compassURL)); err != nil {
+		log.C(ctx).WithError(err).Errorf("Failed to write response body for tenant request creation for tenant %s: %v", regionalTenant.ExternalTenant, err)
+	}
```
pankrator, line 152: This functionality is the same as in the `Create` method. Extract it to a common method

pankrator (changes requested):

components/director/internal/tenantfetchersvc/provisioner.go
```diff
@@ -11,78 +11,112 @@ import (
...
+		}
+		subaccountTenant := p.newSubaccountTenant(request.SubaccountTenantID, request.AccountTenantID, request.Subdomain, request.Region)
+		if len(intTenantID) > 0 {
+			log.C(ctx).Infof("Parent tenant with internal ID %s is already present", intTenantID)
+			subaccountTenant.Parent = intTenantID
+			return []model.BusinessTenantMappingInput{subaccountTenant}, nil
+		}
+
+		accountTenant.Subdomain = ""
+		accountTenant.Region = ""
```
desislavaa, line 81: Or it could get the region and subdomain from the SA, but I tought that later way we won't be sure if the labels are real, or are "inherited" from the SA

components/director/internal/tenantfetchersvc/handler.go
```diff
@@ -114,44 +100,44 @@ func (h *handler) DeleteByExternalID(writer http.ResponseWriter, req *http.Reque
...
-	region, ok := vars[h.config.RegionPathParam]
-	if !ok {
-		log.C(ctx).Errorf("Region path parameter is missing from request")
-		http.Error(writer, "Region path parameter is missing from request", http.StatusBadRequest)
+func (h *handler) Dependencies(writer http.ResponseWriter, request *http.Request) {
+	if _, err := writer.Write([]byte("{}")); err != nil {
+		log.C(request.Context()).WithError(err).Errorf("Failed to write response body for dependencies request")
 		return
 	}
+	writer.Header().Set("Content-Type", "application/json")
```
DimitarPetrov, line 108: You should first write the headers and then write the body.

components/director/internal/tenantfetchersvc/handler.go
```diff
@@ -162,29 +148,24 @@ func (h *handler) tenantFromBody(body []byte) (*model.BusinessTenantMappingInput
...
+		SubaccountTenantID: properties[h.config.SubaccountTenantIdProperty],
+		CustomerTenantID:   properties[h.config.CustomerIdProperty],
+		Subdomain:          properties[h.config.SubdomainProperty],
+		Region:             region,
 	}, nil
 }
 
-func (h *handler) provisionTenant(ctx context.Context, tenant model.BusinessTenantMappingInput, region string) error {
-	externalTenantID := tenant.ExternalTenant
+func getMainTenantID(request *TenantProvisioningRequest) string {
```
DimitarPetrov, line 160: This can be method of TenantProvisioningRequest

components/director/internal/tenantfetchersvc/provisioner.go
```diff
@@ -12,121 +11,112 @@ import (
 	"github.com/pkg/errors"
 )
 
-const regionLabelKey = "region"
+const autogeneratedTenantProvider = "autogenerated"
```
DimitarPetrov, line 14: What is the purpose of using "autogenerated" for Customers?

components/director/internal/tenantfetchersvc/provisioner.go
```diff
@@ -12,121 +11,112 @@ import (
...
-	label := &model.LabelInput{
-		Key:        regionLabelKey,
-		Value:      region,
-		ObjectID:   intTenantID,
-		ObjectType: model.TenantLabelableObject,
-	}
-	if err := p.tenantSvc.SetLabel(ctx, label); err != nil {
-		return errors.Wrapf(err, "while setting region label for tenant with external ID %s", tenant.ExternalTenant)
-	}
+func (p *provisioner) tenantsFromRequest(ctx context.Context, request TenantProvisioningRequest) ([]model.BusinessTenantMappingInput, error) {
```
DimitarPetrov, line 63: Suggestion:

This way it is a bit easier to follow for me, however, it makes more DB calls. I am not sure which is better, so I am fine with both approaches.

```go
func (p *provisioner) tenantsFromRequest(ctx context.Context, request TenantProvisioningRequest) ([]model.BusinessTenantMappingInput, error) {
	tenants := make([]model.BusinessTenantMappingInput, 0, 3)
	customerID := request.CustomerTenantID
	accountID := request.AccountTenantID

	if len(request.CustomerTenantID) > 0 {
		customerInternalID, err := p.tenantSvc.GetInternalTenant(ctx, request.CustomerTenantID)
		if err != nil && !apperrors.IsNotFoundError(err) {
			return nil, errors.Wrapf(err, "while checking if parent of type customer already exists")
		}
		if len(customerInternalID) == 0 {
			tenants = append(tenants, p.newCustomerTenant(request.CustomerTenantID))
		} else {
			customerID = customerInternalID
		}
	}

	if len(request.AccountTenantID) > 0 {
		accountInternalID, err := p.tenantSvc.GetInternalTenant(ctx, request.AccountTenantID)
		if err != nil && !apperrors.IsNotFoundError(err) {
			return nil, errors.Wrapf(err, "while checking if parent of type customer already exists")
		}
		if len(accountInternalID) == 0 {
			accToInsert := p.newAccountTenant(request.AccountTenantID, customerID, request.Subdomain, request.Region)
			if len(request.SubaccountTenantID) > 0 { // This means that the request is for Subaccount provisioning, therefore the subdomain and the region are for the subaccount and not for the GA
				accToInsert.Subdomain = ""
				accToInsert.Region = ""
			}
			tenants = append(tenants, accToInsert)
		} else {
			accountID = accountInternalID
		}
	}

	if len(request.SubaccountTenantID) > 0 {
		tenants = append(tenants, p.newSubaccountTenant(request.SubaccountTenantID, accountID, request.Subdomain, request.Region))
	}
	return tenants, nil
}
```

components/director/cmd/tenantfetcher-job/main.go
```diff
@@ -35,6 +35,7 @@ type config struct {
 	Features features.Config
 
 	TenantProvider       string        `envconfig:"APP_TENANT_PROVIDER"`
+	TenantsRegion        string        `envconfig:"default=central,APP_TENANTS_REGION"`
```
DimitarPetrov, line 38: Is `central` the desired value or we should override it per universe?

tests/tenant-fetcher/tests/handler_test.go
```diff
@@ -236,6 +236,40 @@ func TestRegionalOnboardingHandler(t *testing.T) {
 			require.NoError(t, err)
 			assertTenant(t, tenant, childTenant.SubaccountID, childTenant.Subdomain)
 			require.Equal(t, regionPathParamValue, tenant.Labels["region"])
+
+			parentTenantAfterInsert, err := fixtures.GetTenantByExternalID(dexGraphQLClient, parentTenant.TenantID)
+			require.NoError(t, err)
+			assertTenant(t, parentTenantAfterInsert, parentTenant.TenantID, parentTenant.Subdomain)
+			require.Equal(t, regionPathParamValue, parentTenantAfterInsert.Labels["region"])
```
DimitarPetrov, line 243: Add this check to the checks directly after inserting the parent (before inserting the child) as well.

DimitarPetrov (changes requested):

components/director/internal/domain/tenant/service.go
```diff
@@ -120,51 +122,68 @@ func (s *service) MultipleToTenantMapping(tenantInputs []model.BusinessTenantMap
...
+	tenantFromDB, err := s.tenantMappingRepo.GetByExternalTenant(ctx, tenant.ExternalTenant)
+	if err != nil && !apperrors.IsNotFoundError(err) {
 		return errors.Wrapf(err, "while checking the existence of tenant with external ID %s", tenant.ExternalTenant)
 	}
-	if exists {
-		return nil
+	if tenantFromDB != nil {
+		log.C(ctx).Infof("Tenant with external ID %s already exists", tenantFromDB.ExternalTenant)
+		return s.upsertLabels(ctx, tenantFromDB.ID, subdomain, region)
 	}
```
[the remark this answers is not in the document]
desislavaa: We save ourselves one DB call that way

components/director/cmd/tenantfetcher-job/main.go
desislavaa: We should override it, but I left one `central` here for the sake of open source :D I'll move the default value in the charts

components/director/internal/tenantfetchersvc/provisioner.go
desislavaa: In case we start gathering more info for the customer tenants (for example we get notified when one is created, just like with accounts), then we'll know which tenants should be updated and "enriched"

components/director/internal/domain/tenant/service.go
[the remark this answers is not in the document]
DimitarPetrov: In the current implementation you do Get + Upsert Labels. If you delete it you do Update + Upsert Labels.
It is the same -> 2 DB calls?

components/director/internal/domain/tenant/service.go
```diff
@@ -132,29 +132,44 @@ func (s *labeledService) CreateManyIfNotExists(ctx context.Context, tenantInputs
...
+	} else if apperrors.IsNotUniqueError(err) {
+		tenantFromDB, err := s.tenantMappingRepo.GetByExternalTenant(ctx, tenant.ExternalTenant)
+		if err != nil {
+			return "", errors.Wrapf(err, "while getting internal ID of tenant %s", tenant.ExternalTenant)
+		}
+		tenantID = tenantFromDB.ID
 	}
 
-	return s.upsertLabels(ctx, tenant.ID, subdomain, region)
+	return tenantID, s.upsertLabels(ctx, tenant.ID, subdomain, region)
```
DimitarPetrov, line 172: This upsert should use tenantID instead of tenant.ID in case of Conflict?

DimitarPetrov (changes requested):

components/director/internal/domain/tenant/service.go
desislavaa: Yes. For example, we have an Account tenant with external ID "x", then we want to create it with internal ID "y", however, it already exists with internal ID "z" - in that case we only want to update its labels, and w need the internal ID there - "z"

components/director/internal/domain/tenant/service.go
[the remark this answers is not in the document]
desislavaa: I just saw that you've marked all lines 144-150, so yes, you are right, I'm removing that logic right away

DimitarPetrov (dismissed):

scripts/generic_make_go.mk
```diff
@@ -138,6 +138,10 @@ check-fmt-local:
 fmt-local:
 	go fmt $$($(DIRS_TO_CHECK))
 
+format-local: imports-local fmt-local
+
+verify-local: test-local check-imports-local check-fmt-local errcheck-local
```
DimitarPetrov, line 143: Add lint local here

DimitarPetrov (approved):

pankrator (changes requested):
Change tenantfetcher logic for parents creation and use the new function for tenant creation ordering

pankrator (approved):

PR #2010, opened 2021-09-08
Allow extensive ord querying

nyordanoff:
**Description**

This PR introduces some ORD-related changes in order to enable the ORD Service to support some odata filtering queries (`any`, `all` operators).

- in `tenants_apis` view the `apiProtocol` field is casted to `text` so that the ORD Service can use it for filtering
- `api_product`, `event_product` and `package_product` views are changed so that only the "internal" ids are used for relation between the entities instead of the old approach where the `ord_id` was used. This allowed us to change the primary key of these entities in the ORD Service to be the "internal" id as it is in the compass already.
- a new column `id` is added to the `bundle_references` table and becomes a PK. Its only purpose is to be able to specify this `id` field in the ORD Service as PK for that entity.
- e2e test is added

**Related issue(s)**
- https://github.com/kyma-incubator/ord-service/pull/41

**Pull Request status**

- [x] Implementation
- [x] Integration tests


components/schema-migrator/migrations/director/20210907181222_ord_extensive_querying.up.sql
```diff
@@ -0,0 +1,88 @@
+BEGIN;
```
pankrator, line 1: As the migrations are becoming more and more complex and long to check. Can you add as a comment above each view what is the change since the previous db schema

nyordanoff: I described it in the PR description, but probably for the long run, it could be useful to include the diff in the migration as well. 

We can sync with the team and probably introduce it as a practice for the bigger migrations from now on.

components/director/internal/domain/bundlereferences/service.go
```diff
@@ -20,6 +20,7 @@ type BundleReferenceRepository interface {
 	ListByBundleIDs(ctx context.Context, objectType model.BundleReferenceObjectType, tenantID string, bundleIDs []string, pageSize int, cursor string) ([]*model.BundleReference, map[string]int, error)
 }
 
+// UIDService missing godoc
```
DimitarPetrov, line 23: Consider adding real go docs as part of linter issues fixing. This is a default value and it is not intended to be copied adding more and more technical debt. Its actual intent was to trigger a boy scout principal feeling in people and make them fix a few comments around the place where the task is involved instead of adding more of them.

tests/ord-service/tests/api_test.go
```diff
@@ -66,12 +68,15 @@ func TestORDService(stdT *testing.T) {
...
 	secondaryTenant := tenant.TestTenants.GetIDByName(t, tenant.ApplicationsForRuntimeTenantName)
 	tenantFilteringTenant := tenant.TestTenants.GetIDByName(t, tenant.TenantSeparationTenantName)
+	tenantAPIProtocolFiltering := tenant.TestTenants.GetIDByName(t, tenant.ListLabelDefinitionsTenantName)
 
 	// Cannot use tenant constants as the names become too long and cannot be inserted
 	appInput := fixtures.CreateApp("tenant1")
 	appInput2 := fixtures.CreateApp("tenant2")
 	appInputInScenario := fixtures.CreateApp("tenant3-in-scenario")
 	appInputNotInScenario := fixtures.CreateApp("tenant3-no-scenario")
+	appInputAPIProtocolFiltering := fixtures.CreateApp("tenant3")
```
DimitarPetrov, line 78: `tenant4` ?

DimitarPetrov (dismissed):

DimitarPetrov (approved):

PR #2013, opened 2021-09-13
Tenant Fetcher job improvements

desislavaa:
**Description**

- Instead of separately creating parent tenants, and associating them with their children, just use the [functionality provded in the tenant service](https://github.com/kyma-incubator/compass/blob/main/components/director/internal/domain/tenant/service.go#L112-L134) where the provided input slice is sorted, so the parent tenants are created first

**Pull Request status**

- [x] Implementation
- [x] Unit tests


components/director/internal/tenantfetcher/service.go
```diff
@@ -197,10 +197,6 @@ func (s Service) SyncTenants() error {
 
 	// Order of event processing matters
 	if len(tenantsToCreate) > 0 {
-		if err := s.createParents(ctx, currentTenants, tenantsToCreate); err != nil {
-			return errors.Wrap(err, "while storing parents")
-		}
-		currentTenants, err = s.getCurrentTenants(ctx)
 		if err != nil {
 			return err
 		}
```
DimitarPetrov, line 202: Leftover?

components/director/internal/tenantfetcher/service.go
```diff
@@ -230,14 +226,8 @@ func (s Service) SyncTenants() error {
...
 func (s Service) createTenants(ctx context.Context, currTenants map[string]string, eventsTenants []model.BusinessTenantMappingInput) error {
-	tenantsToCreate := make([]model.BusinessTenantMappingInput, 0)
+	tenantsToCreate := s.parents(currTenants, eventsTenants)
 	for _, eventTenant := range eventsTenants {
-		if _, ok := currTenants[eventTenant.ExternalTenant]; ok {
-			continue
-		}
-		if len(eventTenant.Parent) > 0 {
-			eventTenant.Parent = currTenants[eventTenant.Parent]
-		}
```
DimitarPetrov, line 240: I think that here we should still do some checks:

```go
if _, ok := currTenants[eventTenant.ExternalTenant]; ok {
 	continue
 }
if len(eventTenant.Parent) > 0 {
        if parentInternalID, ok := currTenants[eventTenant.Parent]; ok {
 	        eventTenant.Parent = parentInternalID
        }
}
```

1. If the tenant already exists -> skip it
2. If the tenant has parent and the parent already exists -> convert the ID, since the common repo logic expects already converted parent ID if the parent is not being created in the same request. https://github.com/kyma-incubator/compass/blob/main/components/director/internal/domain/tenant/service.go#L122

DimitarPetrov (changes requested):

components/director/internal/tenantfetcher/service.go
desislavaa: I agree for point 2, but for point one it won't fix the GA labels (this was supposed to be part of the previous PR)

DimitarPetrov (approved):

PR #2023, opened 2021-09-15
Subscription flow labelling

ataleksandrov:
**Description**

Introduce functionality for subscriptions. Tenants could subscribe for runtime with given labels. Non existing tenants will be created in the subscription flow. After a subscription event the consumer runtime will be labeled with the tenant id of the subscriber. After an unsubscription event the unsubscribing tenant id will be removed from the runtime label.Runtime label is an array of tenant ids.

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/tenantfetchersvc/subscriber.go
```diff
@@ -0,0 +1,124 @@
...
+
+		return errors.Wrap(err, fmt.Sprintf("Failed to get runtimes for labels %s: %s and %s: %s", s.RegionLabelKey, region, s.SubscriptionConsumerLabelKey, subscriptionConsumerID))
+	}
+
+	for _, runtime := range runtimes {
+		ctx = tenant.SaveToContext(ctx, runtime.Tenant, "")
+
+		labelOldValue := make([]string, 0)
+		label, err := s.GetLabel(ctx, runtime.ID, s.ConsumerSubaccountIDsLabelKey)
+		if err != nil {
```
desislavaa, line 86: `if err != nil && !apperrors.IsNotFoundError(err)` will save us from the nested `if`

components/director/internal/tenantfetchersvc/subscriber.go
```diff
@@ -0,0 +1,124 @@
...
+			if !apperrors.IsNotFoundError(err) {
+				return errors.Wrap(err, fmt.Sprintf("Failed to get label for runtime with id: %s and key: %s", runtime.ID, s.ConsumerSubaccountIDsLabelKey))
+			}
+			// if the error is not found, do nothing and continue
+		} else {
```
desislavaa, line 91: Do we need the `else`

components/director/internal/tenantfetchersvc/subscriber.go
```diff
@@ -0,0 +1,124 @@
...
+			if labelOldValue, err = labelutils.ValueToStringsSlice(label.Value); err != nil {
+				return errors.Wrap(err, fmt.Sprintf("Failed to parse label values for label with id: %s", label.ID))
+			}
+		}
+
+		var labelNewValue = mutateSliceFunc(labelOldValue, subaccountTenantID)
```
desislavaa, line 97: Kind of hard to think of a better name but maybe `mutateLabelsFunc`, at least locally in this method

desislavaa (commented):
Looks good, just a few minor ones

components/director/internal/tenantfetchersvc/subscriber.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Unsubscribe unsubscribes tenant from runtime.
+func (s *subscriber) Unsubscribe(ctx context.Context, tenantSubscriptionRequest *TenantSubscriptionRequest, region string) error {
+	return s.applyRuntimesSubscriptionChange(ctx, tenantSubscriptionRequest.SubscriptionConsumerID, tenantSubscriptionRequest.SubaccountTenantID, region, removeElement)
+}
+
+func (s *subscriber) applyRuntimesSubscriptionChange(ctx context.Context, subscriptionConsumerID, subaccountTenantID, region string, mutateSliceFunc sliceMutationFunc) error {
+	filters := []*labelfilter.LabelFilter{
+		labelfilter.NewForKeyWithQuery(s.SubscriptionConsumerLabelKey, fmt.Sprintf("\"%s\"", subscriptionConsumerID)),
```
pankrator, line 68: Shouldn't `subscriptionConsumerID` be `subscriptionProviderId` as the provider is the one having the runtime and the consumer is the one consuming the runtime? 
Perhaps the naming of the properties should be aligned somehow.

pankrator (changes requested):

components/director/internal/tenantfetchersvc/subscriber.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, runtime := range runtimes {
+		ctx = tenant.SaveToContext(ctx, runtime.Tenant, "")
+
+		labelOldValue := make([]string, 0)
+		label, err := s.GetLabel(ctx, runtime.ID, s.ConsumerSubaccountIDsLabelKey)
+		if err != nil {
+			if !apperrors.IsNotFoundError(err) {
+				return errors.Wrap(err, fmt.Sprintf("Failed to get label for runtime with id: %s and key: %s", runtime.ID, s.ConsumerSubaccountIDsLabelKey))
+			}
+			// if the error is not found, do nothing and continue
```
pankrator, line 90: If you come in this comment you will anyway modify the labels of the runtime to empty array because the code continues to the `SetLabel` invocation. Perhaps `continue` should be added.

components/director/internal/tenantfetchersvc/subscriber.go
ataleksandrov: This is not logically equivalent. In case of not found error the label is nil, so that is why the else is present as well.

components/director/internal/tenantfetchersvc/subscriber.go
ataleksandrov: see above comment

chart/compass/charts/tenant-fetcher/templates/deployment.yaml
```diff
@@ -65,6 +65,16 @@ spec:
               value: {{ .Values.global.tenantFetcher.tenantProvider.subdomainProperty }}
             - name: APP_TENANT_PROVIDER
               value: {{ .Values.global.tenantFetcher.tenantProvider.name }}
+            - name: APP_TENANT_PROVIDER_SUBSCRIPTION_CONSUMER_ID_PROPERTY
+              value: {{ .Values.global.tenantFetcher.tenantProvider.subscriptionConsumerIdProperty }}
+            - name: APP_REGION_LABEL_KEY
```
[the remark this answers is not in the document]
ataleksandrov: It is inside the internal folder.. This is not a problem here but in the e2e tests must be hardcoded

components/director/internal/tenantfetchersvc/subscriber.go
```diff
@@ -0,0 +1,124 @@
...
+
+	return nil
+}
+
+func addElement(slice []string, elem string) []string {
+	return append(slice, elem)
+}
+
+func removeElement(slice []string, elem string) []string {
+	result := make([]string, 0)
```
krasish, line 117:
```suggestion
result := make([]string, len(slice))
```

This way **result** won't be resized multiple times during the for loop. In fact, it won't ever resize.

components/director/internal/tenantfetchersvc/subscriber.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// Unsubscribe unsubscribes tenant from runtime.
+func (s *subscriber) Unsubscribe(ctx context.Context, tenantSubscriptionRequest *TenantSubscriptionRequest, region string) error {
+	return s.applyRuntimesSubscriptionChange(ctx, tenantSubscriptionRequest.SubscriptionProviderID, tenantSubscriptionRequest.SubaccountTenantID, region, removeElement)
+}
+
+func (s *subscriber) applyRuntimesSubscriptionChange(ctx context.Context, subscriptionConsumerID, subaccountTenantID, region string, mutateLabelsFunc sliceMutationFunc) error {
+	filters := []*labelfilter.LabelFilter{
+		labelfilter.NewForKeyWithQuery(s.SubscriptionProviderLabelKey, fmt.Sprintf("\"%s\"", subscriptionConsumerID)),
```
pankrator, line 66: `subscriptionConsumerID` perhaps this should be `subscriptionProviderID`

pankrator (changes requested):

components/director/internal/tenantfetchersvc/subscriber.go
PetarTodorovv: If I understand correctly we're using _**s.SubscriptionProviderLabelKey**_ to build a filter which will be used to get runtimes labeled with it, and for each runtime to add or remove(depending on the flow - subscribe/unsubscribe) subaccountID label. 
I was wondering who is responsible to set this _**s.SubscriptionProviderLabelKey**_ label because if it missing no runtimes will be returned for processing

pankrator (approved):

PR #2025, opened 2021-09-16
Adjust views used in ORD service

krasish:
**Description**

Adjust views used in ORD service to lookup provider_tenant_ids. The respective changes in the ORD Service can be found [here](https://github.com/kyma-incubator/ord-service/pull/42).

**Pull Request status**

- [X] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/schema-migrator/migrations/director/20210915134043_add_providers_to_views.up.sql
```diff
@@ -0,0 +1,325 @@
...
+CREATE OR REPLACE VIEW tenants_apis
+            (tenant_id, provider_tenant_id, id, app_id, name, description, group_name, default_auth, version_value, version_deprecated,
+             version_deprecated_since, version_for_removal, ord_id, short_description, system_instance_aware,
+             api_protocol, tags, countries, links, api_resource_links, release_status, sunset_date, changelog_entries,
+             labels, package_id, visibility, disabled, part_of_products, line_of_business, industry, ready, created_at,
+             updated_at, deleted_at, error, implementation_standard, custom_implementation_standard,
+             custom_implementation_standard_description, target_urls, extensible, successors, resource_hash)
+AS
+WITH apps_subaccounts AS (SELECT * FROM apps_subaccounts_func()),
+     consumers_provider_for_runtimes AS (SELECT * FROM consumers_provider_for_runtimes_func())
```
DimitarPetrov, line 60: Are those needed? Can you just Select from the function below?

components/schema-migrator/migrations/director/20210915134043_add_providers_to_views.up.sql
```diff
@@ -0,0 +1,325 @@
...
+            (
+                provider_tenant  jsonb,
+                consumer_tenants jsonb
+            )
+    LANGUAGE plpgsql
+AS
+$$
+BEGIN
+    RETURN QUERY
+        SELECT l1.value -> 0 AS provider_tenant, l2.value AS consumer_tenants
```
DimitarPetrov, line 41: You can use `l1.value ->> 0` to extract text and do not cast below.

DimitarPetrov: Also, keep in mind that our data (especially on dev) is not that consistent.

There are `global_subaccount_id` labels with empty values or even worse with values that are not UUIDs. This will make the cast below fail.

DimitarPetrov (changes requested):

components/schema-migrator/migrations/director/20210915134043_add_providers_to_views.up.sql
krasish: I thought there was a performance benefit since the apps_subaccounts CTE is used more than once so I thought that it would get calculated once and then read multiple times. However, after some reading on the Internet, I came to know that CTEs are not very performant in Postgresql. I executed an `EXPLAIN ANALYZE ...` function and actually got better results with functions only, so yes, I will remove them.

DimitarPetrov (approved):

PR #2026, opened 2021-09-16
Add Upsert logic for Tenants

desislavaa:
**Description**
We currently have a race condition where if the tenant fetcher deployment, and tenant fetcher jobs try to insert the same tenant, one of them will fail with "Unique constraint violation", and even if we ignore that error, the transaction is broken and no further tenants can be inserted (e.g. it fails on insterting the parent tenant,  then the child tenant cannot be inserted).

One way to fix that is by using the "Upsert" logic we already have with labels.
However, we don't really need to update anything from tenant-fetcher deployment side. On the other hand, we can update the tenant names in the tenant-fetcher job, if the tenant was added first. by the deployment (the API does not expect the tenant name in its requests payload).

- Add `UnsafeCreator` which deals with conflicts by simply ignoring them. It's "unsafe", because you cannot be sure that the ID you gave your tenant matches the actual ID in the database.
- Make the tenant fetcher service (deployment) use `CreatetManyIfNotExists`
- Make the tenant fetcher job use `UpsertManyIfNotExists`
- [additional] Run the tenant fetcher localli vua `./run.sh --tenant-fetcher`
- [additional] Expose tenant type and parent in GraphQL <- can help with enhancing E2E tests

**Pull Request status**

- [x] Implementation
- [x] Unit tests


components/director/internal/domain/tenant/service.go
```diff
@@ -164,20 +175,17 @@ func (s *labeledService) CreateManyIfNotExists(ctx context.Context, tenantInputs
...
-		return tenantFromDB.ID, s.upsertLabels(ctx, tenantFromDB.ID, subdomain, region)
+func (s *labeledService) createIfNotExists(ctx context.Context, tenant model.BusinessTenantMapping, subdomain, region string, action func(context.Context, model.BusinessTenantMapping) error) (string, error) {
+	if err := action(ctx, tenant); err != nil {
+		return "", err
 	}
 
-	if err = s.tenantMappingRepo.Create(ctx, tenant); err != nil {
-		return "", errors.Wrapf(err, "while creating tenant with ID %s and external ID %s", tenant.ID, tenant.ExternalTenant)
+	tenantFromDB, err := s.tenantMappingRepo.GetByExternalTenant(ctx, tenant.ExternalTenant)
+	if err != nil && !apperrors.IsNotFoundError(err) {
```
DimitarPetrov, line 184: Should return error on not found as well, because otherwise, you will get NPE in the line below.

Also, it is not expected to have NotFound right after insertion, such thing should not be ignored.

components/director/internal/domain/tenant/service.go
```diff
@@ -136,6 +137,16 @@ func (s *service) MultipleToTenantMapping(tenantInputs []model.BusinessTenantMap
 // CreateManyIfNotExists creates all provided tenants if they do not exist.
 // It creates or updates the subdomain and region labels of the provided tenants, no matter if they are pre-existing or not.
 func (s *labeledService) CreateManyIfNotExists(ctx context.Context, tenantInputs ...model.BusinessTenantMappingInput) error {
+	return s.upsertTenants(ctx, tenantInputs, s.tenantMappingRepo.UnsafeCreate)
+}
+
+// UpsertManyIfNotExists creates all provided tenants if they do not exist. If they do exist, they are internally updated.
+// It creates or updates the subdomain and region labels of the provided tenants, no matter if they are pre-existing or not.
+func (s *labeledService) UpsertManyIfNotExists(ctx context.Context, tenantInputs ...model.BusinessTenantMappingInput) error {
```
DimitarPetrov, line 145: `UpsertManyIfNotExists` is not so intuitive as the word `Uspert` indicates that it is Insert and if exists -> Update.
The `IfNotExists` is redundant and adds only confusion for me.

DimitarPetrov (changes requested):

DimitarPetrov (approved):

PR #2029, opened 2021-09-21
Obfuscate personal data in Prometheus

danielgospodinow:
**Description**
Currently, in Prometheus, we can accidentally store personal data since we have authentication flows in which identification IDs can be used to derive personal data.

- 

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/metrics/collector.go
```diff
@@ -9,7 +11,8 @@ import (
 
 // Config configures the behaviour of the metrics collector.
 type Config struct {
-	EnableClientIDInstrumentation bool `envconfig:"default=true,APP_METRICS_ENABLE_CLIENT_ID_INSTRUMENTATION"`
+	EnableClientIDInstrumentation bool     `envconfig:"default=true,APP_METRICS_ENABLE_CLIENT_ID_INSTRUMENTATION"`
+	CensoredFlows                 []string `envconfig:"APP_METRICS_CENSORED_FLOWS"`
```
PetarTodorovv, line 15: CensoredFlows is an array here but in the values.yaml file is a single string, shouldn't be model there as an array too? Or perhaps as a single value in the configuration?

danielgospodinow: As far as I'm informed, the `envconfig` library can't handle `[]`. It expects to parse arrays represented as a single string composed of comma separated values (reference: [its docs](https://pkg.go.dev/github.com/vrischmann/envconfig)).

PetarTodorovv (approved):

PR #2034, opened 2021-09-27
Retry AuditLog operation if fails

dzahariev:
**Description**

- Retry AL operation if fails

**Related issue(s)**
- Sporadically the AL operations fails, and pass upon retry. We are going to retry the operations in case they are failing on first attempt.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/components/gateway/go.mod
+++ b/components/gateway/go.mod
@@ -33,3 +33,5 @@ require (
 	gopkg.in/check.v1 v1.0.0-20200902074654-038fdea0a05b // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+require github.com/avast/retry-go v3.0.0+incompatible // indirect
--- a/components/gateway/go.mod
+++ b/components/gateway/go.mod
@@ -4,6 +4,7 @@ go 1.17
 
 require (
 	github.com/99designs/gqlgen v0.11.0 // indirect
+	github.com/avast/retry-go v3.0.0+incompatible // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.1.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
@@ -33,5 +34,3 @@ require (
 	gopkg.in/check.v1 v1.0.0-20200902074654-038fdea0a05b // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
-
-require github.com/avast/retry-go v3.0.0+incompatible // indirect
--- a/components/gateway/internal/auditlog/client.go
+++ b/components/gateway/internal/auditlog/client.go
@@ -76,22 +76,37 @@ func (c *Client) LogSecurityEvent(ctx context.Context, event model.SecurityEvent
 
 func (c *Client) sendAuditLog(ctx context.Context, req *http.Request) error {
 	logger := log.C(ctx)
+	responseStatusCode, err := c.sendAuditLogInternal(ctx, req)
+	if responseStatusCode != http.StatusCreated || err != nil {
+		logger.Infof("Will retry auditlog request.")
+		responseStatusCode, err = c.sendAuditLogInternal(ctx, req)
+		if responseStatusCode != http.StatusCreated || err != nil {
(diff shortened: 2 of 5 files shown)


components/gateway/internal/auditlog/client.go
```diff
@@ -76,22 +76,37 @@ func (c *Client) LogSecurityEvent(ctx context.Context, event model.SecurityEvent
 
 func (c *Client) sendAuditLog(ctx context.Context, req *http.Request) error {
 	logger := log.C(ctx)
+	responseStatusCode, err := c.sendAuditLogInternal(ctx, req)
+	if responseStatusCode != http.StatusCreated || err != nil {
+		logger.Infof("Will retry auditlog request.")
+		responseStatusCode, err = c.sendAuditLogInternal(ctx, req)
```
nyordanoff, line 82: Due to the fact that we make a `POST` request, we can't reuse the same `req` object for the second attempt (the retry) because the request body will be read from the first call and here we will try to make a `POST` request with an empty body which will result in a similar error:

` Post <url>: http: <someInfo> with Body length 0`

I faced the same issue and the most straightforward solution is either to build the request object before each http call or in order to reuse the `req` object to set the request body before each call

components/gateway/internal/auditlog/client.go
```diff
@@ -76,22 +76,37 @@ func (c *Client) LogSecurityEvent(ctx context.Context, event model.SecurityEvent
...
+		if responseStatusCode != http.StatusCreated || err != nil {
+			return errors.Wrapf(err, "Write to auditlog failed with status code: %d", responseStatusCode)
+		}
+	}
+	return nil
+}
+
+func (c *Client) sendAuditLogInternal(ctx context.Context, req *http.Request) (int, error) {
+	logger := log.C(ctx)
+	var err error = nil
```
nyordanoff, line 92: the default value is nil so this can be simplified to: `var err error`

components/gateway/internal/auditlog/client.go
```diff
@@ -76,22 +76,37 @@ func (c *Client) LogSecurityEvent(ctx context.Context, event model.SecurityEvent
...
 	response, err := c.httpClient.Do(req)
 	if err != nil {
-		return errors.Wrapf(err, "while sending auditlog to: %s", req.URL.String())
+		return 0, errors.Wrapf(err, "while sending auditlog to: %s", req.URL.String())
 	}
 	defer httpcommon.CloseBody(ctx, response.Body)
 
 	if response.StatusCode != http.StatusCreated {
+		var output []byte = nil
```
nyordanoff, line 100: the default value is nil so this can be simplified to: `var output []byte`

nyordanoff (changes requested):

components/gateway/internal/auditlog/client.go
```diff
@@ -76,22 +76,37 @@ func (c *Client) LogSecurityEvent(ctx context.Context, event model.SecurityEvent
 
 func (c *Client) sendAuditLog(ctx context.Context, req *http.Request) error {
 	logger := log.C(ctx)
+	responseStatusCode, err := c.sendAuditLogInternal(ctx, req)
+	if responseStatusCode != http.StatusCreated || err != nil {
```
DimitarPetrov, line 80: Consider using github.com/avast/retry-go in order to mitigate different http req / body issues.

I think this is the library that is used across the project, you can find some usages for reference.

components/gateway/go.mod
```diff
@@ -33,3 +33,5 @@ require (
 	gopkg.in/check.v1 v1.0.0-20200902074654-038fdea0a05b // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+require github.com/avast/retry-go v3.0.0+incompatible // indirect
```
DimitarPetrov, line 37: Why is this outside of the `require` block above?

DimitarPetrov (approved):

components/gateway/go.mod
dzahariev: Automation :-(

DimitarPetrov (approved):

DimitarPetrov (approved):

nyordanoff (approved):

PR #2044, opened 2021-09-29
Utilize application templates in System Fetcher

desislavaa:
**Description**
We need the additional labels provided by the present templates. In order to use the whole template capability, we need to build the inbut via the template.
- Render the application input by using the template matching the fetched system

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/systemfetcher/service.go
```diff
@@ -227,21 +239,75 @@ func (s *SystemFetcher) processSystemsForTenant(ctx context.Context, tenantMappi
...
+		return nil, errors.Wrapf(err, "while preparing ApplicationRegisterInput model from Application Template with name %s", appTemplate.Name)
+	}
+
+	return enrichAppRegisterInput(appRegisterInput, sc), nil
+}
+
+func enrichAppRegisterInput(input model.ApplicationRegisterInput, sc System) *model.ApplicationRegisterInputWithTemplate {
+	initStatusCond := model.ApplicationStatusConditionInitial
+
+	input.Name = sc.DisplayName
```
PetarTodorovv, line 277: Do we need `input.Name = sc.DisplayName` since the name will be already populated in `s.appConverter.CreateInputJSONToModel` with the `name` placeholder value which is `s.DisplayName`.

And also, in the `input.Description = &sc.ProductDescription` assignment, the description will be populated in `s.appConverter.CreateInputJSONToModel` with the `display-name` placeholder value and if this value is different from `&sc.ProductDescription` it will be overridden. Is this the expected behavior?

components/director/internal/systemfetcher/service.go
```diff
@@ -227,21 +239,75 @@ func (s *SystemFetcher) processSystemsForTenant(ctx context.Context, tenantMappi
...
+	return &model.ApplicationRegisterInputWithTemplate{
+		ApplicationRegisterInput: input,
+		TemplateID:               sc.TemplateID,
+	}
+}
+
+func getTemplateInputs(t *model.ApplicationTemplate, s System) model.ApplicationFromTemplateInputValues {
+	inputValues := model.ApplicationFromTemplateInputValues{}
+	for _, p := range t.Placeholders {
+		if strings.Contains(strings.ToLower(p.Name), "url") {
```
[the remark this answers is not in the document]
desislavaa: yep, those are the templates we have registered in our prod DBs, however, one can register whatever template with whatever placeholders, which makes the "fill-in" here a bit harder.
The dummy thing I tried to do was guess what the placeholder can actually "hold", and what we can provide, so the URL here is just a wild guess. There was also an if that checks if the placeholder contains "name", so it fills it in with the system name, but then I tought "what if its something completely different?". 
We can also do the "if contains "name"" thing, and for other cases just pass "unknown", what do you think?

components/director/internal/systemfetcher/service.go
desislavaa: I assumed we want the data from the external registry, since the templates we currently have provide almost no real data, just system names, and types. We can sync this with the team.

components/director/internal/systemfetcher/service.go
```diff
@@ -227,21 +239,75 @@ func (s *SystemFetcher) processSystemsForTenant(ctx context.Context, tenantMappi
...
-		ProviderName:    &sc.InfrastructureProvider,
-		StatusCondition: &initStatusCond,
-		SystemNumber:    &sc.SystemNumber,
-		Labels: map[string]interface{}{
-			"managed": "true",
-		},
-	}
-
-	return appRegisterInput, sc.TemplateID
+		Name: sc.DisplayName,
```
DimitarPetrov, line 248: This is set in the `enrichAppRegisterInput` below as well.

components/director/internal/systemfetcher/service.go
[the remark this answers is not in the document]
PetarTodorovv: If we have any placeholder and values in the template it will be tricky/hard to map these values to the external registry values, we can think about this scenario and how to handle it. 
About the "if thing" my concern is similar to yours - I'm not sure what value we'll get from that placeholder and how will be fit in the application. If we assume the system name will be in `name` placeholder, it's okay in this case but image we have another placeholder containing `*name*` which is with completely different semantic, then we'll set  the system name on it?

P.S. These are a conceptual questions that came up to my mind when I was reviewing the PR, they may not be relevant at all, or they may not need to be handled at this point.

components/director/internal/systemfetcher/service.go
```diff
@@ -227,21 +239,75 @@ func (s *SystemFetcher) processSystemsForTenant(ctx context.Context, tenantMappi
...
+	}
+	input.Labels["managed"] = "true"
+
+	return &model.ApplicationRegisterInputWithTemplate{
+		ApplicationRegisterInput: input,
+		TemplateID:               sc.TemplateID,
+	}
+}
+
+func getTemplateInputs(t *model.ApplicationTemplate, s System) model.ApplicationFromTemplateInputValues {
```
DimitarPetrov, line 295: This function populates the placeholders just for the sake of the service layer to work and add all the static (directly hardcoded things) like type labels from the template.

As such I think we have two options:

1. Either mark it as a fully dummy/workaround function (and document it as such with a godoc) and then assume that everything will be populated by the enrich function directly.
2. Another option is to let it at least try to resolve the fields it can. The best way I can think of is to parse the json tags of the System struct and then search for them in the placeholder. If a similar placeholder is found we then can use the value to populate it. For all placeholders that do not match any json tag we can use the default value (ex. DisplayName)

If we choose option two we should choose what we do if a template matches. Should we identify this in the enrich or the properties we have there are always overridden as right now.

My impression is regarding the `url` "guess". I think that we should either abstain on any trials for real populating the placeholders or do whatever we can in order to make it usable in the future (if we use json tags we know what a system will contain and we can extend the templates to use it - not sure if it actually matters if we override it anyways in the `enrichAppRegisterInput`)

tests/system-fetcher/tests/handler_test.go
```diff
@@ -687,3 +698,18 @@ func cleanupMockSystems(t *testing.T) {
 	}
 	log.D().Info("Successfully reset mock systems")
 }
+
+func applicationLabels(name string, fromTemplate bool) directorSchema.Labels {
+	labels := directorSchema.Labels{
+		"scenarios":           []interface{}{"DEFAULT"},
```
DimitarPetrov, line 704: The `DEFAULT` scenario looks a little risky. What will happen when we run the system fetcher and it fetched a few hundred systems and add them all in the DEFAULT scenario. Those will make into all the kyma runtimes in the DEFAULT scenario which is unexpected for the clients.

Is this DEFAULT scenario come from the template processing that we use now and we didn't used before?

DimitarPetrov (changes requested):
Overall I think that there are a few things that can be discussed in the team.

Not 100% sure if using the template at all is desired in this use case. It feels more of a workaround than a feature to me.

tests/system-fetcher/tests/handler_test.go
desislavaa: the "DEFAULT" scenario was always there, we just didn't validate the systems labels before

components/director/internal/systemfetcher/service.go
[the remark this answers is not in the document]
desislavaa: You can check my answer to Dimitar below

components/director/internal/systemfetcher/service.go
desislavaa: Or we can improve the placeholders, and add a new field which will hold the mapping to a field of the systems fetched from an external source?

DimitarPetrov: I agree now that the templates make sense to be used in this case. 

What do you think about going all-in for the templates and change the templates to match the pattern of the systems from an external source?
This means that the `enrichAppRegisterInput` will no longer be needed and the template will take care of that.

However, with this approach, we should add one more field for each placeholder: `optional bool`
Where all external system-specific properties will be optional and therefore ignored in the standard case.

This will make the templates a bit more usable because the Cockpit can be extended to generate a form based on the placeholders and if they are optional or required. Then when a person registers a system from a template can add more metadata to the id than the name (like systemNumber etc..) if he knows about this stuff.

The system fetcher will also be simplified because it will work with the template only and will not assume a priority of the information sources.

components/director/internal/model/app_template.go
```diff
@@ -50,6 +50,22 @@ type ApplicationFromTemplateInput struct {
 	Values       ApplicationFromTemplateInputValues
 }
 
+// ApplicationRegisterInputPlaceholderValue is used to add semantics to a placeholder value, and suggest its content.
+type ApplicationRegisterInputPlaceholderValue string
```
DimitarPetrov, line 54: Why is this needed and how it is different from the placeholder's name?

DimitarPetrov (changes requested):

PR #2045, opened 2021-09-30
Fix all system auths are assumed to be OneTimeTokens

DimitarPetrov:
There is an issue with OneTimeToken validation. Certificate system auths has nil Value and this is mistakenly assumed to be OneTimeToken.

In addition, if this was not the case a nil pointer dereference panic happens a few lines below since we access sa.Value.OneTimeToken, but Value is nil for certificates.

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -75,7 +75,7 @@ global:
       version: "PR-2027"
     director:
       dir:
-      version: "PR-2043"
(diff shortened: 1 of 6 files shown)


dzahariev (approved):

dzahariev (changes requested):
Number

dzahariev (approved):

PR #2047, opened 2021-10-01
Store subaccounts as tenants

StefanCholakov:
Store subaccounts as tenants

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/tenantfetcher/service_test.go
```diff
@@ -94,51 +83,33 @@ func TestService_SyncTenants(t *testing.T) {
...
 		tenantFieldMapping.CustomerIDField: parent1,
 		tenantFieldMapping.SubdomainField:  "subdomain-1",
+		tenantFieldMapping.EntityTypeField: "GlobalAccount",
 	}
 	event2Fields := map[string]string{
 		tenantFieldMapping.IDField:         busTenant2.ExternalTenant,
 		tenantFieldMapping.NameField:       "bar",
 		tenantFieldMapping.CustomerIDField: parent2,
 		tenantFieldMapping.SubdomainField:  "subdomain-2",
+		tenantFieldMapping.EntityTypeField: "GlobalAccount",
```
PetarTodorovv, line 93: Shouldn't we have also a tests for the new entity type - Subaccount?

PetarTodorovv (changes requested):

chart/compass/templates/subaccount-fetcher-job.yaml
```diff
@@ -0,0 +1,272 @@
+{{ $outerScope := . }}
```
desislavaa, line 1: This is basically the same chart as the old TF job, except for some missing properties, can we somehow reuse the existing one?

components/director/internal/tenantfetcher/event_page.go
```diff
@@ -95,41 +111,86 @@ func (ep eventsPage) eventDataToTenant(eventType EventsType, eventData []byte) (
...
+		if !regionField.Exists() {
+			return nil, invalidFieldFormatError(ep.fieldMapping.RegionField)
+		}
+		region = regionField.String()
+		parentIDField := gjson.Get(jsonPayload, ep.fieldMapping.ParentIDField)
+		if !parentIDField.Exists() {
+			return nil, invalidFieldFormatError(ep.fieldMapping.ParentIDField)
+		}
+		parentID = parentIDField.String()
+		tenantType = tenant.TypeToStr(tenant.Subaccount)
```
desislavaa, line 166: This one should be configurable - `globalAccountRegex := regexp.MustCompile("^GLOBALACCOUNT_.*|GlobalAccount")`

Also this whole block is a bit larger than we would usually want. My suggestion is to extract two functions here, which can be responsible for building the tenant of the type we want.

PR #2050, opened 2021-10-04
Hydrator token handling

StanislavStefanov:
**Description**

- Add our generic authn as hydrator to the compass-gateway-sap-mtls gateway
- Add a map with tenants to the idtoken
- Configure authentication in which claim to put the tenant identifier

**Pull Request status**

- [X] Implementation
- [ ] Unit tests
- [ ] Integration tests


PetarTodorovv (changes requested):

components/director/internal/oathkeeper/model_test.go
```diff
@@ -42,232 +37,6 @@ func TestAuthFlow(t *testing.T) {
 	})
 }
 
-func TestReqData_GetAuthID(t *testing.T) {
```
krasish, line 45: Deleting might be reasonable for some of these tests but also some flows are probably no longer covered. The test coverage for the _model.go_ file from this PR is about 60%. For comparison, from the main brach it is 85%. Consider adding some tests.

components/gateway/pkg/proxy/transport.go
```diff
@@ -161,10 +179,39 @@ func (t *Transport) getClaims(headers http.Header) (Claims, error) {
...
+		return Claims{}, errors.Wrap(err, "while extracting consumers from token")
 	}
+
+	err = json.Unmarshal([]byte(tokenClaims.TenantString), &tokenClaims.Tenant)
+	if err != nil {
+		return Claims{}, errors.Wrap(err, "while extracting tenants from token")
+	}
+
+	claims := Claims{
+		ConsumerTenant: tokenClaims.Tenant["consumerTenant"],
```
krasish, line 200: "consumerTenant" can be exported to a const or if such one form other package exists - it can be reused

krasish (changes requested):

krasish:
Good job on the object contexts’ keys abstraction :)

components/director/internal/oathkeeper/model_test.go
StanislavStefanov: GetAuthId was split into the Match methods of ObjectContextProviders. All test cases were moved to the coresponding test files.  I will add some tests for ExtractCoordinates.

KirilKabakchiev (changes requested):

components/director/internal/authnmappinghandler/handler.go
```diff
@@ -120,28 +124,41 @@ func (h *Handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 	reqData, err := h.reqDataParser.Parse(req)
 	if err != nil {
 		h.logError(ctx, err, "An error has occurred while parsing the request.")
-		http.Error(writer, "Unable to parse request data", http.StatusUnauthorized)
+		http.Error(writer, "Unable to parse request data", http.StatusOK)
+		return
+	}
+
+	path := req.URL.Path
+	matchedAuthenticator := strings.TrimPrefix(path, "/authn-mapping/")
```
KirilKabakchiev, line 132: why not use the standard way to get the path param ?
```
vars := mux.Vars(r)
    authenticationParamter, ok := vars[":authentication"]
    if !ok {
       ....
    }
```

PR #2051, opened 2021-10-04
Fetch ORD Configuration endpoint with webhook credentials

NickyMateev:
# Fetch ORD Configuration endpoint with webhook credentials

**Description**

- Add configuration `securedApplicationTypes`
- Adopt ORD Aggregator to honour `securedApplicationTypes` and use webhook credentials for applications which are from secured types

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/pkg/http/request.go
```diff
@@ -0,0 +1,62 @@
...
+	"net/http"
+
+	"github.com/kyma-incubator/compass/components/director/internal/model"
+	"github.com/kyma-incubator/compass/components/director/pkg/apperrors"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+func RequestWithCredentials(ctx context.Context, client *http.Client, url string, auth *model.Auth) (*http.Response, error) {
```
DimitarPetrov, line 14: Since those functions `RequestWithCredentials` and `RequestWithoutCredentials` are in a generic context now `httpuntil` and not in the fetch request I think it makes sense to name them:

`GetRequestWithCredentials` and `GetRequestWithoutCredentials`

components/director/pkg/http/request_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+func newTestClient(fn RoundTripFunc) *http.Client {
+	return &http.Client{
+		Transport: fn,
+	}
+}
+
+func TestRequestWithoutCredentials_Success(t *testing.T) {
+	client := newTestClient(func(req *http.Request) *http.Response {
+		return expectedResp
```
DimitarPetrov, line 38: Check here that `Authorization` header is empty.

components/director/pkg/http/request_test.go
```diff
@@ -0,0 +1,132 @@
...
+var expectedResp = &http.Response{
+	StatusCode: http.StatusOK,
+	Body:       nil,
+}
+
+type RoundTripFunc func(req *http.Request) *http.Response
+
+func (f RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f(req), nil
+}
```
DimitarPetrov, line 28: It would be better to check if `f` returns nil and return error from the `RoundTrip` then the `TestRequestWithoutCredentials_FailedRequest` and all the other negative tests will return the actual error.

components/director/pkg/http/request_test.go
```diff
@@ -0,0 +1,132 @@
...
+	})
+
+	_, err := httputil.RequestWithoutCredentials(client, testURL)
+	require.Error(t, err)
+}
+
+func TestRequestWithCredentials_SuccessWithBasicAuth(t *testing.T) {
+	client := newTestClient(func(req *http.Request) *http.Response {
+		return expectedResp
+	})
```
DimitarPetrov, line 58: Check that user:pass are present in the `Authorization` header.

components/director/pkg/http/request_test.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+	})
+	require.Error(t, err)
+}
+
+func TestRequestWithCredentials_SuccessWithOAuth(t *testing.T) {
+	client := newTestClient(func(req *http.Request) *http.Response {
+		if strings.Contains(req.URL.String(), testURL) {
+			return expectedResp
+		}
```
DimitarPetrov, line 92: Check for `test-tkn` Authorization header here.

components/director/pkg/http/request_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+		data, err := json.Marshal(struct {
+			AccessToken string `json:"access_token"`
+		}{
+			AccessToken: "test-tkn",
+		})
```
DimitarPetrov, line 98: Assert the client credentials before issuing the token.

components/director/internal/domain/label/repository.go
```diff
@@ -127,6 +129,27 @@ func (r *repository) ListByKey(ctx context.Context, tenant, key string) ([]*mode
 	return labels, nil
 }
 
+// ListGlobalByKeyAndObjects missing godoc
```
DimitarPetrov, line 132: 👎

components/director/internal/open_resource_discovery/service.go
```diff
@@ -88,6 +94,26 @@ func (s *Service) listAppPage(ctx context.Context, pageSize int, cursor string)
...
+		return nil, err
+	}
+
+	appLabelsMap := make(map[string]interface{}, page.TotalCount)
+	for i := range labels {
+		appLabelsMap[labels[i].ObjectID] = labels[i].Value
+	}
+
+	for i, _ := range page.Data {
+		page.Data[i].Labels = []byte(fmt.Sprintf(`{"%s": "%s"}`, applicationTypeLabel, appLabelsMap[page.Data[i].ID]))
```
DimitarPetrov, line 114: Those are the ORD Labels and not CMP Lables.
Currently, it is safe to override them because we are not updating the app when we fetch the systemInstance property from the document. However, I think it is better not to mess with them as we will do it in the future probably.

I think a better approach (and easier considering this JSON manipulations) is to add `applicatioType` field in the system which is `json:"-"` ignored and use it as a temporary metadata field for this flow.

components/director/internal/features/features.go
```diff
@@ -2,6 +2,7 @@ package features
 
 // Config missing godoc
 type Config struct {
-	DefaultScenarioEnabled bool   `envconfig:"default=true,APP_DEFAULT_SCENARIO_ENABLED"`
-	ProtectedLabelPattern  string `envconfig:"default=.*_defaultEventing,APP_PROTECTED_LABEL_PATTERN"`
+	DefaultScenarioEnabled  bool     `envconfig:"default=true,APP_DEFAULT_SCENARIO_ENABLED"`
+	ProtectedLabelPattern   string   `envconfig:"default=.*_defaultEventing,APP_PROTECTED_LABEL_PATTERN"`
+	SecuredApplicationTypes []string `envconfig:"default=empty,APP_SECURED_APPLICATION_TYPES"`
```
DimitarPetrov, line 7: Is the envconfig library works with slices and how?

Is it comma-separated in the chart (currently we have only one value, but I am curious how it will work in the future)

Probably it is a good idea to document it with a comment in the chart or config repo in order to streamline the maintance in the future.

components/director/internal/open_resource_discovery/client_test.go
```diff
@@ -26,39 +29,120 @@ func NewTestClient(fn RoundTripFunc) *http.Client {
...
+		} else if strings.Contains(string(reqBody), "client_secret") {
+			statusCode = http.StatusOK
+			data, err = json.Marshal(struct {
+				AccessToken string `json:"access_token"`
+			}{
+				AccessToken: "test-tkn",
+			})
+			require.NoError(t, err)
+		} else if strings.Contains(string(reqBody), "grant_type=client_credentials") {
+			statusCode = http.StatusOK
```
DimitarPetrov, line 57: I think this case will match the if above and never reach this one?

components/director/internal/open_resource_discovery/interfaces.go
```diff
@@ -6,6 +6,12 @@ import (
 	"github.com/kyma-incubator/compass/components/director/internal/model"
 )
 
+// LabelRepository missing godoc
```
DimitarPetrov, line 9: 👎

You can make the interface unexported as well.

components/director/internal/open_resource_discovery/service_test.go
```diff
@@ -48,6 +56,17 @@ func TestService_SyncORDDocuments(t *testing.T) {
 		return persistTx, transact
 	}
 
+	successfulLabelRepo := func() *automock.LabelRepository {
+		labelRepo := &automock.LabelRepository{}
+		labelRepo.On("ListGlobalByKeyAndObjects", txtest.CtxWithDBMatcher(), model.ApplicationLabelableObject, mock.Anything, applicationTypeLabel).Return([]*model.Label{
```
DimitarPetrov, line 61: You can make a custom mockery matcher that matches slices with just one element instead of mock.Anything. Because you return just one label. This will make the test a little bit more exhaustive.

components/director/internal/open_resource_discovery/service_test.go
```diff
@@ -1561,6 +1641,10 @@ func TestService_SyncORDDocuments(t *testing.T) {
 	for _, test := range testCases {
 		t.Run(test.Name, func(t *testing.T) {
 			_, tx := test.TransactionerFn()
+			labelRepo := &automock.LabelRepository{}
```
DimitarPetrov, line 1644: Add it to AssertExpectations below or it won't do anything.

components/external-services-mock/internal/ord-aggregator/handler.go
```diff
@@ -1,21 +1,91 @@
...
+type ordHandler struct {
+	mutex    sync.Mutex
+	secured  bool
+	username string
+	password string
+}
+
+func NewORDHandler() *ordHandler {
+	return &ordHandler{
+		mutex:   sync.Mutex{},
```
DimitarPetrov, line 24: You can make it Read Write Lock and lock `HandleFuncOrdConfig` for Read and `HandleFuncOrdConfigSecurity` for Write in order to mitigate many readers lock each other.

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -72,205 +77,251 @@ const (
...
+				secondAppInput.Labels = directorSchema.Labels{applicationTypeLabelKey: []interface{}{testConfig.SecuredApplicationTypes[0]}}
+				thirdAppInput = fixtures.FixSampleApplicationRegisterInputWithORDSecuredWebhooks(expectedThirdSystemInstanceName, expectedThirdSystemInstanceDescription, testConfig.ExternalServicesMockAbsoluteURL, ordConfigSecurity.Username, ordConfigSecurity.Password)
+				thirdAppInput.Labels = directorSchema.Labels{applicationTypeLabelKey: []interface{}{testConfig.SecuredApplicationTypes[0]}}
+			} else {
+				appInput = fixtures.FixSampleApplicationRegisterInputWithORDWebhooks(expectedSystemInstanceName, expectedSystemInstanceDescription, testConfig.ExternalServicesMockBaseURL)
+				secondAppInput = fixtures.FixSampleApplicationRegisterInputWithORDWebhooks(expectedSecondSystemInstanceName, expectedSecondSystemInstanceDescription, testConfig.ExternalServicesMockBaseURL)
+				thirdAppInput = fixtures.FixSampleApplicationRegisterInputWithORDWebhooks(expectedThirdSystemInstanceName, expectedThirdSystemInstanceDescription, testConfig.ExternalServicesMockAbsoluteURL)
+			}
```
DimitarPetrov, line 129: Can we encapsulate this `secured: true/false` in `fixtures.FixSampleApplicationRegisterInputWithORDWebhooks`
And don't have this branching here. The fixture can accept a bool parameter and create the relevant Webhook and labels based on the sceurity.

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -72,205 +77,251 @@ const (
...
 
-	bundlesAPIsData := make(map[string][]string)
-	bundlesAPIsData[expectedBundleTitle] = []string{firstAPIExpectedTitle}
-	bundlesAPIsData[secondExpectedBundleTitle] = []string{firstAPIExpectedTitle}
+	defer func() { // Cleanup just in case (remove security)
+		ordConfigSecurity.Enabled = false
+		toggleORDConfigSecurity(t, testConfig.ExternalServicesMockBaseURL, ordConfigSecurity)
+	}()
+
+	for _, secured := range []bool{false, true} {
```
DimitarPetrov, line 112: Not sure if needed, but currently this will create the test for unsecured well-known and secured with basic.

However, based on the implementation, technically we support OAuth secured well-known endpoint as well. 

Should we add an e2e test for it as well?

Another thing to consider here is that ord_aggregator tests a pretty slow (~5 mins) now you make it double. If you address this comment - triple. And with the certificate security - maybe 4 times slower ~20min

DimitarPetrov (changes requested):

components/director/internal/features/features.go
NickyMateev: Yes, multiple values are passed by setting a comma-separated string.

components/director/internal/open_resource_discovery/client_test.go
NickyMateev: The oauth lib sends two requests, the first one is matched by the above case, and the second one by this here

tests/ord-aggregator/tests/handler_test.go
NickyMateev: Yeah, I had a similar concern about re-running everything with the only difference being that the 3 apps are once secured and the other iteration they are not.
The best solution IMO would be for the external services mock to have different servers (ports probably) serving the ord content of each app. This way we can customize only specific ord handlers and not the one and only one that represents all 3 apps. So I'll see how to do that and I imagine having again 3 apps:
- 1 unsecured
- 1 basic
- 1 oauth

Eventually 1 for certs.

DimitarPetrov: Maybe we should have 2 unsecured - one with baseURL and one with Absolute.

And yes I think this will solve the issue as the tests will still need 1 aggregator run.

For each authentication mechanism, we can just register another system with another type of webhook.

From external services mock PoV this will be just different handlers.

components/director/internal/open_resource_discovery/service_test.go
NickyMateev: Good catch 👍

PR #2057, opened 2021-10-06
Aggregate ORD metadata with Cert Auth

DimitarPetrov:
**Description**

- Implementation of mTLS ORD Access Strategy based on external certificate issuer service.
- Certificate issuer mock in external services mock for local testing.
- Configuration and client for getting client certificates from external issuer.

- [x] Implementation
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/components/external-services-mock/internal/ord-aggregator/handler.go
+++ b/components/external-services-mock/internal/ord-aggregator/handler.go
@@ -57,8 +57,13 @@ func (oh *ordHandler) HandleFuncOrdConfig(baseURL, accessStrategy string) func(r
 			}
 		}
 
+		var baseURLFormat string
+		if len(baseURL) > 0 {
+			baseURLFormat = fmt.Sprintf(`"baseUrl": "%s",`, baseURL)
+		}
+
 		rw.WriteHeader(http.StatusOK)
-		_, err := rw.Write([]byte(fmt.Sprintf(ordConfig, baseURL, accessStrategy)))
+		_, err := rw.Write([]byte(fmt.Sprintf(ordConfig, baseURLFormat, accessStrategy)))
 		if err != nil {
 			httphelpers.WriteError(rw, errors.Wrap(err, "error while writing response"), http.StatusInternalServerError)
 		}


components/director/internal/open_resource_discovery/accessstrategy/cmp_mtls_access_strategy.go
```diff
@@ -0,0 +1,89 @@
...
+		if err := as.initialize(ctx, baseClient); err != nil {
+			return nil, errors.Wrap(err, "while initializing access strategy sap:cmp-mtls:v1")
+		}
+	}
+	return as.client.Get(documentURL)
+}
+
+// SetConfig sets the Access Strategy config. This is used when reading from environment is not suitable in that context.
+func (as *cmpMTLSAccessStrategyExecutor) SetConfig(config cert.CertSvcConfig) {
+	as.config = &config
```
danielgospodinow, line 43: Shouldn't this be also synchronized?

components/director/pkg/cert/certsvc_client.go
```diff
@@ -0,0 +1,265 @@
...
+		AuthStyle:    oauth2.AuthStyleAutoDetect,
+	}
+
+	client := ccConf.Client(ctxWithClient)
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
```
KirilKabakchiev, line 180: resource leak/defer close the response body

components/director/internal/open_resource_discovery/accessstrategy/access_strategy.go
```diff
@@ -0,0 +1,72 @@
...
+type AccessStrategyType string
+
+// IsSupported checks if the given AccessStrategy is supported by CMP
+func (a AccessStrategyType) IsSupported() bool {
+	_, ok := supportedAccessStrategies[a]
+	return ok
+}
+
+// Execute executes given access strategy if supported, UnsupportedAccessStrategyErr otherwise
+func (a AccessStrategyType) Execute(ctx context.Context, client *http.Client, url string) (*http.Response, error) {
```
KirilKabakchiev, line 30: this should be a method on the strategy, atleast logically you dont execute the type. I understand that here you've tried to hide looking within the map of strategies and checking if executor for type exists - probably the correct flow should be ExecutorProvider.Provide(strategy string) and then Executor.Execute (rather than strategyTypes.IsSupported and then strategyType.Execute)

components/external-services-mock/cmd/main.go
```diff
@@ -56,9 +71,40 @@ func main() {
...
+			ClientCAs:  caCertPool,
+			ClientAuth: tls.RequireAndVerifyClientCert,
+		},
+	}
+
+	wg := &sync.WaitGroup{}
+	wg.Add(2)
+
+	go startServer(ctx, server, wg)
+	go startServer(ctx, certSecuredServer, wg)
```
pankrator, line 105: The secured server should be started with `ListenAndServerTLS` in order for the client cert authentication to work

components/director/internal/open_resource_discovery/accessstrategy/cmp_mtls_access_strategy.go
```diff
@@ -0,0 +1,92 @@
...
+	if !as.isInitialized() {
+		if err := as.initialize(ctx, baseClient); err != nil {
+			return nil, errors.Wrap(err, "while initializing access strategy sap:cmp-mtls:v1")
+		}
+	}
+	return as.client.Get(documentURL)
+}
+
+// SetConfig sets the Access Strategy config. This is used when reading from environment is not suitable in that context.
+func (as *cmpMTLSAccessStrategyExecutor) SetConfig(config cert.CertSvcConfig) {
```
KirilKabakchiev, line 42: thats a weird one - whats the purpose ? usually if you have to reconfigure something during runtime, its better to instantiate a new object every time

components/director/internal/open_resource_discovery/accessstrategy/access_strategy.go
DimitarPetrov: But the AccessStrategy itself is just a struct with only a type field. Then in the end it is a more complicated way to execute something based on the acessStrategy's type.

chart/compass/values.yaml
```diff
@@ -380,6 +380,20 @@ global:
     dbPool:
       maxOpenConnections: 2
       maxIdleConnections: 2
+    accessStrategies:
+      cmpMtls:
```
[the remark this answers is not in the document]
DimitarPetrov: This has nothing to do with CAs. This is an ordAggregator specific config for each access strategy.

Then each access strategy defines its own Env Prefix and the configuration is wired in the job. Can you elaborate a bit?

PR #2063, opened 2021-10-11
Mutation for application unpair

alextargov:
**Description**

- Support Unpair application flow
- Introduce async operation for unpair which should use a provided external system webhook that deletes the pairing information in the LoB application. We also do not delete the application's data in our database. When the unpairing is successful, the status condition for the application is set to **INITIAL**, and if it's not successful there is a new status condition provided for that - **UNPAIRING_FAILED**.
- Delete hydra oauth client for the application as well as System Auth credentials data

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/components/director/internal/domain/application/service.go
+++ b/components/director/internal/domain/application/service.go
@@ -503,7 +503,7 @@ func (s *service) DeleteLabel(ctx context.Context, applicationID string, key str
 	return nil
 }
 
-// ensureApplicationNotPartOfScenarioWithRuntime Checks if an application has scenarios associated with it. If the scenarios are part of a runtime, then the application is considered being used by that runtime.
+// ensureApplicationNotPartOfScenarioWithRuntime Checks if an application has scenarios associated with it. if a runtime is part of any scenario, then the application is considered being used by that runtime.
 func (s *service) ensureApplicationNotPartOfScenarioWithRuntime(ctx context.Context, tenant, appID string) error {
 	scenarios, err := s.getScenarioNamesForApplication(ctx, appID)
 	if err != nil {
--- a/components/director/internal/domain/application/service_test.go
+++ b/components/director/internal/domain/application/service_test.go
@@ -2466,7 +2466,7 @@ func TestService_Delete(t *testing.T) {
 			ExpectedErrMessage: testErr.Error(),
 		},
 		{
-			Name: "Returns error when application is part of a scenario and a runtime",
+			Name: "Returns error when application is part of a scenario with runtime",
 			AppRepoFn: func() *automock.ApplicationRepository {
 				repo := &automock.ApplicationRepository{}
 				repo.AssertNotCalled(t, "Delete")
@@ -2600,8 +2600,7 @@ func TestService_Unpair(t *testing.T) {
 
 				return ctx
 			},
-			InputID:            id,
-			ExpectedErrMessage: "",
+			InputID: id,
 		},
 		{
 			Name: "Success when application is part of a scenario but not with runtime",
@@ -2627,8 +2626,7 @@ func TestService_Unpair(t *testing.T) {
 
 				return ctx
 			},
-			InputID:            id,
-			ExpectedErrMessage: "",
+			InputID: id,
 		},
 		{
 			Name: "Success when operation type is SYNC and sets the application status to INITIAL",
@@ -2649,8 +2647,7 @@ func TestService_Unpair(t *testing.T) {
 				repo.On("ListAll", ctx, applicationModel.Tenant, mock.Anything).Return(scenarioLabel).Return([]*model.Runtime{}, nil)
 				return repo
 			},
-			InputID:            id,
-			ExpectedErrMessage: "",
+			InputID: id,
 			ContextFn: func() context.Context {
 				backgroundCtx := context.Background()
 				return backgroundCtx
@@ -2675,8 +2672,7 @@ func TestService_Unpair(t *testing.T) {
 				repo.On("ListAll", mock.Anything, applicationModel.Tenant, mock.Anything).Return(scenarioLabel).Return([]*model.Runtime{}, nil)
 				return repo
 			},
-			InputID:            id,
-			ExpectedErrMessage: "",
+			InputID: id,
 			ContextFn: func() context.Context {
 				backgroundCtx := context.Background()
 				backgroundCtx = operation.SaveModeToContext(backgroundCtx, graphql.OperationModeAsync)
@@ -2711,7 +2707,7 @@ func TestService_Unpair(t *testing.T) {
 			ExpectedErrMessage: testErr.Error(),
 		},
 		{
-			Name: "Returns error when application is part of a scenario and a runtime",
+			Name: "Returns error when application is part of a scenario with runtime",
 			AppRepoFn: func() *automock.ApplicationRepository {
 				repo := &automock.ApplicationRepository{}
 				repo.AssertNotCalled(t, "Delete")


components/director/internal/domain/application/repository.go
```diff
@@ -113,6 +113,15 @@ func (r *pgRepository) DeleteGlobal(ctx context.Context, id string) error {
 	return r.globalDeleter.DeleteOneGlobal(ctx, repo.Conditions{repo.NewEqualCondition("id", id)})
 }
 
+// Unpair Updates application's UpdatedAt property by doing an "empty" application update when the operation type is graphql.OperationModeAsync
```
KirilKabakchiev, line 116: maybe add a note that we are looking to update the update timestamp

components/director/internal/domain/application/resolver.go
```diff
@@ -374,6 +376,41 @@ func (r *Resolver) UnregisterApplication(ctx context.Context, id string) (*graph
 	return deletedApp, nil
 }
 
+// UnpairApplication missing godoc
```
KirilKabakchiev, line 379: Maybe add godoc to your new methods

components/director/internal/domain/application/resolver.go
```diff
@@ -374,6 +376,41 @@ func (r *Resolver) UnregisterApplication(ctx context.Context, id string) (*graph
...
+	if err != nil {
+		return nil, err
+	}
+
+	auths, err := r.sysAuthSvc.ListForObject(ctx, model.ApplicationReference, app.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	err = r.sysAuthSvc.DeleteMultipleByIDForObject(ctx, auths)
```
KirilKabakchiev, line 393: you could inline the method call in the if :) same for the calls below

components/director/internal/domain/application/resolver.go
```diff
@@ -374,6 +376,41 @@ func (r *Resolver) UnregisterApplication(ctx context.Context, id string) (*graph
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = r.oAuth20Svc.DeleteMultipleClientCredentials(ctx, auths)
+	if err != nil {
+		return nil, err
+	}
+
+	err = r.appSvc.Unpair(ctx, id)
```
KirilKabakchiev, line 403: you are currently doing the scenario check in there - but logically if you are going to fail the whole thing because something is in formation/scenario, its better to "fail-fast" - basically do this check and fail early, rather than calling some deletes on systemauths and hydra clients first. Also DeleteMultipleClientCredentials does a rest call which we cant really rollback in case of unpair failure so right now if scenario exists unpair will fail and hydra creds will be just gone

components/director/internal/domain/application/service.go
```diff
@@ -488,6 +497,35 @@ func (s *service) DeleteLabel(ctx context.Context, applicationID string, key str
 	return nil
 }
 
+// IsBeingUsed Checks if an application has scenarios associated with it. If the scenarios are part of a runtime, then the application is considered being used by that runtime.
+func (s *service) IsBeingUsed(ctx context.Context, tenant, id string) (bool, error) {
```
KirilKabakchiev, line 501: thats a weird name - maybe something like IsApplicationInScenarios? and then id arg to be called appID ?

components/director/internal/domain/application/service.go
```diff
@@ -488,6 +497,35 @@ func (s *service) DeleteLabel(ctx context.Context, applicationID string, key str
...
+		if err != nil {
+			return false, err
+		}
+
+		if len(runtimes) > 0 {
+			application, err := s.appRepo.GetByID(ctx, tenant, id)
+			if err != nil {
+				return true, errors.Wrapf(err, "while getting application with id %s", id)
+			}
+			msg := fmt.Sprintf("System %s is still used and cannot be deleted. Unassign the system from the following formations first: %s. Then, unassign the system from the following runtimes, too: %s", application.Name, strings.Join(validScenarios, ", "), strings.Join(runtimes, ", "))
```
KirilKabakchiev, line 519: im pretty sure i've seen this code somewhere else, maybe in the delete app flow ? If its copied can we rather try to extract and reuse in order to avoid double maintanance in case this  code evolves in the future ?

components/director/pkg/graphql/schema.graphql
```diff
@@ -1208,6 +1211,7 @@ type Mutation {
 	- [unregister application](examples/unregister-application/unregister-application.graphql)
 	"""
 	unregisterApplication(id: ID!, mode: OperationMode = SYNC): Application! @hasScopes(path: "graphql.mutation.unregisterApplication") @async(operationType: DELETE, idField: "id", webhookType: UNREGISTER_APPLICATION)
+	unpairApplication(id: ID!, mode: OperationMode = SYNC): Application! @hasScopes(path: "graphql.mutation.unpairApplication") @async(operationType: UNPAIR, idField: "id", webhookType: UNREGISTER_APPLICATION)
```
KirilKabakchiev, line 1214: webhook type should be UNPAIR_APPLICATION so that we dont confuse them

components/operations-controller/api/v1alpha1/operation_types.go
```diff
@@ -31,13 +31,14 @@ func init() {
...
 
-// +kubebuilder:validation:Enum=Create;Update;Delete
+// +kubebuilder:validation:Enum=Create;Update;Delete;Unpair
 type OperationType string
 
 const (
 	OperationTypeCreate OperationType = "Create"
 	OperationTypeUpdate OperationType = "Update"
 	OperationTypeDelete OperationType = "Delete"
+	OperationTypeUnpair OperationType = "Unpair"
```
KirilKabakchiev, line 41: do we need a separate operation type called unpair or can we treat it as a operation of type update - i noticed that it most switch cases we are anyway using the "update" case ?

components/schema-migrator/migrations/director/20211015154700_unpairing-application.down.sql
```diff
@@ -0,0 +1,84 @@
...
+        USING status_condition::text::application_status_condition;
+
+ALTER TABLE applications
+    ALTER COLUMN status_condition
+        SET DEFAULT 'INITIAL' ::application_status_condition;
+
+DROP TYPE application_status_condition_old;
+
+CREATE OR REPLACE VIEW tenants_apps
+            (tenant_id, provider_tenant_id, id, name, description, status_condition, status_timestamp, healthcheck_url,
```
KirilKabakchiev, line 36: add a comment around the actual change in the view so that its easier for review

tests/director/tests/application_api_test.go
```diff
@@ -615,6 +615,115 @@ func TestDeleteApplication(t *testing.T) {
...
+
+		//WHEN
+		req := fixtures.FixUnpairApplicationRequest(application.ID)
+		err = testctx.Tc.RunOperationWithCustomTenant(ctx, dexGraphQLClient, tenantID, req, nil)
+
+		//THEN
+		require.NoError(t, err)
+	})
+
+	t.Run("Error when application is in scenario and runtime", func(t *testing.T) {
```
KirilKabakchiev, line 678: "is in scenario WITH runtime"

tests/director/tests/application_api_test.go
```diff
@@ -615,6 +615,115 @@ func TestDeleteApplication(t *testing.T) {
...
+		// WHEN
+		unpairRequest := fixtures.FixUnpairApplicationRequest(actualApp.ID)
+		saveExample(t, unpairRequest.Query(), "unpair application")
+		err = testctx.Tc.RunOperation(ctx, dexGraphQLClient, unpairRequest, &actualApp)
+
+		//THEN
+		require.NoError(t, err)
+	})
+
+	t.Run("Success when application is in scenario but not in runtime", func(t *testing.T) {
```
KirilKabakchiev, line 645: "but not WITH runtime" - apps are not in runtimes

tests/external-services-mock/tests/async_api_unpair_test.go
```diff
@@ -0,0 +1,120 @@
...
+		return isOperationUnpairCompleted(operation)
+	}, time.Minute*3, time.Second*10, "Waiting for operation unpair timed out.")
+
+	t.Log("Verify the unpaired application still exists in director")
+	existingApp := fixtures.GetApplication(t, ctx, gqlClient, testConfig.DefaultTestTenant, app.ID)
+	require.NotEmpty(t, existingApp.Name, "Application is deleted")
+}
+
+func isOperationUnpairCompleted(operation *v1alpha1.Operation) bool {
+	if operation.Status.Phase == v1alpha1.StateSuccess || operation.Status.Phase == v1alpha1.StateFailed {
```
desislavaa, line 116: Don't we want the operation to be successful at the end?

components/director/internal/domain/application/service_test.go
```diff
@@ -2509,6 +2509,189 @@ func TestService_Delete(t *testing.T) {
...
+			RuntimeRepoFn: func() *automock.RuntimeRepository {
+				repo := &automock.RuntimeRepository{}
+				repo.On("ListAll", ctx, applicationModel.Tenant, mock.Anything).Return(scenarioLabel).Return([]*model.Runtime{}, nil)
+				return repo
+			},
+			InputID:            id,
+			ExpectedErrMessage: "",
+		},
+		{
+			Name: "Success when application is part of a scenario but not in runtime",
```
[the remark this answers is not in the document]
alextargov: https://github.com/kyma-incubator/compass/pull/2063#discussion_r731786396 `With` or `In`? :D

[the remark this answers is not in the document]
desislavaa: we're both team "with" :D

components/director/internal/domain/application/repository.go
```diff
@@ -113,6 +113,16 @@ func (r *pgRepository) DeleteGlobal(ctx context.Context, id string) error {
...
 }
 
+// Unpair Does an "empty" Update so we can update application's UpdatedAt property when the operation type is graphql.OperationModeAsync.
+// Unpairs can be multiple on a single application and we we want to track when are unpairings happening. That's why we are doing this update.
+func (r *pgRepository) Unpair(ctx context.Context, app *model.Application) error {
+	opMode := operation.ModeFromCtx(ctx)
+	if opMode == graphql.OperationModeAsync {
+		return r.Update(ctx, app)
+	}
+	return nil
```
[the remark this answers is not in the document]
desislavaa: That's a point that I have missed actually, shouldn't we set it back to "Initial" in that case? I mean, by deleting its system auths, we've basically invalidated all certificates, hence no communication from the app to Compass will be possible until the app is paired again - being in state "Connected" does not make sense in that case.

PR #2064, opened 2021-10-11
Apps for runtime e2e test

nyordanoff:
**Description**

This PR introduces an e2e test regarding the "ApplicationsForRuntime" flow. The setup is as follows:

- register first Application and get a certificate for it
- register Runtime and get a certificate for it
- register second Application and get a certificate for it
- assign both Apps and the Runtime into a formation
- call Director with "ApplicationsForRuntime" query with the Runtime certificate
- verify that the `system_auth_id` value for the Apps from the query result matches the CN from that App's certificate

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [x] Integration tests


PetarTodorovv (changes requested):

tests/connector/tests/apps_for_runtime_test.go
```diff
@@ -0,0 +1,96 @@
...
+	defer fixtures.UpdateScenariosLabelDefinitionWithinTenant(t, ctx, directorClient.DexGraphqlClient, cfg.Tenant, []string{DefaultScenario})
+
+	appIdToCommonName := make(map[string]string)
+
+	// Register first Application
+	firstApp, err := fixtures.RegisterApplicationFromInput(t, ctx, directorClient.DexGraphqlClient, cfg.Tenant, graphql.ApplicationRegisterInput{
+		Name:   "test-first-app",
+		Labels: map[string]interface{}{ScenariosLabel: []string{TestScenario}},
+	})
+	defer fixtures.CleanupApplication(t, ctx, directorClient.DexGraphqlClient, cfg.Tenant, &firstApp)
```
ataleksandrov, line 31: It's a minor,but could you please do:
```
defer func(){
  fixtures.UnassignApplicationFromScenarios(t, ctx, directorClient.DexGraphqlClient, cfg.Tenant, firstApp.ID, true)
  fixtures.CleanupApplication(t, ctx, directorClient.DexGraphqlClient, cfg.Tenant, &firstApp)
}
```
This way it is easier to track the execution of the defer statements

ataleksandrov (commented):
Maybe I am missing something, but could we add this test as a test case here: https://github.com/kyma-incubator/compass/blob/main/tests/director/tests/application_api_test.go#L925

tests/connector/tests/apps_for_runtime_test.go
nyordanoff: Regarding adding the test in the director ones:

Initially, I wanted to add it there because it is more intuitive. However, as this test requires issuing of certificates and all current director tests do not, the whole "setup" from the connector tests should have been copy-pasted to the director. 

I thought this was unnecessary overhead for a single test and that it will fit better in the connector for the certificate related stuff. In addition, nothing major was added/changed to the connector tests setup to make this test work, so I decided to leave it there.

PR #2065, opened 2021-10-12
Use templates in System Fetcher

desislavaa:
**Description**

Currently, when the System Fetcher retrieves an application with a corresponding application template ID, it only sets the template ID, instead of using the template mechanism provided in Compass.
This PR improves the application fetching by rendering its input via its matching template - that way we'll have all additional labels set for our fetched applications.

- Render the application input by using the template matching the fetched application in System Fetcher
- Introduce application input overrides in System Fetcher which can be used to replace the existing application input of the template, and override all common fields except labels - they are merged.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -29,7 +29,7 @@ require (
 	github.com/kr/pretty v0.2.1 // indirect
 	github.com/kyma-incubator/compass/components/connectivity-adapter v0.0.0-20210922143733-cb93bcfd3cd6
 	github.com/kyma-incubator/compass/components/connector v0.0.0-20210922140315-d2f37c2e160d
-	github.com/kyma-incubator/compass/components/director v0.0.0-20211015131944-501a5435ac7a
+	github.com/kyma-incubator/compass/components/director v0.0.0-20211020093310-b929e49b0efc
 	github.com/kyma-incubator/compass/components/external-services-mock v0.0.0-20210922144742-89f134add94f
 	github.com/kyma-incubator/compass/components/gateway v0.0.0-20210922142032-ccf81fabdd36
 	github.com/kyma-incubator/compass/components/operations-controller v0.0.0-20210922140924-1dc663c3ed24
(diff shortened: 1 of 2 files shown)


components/director/internal/systemfetcher/template.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (r *renderer) getTemplateInputs(s System) (*model.ApplicationFromTemplateInputValues, error) {
+	systemJSON, err := json.Marshal(s)
+	if err != nil {
+		return nil, err
+	}
+
+	inputValues := model.ApplicationFromTemplateInputValues{}
+	for _, pm := range r.placeholdersMapping {
+		placeholderInput := gjson.GetBytes(systemJSON, pm.SystemKey).String()
```
DimitarPetrov, line 96: Should we fail if this does not exists?

components/director/internal/systemfetcher/template.go
```diff
@@ -0,0 +1,159 @@
...
+		return "", errors.Wrapf(err, "while unmarshaling original application input")
+	}
+
+	if err := json.Unmarshal([]byte(overrideAppInputJSON), &overrideAppInput); err != nil {
+		return "", errors.Wrapf(err, "while unmarshaling override application input")
+	}
+
+	for field, overrideV := range overrideAppInput {
+		if originalV, ok := originalAppInput[field]; ok && originalV != nil {
+			if reflect.TypeOf(originalV) != reflect.TypeOf(overrideV) {
```
DimitarPetrov, line 120: What does this checks for? Isn't always strings? Or one can be a string and another object or array?

components/director/internal/systemfetcher/service.go
```diff
@@ -219,29 +223,50 @@ func (s *SystemFetcher) processSystemsForTenant(ctx context.Context, tenantMappi
...
-			"managed": "true",
-		},
+
+	input.StatusCondition = &initStatusCond
+	input.SystemNumber = &sc.SystemNumber
+
+	if len(input.Labels) == 0 {
+		input.Labels = make(map[string]interface{}, 1)
 	}
+	input.Labels["managed"] = "true"
```
DimitarPetrov, line 266: You can add the managed label to the template input override and make use of the enrichLabels functionality.

Maybe the same applies for the StatusCondition and the SystemNumber as well and get rid of this function entirely.

components/director/internal/systemfetcher/template_test.go
```diff
@@ -0,0 +1,232 @@
...
+		Description:     &system.ProductDescription,
+		BaseURL:         &system.BaseURL,
+		ProviderName:    &system.InfrastructureProvider,
+		SystemNumber:    &system.SystemNumber,
+		StatusCondition: &initStatusCond,
+		Labels: map[string]interface{}{
+			"managed": "true",
+		},
+	}
+}
```
DimitarPetrov, line 232: Add those to fixtures_test.go

components/director/pkg/graphql/schema.graphql
```diff
@@ -253,9 +253,14 @@ input ApplicationRegisterInput {
 	**Validation:** valid URL, max=256
 	"""
 	healthCheckURL: String
-	bundles: [BundleCreateInput!]
+	"""
+	**Validation:** valid URL, max=256
+	"""
+	baseUrl: String
 	integrationSystemID: ID
+	systemNumber: String
```
DimitarPetrov, line 261: Why do we need to allow setting systemNumber from outside? Can we trust the client for such information?

DimitarPetrov (approved):

PR #2073, opened 2021-10-14
Add check for Extra fields length in SystemAuth Provider matcher

desislavaa:
**Description**

For JWT tokens coming from external auth provider, which is not ORY Hydra, we also fall in the case where we have `client_id` in the JWT Extra fields, hence we try to provide auth context with SystemAuthObjectContextProvider, which results in SQL error. After that error, we're no longer able to execute DB requests, because the transaction is brokern. 

Possible fixes:
* Open a new transaction for each context provider. Looks like an overhead to me.
* Strip the `client_id` in the authenticator authnmappinghandler
* [current] Verify that the Extra attributes do not exceed 3 :) 
* [in progress] Strip configurable amount of extra fields in authnmappinghandler

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/internal/tenantmapping/system_auth_context_provider.go
+++ b/components/director/internal/tenantmapping/system_auth_context_provider.go
@@ -19,6 +19,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+// hydraTokenMaxExtraDataLength is the maximum properties count that we can get from a Hydra token in the OAuth flow.
+const hydraTokenMaxExtraDataLength = 3
+
 // NewSystemAuthContextProvider missing godoc
 func NewSystemAuthContextProvider(systemAuthSvc systemauth.SystemAuthService, scopesGetter ScopesGetter, tenantRepo TenantRepository) *systemAuthContextProvider {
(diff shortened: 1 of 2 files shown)


chart/compass/charts/external-services-mock/templates/oauth-credentials-secret.yaml
```diff
@@ -10,6 +10,7 @@ data:
   auditlog-client-secret: {{ "client_secret" | b64enc | quote }}
   auditlog-oauth-url: {{ printf "http://compass-external-services-mock.%s.svc.cluster.local:%s/audit-log/v2/oauth/token" .Release.Namespace (.Values.service.port | toString) | b64enc | quote }}
 {{end}}
+{{- if eq .Values.global.ordAggregator.accessStrategies.cmpMtls.oauthSecret.managed true }}
```
NickyMateev, line 13: Why?

NickyMateev (approved):

PR #2074, opened 2021-10-14
Improve schema-migrator

PetarTodorovv:
**Description**
In our PR if we try to use the schema-migrator version(PR-XX) from another PR and we run local installation with DB_DUMP without adding the new "migrations" in our local setup the director won't start because of the readiness probe. That happens because we have two jobs - compass-migration and update-expected-schema-version. The second one will use the image of the "remote" PR but the image of the compass-migration job will be build/created in the run.sh of the director during local installation using the migrations currently present in the local setup. So the two versions of the migrations in the jobs will be different and the readiness probe won't be successful.

- Update the image of the update-expected-schema-version job in case we use DB_DUMP option. So we can have same migrations in the both jobs

- [ ] Implementation


desislavaa:
Can we set the image of `update-expected-schema-version` to `schema-migrator:latest` when using DB dump?

nkanchev (dismissed):

PR #2079, opened 2021-10-15
Make subscription test skippable in E2E tests

desislavaa:
**Description**
It's currently not possible to execute tests involving the Tenant Fetcher on our prod landscapes, hence this test should be skipped there. In order to skip a test, one can add it to the `SKIP_TESTS_REGEX` env variable. It is later used in [this utility](https://github.com/kyma-incubator/compass/blob/main/tests/pkg/testing/testing.go#L24).

_PS: >95 lines of code are just identation, so the PR is actually really small._

- Use the skip mechanism in the ORD Service E2E tests for the subscruption flow test

### DIFF
--- a/tests/ord-service/tests/subscription_flow_test.go
+++ b/tests/ord-service/tests/subscription_flow_test.go
@@ -41,7 +41,7 @@ import (
 
 func TestSubscriptionFlow(stdT *testing.T) {
 	t := testingx.NewT(stdT)
-	t.Run("Consuming provider resources after successful subscription", func(t *testing.T) {
+	t.Run("ConsumerProvider flow: calls with provider certificate and consumer token are successful when valid subscription exists", func(t *testing.T) {
 		ctx := context.Background()
 		defaultTenantId := tenant.TestTenants.GetDefaultTenantID()
 		secondaryTenant := tenant.TestTenants.GetIDByName(t, tenant.ApplicationsForRuntimeTenantName)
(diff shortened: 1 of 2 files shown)


tests/ord-service/tests/subscription_flow_test.go
```diff
@@ -38,102 +39,105 @@ import (
...
-	headers := map[string][]string{"Authorization": {fmt.Sprintf("Bearer %s", token.FromExternalServicesMock(t, testConfig.ExternalServicesMockURL, claims))}}
-
-	// Make a request to the ORD service with http client containing certificate with provider information and token with the consumer data.
-	respBody := makeRequestWithHeaders(t, extIssuerCertHttpClient, testConfig.ORDExternalCertSecuredServiceURL+"/systemInstances?$format=json", headers)
-
-	require.Equal(t, 1, len(gjson.Get(respBody, "value").Array()))
-	require.Equal(t, consumerApp.Name, gjson.Get(respBody, "value.0.title").String())
+func TestSubscriptionFlow(stdT *testing.T) {
+	t := testingx.NewT(stdT)
+	t.Run("Consuming provider resources after successful subscription", func(t *testing.T) {
```
desislavaa, line 44: This nesting is required in order to use the skip functionallity, provided by the new `t` we create.
All changed lines below are just identation.

DimitarPetrov: Just to note that this is the value that should be added in the regex, maybe we need to make it a bit more specific / unique.

PR #2083, opened 2021-10-18
Remove Secured App Type configuration

NickyMateev:
# Remove Secured App Type configuration

**Description**

- Remove the env configuration for secured app type introduced here: https://github.com/kyma-incubator/compass/pull/2051

No need to have such a config, we can only look if the webhook has creds or not.

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/components/director/internal/open_resource_discovery/client.go
+++ b/components/director/internal/open_resource_discovery/client.go
@@ -11,7 +11,6 @@ import (
 	"github.com/kyma-incubator/compass/components/director/internal/open_resource_discovery/accessstrategy"
 
 	"github.com/kyma-incubator/compass/components/director/internal/model"
-	"github.com/kyma-incubator/compass/components/director/pkg/str"
 
 	httputil "github.com/kyma-incubator/compass/components/director/pkg/http"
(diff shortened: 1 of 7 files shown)


components/director/internal/open_resource_discovery/client.go
```diff
@@ -117,7 +114,7 @@ func (c *client) fetchConfig(ctx context.Context, app *model.Application, webhoo
 	}
 
 	var resp *http.Response
-	if _, secured := c.securedApplicationTypes[app.Type]; secured {
+	if webhook.Auth != nil {
```
PetarTodorovv, line 117: Here we check the webhook.Auth if it's nil so in the `httputil.GetRequestWithCredentials` we can remove the first condition check which is the same? - https://github.com/kyma-incubator/compass/blob/main/components/director/pkg/http/request.go#L16

NickyMateev: Yes, but it's possible for someone to call `httputil.GetRequestWithCredentials` without checking it and so I think it's a good idea to leave it there as well

PR #2084, opened 2021-10-19
Fix operation processing for already deleted application

desislavaa:
**Description**

The operations controller:
 - Calls the delete webhook, and polls it until success.
 - Calls Director that the operation is finished, which results in Director deleting the application.
 - Fails to update the state of the operation CRD with the following error:
    ```
    Operation cannot be fulfilled on operations.operations.compass \"application-2913742a-a244-4f80-a9af-7851246b2fe5\":
    the object has been modified; please apply your changes to the latest version and try again
   ```

Then on the next run of the reconcile loop, it tries to fetch the application, but it's no longer available, and the operation stucks in `In Progress` state, which results in failing E2E tests.
 The error is completely normal and we should be able to handle that scenario gracefully - https://alenkacz.medium.com/kubernetes-operators-best-practices-understanding-conflict-errors-d05353dff421

- Do not return error when the app is not found in the Director client lib in System Broker (used by Operations Controller)
- In the reconcile loop, when the application cannot be found, the operation is of type `Delete`, and the current status is `in Progress`, then set the operation status to "Successful"
- Small improvement in handling of reconciled finished operations with applications missing in director - skip requeue, instead of requeue with error

🐞  **RCA**
The K8s client we use at the beginning of the reconcile loop is backed by cache, which tends to get left behind a bit, which seems to be normal.
In conslusion - there's no unexpected updates, just cache. We can mitigate this by [polling the resource after updating it](https://github.com/kubernetes/test-infra/blob/8f0f19a905a20ed6f76386e5e11343d4bc2446a7/prow/plank/reconciler.go#L516-L520), until the resourse versions are different - that would mean that on the next reconcile, the version won't be outdated (at least in case the operator itself made the update).
Resource: https://github.com/kubernetes-sigs/controller-runtime/issues/1464

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ]  Integration tests <- ❓ not really testable on purpose


components/system-broker/pkg/director/client.go
```diff
@@ -82,17 +83,15 @@ func (c *GraphQLClient) FetchApplication(ctx context.Context, id string) (*Appli
...
 	req := gcli.NewRequest(query)
 
-	err := c.gcli.Do(ctx, req, &app)
-	if err != nil {
+	if err := c.gcli.Do(ctx, req, &app); err != nil {
 		return nil, errors.Wrap(err, "while fetching application in gqlclient")
 	}
 	if app.Result == nil {
-		return nil, errors.New("failed to fetch application")
+		log.C(ctx).Warnf("Application with ID %s not found", id)
```
KirilKabakchiev, line 93: probably its better to keep the client behavour as is - with the change the op will never be deleted

pankrator (approved):

PR #2095, opened 2021-10-29
Remove implicit label def creation

ataleksandrov:
### Description

After introducing the Self Register flow new label definition is created within the child tenant. This new label definition is visible to the parent tenant, because of the recursive tenant isolation. This causes unexpected behaviour of the LabelDefintion API. The proposed solution is to remove the ensurance that label definition is created during runtime/application registration. 
### Changes:
- Adjust existing unit and e2e tests
- Add new e2e test for Self Register flow


desislavaa (dismissed):

components/director/internal/domain/labeldef/service.go
```diff
@@ -234,11 +240,22 @@ func (s *service) GetAvailableScenarios(ctx context.Context, tenantID string) ([
...
 
 // AddDefaultScenarioIfEnabled adds DEFAULT scenario if defaultScenarioEnabled
-func (s *service) AddDefaultScenarioIfEnabled(ctx context.Context, labels *map[string]interface{}) {
+func (s *service) AddDefaultScenarioIfEnabled(ctx context.Context, tnt string, labels *map[string]interface{}) {
 	if labels == nil || !s.defaultScenarioEnabled {
 		return
 	}
 
+	// do not add scenario to subaccounts
+	btm, err := s.tenantRepo.Get(ctx, tnt)
```
KirilKabakchiev, line 249: can we do this a bit more generically - for example instead of check for tenant type, check if the label definition for this scenario has a schema that contains the Default scenario

tests/ord-service/tests/subscription_flow_test.go
```diff
@@ -42,6 +42,58 @@ import (
 	"github.com/tidwall/gjson"
 )
 
+func TestSelfRegisterFlow(stdT *testing.T) {
+	t := testingx.NewT(stdT)
+	t.Run("TestSelfRegisterFlow flow: label definitions of the parent tenant are not overwritten", func(t *testing.T) {
+		ctx := context.Background()
+
+		// defaultTenantId is the parent of the subaccountID
```
PetarTodorovv, line 50: Currently the `tenant.TestTenants.GetDefaultTenantID()` is not parent of the `tenant.TestProviderSubaccount`. We can set it in the values.yaml file

PetarTodorovv:
For the provider subaccount tenant add the "default tenant" as parent - https://github.com/kyma-incubator/compass/blob/main/chart/compass/values.yaml#L63 add new field `parent: 5577cf46-4f78-45fa-b55f-a42a3bdba868` and also at https://github.com/kyma-incubator/compass/blob/main/tests/pkg/tenant/tenant_setup.go#L179 for the `TestProviderSubaccount` add `Parent:         testDefaultTenant,`

components/director/internal/domain/labeldef/service.go
ataleksandrov: As discussed, we will remain the following check for the hotfix. In the next sprint we will evaluate whether we could drop this piece of code.

PetarTodorovv (approved):

PR #2097, opened 2021-11-03
Fix tenant fetcher not updating configmap if no events are present

pankrator:
**Fix tenant fetcher not updating configmap if no events are present**

The bug causes the problem that if there are no events for a region the fetcher exits without updating the configmap and can even skip events for other regions as well.

- Do not exit when there are no tenant events for a region


ataleksandrov (approved):

ataleksandrov:
this fix is also part of: #2098

ataleksandrov (approved):

PR #2105, opened 2021-11-12
Add CorrelationID to Bundles

StefanCholakov:
**Description**

- Add CorrelationID to Bundles

- [X] Implementation
- [X] Unit tests
- [X] Integration tests

### DIFF
--- a/components/director/internal/domain/bundle/fixtures_test.go
+++ b/components/director/internal/domain/bundle/fixtures_test.go
@@ -175,6 +175,7 @@ const (
 	tenantID         = "ttttttttt-tttt-tttt-tttt-tttttttttttt"
 	externalTenantID = "eeeeeeeee-eeee-eeee-eeee-eeeeeeeeeeee"
 	ordID            = "com.compass.v1"
+	correlationIDs   = `["id1", "id2"]`
 )
 
 func fixBundleModel(name, desc string) *model.Bundle {
@@ -190,6 +191,7 @@ func fixBundleModel(name, desc string) *model.Bundle {
 		Links:                          json.RawMessage("[]"),
 		Labels:                         json.RawMessage("[]"),
 		CredentialExchangeStrategies:   json.RawMessage("[]"),
+		CorrelationIDs:                 json.RawMessage(correlationIDs),
 		BaseEntity: &model.BaseEntity{
 			ID:        bundleID,
 			Ready:     true,
@@ -397,6 +399,10 @@ func fixEntityBundle(id, name, desc string) *bundle.Entity {
 		Links:                         repo.NewValidNullableString("[]"),
 		Labels:                        repo.NewValidNullableString("[]"),
 		CredentialExchangeStrategies:  repo.NewValidNullableString("[]"),
+		CorrelationIDs: sql.NullString{
+			String: correlationIDs,
+			Valid:  true,
+		},
 		BaseEntity: &repo.BaseEntity{
 			ID:        id,
 			Ready:     true,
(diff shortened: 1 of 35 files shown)


components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -1587,6 +1611,33 @@ func TestDocuments_ValidateBundle(t *testing.T) {
 				doc := fixORDDocument()
 				doc.ConsumptionBundles[0].CredentialExchangeStrategies = json.RawMessage("[]")
 
+				return []*ord.Document{doc}
+			},
+		},
+		{
+			Name: "Invalid `correlationIds` field when it is invalid JSON for Bundle",
```
nyordanoff, line 1618: Will you add one more test where a valid JSON is passed but the value of one of the correlationIDs is invalid (does not match the required regex)

tests/pkg/assertions/assertions.go
```diff
@@ -519,6 +516,25 @@ func AssertMultipleEntitiesFromORDService(t *testing.T, respBody string, entitie
...
+		require.NotEmpty(t, entityTitle)
+
+		correlationIds := gjson.Get(respBody, fmt.Sprintf("value.%d.correlationIds", i)).Array()
+		bundleCorrelationIdsAsString := make([]string, 0)
+		for _, id := range correlationIds {
+			idAsString := gjson.Get(id.String(), "value").String()
+			bundleCorrelationIdsAsString = append(bundleCorrelationIdsAsString, idAsString)
+		}
+		expectedCorrelationIds := entitiesMap[entityTitle]
+		assert.True(t, unorderedEqualStringArrays(expectedCorrelationIds, bundleCorrelationIdsAsString))
```
nyordanoff, line 534: here you can use the built-in `assert.ElementsMatch(t, expectedCorrelationIds, bundleCorrelationIdsAsString)`

you can double-check its documentation but from what I saw, the order of the given slices is ignored when comparing them

components/director/internal/domain/bundle/fixtures_test.go
```diff
@@ -397,6 +399,10 @@ func fixEntityBundle(id, name, desc string) *bundle.Entity {
 		Links:                         repo.NewValidNullableString("[]"),
 		Labels:                        repo.NewValidNullableString("[]"),
 		CredentialExchangeStrategies:  repo.NewValidNullableString("[]"),
+		CorrelationIDs: sql.NullString{
```
nyordanoff, line 402: Isn't that cleaner:

`CorrelationIDs:  repo.NewValidNullableString(correlationIDs)`

components/schema-migrator/migrations/director/20211111122118_bundles_correlation_id.up.sql
```diff
@@ -0,0 +1,81 @@
...
+               SELECT *
+               FROM apps_subaccounts_func()
+               UNION ALL
+               SELECT a_s.id, (SELECT id::text FROM business_tenant_mappings WHERE external_tenant = a_s.tenant_id::text), cpr.provider_tenant::text AS provider_tenant_id
+               FROM apps_subaccounts_func() a_s
+                        JOIN consumers_provider_for_runtimes_func() cpr
+                             ON cpr.consumer_tenants ? a_s.tenant_id::text) t_apps ON b.app_id = t_apps.id;
+
+
+CREATE OR REPLACE VIEW correlation_ids(application_id, product_id, bundle_id, value) as
```
nyordanoff, line 55: I think the original `correlation_ids` (latest version can be seen [here](https://github.com/kyma-incubator/compass/blob/main/components/schema-migrator/migrations/director/20210622095252_vendor_product_internal_ids.up.sql#L148)) should be extended with the new bundle logic.

My main concern with the current approach is that the newly introduced `LATERAL` functionality does not add much value because the view is not complex at all.

Moreover, I find the aliases at the beginning: `SELECT app_correlation_ids.product_id, app_correlation_ids.bundle_id` very confusing because product and bundle ids have nothing to do with the app_correlation_ids.

The down migration should use the original `correlation_ids` view as well.

nyordanoff (approved):

nyordanoff (approved):

PR #2114, opened 2021-11-22
Move gke overrides to compass

pankrator:
**Move gke overrides to compass**

Related to this https://github.com/kyma-project/test-infra/pull/4485 which removes the provided overrides with this change.

- now gke and benchmark overrides are provided from compass repo

### DIFF
--- a/installation/resources/installer-config-gke-integration.yaml.tpl
+++ b/installation/resources/installer-config-gke-integration.yaml.tpl
@@ -1,0 +1,35 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: compass-installation-gke-integration-overrides
+  namespace: compass-installer
+  labels:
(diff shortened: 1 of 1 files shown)


installation/resources/installer-config-gke-integration.yaml.tpl
```diff
@@ -0,0 +1,35 @@
+apiVersion: v1
+kind: ConfigMap
```
KirilKabakchiev, line 2: from where were these set before you added https://github.com/kyma-project/test-infra/pull/4477/files ? I dont see anything removed in these PRs ?

pankrator: they are not removed in the other PR for backwards compatibility

desislavaa (approved):

PR #2115, opened 2021-11-24
Storage Redesign

DimitarPetrov:
#### Changes proposed in this pull request:

- Redesign the database based on split views approach.
- Adapt the generic repo creater/updater/deleter/getter/lister/exister to work with the new tenancy concept
- Redesign each entity's repo layer tests to use extracted and generic delete/update/create/get... suites.
- Redesign Automatic Scenario Assignements to use target_tenant_id instead of label selector and restrict it to *global_subaccount_id* key 
  - The scenario label produced by the ASA is set to each runtime in the target_tenant_id, but the label itself lives in the tenant that the ASA is created in.
  - target_tenant_id should be direct child of the tenant_id of the ASA
- Restrict Label Definitions to key = *scenarios* only
- Register runtime with *global_subaccount_id* label will register the runtime in the subaccount and not in the caller tenant (only if the caller is the GA parent of the SA provided in the label)
- Move subaccount feature that moves all tenant accesses based on the changed parent.
- Adapted and simplified ORD Service views
  - Use internal tenant id everywhere
  - Add views for packages/vendors/products/tombstones (see related PR: https://github.com/kyma-incubator/ord-service/pull/48)
- Various adaptations across the project in order to use the tenant isolation views.
  - Add app_id to the document entity.
  - Drop materialised view and the owner_tenant middleware alongside with the recursive tenant isolation query. 
- Regenerate and re-format examples.

## Pull Request status

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/scenarioassignment/engine.go
```diff
@@ -83,45 +82,12 @@ func (e *engine) RemoveAssignedScenarios(ctx context.Context, in []*model.Automa
...
-// GetScenariosForSelectorLabels missing godoc
-func (e engine) GetScenariosForSelectorLabels(ctx context.Context, inputLabels map[string]string) ([]string, error) {
-	tenantID, err := tenant.LoadFromContext(ctx)
-	if err != nil {
-		return nil, err
-	}
-
+// MergeScenariosFromInputLabelsAndAssignments merges all the scenarios that are part of the resource labels (already added + to be added with the current operation)
+// with all the scenarios that should be assigned based on ASAs.
+func (e engine) MergeScenariosFromInputLabelsAndAssignments(ctx context.Context, inputLabels map[string]interface{}, runtimeID string) ([]interface{}, error) {
```
danielgospodinow, line 87: Shouldn't we unify the receivers to all be pointer receivers or all be value receivers? I think that this is a good practice.

components/director/internal/domain/api/service.go
```diff
@@ -147,9 +147,9 @@ func (s *service) Create(ctx context.Context, appID string, bundleID, packageID
 	}
 
 	id := s.uidService.Generate()
-	api := in.ToAPIDefinition(id, appID, packageID, tnt, apiHash)
+	api := in.ToAPIDefinition(id, appID, packageID, apiHash)
 
-	err = s.repo.Create(ctx, api)
+	err = s.repo.Create(ctx, tnt, api)
```
la4ezar, line 152: if err = ... ; err != nil {..}

components/director/internal/domain/bundleinstanceauth/repository_test.go
```diff
@@ -85,487 +84,203 @@ func TestRepository_Create(t *testing.T) {
...
+					}
+				},
+				InvalidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{
+						fixSQLRows([]sqlRow{}),
+					}
+				},
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 125: You can extract this function as above.

components/director/internal/domain/document/repository_test.go
```diff
@@ -2,89 +2,69 @@ package document_test
...
+				},
+			},
+			{
+				Query: regexp.QuoteMeta("INSERT INTO public.documents ( id, bundle_id, app_id, title, display_name, description, format, kind, data, ready, created_at, updated_at, deleted_at, error ) VALUES ( ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ? )"),
+				Args: []driver.Value{givenID(), refID, appID, docEntity.Title, docEntity.DisplayName, docEntity.Description, docEntity.Format, docEntity.Kind, docEntity.Data,
+					docEntity.Ready, docEntity.CreatedAt, docEntity.UpdatedAt, docEntity.DeletedAt, docEntity.Error},
+				ValidResult: sqlmock.NewResult(-1, 1),
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 51: You can extract this function as above.

components/director/internal/domain/eventdef/repository_test.go
```diff
@@ -2,343 +2,225 @@ package eventdef_test
...
+					}
+				},
+				InvalidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{
+						sqlmock.NewRows(fixEventDefinitionColumns()),
+					}
+				},
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 45: You can extract this function as above.

components/director/internal/domain/fetchrequest/repository_test.go
```diff
@@ -3,177 +3,311 @@ package fetchrequest_test
...
+					return []*sqlmock.Rows{testdb.RowWhenObjectDoesNotExist()}
+				},
+			},
+			{
+				Query:       regexp.QuoteMeta("INSERT INTO public.fetch_requests ( id, document_id, url, auth, mode, filter, status_condition, status_message, status_timestamp, spec_id ) VALUES ( ?, ?, ?, ?, ?, ?, ?, ?, ?, ? )"),
+				Args:        []driver.Value{givenID(), sql.NullString{}, "foo.bar", apiFREntity.Auth, apiFREntity.Mode, apiFREntity.Filter, apiFREntity.StatusCondition, apiFREntity.StatusMessage, apiFREntity.StatusTimestamp, refID},
+				ValidResult: sqlmock.NewResult(-1, 1),
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 52: You can extract this function as above.

components/director/internal/domain/label/repository_test.go
```diff
@@ -3,110 +3,131 @@ package label_test
...
-				Valid:  true,
+			{
+				Query:       regexp.QuoteMeta("INSERT INTO public.labels ( id, tenant_id, app_id, runtime_id, runtime_context_id, key, value, version ) VALUES ( ?, ?, ?, ?, ?, ?, ?, ? )"),
+				Args:        []driver.Value{appLabelEntity.ID, appLabelEntity.TenantID, appLabelEntity.AppID, appLabelEntity.RuntimeID, appLabelEntity.RuntimeContextID, appLabelEntity.Key, appLabelEntity.Value, appLabelEntity.Version},
+				ValidResult: sqlmock.NewResult(-1, 1),
 			},
-			Version: version,
-		}
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 52: You can extract this function as above.

components/director/internal/domain/ordvendor/repository_test.go
```diff
@@ -1,237 +1,179 @@
...
+					return []*sqlmock.Rows{testdb.RowWhenObjectDoesNotExist()}
+				},
+			},
+			{
+				Query:       `^INSERT INTO public.vendors \(.+\) VALUES \(.+\)$`,
+				Args:        fixVendorRow(),
+				ValidResult: sqlmock.NewResult(-1, 1),
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 35: You can extract this function as above.

components/director/internal/domain/package/repository_test.go
```diff
@@ -1,238 +1,181 @@
...
+					return []*sqlmock.Rows{testdb.RowWhenObjectDoesNotExist()}
+				},
+			},
+			{
+				Query:       `^INSERT INTO public.packages \(.+\) VALUES \(.+\)$`,
+				Args:        fixPackageRow(),
+				ValidResult: sqlmock.NewResult(-1, 1),
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 36: You can extract this function as above.

components/director/internal/domain/product/repository_test.go
```diff
@@ -1,237 +1,180 @@
...
+					return []*sqlmock.Rows{testdb.RowWhenObjectDoesNotExist()}
+				},
+			},
+			{
+				Query:       `^INSERT INTO public.products \(.+\) VALUES \(.+\)$`,
+				Args:        fixProductRow(),
+				ValidResult: sqlmock.NewResult(-1, 1),
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 36: Extract this function as above.

components/director/internal/domain/runtime/repository_test.go
```diff
@@ -2,150 +2,104 @@ package runtime_test
...
+				IsSelect: true,
+				ValidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixColumns).AddRow(rtModel.ID, rtModel.Name, rtModel.Description, rtModel.Status.Condition, rtModel.Status.Timestamp, rtModel.CreationTimestamp)}
+				},
+				InvalidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixColumns)}
+				},
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 42: You can extract this function as above.

components/director/internal/domain/runtime_context/repository_test.go
```diff
@@ -2,425 +2,294 @@ package runtimectx_test
...
+				IsSelect: true,
+				ValidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixColumns).AddRow(runtimeCtxModel.ID, runtimeCtxModel.RuntimeID, runtimeCtxModel.Key, runtimeCtxModel.Value)}
+				},
+				InvalidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixColumns)}
+				},
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 40: You can extract this function as above.

components/director/internal/domain/spec/repository_test.go
```diff
@@ -12,448 +10,488 @@ import (
...
+				IsSelect: true,
+				ValidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixSpecColumns()).AddRow(fixAPISpecRow()...)}
+				},
+				InvalidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixSpecColumns())}
+				},
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 36: You can extract this function as above.

components/director/internal/domain/tombstone/repository_test.go
```diff
@@ -1,237 +1,186 @@
...
+					return []*sqlmock.Rows{testdb.RowWhenObjectDoesNotExist()}
+				},
+			},
+			{
+				Query:       `^INSERT INTO public.tombstones \(.+\) VALUES \(.+\)$`,
+				Args:        fixTombstoneRow(),
+				ValidResult: sqlmock.NewResult(-1, 1),
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 41: You can extract this function as above.

components/director/internal/domain/webhook/repository_test.go
```diff
@@ -3,561 +3,393 @@ package webhook_test
...
+				IsSelect: true,
+				ValidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixColumns).AddRow(whModel.ID, givenApplicationID(), nil, whModel.Type, whModel.URL, fixAuthAsAString(t), nil, nil, whModel.Mode, whModel.CorrelationIDKey, whModel.RetryInterval, whModel.Timeout, whModel.URLTemplate, whModel.InputTemplate, whModel.HeaderTemplate, whModel.OutputTemplate, whModel.StatusTemplate)}
+				},
+				InvalidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixColumns)}
+				},
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
la4ezar, line 44: You can extract this function as above.

components/director/internal/domain/api/repository_test.go
```diff
@@ -2,343 +2,217 @@ package api_test
...
+				IsSelect: true,
+				ValidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixAPIDefinitionColumns()).AddRow(fixAPIDefinitionRow(apiDefID, "placeholder")...)}
+				},
+				InvalidRowsProvider: func() []*sqlmock.Rows {
+					return []*sqlmock.Rows{sqlmock.NewRows(fixAPIDefinitionColumns())}
+				},
+			},
+		},
+		ConverterMockProvider: func() testdb.Mock {
```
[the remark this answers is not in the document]
DimitarPetrov: The idea here in the generic suits is a bit different. This functions is used as a mock provider. Then in the actual suite each test uses this function to generate itself a mock from the specified type (since tests are generic we need to do define it outside where the suite is created and we have the mock type). Then each test of the suite calls this function to create a mock and then configures the expected invocations based on the test scenario.

In this case this is not an expected unused state and more of a mock generator that can be used to create empty (new) mocks for the specific type.

StanislavStefanov (changes requested):

components/director/internal/repo/testdb/generic_repo_update_tests.go
```diff
@@ -0,0 +1,206 @@
...
+	DBEntity                  interface{}
+	NilModelEntity            interface{}
+	TenantID                  string
+	DisableConverterErrorTest bool
+	UpdateMethodName          string
+}
+
+// Run runs the generic repo update test suite
+func (suite *RepoUpdateTestSuite) Run(t *testing.T) bool {
+	if len(suite.UpdateMethodName) == 0 {
```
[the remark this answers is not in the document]
DimitarPetrov: It seems like unneeded boilerplate in the suite for most of the entities. Most of the entities has this vanilla CRUD operations which are the main focus of the generic suites. However, some entities has slightly modified Read Update or Delete operations based on any set of conditions. This results in a different SQL query and the test suites can be reused, in this case you can override the methodName.

I think that it is useful to have a reasonable default name that works well for most of the entities and override it where it is not suitable instead of redelcareing it everytime.

components/director/internal/repo/testdb/generic_repo_delete_tests.go
```diff
@@ -0,0 +1,230 @@
...
+	RepoConstructorFunc   interface{}
+	MethodName            string
+	MethodArgs            []interface{}
+	IsTopLeveEntity       bool
+	IsDeleteMany          bool
+}
+
+// Run runs the generic repo delete test suite
+func (suite *RepoDeleteTestSuite) Run(t *testing.T) bool {
+	if len(suite.MethodName) == 0 {
```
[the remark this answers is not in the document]
DimitarPetrov: Same as below.

components/director/internal/repo/testdb/generic_repo_list_tests.go
```diff
@@ -0,0 +1,188 @@
...
+	return results[0].Interface(), err
+}
+
+// assertEqualElements checks that elements of two slices are equal.
+// It expects the first slice to contain pointer elements and converts each element of the second slice to pointer if it is a struct value before comparison.
+func assertEqualElements(t *testing.T, arr1, arr2 interface{}) {
+	array1 := reflect.ValueOf(arr1)
+	array2 := reflect.ValueOf(arr2)
+
+	require.Len(t, arr2, array1.Len())
```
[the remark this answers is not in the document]
DimitarPetrov: No, the require.Len method accepts an array and int. In this case we want to assert that two arrays has same length.
Therefore we just require the length of the second to be the length of the first.

The other way of doing it is `require.Equal(t, array2.Len(), array1.Len())`
Which do you think is clearer?

components/director/internal/repo/update.go
```diff
@@ -126,21 +176,28 @@ func (u *universalUpdater) unsafeUpdateSingleWithFields(ctx context.Context, dbE
...
-	if err = persistence.MapSQLError(ctx, err, u.resourceType, resource.Update, "while updating single entity from '%s' table", u.tableName); err != nil {
+	if err = persistence.MapSQLError(ctx, err, resourceType, resource.Update, "while updating single entity from '%s' table", u.tableName); err != nil {
 		return err
 	}
 
 	affected, err := res.RowsAffected()
 	if err != nil {
 		return errors.Wrap(err, "while checking affected rows")
 	}
+	if affected == 0 && (len(tenant) > 0 || u.tenantColumn != nil) {
```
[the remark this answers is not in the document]
DimitarPetrov: This way you will just hide the boolean expression in a method which should either way be opened to see what it is doing. I don't think that the name and the arguments provided will give more clarity as the expression is easy enough anyways.

I think that instead of func we at-least can create a Boolean variable above the if, but still don't see any benefit, do you think it is not clear enough and naming the expression somehow will help?

components/director/internal/domain/api/repository_test.go
```diff
@@ -2,343 +2,217 @@ package api_test
...
-		assert.Equal(t, apiDefID, modelAPIDef.ID)
-		assert.Equal(t, tenantID, modelAPIDef.Tenant)
-		convMock.AssertExpectations(t)
-		sqlMock.AssertExpectations(t)
-	})
+	suite := testdb.RepoGetTestSuite{
+		Name: "Get API",
+		SQLQueryDetails: []testdb.SQLQueryDetails{
+			{
+				Query:    regexp.QuoteMeta(`SELECT id, app_id, package_id, name, description, group_name, ord_id, short_description, system_instance_aware, api_protocol, tags, countries, links, api_resource_links, release_status, sunset_date, changelog_entries, labels, visibility, disabled, part_of_products, line_of_business, industry, version_value, version_deprecated, version_deprecated_since, version_for_removal, ready, created_at, updated_at, deleted_at, error, implementation_standard, custom_implementation_standard, custom_implementation_standard_description, target_urls, extensible, successors, resource_hash FROM "public"."api_definitions" WHERE id = $1 AND (id IN (SELECT id FROM api_definitions_tenants WHERE tenant_id = $2))`),
```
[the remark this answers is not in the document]
DimitarPetrov: Yes, I think this is a good further improvement. 

We talked about it, but in order to draw the line anywhere with the tests refactoring we decided not to rework each and every SQL query to use the columns slice.

I think that we should also get rid of the fixtures that include columns and use the columns slice defined in the productive package itself.

Even further, I don't think we need to have a slice anywhere as we have db tags on the Entity itself and this information can be derived from there via a method on a common interface.

Currently, all tests maintain the db column names 4 times:
1. In the db tags of the entity
2. In the columns slice in the repo
3. In the fixtures as a columns slice again
4. In the SQL query as a literal

This of course needs to be simplified, however I am not sure that this PR is the right place to do it.

components/director/internal/repo/list_union.go
```diff
@@ -5,18 +5,28 @@ import (
...
 )
 
-// UnionLister missing godoc
+// UnionLister is an interface for listing tenant scoped entities with either externally managed tenant accesses (m2m table or view) or embedded tenant in them.
+// It lists entities based on multiple parent queries. For each parent a separate list with a separate tenant isolation subquery is created and the end result is a union of all the results.
 type UnionLister interface {
 	// List stores the result into dest and returns the total count of tuples for each id from ids
-	List(ctx context.Context, tenant string, ids []string, idsColumn string, pageSize int, cursor string, orderBy OrderByParams, dest Collection, additionalConditions ...Condition) (map[string]int, error)
+	List(ctx context.Context, resourceType resource.Type, tenant string, ids []string, idsColumn string, pageSize int, cursor string, orderBy OrderByParams, dest Collection, additionalConditions ...Condition) (map[string]int, error)
+	SetSelectedColumns(selectedColumns []string)
```
la4ezar, line 20: Method is not used.

components/director/internal/repo/list_union.go
```diff
@@ -5,18 +5,28 @@ import (
...
+	Clone() *unionLister
```
la4ezar, line 21: Method is not used.

components/director/internal/domain/runtime/repository.go
```diff
@@ -267,11 +259,21 @@ func (r *pgRepository) GetOldestForFilters(ctx context.Context, tenant string, f
...
 		return nil, err
 	}
 
-	runtimeModel := runtimeEnt.ToModel()
+	runtimeModel := r.conv.FromEntity(&runtimeEnt)
 
 	return runtimeModel, nil
 }
+
+func (r *pgRepository) multipleFromEntities(entities RuntimeCollection) []*model.Runtime {
```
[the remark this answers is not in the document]
DimitarPetrov: I know that most of the times such methods are added to the converter, however in my opinion the converter should be aware of how to convert an entity to model and vice versa. 

The main difference is that if you add it to the converter, then this method becomes part of the interface of the converter -> should be mocked and therefore used in all tests as black box. I don't think that this is conceptually what a converter should do and I would like to be able to configure each of the FromEntity invocations with different mock setup in order to test different scenarios (this is not critical in this example but in others it is).

That's personal preference, if you think that it is better (it will be more consistent with the other entities) we can move it to the converter I am also fine with it.

components/director/internal/repo/testdb/generic_repo_list_tests.go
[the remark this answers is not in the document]
StanislavStefanov: I think require.Equal(t, array2.Len(), array1.Len()) is cleaner

components/director/internal/repo/update.go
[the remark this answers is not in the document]
StanislavStefanov: Mixing && and || in one expression can be confusing especially in case you are not confident on the concept - when use externally provided tenant and when the embeded one and after that when you should say the issue is that you dont have the rights to perform the operation or the issue is somwhere else. It is not just the straightforward read and understand but you should put some thought into it. Maybe boolean variable named checkedAuthorization or usedAuthorization will make it easier to understand.

components/director/internal/repo/delete.go
```diff
@@ -111,3 +142,125 @@ func (g *universalDeleter) unsafeDelete(ctx context.Context, conditions Conditio
...
+	stmtBuilder.WriteString(fmt.Sprintf("SELECT id FROM %s WHERE", g.tableName))
+
+	tenantIsolation, err := NewTenantIsolationCondition(resourceType, tenant, true)
+	if err != nil {
+		return err
+	}
+
+	conditions = append(conditions, tenantIsolation)
+
+	err = writeEnumeratedConditions(&stmtBuilder, conditions)
```
la4ezar, line 168: if err = ...; err != nil {...}

components/director/internal/repo/delete.go
```diff
@@ -111,3 +142,125 @@ func (g *universalDeleter) unsafeDelete(ctx context.Context, conditions Conditio
...
+	stmtBuilder.WriteString(fmt.Sprintf("DELETE FROM %s WHERE", g.tableName))
+
+	tenantIsolation, err := NewTenantIsolationCondition(resourceType, tenant, true)
+	if err != nil {
+		return err
+	}
+
+	conditions = append(conditions, tenantIsolation)
+
+	err = writeEnumeratedConditions(&stmtBuilder, conditions)
```
la4ezar, line 230: if err = ...; err != nil {...}

components/schema-migrator/migrations/director/20211122114401_storage-redesign.up.sql
```diff
@@ -0,0 +1,714 @@
+BEGIN;
+
+CREATE TABLE tenant_applications
+(
+    tenant_id   uuid NOT NULL,
+    id uuid NOT NULL,
```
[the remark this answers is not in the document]
DimitarPetrov: Initially we started with the resource_id as the name of the column.

However, this does not work with the current way views are created. As can be seen below views are created in the form of `SELECT api.*, ta.tenant_id, ta.owner FROM ...`. Therefore the views has an id in them and the m2m tables has resource_id in them.

There are four ways to solve this that we evaluated:
1. Use id in the m2m tables to unify it with the views
2. Make a distinction in the code between tables and views and create different tenant isolation -> with `SELECT resource_id ...` when there is a m2m table and with `SELECT id ...` when there is a view.
3. Make the views exhaustive for all the columns (not an option since this will require recreating the views when new columns are added to the table).
4. Make the views with three columns only `SELECT id as resource_id, ta.tenant_id, ta.owner FROM ...`, but I think we had an idea to use those views in the ORD Service to read directly from them and we wanted all the columns in there.

We choose 1. as the easiest and least confusing option from our PoV. What do you think?

components/schema-migrator/migrations/director/20211122114401_storage-redesign.up.sql
```diff
@@ -0,0 +1,714 @@
...
+    END IF;
+    RETURN NULL;
+END
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER insert_parent AFTER INSERT OR UPDATE ON business_tenant_mappings FOR EACH ROW EXECUTE PROCEDURE on_insert_new_parent('tenant_applications', 'tenant_runtimes');
+
+-- this trigger replaces the cascade on tenant_id foreign key in order to execute it BEFORE the tenant is deleted.
+-- that is needed because in order for delete_parent_chain select parent to work (before the tenant is actually deleted)
+CREATE OR REPLACE FUNCTION on_delete_tenant() RETURNS TRIGGER AS
```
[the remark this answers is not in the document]
DimitarPetrov: If we remove the delete_parent_chain trigger, yes. If we leave it, no because the cascade is triggered AFTER delete and the select from business_tenant_mappings in delete_parent_chain fail. Therefore we need the cascade to happen BEFORE the actual delete -> this trigger.

components/schema-migrator/migrations/director/20211122114401_storage-redesign.up.sql
```diff
@@ -0,0 +1,714 @@
...
+        EXECUTE format('DELETE FROM %I WHERE id = %L;', resource_table, OLD.id);
+    END IF;
+    RETURN NULL;
+END
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER delete_application_resource AFTER DELETE ON tenant_applications FOR EACH ROW EXECUTE PROCEDURE delete_resource('applications');
+CREATE TRIGGER delete_runtime_resource AFTER DELETE ON tenant_runtimes FOR EACH ROW EXECUTE PROCEDURE delete_resource('runtimes');
+
+CREATE OR REPLACE FUNCTION insert_parent_chain() RETURNS TRIGGER AS
```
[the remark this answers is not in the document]
DimitarPetrov: We will give it a try.

components/schema-migrator/migrations/director/20211122114401_storage-redesign.up.sql
```diff
@@ -0,0 +1,714 @@
+BEGIN;
```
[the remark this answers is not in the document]
DimitarPetrov: We are aware of that and we keep it in mind when rebasing.

components/schema-migrator/migrations/director/20211122114401_storage-redesign.up.sql
```diff
@@ -0,0 +1,714 @@
...
+ALTER TABLE automatic_scenario_assignments DROP COLUMN selector_value;
+
+CREATE OR REPLACE FUNCTION check_tenant_id_is_direct_parent_of_target_tenant_id() RETURNS TRIGGER AS
+$$
+DECLARE
+    count INTEGER;
+BEGIN
+    EXECUTE format('SELECT COUNT(1) FROM business_tenant_mappings WHERE id = %L AND parent = %L', NEW.target_tenant_id, NEW.tenant_id) INTO count;
+    IF count = 0 THEN
+        RAISE EXCEPTION 'target_tenant_id should be direct child of tenant_id';
```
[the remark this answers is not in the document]
DimitarPetrov: In the logs yes:
```
SQL Error. Caused by: target_tenant_id should be direct child of tenant_id.
```
In the API it is internal server error as always :D 

```
Internal Server Error [errorID=cf051a6d-5c68-43ef-84c8-e9507da34c8e]
```

components/director/internal/domain/runtime/service.go
```diff
@@ -245,6 +250,34 @@ func (s *service) Create(ctx context.Context, in model.RuntimeInput) (string, er
...
+	tnt, err := s.tenantSvc.GetTenantByID(ctx, rtmTenant)
+	if err != nil {
+		return "", errors.Wrapf(err, "while getting tenant with id %s", rtmTenant)
+	}
+
+	if len(tnt.Parent) == 0 {
+		return id, nil
+	}
+
+	ctxWithParentTenant := tenant.SaveToContext(ctx, tnt.Parent, "")
```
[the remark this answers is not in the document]
DimitarPetrov: The formation is in the tenant_id and the labels produced by it should be there as the LabelDef (the formation) is there.
It will be harder if we don't have the information in place (we will need it anyways and we should derive it from the tenant - more confusing)

Maybe it can be simplified, but it will change the interface much more, currently it is not that major as there are global_subaccount_id key and value which we just convert and store, if we do this then the client won't be able to read it itself after register it (not that ASAs are with embedded tenant).

Anyways, this whole entity will be removed with the formation redesign.

components/schema-migrator/migrations/director/20211122114401_storage-redesign.up.sql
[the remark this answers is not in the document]
KirilKabakchiev: ok so maybe it makes sense to remove the delete trigger i think, together with it this one goes away as well as the delete_resource one, right ?

components/director/pkg/resource/resource.go
```diff
@@ -1,77 +1,179 @@
...
-	TenantIndex Type = "tenantIndex"
-	// Schema missing godoc
+	// TenantAccess type represents tenant access resource.
+	TenantAccess Type = "tenantAccess"
+	// Schema type represents schema resource.
 	Schema Type = "schemaMigration"
 )
 
-// SQLOperation missing godoc
+var tenantAccessTable = map[Type]string{
```
KirilKabakchiev: these postgre specific table mappings are placed in a pkg/resource/resource.go - it really does not make much sense in terms of wording atleast :)

components/director/pkg/resource/resource.go
```diff
@@ -1,77 +1,179 @@
...
+	Document:              "documents_tenants",
+	BundleInstanceAuth:    "bundle_instance_auths_tenants",
+	API:                   "api_definitions_tenants",
+	EventDefinition:       "event_api_definitions_tenants",
+	Webhook:               "webhooks_tenants",
+	AppWebhook:            "application_webhooks_tenants",
+	RuntimeWebhook:        "runtime_webhooks_tenants",
+}
+
+var parentRelation = map[Type]Type{
```
KirilKabakchiev: this mapping you can probably autodiscover via the information schema, https://www.postgresql.org/message-id/40C500D8.9060009@actarg.com where you can explore contraints, fks and metadata about them

components/director/pkg/resource/resource.go
```diff
@@ -1,77 +1,179 @@
...
-// SQLOperation missing godoc
+var tenantAccessTable = map[Type]string{
+	// Tables
+
+	Application: "tenant_applications",
+	Runtime:     "tenant_runtimes",
+
+	// Views
+
+	RuntimeContext:        "runtime_contexts_tenants",
```
KirilKabakchiev: have you thought about making this a method on the storage entity type ? sql_types.go/entity

components/director/pkg/resource/resource.go
```diff
@@ -1,77 +1,179 @@
...
+}
+
+// Parent returns the parent type of the given type.
+func (t Type) Parent() (Type, bool) {
+	parent, ok := parentRelation[t]
+	return parent, ok
+}
+
+// TopLevelEntities is a set of entities that has a many-to-many relationship with the tenants.
+var TopLevelEntities = []Type{Application, Runtime}
```
KirilKabakchiev: this mapping seems unneeded - you can derive it from the parent-relation information (if someone has no parent "he is" top level)

components/director/internal/domain/runtime/service.go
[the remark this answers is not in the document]
KirilKabakchiev: yes true, no need to invest here, probably it makes sense to add some markers in the code where it  will be removed soon due to an upcoming change

components/director/pkg/resource/resource.go
DimitarPetrov: This is just a mapping between resource and it's tenant access table/view. Generally I see this package as a metadata stuff for each entity.

Can you elaborate why this does not make sense for you?

Even the SQL operations constants are here :D

components/director/pkg/resource/resource.go
DimitarPetrov: Yes, of course you can, but does it give us anything to be worth it the complexity?

components/director/pkg/resource/resource.go
DimitarPetrov: Yes initially we were thinking where to put it and the reasoning was that this is a static metadata for the type and not for a specific instance of the entity. It is more of a metadata of the table than of a row of the table.

components/director/pkg/resource/resource.go
DimitarPetrov: There are entities which have ref specific types and generic types. For example you have ApplicationWebhook and you have ApplicationTemplateWebhook. But for some cases where you don't know which type is the parent you have generic just Webhook. Those times, the entity does not have parent, but is not top level, it is child entity however, the parent is unknown.

It should be treated as child entity, but the view that is used in those cases is aggregated for all possible parents.

One more use-case for this is when a subaccount is moved. You need to process each and every top level entity and reconcile the accesses of their m2m tables, so you need to know which of them are top level and which of them not.

components/director/internal/repo/update.go
```diff
@@ -151,36 +208,39 @@ func (u *universalUpdater) unsafeUpdateSingleWithFields(ctx context.Context, dbE
...
 		for _, idCol := range u.idColumns {
 			preparedIDColumns = append(preparedIDColumns, fmt.Sprintf("%s = :%s", idCol, idCol))
 		}
 		stmtBuilder.WriteString(fmt.Sprintf(" %s", strings.Join(preparedIDColumns, " AND ")))
+		if u.tenantColumn != nil || len(tenant) > 0 {
+			stmtBuilder.WriteString(" AND")
+		}
+	}
+
+	if u.tenantColumn != nil { // if embedded tenant
```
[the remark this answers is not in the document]
DimitarPetrov: We did it, there is slightly more code duplication, but it is overall simpler and easier to digest :)

components/director/pkg/resource/resource.go
DimitarPetrov: One more thing is that only the create and update operations has an actual entity, all other operations has just a condition. If we add it to the entity we need to provide an empty entity in Get/Select/Delete instead of resourceType right now.

components/director/internal/repo/delete.go
```diff
@@ -111,3 +142,123 @@ func (g *universalDeleter) unsafeDelete(ctx context.Context, conditions Conditio
...
+	if err = writeEnumeratedConditions(&stmtBuilder, conditions); err != nil {
+		return errors.Wrap(err, "while writing enumerated conditions")
+	}
+	allArgs := getAllArgs(conditions)
+
+	query := getQueryFromBuilder(stmtBuilder)
+	log.C(ctx).Debugf("Executing DB query: %s", query)
+
+	var ids IDs
+	err = persist.SelectContext(ctx, &ids, query, allArgs...)
```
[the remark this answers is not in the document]
DimitarPetrov: I think we should because this + CTE approach will remove all the triggers which is great 🎉

components/director/pkg/resource/resource.go
KirilKabakchiev: well, thats easy to answer :) - usually when you write code, the code should hint you that such a mapping is needed - the interface would do so by asking you to provide the mapping value. A static mapping does not hint you, probably you find out that exists the first time you run your code and you get an error. Best case scenario we get rid of the centralized mappings

components/director/cmd/tenantfetcher-job/main.go
```diff
@@ -30,13 +30,13 @@ import (
...
-	QueryConfig                     tenantfetcher.QueryConfig
-	TenantFieldMapping              tenantfetcher.TenantFieldMapping
-	MovedRuntimeByLabelFieldMapping tenantfetcher.MovedRuntimeByLabelFieldMapping
+	Database                    persistence.DatabaseConfig
+	KubernetesConfig            tenantfetcher.KubeConfig
+	OAuthConfig                 tenantfetcher.OAuth2Config
+	APIConfig                   tenantfetcher.APIConfig
+	QueryConfig                 tenantfetcher.QueryConfig
+	TenantFieldMapping          tenantfetcher.TenantFieldMapping
+	MovedSubaccountFieldMapping tenantfetcher.MovedSubaccountsFieldMapping
```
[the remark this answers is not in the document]
DimitarPetrov: I think that only subaccount tenants can be moved, at least the previous logic expected it hardly.

components/director/pkg/resource/resource.go
KirilKabakchiev: ok, that makes sense. Another idea - we can reuse the idea for the top level mapping - we could put the m2m ref table/view name in the repo constuctor (example check func NewRepository(conv APIDefinitionConverter) *pgRepository) - there we already pass some table names to the unified repositories below, we could pass this one also. This way you distribute the cenralized (static) mapping to the locations that care about it and they dont have to look up the static mapping.
In the end you would end up with a repo instantiation that carries all the details (what is the table, what is the tenant isolation/m2m table, is it top level, etc)

components/director/pkg/resource/resource.go
KirilKabakchiev: ok thats a good point. Then another idea - we could put it in the constuctor (example check func NewRepository(conv APIDefinitionConverter) *pgRepository) - there we already pass some table names to the unified repositories below, we could pass this one also. Then you can remove the

components/director/internal/repo/delete.go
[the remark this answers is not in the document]
DimitarPetrov: Agreed offline to just delete the actual entity for now as it is exactly how it worked until now and we will extend it in the future when we have more information about the flow.

nkanchev (approved):

KirilKabakchiev (dismissed):

StanislavStefanov (approved):

nkanchev (approved):

StanislavStefanov (dismissed):

nkanchev (approved):

StanislavStefanov (approved):

PR #2118, opened 2021-11-29
Enable mtls in operations controller and pairing-adapter

krasish:
**Description**

This PR utilizes the certificate library introduced in [PR-2112](https://github.com/kyma-incubator/compass/pull/2112) to execute webhooks from the operations controller with mtls auth using the certificate cached in the k8s secret. Also the pairing adapter now has a configurable auth for getting tokens. When configured to use mtls it will use that same cert from the secret.

- Load CMP cert from the secret in operations controller and pairing adapter 
- Extend operations controller to execute mtls webhooks using the preloaded cert
- Extend pairing adapter to use mtls with the preloaded cert for token getting when configured to do so
- Export the accessstrategy package in the director component (moved to pkg from internal)
- Adapt helm charts
- Fix and add unit test
- Add e2e tests for both flows (mtls webhook and pairing adapter with mtls)

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [x] Integration tests


components/pairing-adapter/cmd/main.go
```diff
@@ -26,23 +29,36 @@ func main() {
...
+	client := &http.Client{
+		Transport: httputil.NewCorrelationIDTransport(transport),
+		Timeout:   conf.ClientTimeout,
 	}
-
-	baseClient := &http.Client{
-		Transport: httputil.NewCorrelationIDTransport(http.DefaultTransport),
+	ctx = context.WithValue(ctx, oauth2.HTTPClient, client)
+
+	if conf.Auth.Type == adapter.AuthTypeOauth {
```
desislavaa, line 42: Minor one - switch instead of if?

desislavaa (commented):
Looks great! Just two reeeally minor comments

desislavaa (approved):

PR #2119, opened 2021-12-01
Adapt cert loader for local director installation

PetarTodorovv:
**Description**
As part of https://github.com/kyma-incubator/compass/pull/2112 local director installation was not adapter. This PR fix this.
Certificate cache in local director installation will be populated if a certificate and key are provided otherwise it will be empty

- Adaptation for local director installation and unit tests
- Extract the secret name containing the client certificate as environment variable
- Increase tenant-fetcher tests memory

**Related issue(s)**
- https://github.com/kyma-incubator/compass/pull/2112

**Pull Request status**
- [x] Implementation
- [x] Unit tests


components/director/pkg/certloader/loader.go
```diff
@@ -61,23 +89,34 @@ func StartCertLoader(ctx context.Context, externalClientCertSecret string) (Cach
...
 	}
-	certCache := NewCertificateCache()
-	certLoader := NewCertificatesLoader(certCache, k8sClientSet.CoreV1().Secrets(parsedCertSecret.Namespace), parsedCertSecret.Name, time.Second)
+	certLoader := NewCertificateLoader(certLoaderConfig, certCache, k8sClientSet.CoreV1().Secrets(parsedCertSecret.Namespace), parsedCertSecret.Name, time.Second)
 	go certLoader.Run(ctx)
 
 	return certCache, nil
 }
 
+// Run populates the certificate cache on local setup with provided certificate and key
```
KirilKabakchiev, line 98: its normally not ok to have local dev code in the prod code base, but i havent looked at the prev PR. Is this the only way you can think of for this to be achieved ?

KirilKabakchiev: One idea - you could try to install https://github.com/k3s-io/k3s#quick-start---install-script as part of the local run.sh script and then point the kubeconfig to point to it. And in the run.sh also always create a certificate in the relevant secret

PetarTodorovv: That's the thing that I came up with. Most probably there is another way to do it but in the productive setup(and local installation) we use k8s cluster with watch mechanism. The "problem" with the director is that we run it standalone and don't have k8s cluster. I can check what are the alternatives

PetarTodorovv: I've missed the previous comment. About the disabled certloader locally, that was the initial idea to return empty cache. And currently in the PR that's the default behaviour but I introduce a functionality where we can add custom certificate in the director environment(run.sh script) so we can populate the cache with it. The motivation behind this was if we want to run an ord app locally + director(run.sh) which is configured with client cert issued by CA used in the ord app. I'm not sure if we have such case though.

components/director/run.sh
```diff
@@ -89,10 +92,17 @@ function cleanup() {
...
+    echo -e "${GREEN}Destroying k3d cluster..."
+    k3d cluster delete k3d-cluster
 }
 
 trap cleanup EXIT
 
+echo -e "${GREEN}Creating k3d cluster..."
+curl -s https://raw.githubusercontent.com/rancher/k3d/main/install.sh | TAG=v5.2.0 bash
+k3d cluster create k3d-cluster --api-port 6550 --servers 1 --port 443:443@loadbalancer --image rancher/k3s:v1.22.4-k3s1 --kubeconfig-update-default --wait
+
```
DimitarPetrov, line 105: I personally don't like this approach as it will make the run.sh significantly slower and heavier while the benefit is nearly zero (no one is actually using local run of director for testing externally issued certs).

If we would like to leave it like this at least we need a way to make it optional, otherwise, the whole purpose of the script is gone as it will become more and more like minikube installation with the director only.

PetarTodorovv: That was my concern as well. But we've discussed it on our daily and decide to check how effort it will be and how much time it will needs to start. This cluster creation takes around 15 seconds which I think it's acceptable so far? But we can discuss it again?

PetarTodorovv: We've discussed it offline and agreed to use this approach because the cluster creation time is minimal and we can use it for further improvements/testing/debugging

nkanchev (dismissed):

nkanchev (approved):

KirilKabakchiev (dismissed):
looks ok to me

nkanchev (dismissed):

DimitarPetrov (dismissed):

KirilKabakchiev (approved):

nkanchev (approved):

DimitarPetrov (approved):

PR #2120, opened 2021-12-02
Secure graphql with external certificate for configured consumers

la4ezar:
**Description**

- add new subject to consumer type configuration in connector
- add scopes based on consumer type in certificate service context provider(tenant-mapping hydrator)

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/tenantmapping/cert_service_context_provider.go
```diff
@@ -49,39 +50,74 @@ func (m *certServiceContextProvider) GetObjectContext(ctx context.Context, reqDa
...
+		var err error
+		externalTenantID, err = reqData.GetExternalTenantID() // will return tenant ID from header if it exists
+		if err != nil {
+			log.C(ctx).WithError(err).Errorf("Failed to get external tenant ID: %v", err)
+			return ObjectContext{}, err
 		}
+	}
 
-		objCtx := NewObjectContext(NewTenantContext(externalTenantID, tenantMapping.ID), m.tenantKeys, scopes, authDetails.Region, "", authDetails.AuthID, authDetails.AuthFlow, consumer.Runtime, CertServiceObjectContextProvider)
+	scopes, err := p.directorScopes(reqData.GetConsumerTypeFromExtra())
```
krasish, line 79: We already have the conusmer type in extradata.ConsumerType. The way it was populated there makes sure that the consumer type is never empty (it will be runtime if empty initially). Getting it with this method here might return an empty string while the one above will not be. Is that desired? If not, use extradata.ConsumerType instead of calling this method.

components/director/internal/tenantmapping/cert_service_context_provider.go
```diff
@@ -49,39 +50,74 @@ func (m *certServiceContextProvider) GetObjectContext(ctx context.Context, reqDa
...
+			log.C(ctx).Warningf("Could not find tenant with external ID: %s, error: %s", externalTenantID, err.Error())
+			log.C(ctx).Infof("Returning tenant context with empty internal tenant ID and external ID %s", externalTenantID)
+			return NewObjectContext(NewTenantContext(externalTenantID, ""), p.tenantKeys, scopes, authDetails.Region, "", authDetails.AuthID, authDetails.AuthFlow, consumer.Runtime, CertServiceObjectContextProvider), nil
+		}
+		return ObjectContext{}, errors.Wrapf(err, "while getting external tenant mapping [ExternalTenantID=%s]", externalTenantID)
+	}
 
-		return objCtx, nil
+	if extraData.AccessLevel != "" && tenantMapping.Type != extraData.AccessLevel {
+		// TODO improve message
```
krasish, line 98: Improve or remove TODO if already done.

components/director/internal/tenantmapping/cert_service_context_provider.go
```diff
@@ -91,3 +127,15 @@ func (m *certServiceContextProvider) Match(_ context.Context, data oathkeeper.Re
 
 	return false, nil, nil
 }
+
+func (p *certServiceContextProvider) directorScopes(consumerType model.SystemAuthReferenceObjectType) (string, error) {
+	if consumerType == "" { // TODO improve
```
krasish, line 132: Improve or remove TODO if already done.

components/connector/internal/subject/subject_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+func TestAuthIDFromSubjectFunc(t *testing.T) {
+	t.Run("Success when internal consumer id is provided in subject", func(t *testing.T) {
+		p, err := subject.NewProcessor(validConfig, "")
+		require.NoError(t, err)
+
+		res := p.AuthIDFromSubjectFunc()(validSubject)
+		require.Equal(t, validInternalConsumerID, res)
+	})
+	t.Run("Success when internal consumer id is not provided", func(t *testing.T) {
```
krasish, line 78: The messages are the same for the next 3 subtests.

components/connector/internal/subject/subject.go
```diff
@@ -0,0 +1,123 @@
...
+func subjectsMatch(actualSubject, expectedSubject string) bool {
+	actualSubjectComponents := strings.Split(actualSubject, ",")
+	expectedSubjectComponents := strings.Split(expectedSubject, ",")
+
+	if len(actualSubjectComponents) != len(expectedSubjectComponents) {
+		return false
+	}
+
+	for _, expectedSubjectComponent := range expectedSubjectComponents {
+		if !strings.Contains(actualSubject, strings.TrimSpace(expectedSubjectComponent)) {
```
krasish, line 117: This flow is not covered by the unit tests but it seems like quite an important one because we don't want to match unexpected subjects. I'd add a unit test if it's not much effort for some reason or is covered by the e2e tests.

krasish (changes requested):

components/director/cmd/director/main.go
```diff
@@ -437,11 +439,24 @@ func getTenantMappingHandlerFunc(transact persistence.Transactioner, authenticat
 	tenantConverter := tenant.NewConverter()
 	tenantRepo := tenant.NewRepository(tenantConverter)
 
+	consumerExistsCheckers := map[model.SystemAuthReferenceObjectType]func(ctx context.Context, id string) (bool, error){
```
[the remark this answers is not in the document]
desislavaa: it's not the same - we have integration systems once, then runtimes, and we can also have applications - it's basically a wrapper of the service layer logic - instead of three different Svc fields in the CtxProvider we have this map

components/director/internal/tenantmapping/cert_service_context_provider.go
desislavaa: we do this in order to get the `default` scopes, instead of the runtime ones

components/director/internal/tenantmapping/cert_service_context_provider.go
```diff
@@ -49,39 +50,74 @@ func (m *certServiceContextProvider) GetObjectContext(ctx context.Context, reqDa
...
+			log.C(ctx).Infof("Returning tenant context with empty internal tenant ID and external ID %s", externalTenantID)
+			return NewObjectContext(NewTenantContext(externalTenantID, ""), p.tenantKeys, scopes, authDetails.Region, "", authDetails.AuthID, authDetails.AuthFlow, consumer.Runtime, CertServiceObjectContextProvider), nil
+		}
+		return ObjectContext{}, errors.Wrapf(err, "while getting external tenant mapping [ExternalTenantID=%s]", externalTenantID)
+	}
 
-		return objCtx, nil
+	if extraData.AccessLevel != "" && tenantMapping.Type != extraData.AccessLevel {
+		// TODO improve message
+		return ObjectContext{}, apperrors.NewUnauthorizedError(fmt.Sprintf("Certificate with auth ID %s has no access to tenant with ID %s", authDetails.AuthID, tenantMapping.ExternalTenant))
```
pankrator, line 98: Add the tenant types in the error message, as this is what is compared

components/director/internal/tenantmapping/cert_service_context_provider.go
```diff
@@ -49,39 +50,73 @@ func (m *certServiceContextProvider) GetObjectContext(ctx context.Context, reqDa
...
-			if apperrors.IsNotFoundError(err) {
-				log.C(ctx).Warningf("Could not find tenant with external ID: %s, error: %s", externalTenantID, err.Error())
+	// the authID in this flow is an OU selected by the Connector
+	externalTenantID := authDetails.AuthID
+
+	if matchedComponentName[0] != "director" { // ORD Flow, set the external tenant ID both for internal and external tenants
+		objCtx := NewObjectContext(NewTenantContext(externalTenantID, externalTenantID), p.tenantKeys, "", authDetails.Region, "", authDetails.AuthID, authDetails.AuthFlow, consumer.Runtime, CertServiceObjectContextProvider)
+		log.C(ctx).Infof("Successfully got object context: %+v", objCtx)
+		return objCtx, nil
+	}
```
pankrator, line 61: The logic of the `GetObjectContext` up to this line seems like it is related to ORD only.
My proposal is to try and extract it either in another ObjectContextProvider with matcher for the component or at least move it in a separate method in this context provider.

components/director/internal/tenantmapping/cert_service_context_provider.go
```diff
@@ -49,39 +50,73 @@ func (m *certServiceContextProvider) GetObjectContext(ctx context.Context, reqDa
...
 
-		log.C(ctx).Infof("Getting the tenant with external ID: %s", externalTenantID)
-		tenantMapping, err := m.tenantRepo.GetByExternalTenant(ctx, externalTenantID)
-		if err != nil {
-			if apperrors.IsNotFoundError(err) {
-				log.C(ctx).Warningf("Could not find tenant with external ID: %s, error: %s", externalTenantID, err.Error())
+	// the authID in this flow is an OU selected by the Connector
+	externalTenantID := authDetails.AuthID
+
+	if matchedComponentName[0] != "director" { // ORD Flow, set the external tenant ID both for internal and external tenants
```
pankrator, line 57: It would be better to check explicitly for "ord" instead of not director. Otherwise if another component is introduced, this logic might not work.

components/director/internal/tenantmapping/cert_service_context_provider.go
```diff
@@ -91,3 +126,15 @@ func (m *certServiceContextProvider) Match(_ context.Context, data oathkeeper.Re
 
 	return false, nil, nil
 }
+
+func (p *certServiceContextProvider) directorScopes(consumerType model.SystemAuthReferenceObjectType) (string, error) {
+	if consumerType == "" {
+		consumerType = "default"
```
pankrator, line 132: What is the use-case for the default after this change?

pankrator (dismissed):

components/director/internal/tenantmapping/cert_service_context_provider.go
desislavaa: It's removed in the "Storage Redesign" PR, we'll remove it here as well

components/director/internal/tenantmapping/cert_service_context_provider.go
desislavaa: BAS flow :alien: - those scopes were hardcoded before as default

components/director/cmd/director/main.go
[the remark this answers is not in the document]
desislavaa: Thanks for spotting this! 💯 As discussed, we shouldn't check the system auths, but the int systems

krasish (approved):

PR #2123, opened 2021-12-03
Introduce Write-only Protected Labels on Runtimes

StefanCholakov:
**Description**

- Introduce immutable labels(labels which are returned on get/list, but cannot be modified)

**Pull Request status**
-  [X] Implementation
-  [X] Unit tests
-  [X] Integration tests
-  [X] `chart/compass/values.yaml` is updated
-  [X] Mocks are regenerated, using the automated script


la4ezar (dismissed):

components/director/internal/domain/runtime/service.go
```diff
@@ -608,26 +617,30 @@ func extractUnProtectedLabels(labels map[string]*model.Label, protectedLabelsKey
...
+	protected, err := regexp.MatchString(protectedLabelsKeyPattern, labelKey)
+	if err != nil {
+		return false, err
+	}
+	immutable, err := regexp.MatchString(immutableLabelsKeyPattern, labelKey)
 	if err != nil {
 		return false, err
 	}
-	return matched, nil
+	return !protected && !immutable, err
```
PetarTodorovv, line 643: About this function we have two checks/conditions and my concert is if someone pass a string label that match both of them because the "immutable" pattern is just `xsappnameCMPClone`. For example - `someTextxsappnameCMPCloneAnotherText_defaultEventing`. I assume it will match and in this case we will return error?

StefanCholakov: It will match because of the "_defaultEventing" suffix

components/director/internal/features/features.go
```diff
@@ -4,7 +4,7 @@ package features
 type Config struct {
 	DefaultScenarioEnabled        bool   `envconfig:"default=true,APP_DEFAULT_SCENARIO_ENABLED"`
 	ProtectedLabelPattern         string `envconfig:"default=.*_defaultEventing|^consumer_subaccount_ids$"`
-	ImmutableLabelPattern         string `envconfig:"default=xsappnameCMPClone,APP_SELF_REGISTER_LABEL_KEY"`
+	ImmutableLabelPattern         string `envconfig:"default=^xsappnameCMPClone$,APP_SELF_REGISTER_LABEL_KEY"`
```
PetarTodorovv, line 7: The way it's written now, we'll get the "immutable" pattern from `APP_SELF_REGISTER_LABEL_KEY` and if there isn't such environment variable we'll use default value which is `^xsappnameCMPClone$` right? But `APP_SELF_REGISTER_LABEL_KEY` is configure to `xsappnameCMPClone` not `^xsappnameCMPClone$`. Maybe it's better to create a new environment variable dedicated to immutable pattern which value will be the value from `APP_SELF_REGISTER_LABEL_KEY` but with `^` in the begging and `$` in the end and remove the default value? That way if for some reason we change the label key it should work out of the box. Because the immutable pattern is that label key but with the special symbols

components/director/internal/domain/runtime/service.go
PetarTodorovv: Yep, that's the "protected" label pattern but it will also match the "immutable" pattern which is `xsappnameCMPClone`

PetarTodorovv (dismissed):

PetarTodorovv (approved):

la4ezar (approved):

PR #2124, opened 2021-12-07
Lazily store missing subaccounts

DimitarPetrov:
**Description**

- When creating runtime with global_subaccount_id label, first ensure that the provided subaccount exists.
- When creating ASA, first ensure target_tenant_id exists.
- When assigning formation, first ensure a subaccount exists.

- Fix e2e tests on real env where default scenario is disabled.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


desislavaa (dismissed):

ataleksandrov:
Seems good to me, just wanted to point out something. If the tenant fetcher starts a transaction and inserts the subaccount first, then we will block in director till the transaction is committed or rolled back. As we know, tenant fetcher could hold transactions up to some seconds, this may cause (client)timeouts in this flows in director.

desislavaa (approved):

PR #2128, opened 2021-12-08
Introduce ORD defaultBundle for APIs/Events

nyordanoff:
**Description**

This PR introduces a new boolean field to the BundleReference entity which indicates whether or not the specific Bundle is default for the referenced entity (API/Event)

- implementation
- adapting and extending unit tests
- add e2e tests

**Related issue(s)**

- https://github.com/kyma-incubator/ord-service/pull/50

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/open_resource_discovery/validation.go
```diff
@@ -1030,6 +1036,30 @@ func validateAPIPartOfConsumptionBundles(value interface{}, targetURLs json.RawM
...
+	var isFound bool
+	for _, bundleRef := range partOfConsumptionBundles {
+		if *defaultConsumptionBundle == bundleRef.BundleOrdID {
+			isFound = true
+			break
+		}
+	}
+
+	if !isFound {
+		return errors.New("defaultConsumptionBundle MUST be an existing option in the corresponding partOfConsumptionBundles array")
```
desislavaa, line 1058: I'm having some UX flashbacks with that uppercase MUST, we should probably lowecase it, since it's just _"must"_ in the other error messages

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -1030,6 +1036,30 @@ func validateAPIPartOfConsumptionBundles(value interface{}, targetURLs json.RawM
 	return nil
 }
 
+func validateDefaultConsumptionBundle(value interface{}, partOfConsumptionBundles []*model.ConsumptionBundleReference) error {
+	defaultConsumptionBundle, ok := value.(*string)
+	if !ok {
+		return errors.New("error while casting to defaultConsumptionBundle")
```
desislavaa, line 1042: maybe "expected string value for defaultConsumptionBundle, found %T"?

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -1030,6 +1036,30 @@ func validateAPIPartOfConsumptionBundles(value interface{}, targetURLs json.RawM
...
+	}
+
+	if defaultConsumptionBundle == nil {
```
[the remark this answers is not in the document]
nyordanoff: The field is optional by the spec that's why we store it as a pointer in our codebase. If it isn't provided in the document the field will be nil which is fine. 

If it is provided but is an empty string, that will be caught by the validation below because the empty string will not be part of the array `partOfConsumptionBundles` (i.e there is no bundle with ID empty string so your defaultBundle can not be an empty one).

components/director/internal/open_resource_discovery/validation.go
nyordanoff: Makes sense, totally.

desislavaa (approved):

PR #2132, opened 2021-12-08
System Fetcher dual oauth mode

ataleksandrov:
## Description
Enable dual client credentials - pure and mTLS. System fetcher by default does pure client credentials, but in case of missing `client_secret` from the oauth config it will try to perform mTLS cc flow.

#### Changes proposed in this pull request:
- `tokenProvider` is extended to respect additional headers
- `mtlsOauthToken` provider is added
- enable issuing access tokens via certificate in ext services mock
- add authorization middleware on mocked system api
- `gke_integration` job tests mTLS cc and `compass-integration-no-dump` tests pure cc flow


components/director/pkg/auth/mtls_token_provider.go
```diff
@@ -0,0 +1,168 @@
...
+
+// mtlsTokenAuthorizationProvider presents a AuthorizationProvider implementation which crafts OAuth Bearer token values for the Authorization header using mtls http client
+type mtlsTokenAuthorizationProvider struct {
+	httpClient *http.Client
+}
+
+// NewMtlsTokenAuthorizationProvider constructs an TokenAuthorizationProvider
+func NewMtlsTokenAuthorizationProvider(oauthCfg oauth.Config, cache CertificateCache, creator MtlsClientCreator) *mtlsTokenAuthorizationProvider {
+	return &mtlsTokenAuthorizationProvider{
+		httpClient: creator(cache, oauthCfg.SkipSSLValidation, oauthCfg.TokenRequestTimeout),
```
KirilKabakchiev, line 71: is there any reason to pass a creator here and not just create the client before creating the tokenprovider and pass a client interface to the tokenprovider

ataleksandrov: Yes absolutely, I had to choose whether to create` mtlsClient interface`(+ implementation) or use `creatorFunc` + standard http.Client implementation. For the sake of simplicity I chose creatorFunc. Do you see any drawbacks besides that there will always be just one default creator?

components/director/cmd/systemfetcher/main.go
```diff
@@ -215,7 +213,26 @@ func createSystemFetcher(cfg config, cfgProvider *configprovider.Provider, tx pe
 	appTemplateRepo := apptemplate.NewRepository(appTemplateConv)
 	appTemplateSvc := apptemplate.NewService(appTemplateRepo, webhookRepo, uidSvc)
 
-	systemsAPIClient := systemfetcher.NewClient(cfg.APIConfig, cfg.OAuth2Config, systemfetcher.DefaultClientCreator)
+	var authProvider httputil.AuthorizationProvider
+
+	if cfg.OAuth2Config.ClientSecret == "" {
```
pankrator, line 218: This option to use mtls client is a bit hidden. Perhaps introducing a new option specifically for mtls would be better and more meaningful

ataleksandrov: In the near future(~ 1 month) the system fetcher will use only the mTLS flow. Having that in mind, I thought it would be easier to delete the not needed code.

pankrator (approved):

PR #2158, opened 2021-12-22
Rework tenant search

alextargov:
**Description**
Due to the high volume of tenants available, querying them seems like a costly operation. In order to avoid high CPU/memory consumption on our pods, we introduce paging for tenants. The default page would max to 500 tenants.
Also, for the purpose of our UI, we introduce tenant search possibilities. Tenants could be queried by their ID, external ID or external name.

- Support search keyword for tenant query
- Add paging on the tenant query

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/tenant/repository.go
```diff
@@ -155,6 +159,51 @@ func (r *pgRepository) List(ctx context.Context) ([]*model.BusinessTenantMapping
...
+			WHERE %s = $1 AND (%s)`, selectColumns, tableName, statusColumn, r.buildSearchCondition(searchColumns, searchTerm))
+
+	page, totalCount, err := r.pageableQuerierGlobal.ListWithQueryGlobal(ctx, query, []interface{}{tenant.Active}, externalNameColumn, pageSize, cursor, &entityCollection)
+	if err != nil {
+		return nil, errors.Wrap(err, "while listing tenants from DB")
+	}
+
+	items := make([]*model.BusinessTenantMapping, 0, len(entityCollection))
+
+	for _, entity := range entityCollection {
```
StefanCholakov, line 195: You can extract this code in a separate function. You can see multipleFromEntities functions in application and runtime repositories for reference.

components/director/pkg/graphql/schema.graphql
```diff
@@ -1203,7 +1209,7 @@ type Query {
 	**Examples**
 	- [query tenants](examples/query-tenants/query-tenants.graphql)
```
StefanCholakov, line 1210: This example should be re-generated

StefanCholakov (approved):

StefanCholakov (dismissed):

components/director/internal/domain/tenant/repository.go
```diff
@@ -146,13 +150,48 @@ func (r *pgRepository) List(ctx context.Context) ([]*model.BusinessTenantMapping
...
+	for idx, field := range fields {
+		if idx != lastIdx {
+			stmt := fmt.Sprintf(`%s ILIKE '%%%s%%' OR `, field, searchTerm)
+			stmtBuilder.WriteString(stmt)
+			continue
+		}
+
+		stmt := fmt.Sprintf(`%s ILIKE '%%%s%%'`, field, searchTerm)
+		stmtBuilder.WriteString(stmt)
+	}
```
DimitarPetrov, line 170: The arguments in the like operator must not be crafted by hand (via sprintf) in order to avoid SQL injection.
The right way to do this is to replace the `%s` with `$%d` and add the idx which will result in positional arguments.
Then you need an []interfaces representing the args and add all the searchTerms there.

components/director/internal/repo/list_pageable.go
```diff
@@ -22,6 +22,7 @@ type PageableQuerier interface {
 // PageableQuerierGlobal is an interface for listing with paging of global entities.
 type PageableQuerierGlobal interface {
 	ListGlobal(ctx context.Context, pageSize int, cursor string, orderByColumn string, dest Collection, additionalConditions ...Condition) (*pagination.Page, int, error)
+	ListWithQueryGlobal(ctx context.Context, query string, args []interface{}, orderByColumn string, pageSize int, cursor string, dest Collection) (*pagination.Page, int, error)
```
DimitarPetrov, line 25: I don't think this is needed at all. You can just introduce `LIKE` condition and use the standard querier.
This way we will have a few added benefits:
- No SQL injection mistakes can happen as the logic is encapsulated and abstracted with conditions as with the other entities.
- No hardcoded SQL in the tenant repo.

DimitarPetrov (dismissed):

StefanCholakov (dismissed):

StefanCholakov (approved):

PR #2160, opened 2021-12-23
Change label value used during self-registration

krasish:
**Description**
As it is implemented at the moment, CMP uses the distinguishing label's value for input during the self-registration preparation process. However this value might include unsupported characters. This PR changes the input to use the runtime ID.

- Change clone suffix to runtime id

**Pull Request status**

- [X] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/runtime/service_test.go
```diff
@@ -192,9 +192,7 @@ func TestService_CreateWithMandatoryLabels(t *testing.T) {
 				return tenantSvc
 			},
 			UIDServiceFn: func() *automock.UidService {
-				svc := &automock.UidService{}
-				svc.On("Generate").Return(runtimeID)
-				return svc
+				return &automock.UidService{}
```
la4ezar, line 195: You can extract 
`func() *automock.UidService {
				return &automock.UidService{}
}`
in separate function named 'unusedUIDService' for example.

PetarTodorovv:
Looks okay to me ✅ Only need adaptation in the e2e tests because we add the "distinguish label" in the runtime input and after that validate/assert the expected result. But now the "distinguish label" contains runtimeID which we'll have after registration(we don't have it beforehand when build runtime input)

components/director/internal/domain/runtime/resolver.go
```diff
@@ -221,7 +224,7 @@ func (r *Resolver) RegisterRuntime(ctx context.Context, in graphql.RuntimeInput)
 
 	ctx = persistence.SaveToContext(ctx, tx)
 
-	id, err := r.runtimeService.CreateWithMandatoryLabels(ctx, convertedIn, labels)
+	err = r.runtimeService.CreateWithMandatoryLabels(ctx, convertedIn, id, labels)
```
la4ezar, line 227: if err = ...; err != nil {...}

PetarTodorovv (approved):

PR #2169, opened 2022-01-06
ORD Global Registry for global products / vendors

DimitarPetrov:
**Description**

Integration with ORD Global Registry that provides global product and vendors centrally.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/open_resource_discovery/service.go
```diff
@@ -57,12 +58,22 @@ func NewAggregatorService(transact persistence.Transactioner, labelRepo labelRep
...
 		tombstoneSvc:       tombstoneSvc,
 		tenantSvc:          tenantSvc,
+		globalRegistrySvc:  globalRegistrySvc,
 		ordClient:          client,
 	}
 }
 
 // SyncORDDocuments performs resync of ORD information provided via ORD documents for each application
 func (s *Service) SyncORDDocuments(ctx context.Context) error {
+	globalResourcesOrdIDs, err := s.globalRegistrySvc.SyncGlobalResources(ctx)
```
NickyMateev, line 68: Just thoughts - adding this sync here makes the global resources sync a hard dependency in order to process any other app ord documents. From one PoV it makes sense since the app ord docs can reference new things coming from the glob registry and we need to make sure we have them. From another PoV the global registry probably won't change much and even when it does, we'll most likely fetch the new changes in it well before any app ord doc has adopted the new product/vendor changes which might have appeared in the global registry.

I'm saying all this just for us to consider if having this sync as hard dependency can backfire in situations where it could go missing and it would block us from syncing and new app ord documents which might have appeared along the way.
I'm okay with leaving it like that tbh, but we need to be aware of the possible problem it could cause and be ready to change the global resource sync to an independent job if it happens.

components/schema-migrator/migrations/director/20220106093539_ord-global-registry.up.sql
```diff
@@ -0,0 +1,95 @@
...
+CREATE VIEW api_product AS
+SELECT api_definitions.id     AS api_definition_id,
+       p.id                   AS product_id
+FROM api_definitions,
+     jsonb_array_elements_text(api_definitions.part_of_products) AS elements
+         JOIN products p ON elements.value = p.ord_id WHERE p.app_id = api_definitions.app_id OR p.app_id IS NULL;
+
+DROP VIEW event_product;
+
+------
```
NickyMateev, line 31: very minor: these `-----` comment separators seems like are placed to group api_product, event_product,  package_product so that it's more readable, but currently the DELETE VIEW of the event_product and package_product overlap with the previous group

components/director/internal/open_resource_discovery/service.go
DimitarPetrov: This is the reason I made it optional by just logging if something goes wrong instead of failing the whole aggregator execution. However, you are right that some app syncs will fail because we will proceed with empty globalResourcesOrdIDs.

What do you think about the approach where in the error case (when global registry sync fails) we not only log but get all globalResourcesOrdIDs from the database (that were available in the global registry on the last successful execution)? This way we can safely process apps that reference global resources we already know and once the registry is up again we will resync them again.

NickyMateev: Wow, i totally missed that you log only - I thought i saw that it errored.
But generally, yes, I think it would be best to fetch the global resources because otherwise it would almost definitely fail, because our current lob always uses a global resource

chart/compass/values.yaml
```diff
@@ -471,6 +471,7 @@ global:
     dbPool:
       maxOpenConnections: 2
       maxIdleConnections: 2
+    globalRegistryUrl: http://compass-external-services-mock.compass-system.svc.cluster.local:8085/.well-known/open-resource-discovery
```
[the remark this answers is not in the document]
DimitarPetrov: Nice idea! However, it seems like anchors cannot be dereferenced in a composite value like this:
https://stackoverflow.com/questions/40631913/concatenation-of-yaml-anchors-with-another-string-in-keys

PR #2170, opened 2022-01-09
Fix tenant fetcher to read different field from the event

nyordanoff:
**Description**

This PR changes the field from the event response from which the Tenant Fetcher gets the parent of a subaccount.

- changed field
- adapted unit and e2e tests

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/components/director/internal/tenantfetcher/event_page_test.go
+++ b/components/director/internal/tenantfetcher/event_page_test.go
@@ -173,6 +173,7 @@ func Test_getMovedSubaccounts(t *testing.T) {
 func Test_getTenantMappings(t *testing.T) {
 	idField := "id"
 	globalAccountGUIDField := "globalAccountGUID"
+	globalAccountKey := "gaID"
 	id := "1"
 	nameField := "name"
 	name := "test-name"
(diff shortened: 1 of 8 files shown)


components/director/internal/tenantfetcher/event_page_test.go
```diff
@@ -332,7 +333,7 @@ func Test_getTenantMappings(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			events := make([][]byte, 0, len(test.detailsPairs))
 			for i, detailPair := range test.detailsPairs {
-				events = append(events, fixEventWithDetails(fmt.Sprintf("id%d", i), fmt.Sprintf("foo%d", i), "GlobalAccount", constructJSONObject(detailPair...), test.fieldMapping))
+				events = append(events, fixEventWithDetailsWithoutGA(fmt.Sprintf("id%d", i), fmt.Sprintf("foo%d", i), "GlobalAccount", constructJSONObject(detailPair...), test.fieldMapping))
```
DimitarPetrov, line 336: The differentiation between GA events and SA events can be removed here as well.

tests/tenant-fetcher-job/tests/handler_test.go
```diff
@@ -75,11 +75,11 @@ const (
 		"guid": "%s",
 		"displayName": "%s",
 		"subdomain": "%s",
-		"parentGuid": "%s",
```
DimitarPetrov, line 78: Leave it as well with a different value from the globalAccountGUID field to assert that the feature actually works.

DimitarPetrov (changes requested):

components/director/internal/tenantfetcher/event_page_test.go
nyordanoff: Yes, I just adapted it after discussing the other occurrence in the service tests.

DimitarPetrov (approved):

PR #2171, opened 2022-01-09
Clean directories from DB saved as global accounts

StefanCholakov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/dircleaner/service.go
```diff
@@ -0,0 +1,92 @@
...
+func NewCleaner(transact persistence.Transactioner, tenantSvc TenantService, cisSvc CisService) CleanerService {
+	return &service{
+		transact:  transact,
+		tenantSvc: tenantSvc,
+		cisSvc:    cisSvc,
+	}
+}
+
+// Clean missing godoc
+func (s *service) Clean(ctx context.Context) error {
```
DimitarPetrov, line 45: It would be nice to add some logs in order to know what happens while the job is running as success is unrecoverable and hard to validate afterward (therefore you cannot do it again to see what happened).

Information like:
- How many subaccounts were listed
- Which subaccount is being processed
  - does it need an update or not
- For each updated - previous and new values

Then the full log of the first execution can be saved somewhere for reference (and potential recovery) in case something went wrong and we suspect that the job messed it up.

Also, double-check all of the errors if they should be fatal or not. I mean should we fail the whole job execution if one of the subaccounts enrichment fails? We should log it of course but I think we should proceed with next subaccount. Then we can aggregate the IDs of all the subaccounts that processing failed for them and print it at the end and we can proceed with either manual fixing them or execute the job again and validation if they are successfully fixed now.

components/director/internal/dircleaner/service.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return errors.Wrap(err, "while getting all subaccounts")
+	}
+	log.C(ctx).Info("Done listing subaccounts")
+	for _, subaccount := range allSubaccounts {
+		globalAccountGUIDFromCis, err := s.cisSvc.GetGlobalAccount(ctx, subaccount.ID)
+		if err != nil {
+			return errors.Wrapf(err, "while getting global account guid for subaacount with ID %s", subaccount.ID)
+		}
+		err = func() error {
```
DimitarPetrov, line 57: This lambda seems unneeded.

components/director/internal/cis/service.go
```diff
@@ -0,0 +1,73 @@
...
+	"io"
+	"net/http"
+
+	"github.com/kyma-incubator/compass/components/director/pkg/log"
+	"github.com/pkg/errors"
+)
+
+// CISResponse response from the context enrichment api
+type CISResponse struct {
+	GlobalAccountID string `json:"global_account_id"`
```
nyordanoff, line 16: isn't the JSON value from the response with key `globalAccountId`

components/director/internal/dircleaner/service.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	defer s.transact.RollbackUnlessCommitted(ctx, tx)
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	log.C(ctx).Info("Listing all subaccounts provided by the event-service")
+	allSubaccounts, err := s.tenantSvc.ListSubaccounts(ctx)
+	if err != nil {
+		return errors.Wrap(err, "while getting all subaccounts")
+	}
+	log.C(ctx).Infof("Total number of listed subaccounts: %d", len(allSubaccounts))
```
nyordanoff, line 61: maybe we should count how many SAs were successfully processed/updated so in the end we have a summary (X out of Y SAs were processed successfully)

components/director/internal/dircleaner/service.go
```diff
@@ -0,0 +1,125 @@
...
+
+	for _, subaccount := range allSubaccounts {
+		log.C(ctx).Infof("Processing subaccount with ID %s", subaccount.ID)
+		globalAccountGUIDFromCis, err := s.cisSvc.GetGlobalAccount(ctx, subaccount.ExternalTenant)
+		if err != nil {
+			return errors.Wrapf(err, "while getting global account guid for subaacount with ID %s", subaccount.ID)
```
nyordanoff, line 67: Should we log here instead of ending the job

components/director/internal/dircleaner/service.go
```diff
@@ -0,0 +1,125 @@
...
+				return errors.Wrap(err, "failed to begin transaction")
+			}
+			ctx = persistence.SaveToContext(ctx, tx)
+			defer s.transact.RollbackUnlessCommitted(ctx, tx)
+
+			parentFromDB, err := s.tenantSvc.GetTenantByID(ctx, subaccount.Parent)
+			if err != nil {
+				log.C(ctx).Error(err)
+			}
+			if parentFromDB.ExternalTenant != globalAccountGUIDFromCis { // the record is directory and not GA
```
nyordanoff, line 81: if that's false (SA has valid parent) maybe we should log that the given SA is fine and doesn't need an update and increase the counter for successfully processed records

components/director/internal/dircleaner/service.go
```diff
@@ -0,0 +1,125 @@
...
+						Parent:         conflictingGA.ID,
+						Type:           tenant.TypeToStr(subaccount.Type),
+						Provider:       subaccount.Provider,
+					}
+					log.C(ctx).Infof("Updating subaccount with id %s to point to existing GA with id %s", subaccount.ID, conflictingGA.ID)
+					if err := s.tenantSvc.Update(ctx, subaccount.ID, updateSubaccount); err != nil {
+						log.C(ctx).Error(err)
+					}
+					// now delete the directory
+					if err = s.tenantSvc.DeleteByExternalTenant(ctx, parentFromDB.ExternalTenant); err != nil {
```
nyordanoff, line 97: Log the deletion of the Dir

nyordanoff (changes requested):

components/director/internal/dircleaner/service.go
```diff
@@ -0,0 +1,138 @@
...
+	return &service{
+		transact:  transact,
+		tenantSvc: tenantSvc,
+		cisSvc:    cisSvc,
+	}
+}
+
+// Clean missing godoc
+func (s *service) Clean(ctx context.Context) error {
+	tx, err := s.transact.Begin()
```
nyordanoff, line 49: The transaction is not committed.

components/director/internal/dircleaner/service.go
```diff
@@ -0,0 +1,138 @@
...
+	succsessfullyProcessed := 0
+	for _, subaccount := range allSubaccounts {
+		log.C(ctx).Infof("Processing subaccount with ID %s", subaccount.ID)
+		globalAccountGUIDFromCis, err := s.cisSvc.GetGlobalAccount(ctx, subaccount.ExternalTenant)
+		if err != nil {
+			log.C(ctx).Errorf("Could not get globalAccountGuid for subaacount with ID %s from CIS %v", subaccount.ID, err)
+			continue
+		}
+		err = func() error {
+			tx, err := s.transact.Begin()
```
nyordanoff, line 71: The transaction is not committed.

PR #2174, opened 2022-01-13
Enable Auditlog tests on real env

NickyMateev:
# Enable Auditlog tests on real env

**Description**

- Make Auditlog e2e tests configurable with custom TokenURL, client creds and read auditlog API for log assertion
- Refactor ext svc mock implementation of auditlog api to have a search by timestamp API instead of search by string

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/external-services-mock/internal/auditlog/configurationchange/config.go
```diff
@@ -7,10 +7,10 @@ import (
...
 
 func InitConfigurationChangeHandler(router *mux.Router, handler *ConfigChangeHandler) {
-	router.HandleFunc("/search", handler.SearchByString).Methods(http.MethodGet)
+	router.HandleFunc("/search", handler.SearchByTimestamp).Methods(http.MethodGet)
 	router.HandleFunc("", handler.Save).Methods(http.MethodPost)
 	router.HandleFunc("", handler.List).Methods(http.MethodGet)
 	router.HandleFunc("/{id}", handler.Get).Methods(http.MethodGet)
 	router.HandleFunc("/{id}", handler.Delete).Methods(http.MethodDelete)
-	router.Path("/search").Queries("query", "*").HandlerFunc(handler.SearchByString).Methods("GET")
+	router.Path("/search").Queries("timeFrom", "*", "timeTo", "*").HandlerFunc(handler.SearchByTimestamp).Methods("GET")
```
krasish, line 15: I wonder why should we have this combination of path and query params additionally specified? Won't those kind of requests get caught up by the handler on line 10. The HandlerFunc is looking for queryparams which are snake cased ("time_from" and "time_to") so those would not even be processed.

installation/resources/installer-config-gke-benchmark.yaml.tpl
```diff
@@ -9,7 +9,7 @@ metadata:
     kyma-project.io/installation: ""
 data:
   global.externalServicesMock.enabled: "true"
-  global.externalServicesMock.auditlog: "false"
+  global.externalServicesMock.auditlog.applyMockConfiguration: "false"
```
krasish, line 12: Shouldn't this be set to true?

tests/external-services-mock/tests/auditlog_integration_test.go
```diff
@@ -40,26 +40,25 @@ func TestAuditlogIntegration(t *testing.T) {
...
 	require.NoError(t, err)
+	timeTo := timeFrom.Add(1 * time.Minute)
 
 	t.Log("Get auditlog service Token")
-	auditlogToken := fixtures.GetAuditlogMockToken(t, &httpClient, testConfig.ExternalServicesMockBaseURL)
+	auditlogToken := fixtures.GetAuditlogToken(t, &httpClient, testConfig.Auditlog)
 
 	t.Log("Get auditlog from external services mock")
-	auditlogs := fixtures.SearchForAuditlogByString(t, &httpClient, testConfig.ExternalServicesMockBaseURL, auditlogToken, appName)
+	auditlogs := fixtures.SearchForAuditlogByTimestampAndString(t, &httpClient, testConfig.Auditlog, auditlogToken, appName, timeFrom, timeTo)
```
krasish, line 54: Change the log message as well. It will not always be external services mock

installation/resources/installer-config-gke-benchmark.yaml.tpl
NickyMateev: The auditlog e2e tests were disabled for benchmark jobs even before this PR, so this configuration is really obsolete and doesn't have an effect on the end result

components/external-services-mock/internal/auditlog/configurationchange/config.go
NickyMateev: I removed this as it was unnecessary

krasish (approved):

nkanchev (approved):

PR #2177, opened 2022-01-17
Rate limiting

krasish:
**Description**

This PR utilizes Istio's local rate-limit feature to implement rate-limiting in CMP. Additional hydrators were added so that we are able to make more specific rate-limiting. 

**Pull Request status**

- [X] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
+++ b/chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
@@ -157,6 +157,10 @@ spec:
 {{ toYaml .Values.global.oathkeeper.mutators.certificateResolverService | indent 4 }}
   - handler: hydrator
 {{ toYaml .Values.global.oathkeeper.mutators.tenantMappingService | indent 4 }}
+  - handler: header
+    config:
+      headers:
+        {{ .Values.global.gateway.headers.rateLimit }}: "{{ .Values.rateLimit.headerValue }}"
   - handler: id_token
     config:
       claims: {{ .Values.global.oathkeeper.idTokenConfig.claims | quote }}
--- a/chart/compass/charts/gateway/values.yaml
+++ b/chart/compass/charts/gateway/values.yaml
@@ -19,5 +19,11 @@ gateway:
     enabled: false
     authMode: "basic"
 
+rateLimit:
+  headerValue: "{{ print .Extra.consumerType }}"
+  maxTokens: 5000
+  tokensPerFill: 500
+  fillInterval: "60s"
+
 metrics:
   port: 3001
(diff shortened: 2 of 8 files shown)


chart/compass/charts/gateway/values.yaml
```diff
@@ -19,5 +19,11 @@ gateway:
     enabled: false
     authMode: "basic"
 
+rateLimit:
+  headerValue: "{{ print .Extra.consumerType }}"
```
DimitarPetrov, line 23: Shouldn't this also contain the consumerID?  Otherwise, if there is a [noisy neighbor](https://searchcloudcomputing.techtarget.com/definition/noisy-neighbor-cloud-computing-performance), the envoy will rate limit everyone instead of only the one generating load.

Is it possible for envoy to map a token_bucket based on two headers or a regex? This way for each request we can have consumerType and consumerId, then we will have bucket configuration per type and for each consumer of that type, a new bucket will be created. (The values, in this case, will be much smaller of course).

krasish: Unroturnatly that (token bucket per customer) is not possible using only Istio local rate limiting. The global rate-limiting supports that functionality and yes, that would solve the noisy neighbor problem but in order to use the global rate-limiting, we would need to deploy an additional service that handles all the token buckets for each customer. This is out of scope for this PR.

DimitarPetrov (approved):

DimitarPetrov (changes requested):
Discussed offline:

We need to remove the rate limit filter from director because in the current setup it also matches for hydrators and cotribute to the director rate limits while calling any component.

Only the graphQL call goes to the Runtime bucket all hydrator calls go to the common bucket again exhausting other clients.

As gateway is after ory we can move this config to the gateway and do not implement rate limits on director directly (this will effectively exclude hydrators as they call FQDN and don't go through gateway).

chart/compass/charts/gateway/templates/oathkeeper-rules.yaml
```diff
@@ -157,6 +157,10 @@ spec:
 {{ toYaml .Values.global.oathkeeper.mutators.certificateResolverService | indent 4 }}
   - handler: hydrator
 {{ toYaml .Values.global.oathkeeper.mutators.tenantMappingService | indent 4 }}
+  - handler: header
+    config:
+      headers:
+        {{ .Values.global.gateway.headers.rateLimit }}: "{{ .Values.rateLimit.headerValueTemplate }}"
```
DimitarPetrov, line 163: You need to add the header to `compass-connector-certs` ory rule as well to cover certificate renewal in the Runtime bucket instead of the common bucket.

DimitarPetrov (approved):

DimitarPetrov (approved):

PR #2180, opened 2022-01-18
Improve CORS handling for compass-console

StefanCholakov:
**Description**

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


DimitarPetrov (changes requested):
I think we can remove the whole corsPolicy object for the virtual services that we assume are not called by browsers (the ones that you removed the origin).

Also compass console does not work:
image

There is a compass-mf virtualservice in cockpit subchart which should be considered as well (potentially causing the first 3 errors about missing frames).

DimitarPetrov (approved):

PR #2191, opened 2022-01-27
Add support for x509 certificates when requesting an access token

SvetlinZarev-SAP:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/cmd/tenantfetcher-job/main.go
+++ b/components/director/cmd/tenantfetcher-job/main.go
@@ -6,6 +6,8 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/kyma-incubator/compass/components/director/pkg/oauth"
+
 	graphqlclient "github.com/kyma-incubator/compass/components/director/pkg/graphql_client"
 
 	httputil "github.com/kyma-incubator/compass/components/director/pkg/http"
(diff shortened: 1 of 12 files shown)


components/director/pkg/oauth/config.go
```diff
@@ -14,3 +23,51 @@ type Config struct {
 	TokenRequestTimeout   time.Duration `envconfig:"APP_OAUTH_TOKEN_REQUEST_TIMEOUT"`
 	SkipSSLValidation     bool          `envconfig:"APP_OAUTH_SKIP_SSL_VALIDATION"`
 }
+
+type X509Config struct {
+	Cert string `envconfig:"APP_OAUTH_X509_CERT,optional"`
+	Key  string `envconfig:"APP_OAUTH_X509_KEY,optional"`
+}
+
+func (c *X509Config) ParseCertificate() (*tls.Certificate, error) {
```
desislavaa, line 32: Almost the same method is used in the certificate loader - https://github.com/kyma-incubator/compass/blob/main/components/director/pkg/certloader/loader.go#L148
Maybe we should extract it somewhere where we can use it from both places.
We have a`cert` package which holds some utils, used by `connector` and `director`, maybe we can put it there?

SvetlinZarev-SAP: done

components/director/cmd/tenantfetcher-job/main.go
```diff
@@ -123,7 +126,10 @@ func createTenantFetcherSvc(cfg config, transact persistence.Transactioner, kube
 
 	runtimeService := runtime.NewService(runtimeRepository, labelRepository, labelDefService, labelService, uidSvc, scenarioAssignEngine, tenantStorageSvc, cfg.Features.ProtectedLabelPattern, cfg.Features.ImmutableLabelPattern)
 
-	eventAPIClient := tenantfetcher.NewClient(cfg.OAuthConfig, cfg.APIConfig, cfg.ClientTimeout)
+	eventAPIClient, err := tenantfetcher.NewClient(cfg.OAuthConfig, cfg.AuthMode, cfg.APIConfig, cfg.ClientTimeout)
+	if nil != err {
+		exitOnError(err, "failed to create event-api client")
+	}
```
DimitarPetrov, line 132: No need of the if can directly invoke exitOnError has the same check inside.

components/director/internal/tenantfetcher/client.go
```diff
@@ -65,20 +70,59 @@ type Client struct {
 }
 
 // NewClient missing godoc
-func NewClient(oAuth2Config OAuth2Config, apiConfig APIConfig, timeout time.Duration) *Client {
+func NewClient(oAuth2Config OAuth2Config, authMode oauth.AuthMode, apiConfig APIConfig, timeout time.Duration) (*Client, error) {
```
DimitarPetrov, line 73: Can we add tests for the constructor that asserts the httpClient's configuration is what is expected based on the AuthMode?

DimitarPetrov (changes requested):

components/director/cmd/tenantfetcher-job/main.go
SvetlinZarev-SAP: The There are several exit points inside `NewClient`. On the other hand the rest of the methods return an `error` and do `exitOnError()` in main. I'll propagate the error just to be consistent with the rest of the project.

components/director/internal/tenantfetcher/client.go
SvetlinZarev-SAP: done

DimitarPetrov (approved):

PR #2192, opened 2022-01-28
Prevent One-Time Token fetching from being audit logged

danielgospodinow:
**Description**

Currently, the Connector component calls the Director component through the Gateway component. And since the request to retrieve a one-time token is a GraphQL `mutation`, the operation is being audit logged. And since this is a frequent operation, we audit log more often than necessary. Preventing audit logging of one-time token fetching requests reduces a lot of load.

- Prevent OTT requests from being audit logged.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


chart/compass/values.yaml
```diff
@@ -141,6 +141,10 @@ global:
     graphql:
       external:
         port: 3000
+    tls:
```
danielgospodinow, line 144: This is influenced by the gateway chart. Unfortunately, I'm not sure whether this host is considered as TLS in any way. The only thing that it's doing is expecting a valid JWT token. Maybe we should drop the `tls` property since it's misleading.

danielgospodinow: Nevermind, it's okay. It was discussed offline.

components/connector/internal/api/certificate_resolver.go
```diff
@@ -65,8 +65,18 @@ func (r *certificateResolver) Configuration(ctx context.Context) (*externalschem
 		return nil, errors.Wrap(err, "Failed to authenticate request, consumer type not found")
 	}
 
+	serviceAccountFile := auth.DefaultServiceAccountTokenPath
+	if saFileString, ok := ctx.Value(authentication.ServiceAccountFile).(string); ok {
+		serviceAccountFile = saFileString
+	}
+	tokenProvider := auth.NewServiceAccountTokenAuthorizationProviderWithPath(serviceAccountFile)
+	authToken, err := tokenProvider.GetAuthorization(ctx)
```
krasish, line 73: An info log would be useful here.

components/connector/internal/tokens/service.go
```diff
@@ -42,22 +42,26 @@ func NewTokenService(cli GraphQLClient) *tokenService {
...
 
-func (svc *tokenService) getOneTimeToken(ctx context.Context, id, consumerType string) (string, error) {
+func (svc *tokenService) getOneTimeToken(ctx context.Context, id, authToken, consumerType string) (string, error) {
 	query, found := consumerTypeToQuery[consumerType]
 	if !found {
 		return "", errors.Errorf("No token generation for consumer type %s", consumerType)
 	}
 	req := gcli.NewRequest(fmt.Sprintf(query, "", id))
 
+	if authToken != "" {
```
krasish, line 61: Why is this now neccesary? It seems to me that even with the previous implementation this internal token was needed. Isn't this done by the serviceAccountTokenTransport used for the httpClient which is used in the director client?

danielgospodinow: Yes, you're totally right. Nice catch, strong eyes 👀 .

krasish (approved):

PR #2196, opened 2022-01-31
Enhance error messages

PetarTodorovv:
**Description**
Enhance error message with correct status code and meaningful message

- Add new graphql error that accept error code and message and use it for self register flow

- [x] Implementation
- [x] Unit tests

### DIFF
--- a/components/director/internal/error_presenter/presenter.go
+++ b/components/director/internal/error_presenter/presenter.go
@@ -38,7 +38,7 @@ func (p *presenter) Do(ctx context.Context, err error) *gqlerror.Error {
 	}
 
 	if apperrors.ErrorCode(customErr) == apperrors.InternalError {
-		log.C(ctx).WithField("errorID", errID).WithError(err).Errorf("Internal Server Error: %v", err)
+		log.C(ctx).WithField("errorID", errID).Errorf("Internal Server Error: %v", err)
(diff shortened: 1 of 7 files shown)


components/director/internal/error_presenter/presenter.go
```diff
@@ -33,12 +33,12 @@ func (p *presenter) Do(ctx context.Context, err error) *gqlerror.Error {
 	errID := p.uuidService.Generate()
 
 	if found := errors.As(err, &customErr); !found {
-		log.C(ctx).WithField("errorID", errID).WithError(err).Errorf("Unknown error: %v", err)
-		return newGraphqlErrorResponse(ctx, apperrors.InternalError, "Internal Server Error [errorID=%s]", errID)
+		log.C(ctx).WithField("errorID", errID).Errorf("Unknown error: %v", err)
+		return newGraphqlErrorResponse(ctx, apperrors.InternalError, "Internal Server Error [errorID=%s, reason: %s]", errID, err.Error())
```
PetarTodorovv, line 37: I'm not sure for the additional reason message here? I've added it to prevent future cases similar to "self register" flow where the error is from the external call or the error returned by the code is not "our custom" error(graphql error). Where we can have additional information for the failure.

dzahariev (approved):

dzahariev (approved):

PR #2200, opened 2022-02-02
Call global registry in ORD aggregator tests

dzahariev:
**Description**

- Currently ord-aggregator e2e tests are failing on real env, because the products and vendors, returned by the ORD service are more than expected (global products/vendors are also returned). This implementation adds logic for calling the global registry in order to determine the correct number of products/vendors.

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


tests/ord-aggregator/tests/handler_test.go
```diff
@@ -89,6 +87,12 @@ const (
 	documentationLabelSecondValue = "With multiple values"
 )
 
+var (
+	// The expected number is increased with initial number of global vendors/products before test execution
+	expectedNumberOfProducts = 7
+	expectedNumberOfVendors  = 7
+)
```
DimitarPetrov, line 94: There should be 6 as there is 1 global product and 1 global vendor in the local setup.

If the test fails with 6 I suspect that the getting of global vendors & products is the problem. Otherwise, it will fail again on real env.

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -223,6 +203,41 @@ func TestORDAggregator(t *testing.T) {
...
 
+		// Get vendors
+		respBody := makeRequestWithHeaders(t, httpClient, testConfig.ORDServiceURL+"/vendors?$format=json", map[string][]string{tenantHeader: {testConfig.DefaultTestTenant}})
+		numberOfGlobalVendors := len(gjson.Get(respBody, "value").Array())
+		expectedNumberOfVendors += numberOfGlobalVendors
+
+		// Get products
+		respBody = makeRequestWithHeaders(t, httpClient, testConfig.ORDServiceURL+"/products?$format=json", map[string][]string{tenantHeader: {testConfig.DefaultTestTenant}})
+		numberOfGlobalProducts := len(gjson.Get(respBody, "value").Array())
+		expectedNumberOfProducts += numberOfGlobalProducts
```
DimitarPetrov, line 214: Add some logs for easier troubleshooting of failing tests.

How many global products / vendors are found and what is the new expected number.

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -398,3 +411,20 @@ func storeMappingBetweenORDAndInternalBundleID(t *testing.T, respBody string, nu
...
+		return 0, 0
+	}
+
+	accessStrategyExecutorProvider := accessstrategy.NewDefaultExecutorProvider(certCache)
+	ordClient := NewClient(httpClient, accessStrategyExecutorProvider)
+
+	products, vendors, err := ordClient.GetGlobalProductsAndVendorsNumber(ctx, testConfig.GlobalRegistryURL)
+	if err != nil {
+		log.C(ctx).Errorf("while fetching global registry resources from %s %v", testConfig.GlobalRegistryURL, err)
+	}
```
DimitarPetrov, line 428: Pass testing.T in this function and fail the test on error with a meaningful message. Otherwise, it will be hard to figure out why the numbers differ in the asserts.

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -398,3 +411,20 @@ func storeMappingBetweenORDAndInternalBundleID(t *testing.T, respBody string, nu
 
 	return ordAndInternalIDsMapping
 }
+
+func getGlobalResourcesNumber(ctx context.Context, httpClient *http.Client) (int, int) {
+	certCache, err := certloader.StartCertLoader(ctx, testConfig.CertLoaderConfig)
+	if err != nil {
+		log.C(ctx).Errorf("Failed to initialize certificate loader %v", err)
+		return 0, 0
+	}
```
DimitarPetrov, line 420: Fail the test here as well, better early with error with context instead of later.

tests/ord-aggregator/tests/registry_client.go
```diff
@@ -0,0 +1,159 @@
...
+	for _, docDetails := range config.OpenResourceDiscoveryV1.Documents {
+		documentURL, err := buildDocumentURL(docDetails.URL, baseURL)
+		if err != nil {
+			return globalProducts, globalVendors, errors.Wrap(err, "error building document URL")
+		}
+		strategy, ok := docDetails.AccessStrategies.GetSupported()
+		if !ok {
+			log.C(ctx).Warnf("Unsupported access strategies for ORD Document %q", documentURL)
+			continue
+		}
```
DimitarPetrov, line 57: Better return error here as the tests need this. In the aggregator, we continue because in productive use-cases some of the documents that we can fetch are still more valuable than no documents. However, in the test scenario, you need all documents, otherwise, the test will fail later with a difference in expected numbers which will be harder to debug.

chart/compass/templates/tests/ord-aggregator/ord-aggregator-test.yaml
```diff
@@ -27,6 +27,7 @@ spec:
               - "oauth2.{{ .Values.global.ingress.domainName }}"
               - "compass-external-services-mock.compass-system.svc.cluster.local"
       {{ end }}
+      serviceAccountName: {{ $.Chart.Name }}-ord-aggregator
```
DimitarPetrov, line 30: The tests are in kyma-system namespace and this SA is in compass-system.

I am not sure, but it is possible that this will not work and you need to redefine the service account + role & rolebinding in the test chart in kyma-system as well.

DimitarPetrov (changes requested):

DimitarPetrov (approved):

PR #2205, opened 2022-02-07
Environment changes for tenant fetcher x509 support

la4ezar:
**Description**

Environment changes for tenant fetcher x509 support

**Related issue(s)**
https://github.com/kyma-incubator/compass/pull/2201

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


chart/compass/templates/tenant-fetcher-secret.yaml
```diff
@@ -11,5 +11,9 @@ data:
   {{ $config.secret.clientIdKey }}: {{ $config.oauth.client | b64enc | quote }}
   {{ $config.secret.clientSecretKey }}: {{ $config.oauth.secret | b64enc | quote }}
   {{ $config.secret.oauthUrlKey }}: {{ $config.oauth.tokenURL | b64enc | quote }}
+  {{ if eq $config.secret.oauthMode "oauth-mtls" }}
```
PetarTodorovv, line 14: do we need this if? oauthMode seems to be always `oauth-mtls` from values.yaml in local setup

la4ezar: We actually don't.

la4ezar: Update: We actually need it. If the OAuth mode is different than 'oauth-mtls' then the compass-installer will fail, because the clientCertKey and clientKeyKey do not exist.

PetarTodorovv: Do we have oauthMode different than `oauth-mtls` because in the values.yaml I see only this? And if we have different mode then I guess compass-installer fails because in the charts we have env for different modes and he try to populate all of them but in the secret we do not have the mention properties(clientCertKey, clientKeyKey)? If that's the case then we should add/remove them to/from the charts conditionally?
And also it would be semantically closer to the real environment where we have both mtls and non-mtls configuration in one secret

installation/resources/installer-config-gke-benchmark.yaml.tpl
```diff
@@ -29,3 +29,6 @@ data:
...
...096WUNSSVBkNnRUMTkrYkVtanByVnI3OVlkZUp5OExXTgpNSGU2Tlp3REtLWk1FTURaUHNINi9Jczl2TCtoa3pSd0NYRUhWa2dVczVYSWNvSERiWktVTXlZZGQ4cC9ESi8yCndybFNKRi9XOHV3YzVGcUY3SHdvZm1kYSt5ekhVSExDK2hwblFrR0g5dGtTV1VuTVdRMW9mTUVDZ1lFQXpYV1QKQy8yNkpYeWVHRWhBOWJOSTBESDE3NFpxNHpiZFgrSnVISUplaVE5alRvV3lqR3p5ckZvd2tXK3VZM3NqQU1MQQo1SE5TaW5nWnpEZ01jenFUbWJjMG9yQlVUMEF6dFBSZHg0cjRZZUVGL1h2dlJxYnNyeGdHTWliclhyd3BDSjVDCnFZVk82RWtqNXJiUWJKY0NQL1EyR1BFZ1pLR0JMdk0wSzd2UTlJMENnWUJ0aWpyc1orZWNlU0dEMlQ3RlFMbEIKckhZY2VFeVptUERXc0drQjRGUVJ1Zk5uVzdxK2xRNWhDdGR2N05zaklCNC9xeVBKaHVrK1FTRW9XeUR3VHQrYgp5K3gxSVBJY1lkbmp5WXVBaHlBR3JMT21yT0pxdkRTeDNEaGxCWUtzWlYxbEZlRm9ONEZRQkk5Yjdhb29nSnN3CldoNzVCckRaeUVUckpUV09Wck40QVFLQmdBeDZNMi9xL0w4Q0RtZlRHMzdRWUgra1NSYyt4b2I5OGZ1OHVJc3EKcjZzTE1EQzRsZHRKVW9OOUJxNE9aanpNWVpmT1BBQ2pzRU9RZjZDVFZzNDRwSFlWVmpEN0hHT2p0b0FxeHZjegpUVnBFWENURXZnZEFZK2RPUWpJUmd3SEIwNHdlY0ZYekxwT1V2WVZwWE1iN2RMdUZqVE4ra1VVTE9ka3NFK01FCkNQQ1JBb0dBQXEzUUptNUhzVzV1VEhSanlzRGM3UmdWZ2RhUTdRdTJHYVVkZEU0ZnJSMXB6SE1JRHhzclpaeWwKZ2lvcUNwUDd3VFV0Wmo0ZXQ3QmpGWU5nUUkvSEp6L3htNmJXYXRhSFBCR3grc2tQN3NLaG5ab3lsbzdDeThCLwp1cWRtT2FIaW9YSGdQc2lCcks0dlZKZnJ4aTlSbGhkWjJyeFpKNWJxYWdIcmdzdE1LUzg9Ci0tLS0tRU5EIFJTQSBQUklWQVRFIEtFWS0tLS0tCg=="
```
PetarTodorovv, line 33: rebase main and fix merge conflicts because this is already introduced in different PR. Valid for the other local override files as well

installation/resources/installer-config-gke-integration.yaml.tpl
```diff
@@ -79,10 +84,15 @@ data:
...
   global.tenantFetchers.subaccount-fetcher.secret.clientSecretKey: "client-secret"
+  global.tenantFetchers.subaccount-fetcher.secret.oauthMode: "oauth-mtls"
+  global.tenantFetchers.subaccount-fetcher.secret.clientCertKey: "client-cert"
+  global.tenantFetchers.subaccount-fetcher.secret.clientKeyKey: "client-key"
   global.tenantFetchers.subaccount-fetcher.secret.oauthUrlKey: "url"
+  global.tenantFetchers.subaccount-fetcher.secret.skipSSLValidation: "true"
   global.tenantFetchers.subaccount-fetcher.oauth.client: "client_id"
   global.tenantFetchers.subaccount-fetcher.oauth.secret: "client_secret"
-  global.tenantFetchers.subaccount-fetcher.oauth.tokenURL: "http://compass-external-services-mock.compass-system.svc.cluster.local:8080/secured/oauth/token"
+  global.tenantFetchers.subaccount-fetcher.oauth.tokenURL: '{{ printf "https://%s.%s" .Values.global.externalServicesMock.certSecuredHost .Values.global.ingress.domainName }}'
```
PetarTodorovv, line 94: would this be resolved to the correct value? Have you tested it on local installation?

la4ezar: It is tested.

PetarTodorovv (approved):

PR #2206, opened 2022-02-08
[Abandoned] external services mock servers fix

PetarTodorovv:
**Description**
The mock servers in external services mock were misconfigured and the token handler that issue token with cert was on unsecured servers(http://). And we cannot be confident that the new functionality is okay because the newly introduced tests will be executing on unsecured host and will always be passing. Also there were a changes in ORD cert server that were not ORD related.

- Introduce new "default cert server" that can be used for future testing with certificate
- Move changes from ORD cert secured server that were not ORD specific in the newly introduced "default cert server"

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


chart/compass/templates/system-fetcher-job.yaml
```diff
@@ -25,7 +25,7 @@ spec:
               hostnames:
                 # Used for calls to Director
                 - "{{ .Values.global.gateway.tls.secure.internal.host }}.{{ .Values.global.ingress.domainName }}"
-                - "{{ .Values.global.externalServicesMock.certSecuredHost }}.{{ .Values.global.ingress.domainName }}"
+                - "{{ .Values.global.externalServicesMock.ordCertSecuredHost }}.{{ .Values.global.ingress.domainName }}"
```
DimitarPetrov, line 28: This needs to be certSecuredHost as the system fetcher don't do anything ORD related.

components/external-services-mock/cmd/main.go
```diff
@@ -212,34 +210,50 @@ func initDefaultServer(cfg config, key *rsa.PrivateKey) *http.Server {
...
+
+	// Oauth server handlers
+	tokenHandlerWithKey := oauth.NewHandlerWithSigningKey(cfg.ClientSecret, cfg.ClientID, cfg.TenantHeader, key)
+	// TODO The mtls_token_provider sends client id and scopes in url.values form. When the change for fetching xsuaa token
+	// with certificate is merged GenerateWithCredentialsFromReqBody should be used for testing the flows that include fetching
+	// xsuaa token with certificate. APP_SELF_REGISTER_OAUTH_TOKEN_PATH for local env should be adapted.
+	router.HandleFunc("/oauth/token", tokenHandlerWithKey.GenerateWithCredentialsFromReqBody).Methods(http.MethodPost)
+
+	tokenHandler := oauth.NewHandler(cfg.ClientSecret, cfg.ClientID)
+	router.HandleFunc("/cert/token", tokenHandler.GenerateWithoutCredentials).Methods(http.MethodPost)
```
DimitarPetrov, line 224: Why we need both token handlers with cert?

DimitarPetrov (changes requested):

components/external-services-mock/cmd/main.go
PetarTodorovv: We may unify them, they were introduced from another changes/PRs and I transfer them in the "default cert server" but I can check what need to be change to remove one of them

PR #2207, opened 2022-02-09
Adopt APIs/Events filterting based on ORD visibilities

nyordanoff:
**Description**

This PR introduces filtering of APIs/Events based on their `visibility` field (`public`, `internal` or `private`).

If a call is made with a token containing the `internal_visibility:read` scope, all APIs/Events are returned (no matter whether they are `public`, `internal` or `private`). If the scope is not present, only the `public` APIs/Events are returned.

- grant the new scope to integration systems and when accessing Compass from compass console
- introduce new column `visibility` to `bundle_references` table so that we can filter APIs/Events with graphql paging
- adapt the APIs/Events' repo layer to take into account the scopes coming from the `id_token` and filter accordingly
- adapt aggregator e2e test

**Related issue(s)**

- https://github.com/kyma-incubator/ord-service/pull/59

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/schema-migrator/migrations/director/20220203161810_ord_internal_visibility.up.sql
```diff
@@ -0,0 +1,214 @@
+BEGIN;
+
+ALTER TABLE bundle_references
+ADD COLUMN visibility varchar;
```
DimitarPetrov, line 4: Filtering happens only on the visibility column of the bundle_references and not on the visibility column of the apis/events. This is fine since the visibility column of bundle_refs is always a copy of the visibility column of the referenced api/event.

However, we can potentially break this relation in the code while introducing new features (create bundle ref and forget it).

We can discuss in the team whether this is a good candidate for a trigger/cascade and make it "automatically" equal to the visibility of the referenced API/event or we will rely on tests to cover such errors.

Another option which probably I like the most is to use a subquery in the code for the visibility check.

Something like:

`WHERE (SELECT visibility FROM apis WHERE apis.id = br.api_def_id) = 'public'`

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -332,6 +374,9 @@ func TestORDAggregator(t *testing.T) {
 			}
 			t.Log("Successfully verified api spec")
 
+			// verify public apis
+			verifyEntitiesWithPublicVisibility(t, httpClientWithoutVisibilityScope, publicApisMap, apisField, expectedNumberOfPublicAPIs)
```
DimitarPetrov, line 378: What do you think about `verifyVisibilityFiltering` ?

DimitarPetrov (changes requested):

components/schema-migrator/migrations/director/20220203161810_ord_internal_visibility.up.sql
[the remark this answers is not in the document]
nyordanoff: Makes sense, will change it.

components/schema-migrator/migrations/director/20220203161810_ord_internal_visibility.up.sql
```diff
@@ -0,0 +1,214 @@
...
+ALTER COLUMN visibility SET NOT NULL;
+
+
+--- changes to tenants_apis and tenants_events ORD views
+
+DROP VIEW IF EXISTS tenants_specifications; -- drop tenant_specifications because tenant_apis and tenant_events depends on it. There are no changes on this view.
+DROP VIEW IF EXISTS tenants_apis;
+DROP VIEW IF EXISTS tenants_events;
+
+-- cast visibility field to text
```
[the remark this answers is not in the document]
nyordanoff: Due to the introduction of the new claim and the corresponding `@EdmProtectedBy` annotation in the ORD Service. 

When the `WHERE` clause based on the claims is constructed and the `visibility` field is added there, it expects a string (because the `visibility` field is of type string in the Java code) but the matching value from the view is the enum type and error is thrown due to mismatching types.

Part of the error: `No operator matches the given name and argument types. You might need to add explicit type casts.`

components/director/internal/domain/bundlereferences/repository.go
```diff
@@ -190,10 +196,18 @@ func (r *repository) ListByBundleIDs(ctx context.Context, objectType model.Bundl
 		return nil, nil, err
 	}
 
-	conditions := repo.Conditions{
-		repo.NewNotNullCondition(objectFieldName),
+	isInternalVisibilityScopePresent, err := scope.IsGivenScopePresent(ctx, internalVisibilityScope)
```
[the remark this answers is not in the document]
nyordanoff: That's definitely better, thanks.

components/director/internal/domain/bundlereferences/repository.go
```diff
@@ -190,10 +196,18 @@ func (r *repository) ListByBundleIDs(ctx context.Context, objectType model.Bundl
...
+	if err != nil {
+		return nil, nil, err
```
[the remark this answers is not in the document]
nyordanoff: Nice catch, thanks.

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -459,13 +482,64 @@ func storeMappingBetweenORDAndInternalBundleID(t *testing.T, respBody string, nu
...
 
-func verifyEntitiesWithPublicVisibility(t *testing.T, httpClient *http.Client, publicEntitiesMap map[string]string, entity string, expectedNumberOfPublicEntities int) {
+func verifyEntitiesWithPublicVisibilityInORD(t *testing.T, httpClient *http.Client, publicEntitiesMap map[string]string, entity string, expectedNumberOfPublicEntities int) {
 	respBody := makeRequestWithHeaders(t, httpClient, testConfig.ORDServiceURL+fmt.Sprintf("/%s?$format=json", entity), map[string][]string{tenantHeader: {testConfig.DefaultTestTenant}})
 
 	assertions.AssertMultipleEntitiesFromORDService(t, respBody, publicEntitiesMap, expectedNumberOfPublicEntities, descriptionField)
 	t.Logf("Successfully verified public %s", entity)
 }
 
+func verifyEntitiesVisibilityViaGraphql(t *testing.T, clientWithInternalScope, clientWithoutInternalScope *gcli.Client, entitiesMap, publicEntitiesMap map[string]string, expectedNumberOfEntities, expectedNumberOfPublicEntities int, appID string, isAPIDef bool) {
```
DimitarPetrov, line 492: Can we invoke the method just once after events are verified and do both asserts?

This way we can get rid of the bool parameter and the if as well as make 2 fewer calls to director which will speed things up.

tests/ord-aggregator/tests/handler_test.go
DimitarPetrov: It makes sense 👍

components/schema-migrator/migrations/director/20220203161810_ord_internal_visibility.up.sql
nyordanoff: After a brief offline discussion, the approach with the subquery was chosen.

components/director/internal/domain/bundlereferences/repository.go
```diff
@@ -196,14 +201,15 @@ func (r *repository) ListByBundleIDs(ctx context.Context, objectType model.Bundl
 		return nil, nil, err
 	}
 
-	isInternalVisibilityScopePresent, err := scope.IsGivenScopePresent(ctx, internalVisibilityScope)
+	isInternalVisibilityScopePresent, err := scope.Contains(ctx, internalVisibilityScope)
 	if err != nil {
-		return nil, nil, err
+		isInternalVisibilityScopePresent = true
```
DimitarPetrov, line 206: An Info log here and in the if on 212 line would be helpful.

components/director/internal/domain/bundlereferences/repository.go
```diff
@@ -196,14 +201,15 @@ func (r *repository) ListByBundleIDs(ctx context.Context, objectType model.Bundl
...
 	}
 
+	visibilityFilteringSubquery := fmt.Sprintf("(SELECT %s FROM %s WHERE %s.id = %s.%s)", visibilityColumn, objectTable, objectTable, BundleReferenceTable, objectIDCol)
```
DimitarPetrov, line 209: Why don't use the query builder? I think in most cases we use the query builer to generate SQL for subqueries.
This way the SQL crafting will be isolated there.

components/director/internal/domain/bundlereferences/repository_test.go
```diff
@@ -537,24 +530,6 @@ func TestPgRepository_ListAllForBundle(t *testing.T) {
 		sqlMock.AssertExpectations(t)
 	})
 
-	t.Run("returns error when check for internal scope fails", func(t *testing.T) {
```
DimitarPetrov: Maybe you can retrun the tests but with the opposite asserts. Something along the lines of:

```
returns both public and internal/private APIs/events when check for internal scope fails
```

components/director/internal/domain/eventdef/converter.go
```diff
@@ -188,7 +195,7 @@ func (c *converter) ToEntity(eventModel *model.EventDefinition) *Entity {
 		Successors:          repo.NewNullableStringFromJSONRawMessage(eventModel.Successors),
 		ChangeLogEntries:    repo.NewNullableStringFromJSONRawMessage(eventModel.ChangeLogEntries),
 		Labels:              repo.NewNullableStringFromJSONRawMessage(eventModel.Labels),
-		Visibility:          repo.NewNullableString(eventModel.Visibility),
+		Visibility:          repo.NewNullableString(visibility),
```
DimitarPetrov, line 198: You can make the visibility column string instead of NullableString and add not null constraint in the database in order to ensure that the if above will never be missed.

Valid for the API Entity as well.

DimitarPetrov (changes requested):

components/director/internal/domain/bundlereferences/repository.go
nyordanoff: Because the subquery itself is pretty straightforward - it does not contain any arguments that may need rebinding or some additional subquery (e.g. tenant isolation). That's why, I decided the effort of initialising, constructing and building the subquery via builder is not worth it.

DimitarPetrov (approved):

components/director/internal/domain/bundlereferences/converter.go
```diff
@@ -36,6 +36,7 @@ func (c *converter) ToEntity(in model.BundleReference) Entity {
 		APIDefID:            apiDefID,
 		EventDefID:          eventDefID,
 		APIDefaultTargetURL: apiDefaultTargetURL,
+		Visibility:          in.Visibility,
```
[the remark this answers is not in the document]
KirilKabakchiev: Ok, from domain model point of view, it seems much cleaner and more correct to not have visibility in BRs as visibilities have nothing to do with referencing bundles, so i like the new approach 👍

PR #2212, opened 2022-02-16
Build docker images for multiple architectures

NickyMateev:


scripts/generic_make_go.mk
```diff
@@ -95,14 +95,12 @@ endef
 verify:: test check-imports check-fmt errcheck lint
 format:: imports fmt
 
-release: verify build-image push-image
+release: verify build-image
 
-.PHONY: build-image push-image
+.PHONY: build-image
 build-image: pull-licenses
```
KirilKabakchiev, line 101: why did you include the push part in the build step ? I think having separate steps is better

DimitarPetrov: While building multi-arch images we should push it right away (build it in the registry as destination) or load it in the local docker driver (local docker installation).

However, we can load only the image (or the part of it) that is for the current architecture docker is running in. Therefore if we build it, then we need to load it and push it we will push only the architecture that we loaded which is the arch of the machine that built it.

If we push it right away this will push both images for both architectures directly to the repo and each pull will be based on the arch of the puller.

Atleast that's my understanding - reference: https://github.com/docker/buildx/issues/59

KirilKabakchiev (approved):

PR #2216, opened 2022-02-21
Adjust ORD aggregator e2e tests

StefanCholakov:
- Currently there is a problem when trying to execute ORD aggregator's tests on real env, because the test is trying to read a secret from another namespace(kyma-system). The proposed change initially copy this secret in the compass-system namespace, so the test can access it

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


chart/compass/templates/tests/ord-aggregator/ord-aggregator-test.yaml
```diff
@@ -34,7 +34,7 @@ spec:
           image: {{ .Values.global.images.containerRegistry.path }}/{{ .Values.global.images.e2e_tests.dir }}compass-tests:{{ .Values.global.images.e2e_tests.version }}
           imagePullPolicy: IfNotPresent
           command: ["/bin/sh"]
-          args: ["-c", "./ord-aggregator.test -test.v; exit_code=$?; curl -XPOST http://127.0.0.1:15020/quitquitquit; sleep 4; exit $exit_code;"]
+          args: ["-c", "kubectl get secret external-client-certificate -n compass-system -o yaml | grep -v ‘^\s*namespace:\s’ | kubectl apply -n kyma-system -f - && ./ord-aggregator.test -test.v; exit_code=$?; curl -XPOST http://127.0.0.1:15020/quitquitquit; sleep 4; exit $exit_code;"]
```
DimitarPetrov, line 37: This seems like an unreliable approach since the contents of this secret is changed all the time by the job.

You can see here: https://github.com/kyma-incubator/compass/blob/main/tests/director/tests/main_test.go#L43

The same secret is read by the tests in director tests (also running in kyma-system), if the same approach is possible this will be much cleaner solution.

Note: When Dani remove the dexGraphQLClient I believe that we can move the tests in compass-system and get rid of all those copying workarounds.

PR #2223, opened 2022-02-25
Add SupportInfo on ORD Packages

PetarTodorovv:
**Description**
Add supportInfo property to ORD packages

- Add supportInfo property to ORD packages
- adjust tests

**Related issue(s)**
https://github.com/kyma-incubator/ord-service/pull/62

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/schema-migrator/migrations/director/20220224153857_add-support-info-to-packages.up.sql
```diff
@@ -0,0 +1,61 @@
+BEGIN;
+
+ALTER TABLE packages
+    ADD COLUMN support_info VARCHAR(256);
```
nyordanoff, line 4: By the spec this is a free text field which can have an arbitrary length, so I think we can remove the length restriction and just leave it as `varchar`

components/schema-migrator/migrations/director/20220224153857_add-support-info-to-packages.up.sql
```diff
@@ -1,7 +1,7 @@
 BEGIN;
 
 ALTER TABLE packages
-    ADD COLUMN support_info VARCHAR(256);
+    ADD COLUMN support_info VARCHAR;
```
DimitarPetrov, line 4: Text is probably a better option for arbitrarily long strings.

nyordanoff (approved):

PR #2230, opened 2022-03-02
Upgrade Kyma to version 2.0

StefanCholakov:
**Description**

This PR bumps and adapts the kyma from 1.24.8 to 2.0

- adapt usage of k3d instead of minikube on local installation
- delete obsolete 1.24.8 related installer resources
- adapt overrides from `.dot` style to helm native yaml tree format
- install compass with helm
- various adaptations across charts & installation scripts

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


installation/resources/compass-overrides-local.yaml
```diff
@@ -0,0 +1,205 @@
+config:
+  oidc:
+    caFile: "/etc/dex-tls-cert/tls.crt"
```
DimitarPetrov, line 3: Probably a leftover dex reference?

installation/scripts/install-compass.sh
```diff
@@ -0,0 +1,49 @@
...
+            shift # past argument
+        ;;
+    esac
+done
+set -- "${POSITIONAL[@]}" # restore positional parameters
+
+CRDS_FOLDER="${CURRENT_DIR}/../resources/crds"
+kubectl apply -f "${CRDS_FOLDER}"
+helm upgrade --install --wait --timeout "${TIMEOUT}" -f ./mergedOverrides.yaml --create-namespace --namespace compass-system compass "${COMPASS_CHARTS}"
+rm mergedOverrides.yaml
```
DimitarPetrov, line 49: If something fail the file will be left behind.

Add it to gitignore for such cases.

Also you can use trap to minimize such cases: https://github.com/kyma-incubator/compass/blob/e495587b9928d436701650783fb9e453d03b424f/installation/cmd/run.sh#L184

NickyMateev (approved):
Looks good to me - at least I couldn't find anything disturbing 👍 

One question - how do things look for Mac M1 users - does the local k3d cluster compass installation work?

Also I noticed that `--k3d-cpus` was removed - I don't remember the original reason for adding it - maybe because there was an equivalent minikube-like option or that we were trying to stabilize things on M1 machines, by allowing the tweaking of the cpu. Anyway, the parameter is now removed - what were the reasons and thoughts behind that?

nyordanoff:
> Looks good to me - at least I couldn't find anything disturbing 👍
> 
> One question - how do things look for Mac M1 users - does the local k3d cluster compass installation work?
> 
> Also I noticed that `--k3d-cpus` was removed - I don't remember the original reason for adding it - maybe because there was an equivalent minikube-like option or that we were trying to stabilize things on M1 machines, by allowing the tweaking of the cpu. Anyway, the parameter is now removed - what were the reasons and thoughts behind that?

М1 testing and adaptations were left out of scope for the current pass. We will adapt things ad-hoc if something comes up. 

Regarding the `--k3d-cpus` flag: currently, while provisioning minikube one can explicitly specify the number of cpus via flag. However, k3d does not support such flag (it supports only for memory) so the `cpu` flag to the script became obsolete. That's why it was removed.

I searched through k3d docs & code to check if somehow we can pass number of cpus during cluster creation but it seems that it is not supported. If I find a way to do it, I will add back the flag.

nkanchev (dismissed):

NickyMateev (dismissed):

desislavaa (approved):

StanislavStefanov (approved):

desislavaa (approved):

StanislavStefanov (approved):

NickyMateev (approved):

nkanchev (approved):

PR #2231, opened 2022-03-02
Address ORD validation feedback

nyordanoff:
**Description**
In this PR some validation rules are slightly tweaked/refactored.

Note: in  the `.up` migration the only actual change is increasing the size of the `short_description` column. The rest is re-creation of the dropped depending views. In the `.down` migration, however, some columns had to be excluded from some of the views as they were not maintained properly in earlier migrations - detailed explanation can be found as comment on each affected view in the migration.

- limit the length of the `description` field across all entities to 5000 chars
- limit the length of the `short description` field across all entities to 256 chars
- enhance the `date` validation to support both `rfc3339` protocol and some extensions of `iso8601`
- validation unit tests

**Related issue(s)**

- https://github.com/kyma-incubator/ord-service/pull/63

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/open_resource_discovery/validation.go
```diff
@@ -1107,14 +1118,22 @@ func areThereEntryPointDuplicates(entryPoints []gjson.Result) bool {
...
-			return nil
-		} else if _, err = time.Parse("2006-01-02T15:04:05Z07:00", *date); err == nil {
-			return nil
-		}
-		return errors.New("invalid date")
+		return validateDate(date)
+	}
+}
+
+func validateDate(date *string) error {
```
DimitarPetrov, line 1125: You can make the date argument `interface{}` and then cast it to string or *string.

If it is not string or *string you will return the relevant error.

This way you will not need to use crossFieldRule in validateORDChangeLogEntries to workaround type issues. 

Also you will not need to invoke the method manually with the value in the inlined rules like:

```go
validation.By(isValidDate(api.SunsetDate))
```

Will be simplified to:
```go
validation.By(isValidDate)
```

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -1152,9 +1171,10 @@ func validateExtensibleInnerFields(el gjson.Result) error {
 
 	descriptionProperty := el.Get("description")
 	descriptionValue, ok := descriptionProperty.Value().(string)
+	validLength := len(descriptionValue) >= MinDescriptionLength && len(descriptionValue) <= MaxDescriptionLength
 
-	if supportedProperty.Exists() && (supportedValue == "manual" || supportedValue == "automatic") && (descriptionValue == "" || !ok) {
-		return errors.New("if supported is either 'manual' or 'automatic', description should be provided")
+	if supportedProperty.Exists() && (supportedValue == "manual" || supportedValue == "automatic") && (descriptionValue == "" || !ok || !validLength) {
```
DimitarPetrov, line 1176: `descriptionValue == ""` is redundant as it will has a len of 0 which is not validLength.

Same for `!ok` as if the ok is false the descriptionValue would be empty string which will also has a len of 0.

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -975,6 +983,9 @@ func validateCustomDescription(el gjson.Result) error {
 	if el.Get("customDescription").Exists() && el.Get("type").String() != custom {
 		return errors.New("if customDescription is provided, type should be set to 'custom'")
 	}
+	if el.Get("customDescription").Exists() && el.Get("type").String() == custom && len(el.Get("customDescription").String()) > MaxDescriptionLength {
```
DimitarPetrov, line 986: Check minimum length as well as the customDescription can exsist with 0 length. Something like:

```json
{
    "customDescription": ""
}
```

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -115,6 +116,13 @@ var (
           "title": "myTitle"
         }
       ]`
+	invalidLinkDueToInvalidLengthOfDescription = `[
+        {
+          "title": "myTitle",
+		  "url": "https://example2.com/en/legal/terms-of-use.html",
```
DimitarPetrov, line 122: Broken formatting

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -202,6 +210,13 @@ var (
 		  "customType": "ns:credential-exchange:v1",
 		  "customDescription": "foo bar"
         }
+      ]`
+	invalidCredentialsExchangeStrategyDueToExceedingLenOfCustomDescription = `[
+        {
+          "type": "custom",
+		  "customType": "ns:credential-exchange:v1",
+		  "customDescription": "%s"
```
DimitarPetrov, line 218: Broken formatting

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -310,6 +325,15 @@ var (
           "version": "1.0.0"
         }
       ]`
+	invalidChangeLogEntriesDueToInvalidLengthOfDescription = `[
+        {
+		  "date": "2020-04-29",
+		  "description": "%s",
+		  "releaseStatus": "active",
+		  "url": "https://example.com/changelog/v1",
```
DimitarPetrov, line 333: Broken formatting

components/schema-migrator/migrations/director/20220301143225_ord_increase_short_description_length.down.sql
```diff
@@ -0,0 +1,526 @@
...
+ALTER COLUMN short_description TYPE VARCHAR(255);
+
+ALTER TABLE event_api_definitions
+ALTER COLUMN short_description TYPE VARCHAR(255);
+
+
+-- while re-creating this view, the 'support_info' column had to be explicitly omitted (thus selecting the rest columns one by one). This is needed because when it was introduced in
+-- "add-support-info-to-packages.up" migration, this view was not adapted to include it. In the current "up" migration, the 'support_info' column is included
+-- because all package columns are selected there (p.*). If we do not exclude the "support_info" column here, the "add-support-info-to-packages.down"
+-- migration will fail because when trying to drop the column this view will already rely on it.
```
DimitarPetrov, line 36: Does that mean that if we have a `p.*` in the view and we add a new column to the underlying table then the view is not automatically updated to have this column and need to be plain recreated in order to include the new column?

If that is the case and the reason for this headaches then we need to think of a way of simplifying contribution of migrations as this means that we need to recreate each and every view when adding new columns.

DimitarPetrov (changes requested):

components/schema-migrator/migrations/director/20220301143225_ord_increase_short_description_length.down.sql
nyordanoff: To your question: yes, that is the case. In case a new column is added, these ("storage-redesign") views are not automatically updated to include the newly added column. And if in the future, that view is recreated in other migration , then it will include the new column (as we will select all table columns) which will break the initial down migration as it does not expect that view to depend on the new column.

So, for the time being, whenever we introduce a new column, we should make sure that all views that rely on that table are adapted to include it as well. However, as you suggested, in the long run it might be a better idea to come up with some unified solution that will solve this "error/forgetting prone" approach.

DimitarPetrov (approved):

PR #2235, opened 2022-03-07
clean dirty flag before retrying migrations

StanislavStefanov:
**Description**
When migration fails the migrator tool sets the "dirty" property in "schema_migrations" table to "true". When the migrator job retries the migrations they are not actually applied again, as if the "dirty" property is set to "true" the migration tool asks for the flag to be cleaned before it tries to apply the migration.

- clean dirty flag before retrying migrations

- [X] Implementation


components/schema-migrator/run.sh
```diff
@@ -151,6 +151,20 @@ if [[ "${DIRECTION}" == "up" ]]; then
...
 
+  TABLE_EXISTS=$(PGPASSWORD="${DB_PASSWORD}" psql -qtAX -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}" -c "SELECT EXISTS (SELECT FROM pg_tables WHERE schemaname = 'public' AND tablename  = 'schema_migrations');")
+  echo "Table \"schema_migrations\" exists: $TABLE_EXISTS"
+  if [[ "$TABLE_EXISTS" == t ]]; then
+    DATABASE_STATE=$(PGPASSWORD="${DB_PASSWORD}" psql -qtAX -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}" -c "SELECT dirty FROM schema_migrations")
+    echo "DB state is dirty: $DATABASE_STATE"
+    if [[ "$DATABASE_STATE" == t ]]; then
+      CURRENT_VERSION=$(PGPASSWORD="${DB_PASSWORD}" psql -qtAX -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}" -c "SELECT version FROM schema_migrations")
+      echo "Current schema version: $CURRENT_VERSION"
+      LAST_SUCCESSFUL_MIGRATION=$(ls -lr migrations/${MIGRATION_PATH} | grep -i -A 1 ${CURRENT_VERSION} | tail -n 1 | tr -s ' ' | cut -d ' ' -f9 | cut -d '_' -f1)
```
desislavaa, line 162: what's the benefit of using this approach instead of
```
LAST_SUCCESSFUL_MIGRATION=$(currentVersion)
```

StanislavStefanov: Lets say we have migration 11 - a good one and migration 12 - a faulty one. The first time the migrator job tries to apply the migrations it will fail on the second one and in the DB we will have record (version = 12, dirty = true). The current version in that case is the faulty one - 12. We need to forse the schema to the last successfull version - the version before the faulty one - 11. If we set LAST_SUCCESSFUL_MIGRATION=$(currentVersion) this will mean that the last applyed migration is good which is not the case.

desislavaa (approved):
Validated with two sequential migrations that tried to create the same DB. The following logs are from the retried job:
```
Previous migrations exist, verifying that the state is not dirty.                                                                                                                                                                                                            │
│ DB state is dirty: t                                                                                                                                                                                                                                                         │
│ Current schema version: 20220310090002                                                                                                                                                                                                                                       │
│ Forcing DB schema version to last successful migration version - 20220310090001
```

desislavaa (approved):

PR #2236, opened 2022-03-07
Add coverage to unit tests

desislavaa:
**Description**
The new delivery program expects coverage reports for Compass components.

- Lightweight code coverage implementation. It was decided that we'll stick to this approach until more sophisticated code coverage reporting is required.

image


scripts/generic_make_go.mk
```diff
@@ -175,9 +175,16 @@ endif
...
 $(foreach t,$(COPY_TARGETS),$(eval $(call buildpack-cp-ro,$(t))))
 
-test-local:
+test-local-no-coverage:
 	go test ./...
 
+test-local:
+	@go test ./... -coverprofile cover.out.tmp
+	@cat cover.out.tmp | grep -v "_gen.go" | grep -v "hack" > cover.out
+	@echo -n Code Coverage:
```
la4ezar, line 184: The output will look better if it is 
@echo -n 'Code Coverage: '

scripts/generic_make_go.mk
```diff
@@ -175,9 +175,16 @@ endif
...
-test-local:
+test-local-no-coverage:
 	go test ./...
 
+test-local:
+	@go test ./... -coverprofile cover.out.tmp
+	@cat cover.out.tmp | grep -v "_gen.go" | grep -v "hack" > cover.out
+	@echo -n Code Coverage: 
+	@go tool cover -func cover.out | grep total | tr -s "[:blank:]" | tr "[:blank:]" " " | cut -d " " -f 3
+	@rm cover.out.tmp cover.out
```
la4ezar, line 186: If 'go test ...' fails this will not be executed and the 'cover.out.tmp' file will not be deleted.

la4ezar (changes requested):

scripts/generic_make_go.mk
la4ezar: You can change the 'go test...' with `@go test ./... -coverprofile cover.out.tmp || true`

DimitarPetrov: This will prevent prow jobs from failing when tests are failing.

desislavaa: for prow jobs - it doesn't matter, for local dev - I've added it into gitignore - is that ok?

la4ezar (approved):

PR #2253, opened 2022-03-18
Automatic Scenario Assignments documentation updates

desislavaa:
**Description**
The labeling mechanism for Automatic Scenario Assignment is deprecated. The ASA works with subaccount tenants now - the ASA is configured to match all runtimes in a given subaccount, and when a new runtime is created in that subaccount, it is automatically assigned to the scenario.


docs/compass/03-03-automatic-scenario-assignment.md
```diff
@@ -19,26 +19,31 @@ type AutomaticScenarioAssignment {
...
    value: Any!
 }
 
 ```
 
-A condition is defined as a label selector in the **selector** field. If a Runtime is labeled with a label that matches the value of the **selector** parameter, the Runtime is assigned to the given Scenario.
+A condition is defined as a label selector in the **selector** field. 
+The label key is validated to be `global_subaccount_id`. It is also validated, that the subaccount tenant is a child of the tenant where the scenario and ASA resides.
+
+Then, if a Runtime is labeled with `global_subaccount_id`, the Runtime is assigned to the given Scenario, even though they are in different tenants.
```
DimitarPetrov, line 31: There is no more label processing. The scenario assignment is based on the tenant that the runtime is registered in.

We can add a few sentences that previously it was label selector and now it is subaccount_id selector and the interface is kept for backwards compatibility, but the only key possible is global_subaccount_id. Then the tenant of the runtime is checked and not labels anymore.

DimitarPetrov (changes requested):

nkanchev (approved):

DimitarPetrov (approved):

nkanchev (approved):

PR #2259, opened 2022-03-18
Audit log operations in tenant fetcher deployment

StefanCholakov:
**Description**

- Use director internal GraphQL api for storing tenants in the database
- Introduce new mutations in Director responsible for tenant subscription/unsubscription
- Remove non-regional endpoint from tenant fetcher as it is no longer used

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/subscription/service_test.go
```diff
@@ -0,0 +1,641 @@
...
+			} else {
+				assert.NoError(t, err)
+			}
+
+			assert.Equal(t, testCase.IsSuccessful, isUnsubscribeSuccessful)
+		})
+	}
+}
+
+func fixUUID() string {
```
krasish, line 639: Not an issue but why don't we simply define a constant?

krasish (approved):

PR #2263, opened 2022-03-21
Allow self-registration from different regions

StanislavStefanov:
**Description**

- read instance configuration for all regions
- take region into consideration on self-registration and deregistration

**Pull Request status**
- [X] Implementation
- [x] Unit tests
- [X] Integration tests


DimitarPetrov (changes requested):

components/director/internal/domain/runtime/selfreg_config.go
```diff
@@ -0,0 +1,64 @@
...
+	Key          string
+}
+
+// MapInstanceConfigs parses the InstanceConfigs json string to map with key: regin_name and value: InstanceConfig for the instance in the region
+func (c *SelfRegConfig) MapInstanceConfigs() {
+	bindingsResult := gjson.Parse(c.InstanceConfigs)
+	bindingsMap := bindingsResult.Map()
+	c.RegionToInstanceConfig = make(map[string]InstanceConfig)
+	for region, config := range bindingsMap {
+		i := InstanceConfig{
```
PetarTodorovv, line 54: Additional to Dimitar's comment about usage of gjson.Valid, can we ensure somehow(with validate function or something else) that the end result of `c.RegionToInstanceConfig` is what we expected? E.g. `bindigsMap` won't be empty + `gjson.Get` for the properties return non empty value? 

Example: `gjson.Get(config.String(), c.InstanceClientSecretPath).String()` tries from config's json to get client_secret from some path and that returns Result and e.g we can check that client_secret is actually exist with Result's `.Exists` method. 

That may help us to mitigate issues related to wrong path and missing property which from other hand most probably will result in error when getting token

components/director/internal/domain/runtime/selfreg_manager.go
```diff
@@ -134,9 +143,9 @@ func (s *selfRegisterManager) GetSelfRegDistinguishingLabelKey() string {
 	return s.cfg.SelfRegisterDistinguishLabelKey
 }
 
-func (s *selfRegisterManager) createSelfRegPrepRequest(runtimeID, tenant string) (*http.Request, error) {
+func (s *selfRegisterManager) createSelfRegPrepRequest(runtimeID, tenant string, instanceConfig InstanceConfig) (*http.Request, error) {
```
PetarTodorovv, line 146: we can pass only URL string instead of whole config?
P.S. Valid for `createSelfRegDelRequestv` as well

components/director/internal/domain/runtime/resolver.go
```diff
@@ -306,13 +300,32 @@ func (r *Resolver) DeleteRuntime(ctx context.Context, id string) (*graphql.Runti
...
+		if err = tx.Commit(); err != nil {
+			return nil, err
+		}
+
+		log.C(ctx).Infof("Executing clean-up for self-registered runtime with id %q", runtime.ID)
+		if err := r.selfRegManager.CleanupSelfRegisteredRuntime(ctx, runtime.ID, regionLabel.Value.(string)); err != nil {
+			return nil, errors.Wrap(err, "An error occurred during cleanup of self-registered runtime: ")
+		}
+
+		tx, err = r.transact.Begin()
```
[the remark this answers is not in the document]
la4ezar: We are using the same variable(`tx`) when creating the second transaction. If the second transaction fails the defer will be executed with the `tx` which will be the second transaction.
Also, we have a unit test for this scenario: https://github.com/kyma-incubator/compass/blob/f718f767d19bd19f6a3f00fb2afe080e4f5597e9/components/director/internal/domain/runtime/resolver_test.go#L552

chart/compass/values.yaml
```diff
@@ -191,13 +191,23 @@ global:
...
     selfRegister:
       secret:
-        name: "compass-self-registration-credentials"
-        clientIdKey: client-id
-        clientSecretKey: client-secret
-        urlKey: url
-        tokenUrlKey: token-url
-        clientCertKey: client-cert
-        clientKeyKey: client-key
+        name: "compass-instance-to-region"
```
PetarTodorovv, line 194: region-instances-credentials/creds?

chart/compass/charts/external-services-mock/templates/oauth-credentials-secret.yaml
```diff
@@ -50,12 +50,11 @@ metadata:
   namespace: compass-system
 type: Opaque
```
PetarTodorovv, line 51: can we move this secret to be not between ext-svc-mock-oauth-secrets(kyma/compass ns) and if needed to add manage property dedicated to this secret because now it uses the one from ext-svc-mock?

components/director/internal/domain/runtime/resolver.go
```diff
@@ -281,16 +279,12 @@ func (r *Resolver) UpdateRuntime(ctx context.Context, id string, in graphql.Runt
...
 	tx, err := r.transact.Begin()
 	if err != nil {
 		return nil, err
 	}
 	defer func() {
-		didRollback := r.transact.RollbackUnlessCommitted(ctx, tx)
-		if !didRollback { // if we did rollback we should not try to execute the cleanup
-			r.cleanupAndLogOnError(ctx, selfRegCleanup)
-		}
+		r.transact.RollbackUnlessCommitted(ctx, tx)
```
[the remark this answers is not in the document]
la4ezar: If we directly defer the deferred call's arguments are evaluated immediately. In this case, we will have to add another defer for the second transaction that we are creating. If we leave it as a lambda function that won't be needed. The closure captures the variable `tx` and when the code in the closure (later) executes, the variable `tx` has the value of the last opened transaction.

components/director/internal/domain/runtime/selfreg_config.go
```diff
@@ -45,20 +47,25 @@ type InstanceConfig struct {
 	Key          string
 }
 
-// MapInstanceConfigs parses the InstanceConfigs json string to map with key: regin_name and value: InstanceConfig for the instance in the region
-func (c *SelfRegConfig) MapInstanceConfigs() {
+// MapInstanceConfigs parses the InstanceConfigs json string to map with key: region name and value: InstanceConfig for the instance in the region
+func (c *SelfRegConfig) MapInstanceConfigs() error {
+	if ok := gjson.Valid(c.InstanceConfigs); ok == false {
```
DimitarPetrov:
```
if !gjson.Valid(c.InstanceConfigs) {
...
}
```

components/director/internal/domain/runtime/selfreg_config.go
```diff
@@ -47,8 +48,43 @@ type InstanceConfig struct {
...
+	if errorMsg != "" {
+		return errors.New(errorMsg)
+	}
+
+	return nil
+}
+
 // MapInstanceConfigs parses the InstanceConfigs json string to map with key: region name and value: InstanceConfig for the instance in the region
-func (c *SelfRegConfig) MapInstanceConfigs() error {
+func (c *SelfRegConfig) MapInstanceConfigs(oauthMode oauth.AuthMode) error {
```
DimitarPetrov, line 89: No need of the `oauthMode oauth.AuthMode` argument, you have it in the `c` reciever.

PetarTodorovv (approved):

PR #2262, opened 2022-03-21
Code Owners update respecting each team

NickyMateev:
# Description

This PR sets specific restrictions on ownership of directories with respect to our teams - dragon, falcon, raptor.

PR #2267, opened 2022-03-22
Expose ORD Version Through Well-Known Endpoint

SvetlinZarev-SAP:
**Description**

* Expose the implemented ORD version in the /info endpoint

**Related issue(s)**

- CMP-1679

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


dzahariev (approved):

dzahariev (dismissed):

dzahariev:
Looking at history of change https://github.com/kyma-incubator/compass/pull/2262, I can see that job pre-main-compass-integration became unstable. Since then the job needs to be triggered multiple times in order to pass.

components/director/internal/info/handler.go
```diff
@@ -7,21 +7,42 @@ import (
 	"github.com/kyma-incubator/compass/components/director/pkg/httputils"
 )
 
+// Implemented ORD Version
+const ordVersion = "1.1.0"
```
DimitarPetrov, line 11: It is probably better to move this constant to the open_resource_discovery package and export it.

On one hand, it is domain-specific and fits better, on the other hand, if you add it here on each version increasing a review from team-raptor will be needed.

DimitarPetrov (changes requested):

components/director/internal/info/handler.go
SvetlinZarev-SAP: done

DimitarPetrov (approved):

alextargov (approved):

DimitarPetrov (approved):

PR #2281, opened 2022-03-25
process scopes in deterministic way

StanislavStefanov:
**Description**

Process scopes in the following order:
1. If there is Override operation - take the scopes and return
2. Then process all Merge operations
3. Then process all Intersect operation

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/tenantmapping/handler_test.go
```diff
@@ -906,3 +1517,43 @@ func getMockContextProvider() *automock.ObjectContextProvider {
...
+			require.Equal(t, v.Str, actualExtra[k].Str)
+		}
+	}
+
+	expectedScopes := strings.Split(expectedExtra["scope"].String(), " ")
+	actualScopes := strings.Split(actualExtra["scope"].String(), " ")
+	require.True(t, scopesAreEqual(expectedScopes, actualScopes))
+}
+
+func scopesAreEqual(expectedScopes, actualScopes []string) bool {
```
DimitarPetrov, line 1535: Why not use require.Equal?

StanislavStefanov: As the order in which the scopes will be provided is not deterministic we may receive different results on separate executions of the tests. If we set expectedScopes to be []string{"test1", "test2", "test3"} and the actual scopes are []string{"test1", "test3", "test2"} the result from require.Equal will fail as the order of elements is not the same.

DimitarPetrov: Okay, makes sense, then you can use `require.ElementsMatch`

DimitarPetrov (approved):

DimitarPetrov (approved):

PR #2282, opened 2022-03-25
Add logs from jobs started from e2e tests

la4ezar:
**Description**
Add logs from jobs started from e2e tests

- Currently, the e2e tests for system-fetcher, tenant-fetcher, and ord-service(in the self-register flow cert-rotation job is created) create tests as part of the test. Logs from these jobs should be redirected to the standard output so that the logs remain even after the pod has been deleted.

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [x] Integration tests


tests/pkg/k8s/jobs.go
```diff
@@ -127,3 +131,42 @@ func WaitForJob(t *testing.T, ctx context.Context, k8sClient *kubernetes.Clients
 		time.Sleep(time.Second * 5)
 	}
 }
+
+func PrintJobLogs(t *testing.T, ctx context.Context, k8sClient *kubernetes.Clientset, jobName, namespace, containerName string) {
```
PetarTodorovv, line 135: can we somehow get the pod's logs only if it fails and have error? Or the logs from these jobs(system/tenant fetcher, cert-rotation-job, etc..) are not too verbose and it won't be a problem?

tests/system-fetcher/tests/handler_test.go
```diff
@@ -56,6 +56,8 @@ const (
 
 	nameLabelKey    = "displayName"
 	namePlaceholder = "name"
+
+	systemFetcherContainerName = "system-fetcher"
```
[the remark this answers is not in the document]
la4ezar: Yes, I think it will be much better if we extract the container names.

chart/compass/values.yaml
```diff
@@ -588,6 +589,7 @@ global:
       pageSize: 200
       sizeParam: "$top"
       skipParam: "$skip"
+    containerName: "system-fetcher"
```
NickyMateev, line 592: Make use of this in `system-fetcher-job.yaml` where the container name is defined, otherwise this is double maintenance

NickyMateev (changes requested):

NickyMateev (approved):

NickyMateev (approved):

desislavaa (approved):

PR #2284, opened 2022-03-28
Make region label value conigurable

StanislavStefanov:
*Description**

- make region label value configurable

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -127,7 +127,7 @@ global:
       version: "PR-56"
     e2e_tests:
       dir:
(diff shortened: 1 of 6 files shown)


dzahariev (approved):

chart/compass/values.yaml
```diff
@@ -127,7 +127,7 @@ global:
       version: "PR-56"
     e2e_tests:
       dir:
-      version: "PR-2263"
+      version: "PR-2283"
```
la4ezar, line 130: Shouldn't this be 2284?

PR #2292, opened 2022-03-30
Install ns adapter as helm post hook

PetarTodorovv:
**Description**
compass-migration job is failing(sporadically) because during its execution in parallel there is ns-adapter component which register application template, in later stage the migration job introduce unique indexes and data(app template with same name) which violates the unique index and the migration fails

- run the notification adapter as post helm hook when the db is up and running

- [x] Implementation

### DIFF
--- a/chart/compass/charts/ns-adapter/templates/deployment.yaml
+++ b/chart/compass/charts/ns-adapter/templates/deployment.yaml
@@ -7,9 +7,11 @@ metadata:
     app: {{ .Chart.Name }}
     reqlimit: big
     release: {{ .Release.Name }}
+  {{- if .Values.global.isLocalEnv }}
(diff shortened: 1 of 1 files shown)


chart/compass/charts/ns-adapter/templates/deployment.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
     app: {{ .Chart.Name }}
     reqlimit: big
     release: {{ .Release.Name }}
+  annotations:
```
DimitarPetrov, line 10: It is probably better to wrap this with isLocalEnv if.

DimitarPetrov (approved):

NickyMateev (approved):

PR #2298, opened 2022-03-31
Apply system-fetcher findings

kaloyanspiridonov:
**Description**

- the system-fetcher only process tenants which are global accounts
- remove one of the sLIS API filters
- add additional query param to the sLIS request
- use `productId` instead of `ppmsProductLineOfficialName` for system filtering
- store `productId` and `ppmsProductVersionId` as application labels
- map system `mainUrl` to app `baseUrl`
- remove tenant isolation for system-fetcher

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


components/director/internal/repo/upsert.go
```diff
@@ -125,33 +139,24 @@ func (u *upserter) unsafeUpsert(ctx context.Context, resourceType resource.Type,
 		return "", apperrors.NewInternalError("id cannot be empty, check if the entity implements Identifiable")
 	}
 
-	if resourceType.IsTopLevel() && id == upsertedID {
```
NickyMateev, line 128: Why did we have to remove this `id == upsertedID` check?

tests/system-fetcher/tests/handler_test.go
```diff
@@ -47,6 +47,8 @@ const (
 		"systemNumber": "%d",
```
NickyMateev, line 47: How much effort would it be to have the e2e test which ensures that the many to many table is leveraged when one and the same app is bound to multiple tenants?

dzahariev (approved):

desislavaa (approved):

StanislavStefanov (approved):

PR #2302, opened 2022-04-05
Ensure Notification Service Adapter auditlogs don't exceed limit

SvetlinZarev-SAP:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/gateway/pkg/proxy/adaptertransport.go
```diff
@@ -8,87 +8,192 @@ import (
...
+func newAuditLogContext(t *AdapterTransport, req *http.Request) (*auditLogContext, error) {
+	correlationHeaders := correlation.HeadersForRequest(req)
+
+	claims, err := getClaims(req.Context(), req.Header)
+	if err != nil {
+		return nil, errors.Wrap(err, "while parsing JWT")
+	}
+
+	transactionId, err := uuid.NewRandom()
+	if nil != err {
```
alextargov, line 59: Can we use the style `if x != nil` here and on the other places in the pr? The whole codebase uses this style.

components/gateway/pkg/proxy/adaptertransport.go
```diff
@@ -8,87 +8,192 @@ import (
...
+
+		if nil != err {
+			return errors.Wrapf(err, "while sending pre-change auditlog message to auditlog service for %s", transactionId)
+		}
+	}
+
+	return nil
+}
+
+func doPostAuditLog(ctx context.Context, auditLogCtx *auditLogContext, responseBody []byte) error {
```
la4ezar, line 158: Looks like this function is the same as the `doPreAuditLog` func, can we extract the code into a new function? 
The only difference looks like the `PreLog` and `Log` function invocation which is in the for-loop. One idea is to write a new function that accepts a function as a parameter. And on line 94 we will pass the `PreLog` function as an argument and on line 111 we will pass the `Log` function as an argument. 
Also, this new function will not wrap the error instead, we will wrap it on the upper level.

components/gateway/pkg/proxy/adaptertransport.go
```diff
@@ -120,3 +225,15 @@ func (t *AdapterTransport) getClaims(headers http.Header) (Claims, error) {
...
+func min(a int, b int) int {
+	if a <= b {
+		return a
+	}
+
+	return b
+}
+
+func multipleOfFour(i int) int {
+	return ((i + 3) / 4) * 4
```
la4ezar, line 238: The brackets around `(i + 3) / 4` are not actually needed.

chart/compass/charts/gateway/values.yaml
```diff
@@ -38,3 +38,7 @@ rateLimit:
 
 metrics:
   port: 3001
+
+auditLog:
+  maxBodySizeBytes: 998000
```
la4ezar, line 43: Why exactly 998000?

la4ezar (changes requested):

la4ezar:
Should we write an e2e test with a payload that exceeds 1MB?

chart/compass/charts/gateway/values.yaml
SvetlinZarev-SAP: because the current auditlog limit is 1MB and I want to be a bit less than that, becasue the auditlog message contains additional information, not just the body

components/gateway/pkg/proxy/adaptertransport.go
SvetlinZarev-SAP: I always use nil!=xxx, and it would take me a lot of time to find out all the places I've used it

components/gateway/pkg/proxy/adaptertransport.go
```diff
@@ -8,87 +8,192 @@ import (
...
 func (t *AdapterTransport) RoundTrip(req *http.Request) (resp *http.Response, err error) {
 	if req.Body == nil || req.Method == http.MethodGet {
 		return t.RoundTripper.RoundTrip(req)
 	}
+	defer httpcommon.CloseBody(req.Context(), req.Body)
 
-	buf := &bytes.Buffer{}
-	_, err = io.Copy(buf, req.Body)
+	buffer := &bytes.Buffer{}
+	_, err = io.Copy(buffer, req.Body)
```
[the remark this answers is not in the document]
SvetlinZarev-SAP: It's the old code, I've only renamed the variable

components/gateway/pkg/proxy/adaptertransport.go
```diff
@@ -8,87 +8,192 @@ import (
...
 
+func doPreAuditLog(ctx context.Context, auditLogCtx *auditLogContext, requestBody []byte) error {
+	log.C(ctx).Infof("Maximum request body length per audit log message: %d", auditLogCtx.cfg.MsgBodySizeLimit)
+	log.C(ctx).Infof("Length of received request body: %d", len(requestBody))
+
+	transactionId := auditLogCtx.logId.String()
+
+	// Compute how many pre-audit logs we have to send
+	shards := len(requestBody) / auditLogCtx.cfg.MsgBodySizeLimit
+	if shards%auditLogCtx.cfg.MsgBodySizeLimit != 0 {
```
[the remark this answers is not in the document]
SvetlinZarev-SAP: I tried to, but gofmt removes the whitespaces

components/gateway/pkg/proxy/adaptertransport.go
```diff
@@ -8,87 +8,192 @@ import (
...
+		shards += 1
```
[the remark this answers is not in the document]
SvetlinZarev-SAP: I prefer += 1

components/gateway/pkg/proxy/adaptertransport.go
SvetlinZarev-SAP: IMO it's clearer with the brackets

components/gateway/pkg/proxy/adaptertransport.go
SvetlinZarev-SAP: They are indeed very similar, but there are also a lot of small differences, which makes it very difficult to make it only one function. I'll extract some common duplicate code in a separate method

NickyMateev (approved):

StanislavStefanov (approved):

PR #2303, opened 2022-04-05
Self registration secret mount as volume

PetarTodorovv:
**Description**
Currently we load self registration configuration as environment variable and this approach is not robust when this configuration is bigger. It fails with `argument list too long `

- Load the self registration configuration from file not from environment variable
- Adapt local director script

**Related issue(s)**
- https://github.com/kyma-incubator/compass/pull/2263

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/runtime/selfreg_config.go
```diff
@@ -20,6 +21,7 @@ type SelfRegConfig struct {
 	SelfRegisterNameQueryParam      string `envconfig:"APP_SELF_REGISTER_NAME_QUERY_PARAM,optional"`
 	SelfRegisterTenantQueryParam    string `envconfig:"APP_SELF_REGISTER_TENANT_QUERY_PARAM,optional"`
 	SelfRegisterRequestBodyPattern  string `envconfig:"APP_SELF_REGISTER_REQUEST_BODY_PATTERN,optional"`
+	SelfRegisterSecretPath          string `envconfig:"APP_SELF_REGISTER_SECRET_PATH,optional"`
```
DimitarPetrov, line 24: This should not be marked as optional.

DimitarPetrov (changes requested):

DimitarPetrov (approved):

PR #2294, opened 2022-03-30
Introduce Tenant on-demand API

inaand:
**Description**

New API to fetch on-demand newly created tenants and store them 

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests
- [x] `chart/compass/values.yaml` is updated 
- [ ] Mocks are regenerated, using the automated script

PR #2307, opened 2022-04-12
Use tenant on demand API in Director

StefanCholakov:
**Description**

- Call the API introduced in [PR-2294](https://github.com/kyma-incubator/compass/pull/2294) when an automatic scenario assignment is created.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/scenarioassignment/resolver.go
```diff
@@ -54,31 +57,22 @@ type Resolver struct {
 	converter     gqlConverter
 	svc           asaService
 	tenantService tenantService
+	fetcher       tenantFetcher
 }
 
 // CreateAutomaticScenarioAssignment missing godoc
 func (r *Resolver) CreateAutomaticScenarioAssignment(ctx context.Context, in graphql.AutomaticScenarioAssignmentSetInput) (*graphql.AutomaticScenarioAssignment, error) {
+	if err := r.fetcher.FetchOnDemand(in.Selector.Value); err != nil {
+		return nil, errors.Wrapf(err, "while trying to create if not exists subaccount %s", in.Selector.Value)
```
desislavaa, line 66: We should probably add a check here to first fetch the tenant from the Compass DB.
Then, if it does not exist, we "fetch it on demand", and if everything is ok -> continue.

The transaction here is interesting - we can open it in the beginning of the method, if the tenant does not exist -> close it, and open a new one when the call to TF succeeds - something like that, so we don't have ext calls inside of a transaction.

components/director/internal/domain/tenant/fetcher_on_demand.go
```diff
@@ -0,0 +1,43 @@
...
+
+// NewFetchOnDemandService returns object responsible for fetching tenants
+func NewFetchOnDemandService(client Client, url string) *fetchOnDemandService {
+	return &fetchOnDemandService{
+		client:           client,
+		tenantFetcherURL: url,
+	}
+}
+
+// FetchOnDemand calls an API which fetch details for the given tenant and store them in the database
```
desislavaa, line 29: Since we don't process the response body, we can rephrase this like:
> FetchOnDemand calls an API which fetches details for the given tenant from an external tenancy service, stores the tenant in the Compass DB and returns 200 OK if the tenant was successfully created.

components/director/internal/domain/tenant/fetcher_on_demand_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+func TestFetchOnDemand(t *testing.T) {
+	fetchTenantURL := "https://compass-tenant-fetcher.kyma.local/tenants/v1/fetch"
+	tenantID := "b91b59f7-2563-40b2-aba9-fef726037aa3"
+	testErr := errors.New("error")
+
+	testCases := []struct {
+		Name          string
+		Client        func() *automock.Client
+		ExpectedError error
```
desislavaa, line 27: `string` instead of `error` can also be used here (`ExpectedErrorMsg`), since we only need the error message of the error

components/external-services-mock/internal/provider/handler.go
```diff
@@ -101,3 +103,23 @@ func (h *handler) Dependencies(writer http.ResponseWriter, r *http.Request) {
...
+		log.C(ctx).Error("Tenant path parameter is missing from request")
+		http.Error(writer, "Region path parameter is missing from request", http.StatusBadRequest)
+		return
+	}
+	log.C(ctx).Printf("Fetching tenant details for tenant with id %s", tenantId)
+
+	writer.WriteHeader(http.StatusOK)
+	writer.Header().Set("Content-Type", "application/json")
+	if _, err := writer.Write([]byte("{}")); err != nil {
+		log.C(request.Context()).WithError(err).Errorf("Failed to write response body for fetch tenant request")
```
desislavaa, line 122: we have the request context in `ctx`

components/external-services-mock/internal/provider/handler.go
```diff
@@ -101,3 +103,23 @@ func (h *handler) Dependencies(writer http.ResponseWriter, r *http.Request) {
...
+func (h *handler) FetchTenantOnDemand(writer http.ResponseWriter, request *http.Request) {
+	ctx := request.Context()
+	vars := mux.Vars(request)
+	tenantId, ok := vars["tenantId"]
+	if !ok {
+		log.C(ctx).Error("Tenant path parameter is missing from request")
+		http.Error(writer, "Region path parameter is missing from request", http.StatusBadRequest)
+		return
+	}
+	log.C(ctx).Printf("Fetching tenant details for tenant with id %s", tenantId)
```
desislavaa, line 117:
```
log.C(ctx).Infof
```

components/director/internal/domain/scenarioassignment/resolver.go
StefanCholakov: The new tenant fetcher API is constructed in a way that it first checks whether the tenant exists in the database and it won't fetch it if it does. So this check is done there.

desislavaa (dismissed):

desislavaa (approved):

PR #2308, opened 2022-04-14
Derive baseurl for customer managed systems

alextargov:
**Description**

- 

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/internal/domain/application/fixtures_test.go
+++ b/components/director/internal/domain/application/fixtures_test.go
@@ -69,7 +69,7 @@ func fixModelApplication(id, tenant, name, description string) *model.Applicatio
 	}
 }
 
-func fixModelApplicationWithAllUpdatableFields(id, tenant, name, description, url string, conditionStatus model.ApplicationStatusCondition, conditionTimestamp time.Time) *model.Application {
+func fixModelApplicationWithAllUpdatableFields(id, name, description, url string, baseURL *string, conditionStatus model.ApplicationStatusCondition, conditionTimestamp time.Time) *model.Application {
 	return &model.Application{
 		Status: &model.ApplicationStatus{
 			Condition: conditionStatus,
@@ -81,6 +81,7 @@ func fixModelApplicationWithAllUpdatableFields(id, tenant, name, description, ur
 		HealthCheckURL:      &url,
 		ProviderName:        &providerName,
 		BaseEntity:          &model.BaseEntity{ID: id},
+		BaseURL:             baseURL,
 	}
 }
 
--- a/components/director/internal/domain/application/service.go
+++ b/components/director/internal/domain/application/service.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
+
+	"github.com/kyma-incubator/compass/components/director/pkg/str"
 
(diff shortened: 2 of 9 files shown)


components/director/internal/domain/application/fixtures_test.go
```diff
@@ -81,6 +81,7 @@ func fixModelApplicationWithAllUpdatableFields(id, tenant, name, description, ur
 		HealthCheckURL:      &url,
 		ProviderName:        &providerName,
 		BaseEntity:          &model.BaseEntity{ID: id},
+		BaseURL:             baseURL,
```
dzahariev, line 84: Use &baseURL here. 
The Application struct defines BaseURL as pointer.

components/director/internal/domain/application/service.go
```diff
@@ -500,6 +503,35 @@ func (s *service) Upsert(ctx context.Context, in model.ApplicationRegisterInput)
...
+// TryUpdateBaseURL Gets application by ID. If the application does not have a BaseURL set, the API TargetURL is parsed and set as BaseURL
+func (s *service) TryUpdateBaseURL(ctx context.Context, appID, targetURL string) error {
+	appTenant, err := tenant.LoadFromContext(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "while loading tenant from context")
+	}
+
+	app, err := s.Get(ctx, appID)
+	if err != nil {
+		return err
```
dzahariev, line 515: Wrap error:
return errors.Wrapf(err, "while getting Application with id %s", id)

components/director/internal/domain/application/service.go
```diff
@@ -500,6 +503,35 @@ func (s *service) Upsert(ctx context.Context, in model.ApplicationRegisterInput)
...
+	}
+
+	if app.BaseURL != nil && len(*app.BaseURL) > 0 {
```
dzahariev, line 518: len(*app.BaseURL) > 0 is sufficient, as it also handles nil.

components/director/internal/domain/application/service.go
```diff
@@ -500,6 +503,35 @@ func (s *service) Upsert(ctx context.Context, in model.ApplicationRegisterInput)
...
+		log.C(ctx).Infof("BaseURL for Application %s already exists. Will not update it.", appID)
+		return nil
+	}
+
+	log.C(ctx).Infof("BaseURL for Application %s does not exist. Will update it.", appID)
+
+	parsedTargetURL, err := url.Parse(targetURL)
+	if err != nil {
+		return errors.Wrapf(err, "while parsing targetURL")
```
dzahariev, line 527: Add more information about the error. It is good to include also the target URL that was not parsed here:
return errors.Wrapf(err, "while parsing targetURL: %s", targetURL)

components/director/internal/domain/application/service.go
```diff
@@ -500,6 +503,35 @@ func (s *service) Upsert(ctx context.Context, in model.ApplicationRegisterInput)
...
+	}
+
+	app.BaseURL = str.Ptr(fmt.Sprintf("%s://%s", parsedTargetURL.Scheme, parsedTargetURL.Host))
```
dzahariev, line 530: This additional conversion will not be needed if define BaseURL as string, not as pointer in model.Application struct.

dzahariev (approved):

PR #2311, opened 2022-04-14
Metrics for GraphQL endpoints

ivantenevvasilev:
**Description**

- introduce graphql endpoint usage metrics
- introduce graphql mutation and query usage charts
- update Compass/Director dashboard

**Pull Request status**
- [x] Implementation


components/director/internal/metrics/collector.go
```diff
@@ -24,6 +25,7 @@ type Collector struct {
 	hydraRequestTotal      *prometheus.CounterVec
 	hydraRequestDuration   *prometheus.HistogramVec
 	clientTotal            *prometheus.CounterVec
+	mutationCount          *prometheus.CounterVec
```
StanislavStefanov, line 28: We are using this counter for both graphql mutations and queries so `gqlOperationCount` may be mora appropriate name

StanislavStefanov (changes requested):

StanislavStefanov (approved):

chart/compass/values.yaml
```diff
@@ -182,6 +182,7 @@ global:
     metrics:
       port: 3003
       enableClientInstrumentation: true
+      enableGraphqlOperationInstrumentation: false
```
PetarTodorovv, line 185: why this property is false? Shouldn't we use it in production environment as well?

ivantenevvasilev: I was under the impression, that it's better to have it off by default and have it explicitly overridden to be enabled wherever it is needed.

PetarTodorovv: Yep, generally that's the idea but If we're gonna use this functionality in production(everywhere) then I think we can switch it to `true` by default. That way we'll have it configured only at one place, less overrides and cleaner setup.

Usually we use the "default false" approach when we want the productive configuration to be `false` and only override it to `true` for local setup

StanislavStefanov (approved):

StanislavStefanov (approved):

StanislavStefanov (approved):

PR #2318, opened 2022-04-19
Merge systems api

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/domain/application/service.go
```diff
@@ -712,6 +714,60 @@ func (s *service) DeleteLabel(ctx context.Context, applicationID string, key str
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "while getting destination application")
+	}
+
+	srcApp, err := s.Get(ctx, srcID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "while getting source application")
+	}
+
+	srcBaseURL := str.PtrStrToStr(srcApp.BaseURL)
```
NickyMateev, line 735: You probably wanna make sure this works even if one of srcBaseURL or  destBaseURL has a trailing `/` symbol as part of the URL - so maybe trim it for both URLs just to be sure that both URLs are as equal as possible

components/director/internal/domain/application/service.go
```diff
@@ -712,6 +715,60 @@ func (s *service) DeleteLabel(ctx context.Context, applicationID string, key str
...
+
+	if srcApp.Status == nil {
+		return nil, errors.Errorf("Could not determine status of source application with id %s", srcID)
+	}
+
+	if srcApp.Status.Condition == model.ApplicationStatusConditionConnected {
+		return nil, errors.Errorf("Cannot merge application with id %s, because it is in a %s status", srcID, model.ApplicationStatusConditionConnected)
+	}
+
+	if err := mergo.Merge(destApp, *srcApp); err != nil {
```
NickyMateev, line 756: Just realized we'd be having a nasty bug if all we did was merge the fields and forget the labels.

These are the types of labels associated with a autodiscovered app:
image


These are the types of labels associated with a self-reg app:
image

Ideally this should be fixed in the same PR, but since the API is not open for productive use we can think about splitting it or opening a bug. Let's discuss this..

NickyMateev: Also the E2E test should ideally cover this labels scenario as well.

NickyMateev (changes requested):

nkanchev (approved):

components/director/internal/domain/application/service.go
```diff
@@ -749,26 +760,100 @@ func (s *service) Merge(ctx context.Context, destID, srcID string) (*model.Appli
...
+		return nil, errors.Wrapf(err, "while trying to merge labels for applications with ids %s and %s", destID, srcID)
+	}
+
 	log.C(ctx).Infof("Deleting source application with id %s", srcID)
 	if err := s.Delete(ctx, srcID); err != nil {
 		return nil, err
 	}
 
+	log.C(ctx).Infof("Deleting source labels for application with id %s", srcID)
+	if err := s.labelRepo.DeleteAll(ctx, appTenant, model.ApplicationLabelableObject, srcID); err != nil {
```
NickyMateev: Do you need to delete the source labels when the source app will be deleted anyway (the labels should be gone after the app is deleted)?

components/director/internal/domain/application/service.go
```diff
@@ -749,26 +760,100 @@ func (s *service) Merge(ctx context.Context, destID, srcID string) (*model.Appli
...
+	}
+
 	return destApp, nil
 }
 
+// handleMergeLabels merges source labels into destination labels
+func (s *service) handleMergeLabels(srcAppLabels, destAppLabels map[string]*model.Label) (map[string]*model.Label, error) {
+	destIntSysID := fmt.Sprintf("%v", destAppLabels[intSysKey].Value)
+	srcIntSysID := fmt.Sprintf("%v", srcAppLabels[intSysKey].Value)
+	if len(destIntSysID) == 0 && len(srcIntSysID) > 0 {
```
NickyMateev, line 799: Not sure that this case is even possible, but maybe it's a good sanity check

components/director/internal/domain/application/service.go
```diff
@@ -749,26 +760,100 @@ func (s *service) Merge(ctx context.Context, destID, srcID string) (*model.Appli
...
+		}
+	}
+
+	if err := mergo.Merge(&destAppLabels, srcAppLabels); err != nil {
+		return nil, errors.Wrapf(err, "while trying to merge labels")
+	}
+
+	destAppLabels[model.ScenariosKey].Value = destScenariosStrSlice
+	destAppLabels[managedKey].Value = srcAppLabels[managedKey].Value
+
```
NickyMateev, line 845: I was wondering why some of the labels have to be manually assigned instead of just leveraging the mergo lib..?

Maybe there's some limitations to the lib - if so maybe it's worth noting in the function go doc?

components/director/internal/domain/application/service.go
```diff
@@ -1057,3 +1142,12 @@ func (s *service) genericUpsert(ctx context.Context, appTenant string, in model.
 
 	return nil
 }
+
+func containsStr(s []string, str string) bool {
```
NickyMateev, line 1143: add this function to director/pkg/str package or check if such doesn't already exist there and reuse it

NickyMateev (changes requested):

components/director/internal/domain/application/service.go
alextargov: Ooops, that's a leftover

components/director/internal/domain/application/service.go
alextargov: Limitations, yeah. That's why I need to handle the scenarios and intSysID manually. The case for IntSysID is that if any of the IDs (source or dest) is empty string, it ignores the whole field and does not do the merging. You can specify options for the merging there was one "Overwrite" which would do the trick in the case, but it would merge all fields from src in dest even if the src already has them populated. Which is not what we want. Will add more info in the godoc

components/director/internal/domain/application/service.go
```diff
@@ -712,6 +716,141 @@ func (s *service) DeleteLabel(ctx context.Context, applicationID string, key str
...
+			destScenariosStrSlice = append(destScenariosStrSlice, srcScenario)
+		}
+	}
+
+	if err := mergo.Merge(&destAppLabels, srcAppLabels); err != nil {
+		return nil, errors.Wrapf(err, "while trying to merge labels")
+	}
+
+	destAppLabels[model.ScenariosKey].Value = destScenariosStrSlice
+	destAppLabels[managedKey].Value = srcAppLabels[managedKey].Value
```
NickyMateev, line 844: Regarding the final value of the `managed` label:
- if `src` is auto-discovered and `dest` is customer-managed then we want the label `managed` to stay `true` -> you cover that here
- if `src` is customer-managed and `dest` is auto-discovered then we AGAIN want the label `managed` to stay `true`, because the final app will again have all fields (like systemNumber) which constitute an auto-discovered app (managed by system fetcher) -> this is not covered here and the final value for `managed` would actually be `false` with current logic.

Ultimately this will require just a slight tweak of the sort:
```
if destAppLabels[managedKey].Value.(bool) == true ||  srcAppLabels[managedKey].Value.(bool) == true {
  destAppLabels[managedKey].Value = true
}
```

tests/director/tests/application_api_test.go
```diff
@@ -1485,3 +1487,129 @@ func TestApplicationDeletionInScenario(t *testing.T) {
...
+	healthURL := ptr.String("http://health.com")
+	providerName := ptr.String("test-provider")
+	description := ptr.String("app description")
+	tenantId := tenant.TestTenants.GetDefaultTenantID()
+	namePlaceholder := "name"
+	managedLabelValue := "true"
+	sccLabelValue := "cloud connector"
+	expectedProductType := "MergeTemplate"
+	newFormation := "formation-merge-applications-e2e"
+	destScenariosValue := []interface{}{"DEFAULT", "formation-merge-applications-e2e"}
```
NickyMateev, line 1505: You can reuse the `newFormation` variable here to avoid double maintenance

components/director/internal/domain/application/service.go
alextargov: Explanation here: https://github.com/kyma-incubator/compass/pull/2318#discussion_r870238316

NickyMateev: Yes, I just meant that since we have a sanity check beforehand that the `src` and `dest` have one and the same app template, we usually specify one and the same int system (when there is one) when an app template has a corresponding int system. So we kind of have the guarantee that if the app templates match and there is an int system associated with the app template, then both `src` and `dest` should have one and the same value for this label.

I don't see a scenario where two systems from one and the same app template have two different int sys IDs. So either they are the same, or it's missing (and so it's fine for the mergo lib to skip it or whatever it does when the value is empty string). I'm kind of iffy on whether we should keep this code, because practically it should not reach such a situation, theoretically everything is possible, but the trade off is that we have some cryptic incentive with this code part which people who aren't aware of this discussion here will wonder why we have this custom logic.

NickyMateev (approved):

nkanchev (approved):

PR #2322, opened 2022-04-20
Replace K8s jobs with Go routines

inaand:
**Description**

Replace K8s jobs with Go routines

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/pkg/kubernetes/client.go
```diff
@@ -17,9 +17,9 @@ import (
 
 // Config missing godoc
 type Config struct {
-	PollInterval time.Duration `envconfig:"default=2s,APP_KUBERNETES_POLL_INTERVAL"`
-	PollTimeout  time.Duration `envconfig:"default=1m,APP_KUBERNETES_POLL_TIMEOUT"`
-	Timeout      time.Duration `envconfig:"default=2m,APP_KUBERNETES_TIMEOUT"`
+	PollInterval time.Duration `envconfig:"optional,default=2s,APP_KUBERNETES_POLL_INTERVAL"`
```
desislavaa, line 20: does that need `optional` when it already nas a default?

components/director/internal/tenantfetchersvc/tenant_fetcher_job_config_test.go
```diff
@@ -0,0 +1,208 @@
...
+	// GIVEN
+	jobName := "testJob"
+
+	jobIntervalMins := 3 * time.Minute
+	defaultJobIntervalMins := 1 * time.Minute
+
+	tenantProvider := "testProvider"
+	defaultTenantProvider := "external-provider"
+
+	testCases := []struct {
```
desislavaa, line 163: we can have "explicit" test cases here instead of the approach with the loop, it will probably look more simple.
The same applies for the test above

components/director/pkg/kubernetes/client.go
desislavaa: tricky one, got it

desislavaa (approved):

desislavaa (approved):

desislavaa (approved):

PR #2325, opened 2022-04-25
Prepare for Github workflow in compass

Halamix2:
**Description**

- add CODEOWNERS file, updated with the current data

**Related issue(s)**

- https://github.com/kyma-project/test-infra/issues/5322
- https://github.com/kyma-project/test-infra/issues/5312

/assign
/retest


DimitarPetrov:
What is the reason to roll back the prow approach with the OWNERS files?

We are more or less happy with the current workflow and would like to keep it for now. Are the prow plugins going to stay?

dekiel:
@DimitarPetrov We revert that change, because revoking write access to the repository is causing too much negative impact on teams daily work. Main assumption for implementing Prow approval flow was that organisation members has read only access to repositories. Plugins will be disabled because we must restore write access. As far I know for your team it was not revoked at all because it would block your development flow. We will step back and research again possible solutions. You can follow in that topic [here](https://github.com/kyma-project/community/issues/613).

DimitarPetrov:
@dekiel Can we leave the plugins on and just use OWNERS files only in repos that want to leverage the prow functionality?

The reason I am asking this is that we already established a process in our team with the auto-assigning of labels based on  OWNERS files and multi-team approvals. 

Therefore it will be an effort to switch for us (not only the files but in our team processes that we just established around the prow workflow). If there is a way to keep the plugins so that we can keep using them it would be great.

dekiel:
@DimitarPetrov Prow approval flow plugins together with write access allow workaround required reviews for PR. This is serious security and process issue. We will think if we can keep enabled automatic labeling.

DimitarPetrov:
@dekiel Does the issue occurs even for repositories that do not contain any OWNERS files. 

My understanding is that repos that have strict requirements for who can merge what and who have write access (probably repos that many teams contribute to) can use Github CODEOWERNS and if the repo does not contain any prow OWNERS files the repo will not be affected by the issue you mention.

For our workflow, the benefits outweigh the potential impact of the write access problem and that's why we are using prow OWNERS workflow for the compass repo. However, if your decision as prow owners is to remove those plugins from the prow installation we will migrate our process, but we will need some time.

DimitarPetrov:
@Halamix2 Can you run the script that generates CODEOWNERS from current OWNERS files again for up to date version of this PR?

DimitarPetrov (approved):

PR #2326, opened 2022-04-26
Create test definitions conditionally

kaloyanspiridonov:
**Description**

- create test definitions only if global.isForTesting is enabled.

**Pull Request status**
- [X] Implementation
- [ ] Unit tests
- [ ] Integration tests


chart/compass/templates/bench/director/director-bench.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if .Values.global.isForTesting }}
```
PetarTodorovv, line 1: Shouldn't we have override for this regarding the productive setup?

PR #2330, opened 2022-04-28
Stop storing subaccounts lazily

StefanCholakov:
**Description**

- In addition to [PR-2307](https://github.com/kyma-incubator/compass/pull/2307), two more places in the code were identified to lazily store subaccounts. This PR replaces this logic with calling the tenant fetcher API introduced in [PR-2294](https://github.com/kyma-incubator/compass/pull/2294)

Testing:

The logic was tested manually, by running local installation and configuring tenant-fetcher deployment to call a real tenant provider. After that, registerRuntime mutation was executed, using runtime input with a label which specifies missing tenant. In the end, the tenant from the label was successfully fetched and stored in the database.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/formation/resolver.go
```diff
@@ -149,3 +169,32 @@ func (r *Resolver) UnassignFormation(ctx context.Context, objectID string, objec
 
 	return r.conv.ToGraphQL(newFormation), nil
 }
+
+func (r *Resolver) assignFormationForTenant(ctx context.Context, tx *persistence.PersistenceTx, tenantFromContext string, objectID string, formationInput graphql.FormationInput) (*model.Formation, error) {
```
DimitarPetrov, line 173: Can you move this logic in the service instead of the resolver?

We are planning to use the formation service in other flows as well and this logic fits more like a business logic instead of graphQL API logic.

Also, this will remove the if above and the differentiation between the formation object type will be only in the service.

components/director/internal/domain/runtime/resolver.go
```diff
@@ -710,3 +728,45 @@ func (r *Resolver) cleanupAndLogOnError(ctx context.Context, runtimeID, region s
 		log.C(ctx).Errorf("An error occurred during cleanup of self-registered runtime: %v", err)
 	}
 }
+
+func (r *Resolver) extractTenantFromSubaccountLabel(ctx context.Context, tx *persistence.PersistenceTx, value interface{}) (*model.BusinessTenantMapping, error) {
```
DimitarPetrov, line 732: Why move this to the graphQL layer? It fits in the service layer more.

StefanCholakov: If it's moved there, CreateWithMandatoryLabels should receive a pointer to persistence.PersistenceTx in order to manipulate the transaction(possibly close it, make the API call and re-open it). But the CreateWithMandatoryLabels function is also used by Create and I am not sure how we can get the PersistenceTx there in order to pass it further. I just didn't want to change the signatures of Create and CreateWithMandatoryLabels.

desislavaa (approved):

components/director/internal/domain/runtime/service.go
```diff
@@ -222,7 +220,7 @@ func (s *service) Create(ctx context.Context, in model.RuntimeInput) (string, er
 
 // CreateWithMandatoryLabels creates a runtime in a given tenant and also adds mandatory labels to it.
 func (s *service) CreateWithMandatoryLabels(ctx context.Context, in model.RuntimeInput, id string, mandatoryLabels map[string]interface{}) error {
-	if saVal, ok := in.Labels[scenarioassignment.SubaccountIDKey]; ok { // TODO: <backwards-compatibility>: Should be deleted once the provisioner start creating runtimes in a subaccount
+	if saVal, ok := in.Labels[scenarioassignment.SubaccountIDKey]; ok {
```
DimitarPetrov, line 223: The comment is relevant here and can be returned back.

DimitarPetrov (approved):

PR #2374, opened 2022-05-20
Adapt ASA Engine to use Formations API

ivantenevvasilev:
**Description**

- adapt EnsureScenarioAssigned to use AssignFormation
- adapt RemoveAssignedScenario to use UnassignFormation
- move CreateAutomaticScenarioAssignment to Formations Service layer
- move DeleteAutomaticScenarioAssignment to Formations Service layer
- move engine.go to Formations Service layer
- adapt unit tests
 
**Related issue(s)**
- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests


components/director/internal/domain/scenarioassignment/engine.go
```diff
@@ -50,38 +48,6 @@ func NewEngine(labelService labelUpsertService, labelRepo labelRepository, scena
 	}
 }
 
-// EnsureScenarioAssigned ensures that the scenario is assigned to all the runtimes that are in the ASAs target_tenant_id
-func (e *engine) EnsureScenarioAssigned(ctx context.Context, in model.AutomaticScenarioAssignment) error {
```
DimitarPetrov, line 54: Maybe we can moove the other two methods as well as it does not makes much sense to encapsulate only those two in a separate "engine".

If you decided to leave it like this there are labelRepo and labelService which seems unused now.

components/director/internal/domain/scenarioassignment/resolver_test.go
```diff
@@ -72,7 +72,9 @@ func TestResolverCreateAutomaticScenarioAssignment(t *testing.T) {
 		fetcherSvc := &automock.TenantFetcher{}
 		fetcherSvc.On("FetchOnDemand", mock.Anything).Return(testErr).Once()
 
-		sut := scenarioassignment.NewResolver(transact, nil, nil, nil, fetcherSvc)
+		formationSvc := &automock.FormationService{}
```
DimitarPetrov, line 75: You can pass nil to the constructor if it is unused in this particular test.

Valid for tests below as well.

components/director/internal/domain/formation/service.go
```diff
@@ -180,6 +205,160 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+	}
+
+	if err = s.RemoveAssignedScenario(ctx, in); err != nil {
+		return errors.Wrap(err, "while unassigning scenario from runtimes")
+	}
+
+	return nil
+}
+
+func (s *service) EnsureScenarioAssigned(ctx context.Context, in model.AutomaticScenarioAssignment) error {
```
DimitarPetrov, line 281: Missing godoc.

components/director/internal/domain/formation/service.go
```diff
@@ -180,6 +205,160 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+	runtimes, err := s.runtimeRepo.ListAll(ctx, in.TargetTenantID, nil)
+
```
DimitarPetrov, line 283: Unnecessary new line.

components/director/internal/domain/formation/service.go
```diff
@@ -180,6 +205,160 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+	if err != nil {
+		return errors.Wrapf(err, "while fetching runtimes in target tenant: %s", in.TargetTenantID)
+	}
+	for _, runtime := range runtimes {
+		_, err = s.AssignFormation(ctx, in.Tenant, runtime.ID, graphql.FormationObjectTypeRuntime, model.Formation{Name: in.ScenarioName})
+		if err != nil && !apperrors.IsNotUniqueError(err) {
```
DimitarPetrov, line 289: In which scenario an not unique error can be returned?

components/director/internal/domain/formation/service.go
```diff
@@ -180,6 +205,160 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+			return errors.Wrapf(err, "while assigning formation")
```
DimitarPetrov, line 290: while assigning runtime with id %s to formation %s coming from ASA

components/director/internal/domain/formation/service.go
```diff
@@ -180,6 +205,160 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+func (s *service) RemoveAssignedScenario(ctx context.Context, in model.AutomaticScenarioAssignment) error {
+	runtimes, err := s.runtimeRepo.ListAll(ctx, in.TargetTenantID, nil)
+	if err != nil {
+		return errors.Wrapf(err, "while fetching runtimes in target tenant: %s", in.TargetTenantID)
+	}
+
+	for _, runtime := range runtimes {
+		_, err = s.UnassignFormation(ctx, in.Tenant, runtime.ID, graphql.FormationObjectTypeRuntime, model.Formation{Name: in.ScenarioName})
+		if err != nil && !apperrors.IsNotFoundError(err) {
+			return errors.Wrapf(err, "while unassigning formation")
```
DimitarPetrov, line 306: while unassigning runtime with id %s from formation %s coming from ASA

components/director/internal/domain/formation/service.go
```diff
@@ -180,6 +205,160 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+// RemoveAssignedScenario removes all the scenarios that are coming from the provided ASA
+func (s *service) RemoveAssignedScenario(ctx context.Context, in model.AutomaticScenarioAssignment) error {
+	runtimes, err := s.runtimeRepo.ListAll(ctx, in.TargetTenantID, nil)
+	if err != nil {
+		return errors.Wrapf(err, "while fetching runtimes in target tenant: %s", in.TargetTenantID)
+	}
+
+	for _, runtime := range runtimes {
+		_, err = s.UnassignFormation(ctx, in.Tenant, runtime.ID, graphql.FormationObjectTypeRuntime, model.Formation{Name: in.ScenarioName})
+		if err != nil && !apperrors.IsNotFoundError(err) {
```
DimitarPetrov, line 305: In which scenario a not found error will be returned here?

DimitarPetrov (changes requested):

PetarTodorovv (changes requested):

components/director/internal/domain/scenarioassignment/service.go
```diff
@@ -150,60 +83,3 @@ func (s *service) List(ctx context.Context, pageSize int, cursor string) (*model
 
 	return s.repo.List(ctx, tnt, pageSize, cursor)
 }
-
-// DeleteManyForSameSelector missing godoc
-func (s *service) DeleteManyForSameTargetTenant(ctx context.Context, in []*model.AutomaticScenarioAssignment) error {
```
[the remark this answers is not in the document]
ivantenevvasilev: The `Create`, `Delete`, `DeleteForTargetTenant` were moved from the ASA service to the formations service because it would lead to a circular dependency between the formations service, which uses those methods, the ASA engine, which is used in the ASA service and now, the ASA engine which is going to be use the formations service. The first two were renamed to `CreateAutomaticScenarioAssignment` and `DeleteAutomaticScenarioAssignment`. The repo layer is still used in the formations service.

components/director/internal/domain/formation/service.go
```diff
@@ -26,11 +30,19 @@ type labelRepository interface {
 	Delete(context.Context, string, model.LabelableObject, string, string) error
 }
 
+//go:generate mockery --exported --name=runtimeRepository --output=automock --outpkg=automock --case=underscore --disable-version-string
+type runtimeRepository interface {
+	ListAll(ctx context.Context, tenant string, filter []*labelfilter.LabelFilter) ([]*model.Runtime, error)
+	Exists(ctx context.Context, tenant, id string) (bool, error)
```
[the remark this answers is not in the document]
ivantenevvasilev: In the new `EnsureScenarioAssigned` and `RemoveAssignedScenario` methods which are now part of the formations service, rather than the ASA engine.

components/director/internal/domain/formation/service.go
```diff
@@ -47,9 +59,18 @@ type uidService interface {
...
-	Delete(ctx context.Context, in model.AutomaticScenarioAssignment) error
+}
+
+//go:generate mockery --exported --name=automaticFormationAssignmentRepository --output=automock --outpkg=automock --case=underscore --disable-version-string
+type automaticFormationAssignmentRepository interface {
+	Create(ctx context.Context, model model.AutomaticScenarioAssignment) error
+	ListForTargetTenant(ctx context.Context, tenantID string, targetTenantID string) ([]*model.AutomaticScenarioAssignment, error)
+	GetForScenarioName(ctx context.Context, tenantID, scenarioName string) (model.AutomaticScenarioAssignment, error)
+	ListAll(ctx context.Context, tenantID string) ([]*model.AutomaticScenarioAssignment, error)
+	List(ctx context.Context, tenant string, pageSize int, cursor string) (*model.AutomaticScenarioAssignmentPage, error)
```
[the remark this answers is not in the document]
ivantenevvasilev: Same thing as the previous comment, while I was moving the implementation I seem to have used the full interface that was used in the previous ASA repo interface, out of these in the formations service the only used ones are `Create`, `DeleteForTargetTenant` and `DeleteForScenarioName`, because the corresponding service layer methods were moved here. I will remove unused interface methods.

components/director/internal/domain/formation/service.go
ivantenevvasilev: I presume none, while I was debugging before opening a PR I assumed the issue could be caused when trying to delete a scenario that doesn't exists, and tried to make the deletion more robust to find out if this is the cause. I seem to have forgotten to remove it.

components/director/internal/domain/runtime/resolver_test.go
```diff
@@ -3280,3 +3334,9 @@ func UnusedBundleInstanceAuthService() func() *automock.BundleInstanceAuthServic
 		return &automock.BundleInstanceAuthService{}
 	}
 }
+
+func UnusedFormationService() func() *automock.FormationService {
+	return func() *automock.FormationService {
```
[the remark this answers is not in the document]
ivantenevvasilev: I did it the same way the previously available methods were done to keep it consistent as they were all wrapped in a similar way.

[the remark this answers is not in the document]
ivantenevvasilev: I could change it, but I guess I'd need to change all unused mock creating functions to keep it consistent.

components/director/internal/domain/scenarioassignment/resolver_test.go
ivantenevvasilev: Will do, I removed the unused ones in the service_test.go, but initially I added all new mocks to see if and when they are used in each test instead of a nil exception and I forgot to remove these here. Thanks.

components/director/internal/domain/scenarioassignment/engine.go
ivantenevvasilev: I'm not sure whether or not they should be moved or where exactly those methods should reside, I'll think about what will need to change and what might occur before moving them because this could add more dependency issues as the methods are used in the runtime service. For now I'll remove the unused fields and start from there.

components/director/internal/domain/scenarioassignment/service.go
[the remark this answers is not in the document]
PetarTodorovv: My comment was about different methods that the ones you mentioned. I didn't find where `DeleteForScenarioName`, `List` and `Create` are used?

components/director/internal/domain/scenarioassignment/engine.go
ivantenevvasilev: If we move them to the runtime service, as that's where they're primarily used, that would force the formation service to depend on the runtime service, which might become a problem. But if it's moved to the formations service it would require the runtime service to depend on the formations service, which also could prove to be a problem in the future. Out of the two options, the latter makes more sense, but requires more changes. Leaving it as is will still probably require a change in the future.

DimitarPetrov: The runtime service will depend on the formation service anyways once we adapt the runtime mutations as part of a separate task. So, maybe we can move them in the formation service if there are no major obstacles right now.

components/director/internal/domain/scenarioassignment/service.go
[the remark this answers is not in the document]
ivantenevvasilev: I think I misunderstood the initial comment, sorry. Those methods in the interface don't seem to be used here as it was practically split up. Those that are unused here are used in the formations service now, so these methods in the repo interface are unnecessary and will remove them, but the code in the repository layer is still needed as it's now used in the formations service. Only used methods seem to be `ListForTargetTenant`, `GetForScenarioName`, and `List`. Thanks, I was going to miss this.

components/director/internal/domain/scenarioassignment/engine.go
ivantenevvasilev: I will start on moving them to the formations service right now in that case.

components/director/internal/domain/formation/service.go
```diff
@@ -180,8 +195,239 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+		matches, err := s.isASAMatchingRuntime(ctx, scenarioAssignment, runtimeID)
+		if err != nil {
+			return nil, errors.Wrapf(err, "while checkig if asa matches runtime with ID %s", runtimeID)
+		}
+		if matches {
+			matchingASAs = append(matchingASAs, scenarioAssignment)
+		}
+	}
+
+	for _, sa := range matchingASAs {
```
la4ezar, line 414: I know this is not your code, but is this loop actually needed? We can change the loop below to iterate through the `matchingASAs` and append the `sa.ScenarioName`. Also, the `scenariosSet` variable is not needed at all.

ivantenevvasilev: Looks to me as if it's been copied from the `MergeScenariosFromInputLabelsAndAssignments` method above and just adapted the variable names and in the current method the `scenarioSet` and the loop are not needed.

components/director/internal/domain/scenarioassignment/service.go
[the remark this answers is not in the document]
PetarTodorovv: No worries ;) But now it seems the `ListAll` is not used in the `scenarioassignment/service.go file` ?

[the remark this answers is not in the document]
ivantenevvasilev: After moving the rest of the methods for the `engine.go` it isn't used anymore. I just removed it from here, thanks.

PetarTodorovv (approved):

components/director/internal/domain/formation/service.go
```diff
@@ -180,8 +195,189 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+	}
+
+	for _, scenario := range scenariosFromAssignments {
+		scenariosSet[scenario] = struct{}{}
+	}
+
+	scenariosFromInput, isScenarioLabelInInput := inputLabels[model.ScenariosKey]
+
+	if isScenarioLabelInInput {
+		scenariosFromInputInterfaceSlice, ok := scenariosFromInput.([]interface{})
```
la4ezar, line 326: You can use the `label.ValueToStringsSlice` function.

components/director/internal/domain/formation/service.go
```diff
@@ -180,8 +195,189 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+	}
+
+	scenarioAssignments, err := s.repo.ListAll(ctx, tenantID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "while listinng Automatic Scenario Assignments in tenant: %s", tenantID)
+	}
+
+	matchingASAs := make([]*model.AutomaticScenarioAssignment, 0, len(scenarioAssignments))
+	for _, scenarioAssignment := range scenarioAssignments {
+		matches, err := s.isASAMatchingRuntime(ctx, scenarioAssignment, runtimeID)
```
la4ezar, line 359: `if matches, err := ...; err != nil {...} else if matches {...}`

components/director/internal/domain/formation/service.go
```diff
@@ -180,8 +195,189 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+		return errors.Wrap(err, "while unassigning scenario from runtimes")
+	}
+
+	return nil
+}
+
+// MergeScenariosFromInputLabelsAndAssignments merges all the scenarios that are part of the resource labels (already added + to be added with the current operation)
+// with all the scenarios that should be assigned based on ASAs.
+func (s *service) MergeScenariosFromInputLabelsAndAssignments(ctx context.Context, inputLabels map[string]interface{}, runtimeID string) ([]interface{}, error) {
+	scenariosSet := make(map[string]struct{})
```
la4ezar, line 312: Define `scenariosSet` after `scenariosFromAssignments` and with capacity = len(scenariosFromAssignments).

la4ezar (approved):

PR #2382, opened 2022-05-25
clean-up test formation after test

SvetlinZarev-SAP:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/tests/director/tests/application_api_test.go
+++ b/tests/director/tests/application_api_test.go
@@ -1586,6 +1586,7 @@ func TestMergeApplications(t *testing.T) {
 		var response graphql.Formation
 		deleteFormationReq := fixtures.FixDeleteFormationRequest(newFormation)
 		err = testctx.Tc.RunOperation(ctx, certSecuredGraphQLClient, deleteFormationReq, &response)
+		require.NoError(t, err)
 
 		t.Logf("Deleted formation with name: %s", response.Name)
 	}()
--- a/tests/director/tests/application_api_test.go
+++ b/tests/director/tests/application_api_test.go
@@ -1603,8 +1603,6 @@ func TestMergeApplications(t *testing.T) {
 		request := fixtures.FixUnassignFormationRequest(outputSrcApp.ID, "APPLICATION", newFormation)
 		var response graphql.Formation
 		err = testctx.Tc.RunOperation(ctx, certSecuredGraphQLClient, request, &response)
-		// require.NoError(t, err)
-		// require.Equal(t, newFormation, response.Name)
 		if nil == err {
 			t.Logf("Src-app was unassigned from formation %s", newFormation)
 		} else {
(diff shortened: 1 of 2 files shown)


tests/director/tests/application_api_test.go
```diff
@@ -1580,6 +1580,16 @@ func TestMergeApplications(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, newFormation, formation.Name)
 
+	defer func() {
+		t.Logf("Cleaning up formation: %s", newFormation)
+
+		var response graphql.Formation
+		deleteFormationReq := fixtures.FixDeleteFormationRequest(newFormation)
+		err = testctx.Tc.RunOperation(ctx, certSecuredGraphQLClient, deleteFormationReq, &response)
```
PetarTodorovv, line 1588: Add error check for the request execution - `require.NoError(t, err)` and you could also assert the formation name - `require.Equal(newFormation, response.Name)`

tests/director/tests/application_api_test.go
```diff
@@ -1580,6 +1580,16 @@ func TestMergeApplications(t *testing.T) {
...
+
+		t.Logf("Deleted formation: %v", response)
```
PetarTodorovv, line 1590: You could print only the formation name instead of the whole object 
```suggestion
		t.Logf("Deleted formation with name: %s", response.Name)
```

PetarTodorovv (dismissed):
Also, you need to change the "PR version" of the e2e tests in the values.yaml with `PR-2382` to include the new changes

tests/director/tests/application_api_test.go
```diff
@@ -1580,6 +1580,16 @@ func TestMergeApplications(t *testing.T) {
...
+	}()
+
 	t.Logf("Assign application to formation %s", newFormation)
 	assignReq := fixtures.FixAssignFormationRequest(outputSrcApp.ID, "APPLICATION", newFormation)
```
PetarTodorovv, line 1594: I believe you should add one more defer to unassign the application from formation

tests/director/tests/application_api_test.go
SvetlinZarev-SAP: done

tests/director/tests/application_api_test.go
SvetlinZarev-SAP: done

tests/director/tests/application_api_test.go
dzahariev: I think this is not required, as the formation will be deleted, but can be added just for completeness.
The similar flow can be seen here: https://github.com/kyma-incubator/compass/blob/0d0b8296aeb444f2625420b1e3a5c0fe46cfe0cf/tests/ord-service/tests/subscription_flow_test.go#L175
And explicit defer for un-assign was added.

SvetlinZarev-SAP: I've added defers for unassigning the apps from the formation, because the deletion did not work unless the apps were unassigned first.

dzahariev (approved):

dzahariev (approved):

PR #2384, opened 2022-05-26
Forbid merging of systems which represent SaaS Application subscriptions

dragobt:
**Description**

- add checks to see if an application is labeled with "xsappname"
- unit tests

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [ ] Integration tests

### DIFF
--- a/chart/compass/charts/postgresql/values-production.yaml
+++ b/chart/compass/charts/postgresql/values-production.yaml
@@ -14,7 +14,7 @@ global:
 image:
   registry: docker.io
   repository: bitnami/postgresql
-  tag: 11.4.0-debian-9-r12
+  tag: 11.15.0-debian-10-r93
   ## Specify a imagePullPolicy
   ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
--- a/chart/compass/charts/postgresql/values.yaml
+++ b/chart/compass/charts/postgresql/values.yaml
@@ -14,7 +14,7 @@ global:
 image:
   registry: docker.io
   repository: bitnami/postgresql
-  tag: 11.5.0-debian-9-r15
+  tag: 11.15.0-debian-10-r93
   ## Specify a imagePullPolicy
   ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
--- a/installation/scripts/prow/deploy-and-test.sh
+++ b/installation/scripts/prow/deploy-and-test.sh
@@ -36,9 +36,9 @@ set -- "${POSITIONAL[@]}" # restore positional parameters
 
 
 if [[ ${DUMP_DB} ]]; then
-    sudo ${INSTALLATION_DIR}/cmd/run.sh --dump-db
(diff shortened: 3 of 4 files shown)


components/director/internal/domain/application/service.go
```diff
@@ -40,6 +40,7 @@ const (
 	managedKey           = "managed"
 	subaccountKey        = "Subaccount"
 	locationIDKey        = "LocationID"
+	xsappnameKey         = "xsappname"
```
alextargov, line 43: Better use the env var APP_SELF_REGISTER_DISTINGUISH_LABEL_KEY because the key has a very specific SAP name. Also there is an override in the management-plane-config repo for this value.
When you change this to the respective env variable, the unit test will probably fail because the local value for the env variable is `subscriptionProviderId` or something like that - not sure.

components/director/internal/domain/application/service.go
```diff
@@ -764,6 +765,15 @@ func (s *service) Merge(ctx context.Context, destID, srcID string) (*model.Appli
 		return nil, errors.Errorf("Application templates are not the same. Destination app template: %s. Source app template: %s", destTemplateID, srcTemplateID)
 	}
 
+	_, ok := srcAppLabels[xsappnameKey]
```
alextargov, line 768: A cleaner approach would be 
`if _, ok := srcAppLabels[xsappnameKey]; ok {
}`

components/director/internal/domain/application/service.go
```diff
@@ -764,6 +765,15 @@ func (s *service) Merge(ctx context.Context, destID, srcID string) (*model.Appli
...
+	if ok {
+		return nil, errors.Errorf("Source app template: %s has label xsappname", srcTemplateID)
+	}
+	_, ok = destAppLabels[xsappnameKey]
```
alextargov, line 772: Probably we can add an e2e test that registers 2 systems with that label and tries to merge them. We can assert that the merge mutation has failed. Don't forget to use the label key from an env variable again.

alextargov (changes requested):

components/director/internal/domain/application/service.go
```diff
@@ -764,6 +767,13 @@ func (s *service) Merge(ctx context.Context, destID, srcID string) (*model.Appli
 		return nil, errors.Errorf("Application templates are not the same. Destination app template: %s. Source app template: %s", destTemplateID, srcTemplateID)
 	}
 
+	if _, exists := srcAppLabels[s.selfRegisterDistinguishLabelKey]; exists {
```
kaloyanspiridonov, line 770: I think that you have to check apptemplate labels instead of srcAppLabes?

NickyMateev: Indeed, the app template will be the one that can have the distinguished label, so you should check that instead of the srcApp/destApp labels.

alextargov (approved):

dzahariev (approved):

dzahariev (approved):

dzahariev (approved):

dzahariev (approved):

dzahariev (changes requested):
To be reviewed again. Seems like the merge is not done properly.

la4ezar (changes requested):
It looks like there was a wrong merge with the main. Everything from https://github.com/kyma-incubator/compass/pull/2386 is reverted.

NickyMateev (approved):

dzahariev:
Closed in flavour of: https://github.com/kyma-incubator/compass/pull/2415

PR #2392, opened 2022-06-01
Adapt Runtime label APIs to use new Formation APIs underneath

la4ezar:
**Description**
Adapt Runtime label APIs to use new Formation APIs underneath.

- Adapt the following: setRuntimeLabel, registerRuntime, unregisterRuntime, updateRuntime and deleteRuntimeLabel.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/runtime/service.go
```diff
@@ -525,39 +497,97 @@ func (s *service) ensureRuntimeExists(ctx context.Context, tnt string, runtimeID
...
+	currentScenariosSlice, err := label.ValueToStringsSlice(currentScenarios)
+	if err != nil {
+		return errors.Wrapf(err, "while converting current runtime scenarios: %+v to slice of strings", currentScenarios)
+	}
+
+	// Calculate the diff between current scenarios and merged scenarios.
+	// This modifyScenarios maps scenario to bool
+	// if scenario should be assigned we map it to 'true'
+	// if it should be unassigned we map it to 'false'
+	modifyScenarios := make(map[string]bool, 0)
```
ivantenevvasilev, line 567: `make(map[string]bool, len(mergedScenariosSlice))`

components/director/internal/domain/runtime/service.go
```diff
@@ -525,39 +497,97 @@ func (s *service) ensureRuntimeExists(ctx context.Context, tnt string, runtimeID
...
+	for _, scenario := range currentScenariosSlice {
+		if _, exists := modifyScenarios[scenario]; !exists {
+			modifyScenarios[scenario] = false
+		} else {
+			delete(modifyScenarios, scenario)
+		}
+	}
+
+	for scenario, modify := range modifyScenarios {
+		if modify == true {
```
ivantenevvasilev, line 582: if modify

components/director/internal/domain/runtime/service.go
```diff
@@ -525,39 +497,97 @@ func (s *service) ensureRuntimeExists(ctx context.Context, tnt string, runtimeID
...
+			if _, err = s.scenarioAssignmentEngine.AssignFormation(ctx, rtmTenant, rtmId, graphql.FormationObjectTypeRuntime, model.Formation{Name: scenario}); err != nil {
+				return errors.Wrapf(err, "while unassigning formation %q from runtime with ID %q", scenario, rtmId)
```
ivantenevvasilev, line 584: while assigning formation

components/director/internal/domain/runtime/service_test.go
```diff
@@ -226,6 +221,7 @@ func TestService_CreateWithMandatoryLabels(t *testing.T) {
 			},
 			EngineServiceFn: func() *automock.ScenarioAssignmentEngine {
 				svc := &automock.ScenarioAssignmentEngine{}
+				svc.On("AssignFormation", mock.Anything, tnt, runtimeID, graphql.FormationObjectTypeRuntime, model.Formation{Name: model.DefaultScenario}).Return(&model.Formation{Name: model.DefaultScenario}, nil)
```
ivantenevvasilev, line 224: You could use the same `Context` as the test case for it to be more precise in the mocking, as the tests use `ctxWithSubaccount`, `ctx` and others.

la4ezar: Not applicable for all tests, but I will change it where it's possible.

components/director/internal/domain/runtime/service.go
```diff
@@ -377,42 +369,30 @@ func (s *service) Delete(ctx context.Context, id string) error {
...
-	modifiable, err := isLabelModifiable(labelInput.Key, s.protectedLabelPattern, s.immutableLabelPattern)
-	if err != nil {
-		return err
-	}
-	if !modifiable {
-		return apperrors.NewInvalidDataError("could not set unmodifiable label with key %s", labelInput.Key)
-	}
-	if labelInput.Key != model.ScenariosKey {
+	if labelInput.Key == model.ScenariosKey {
+		if err := s.updateScenariosLabel(ctx, rtmTenant, id, map[string]interface{}{model.ScenariosKey: labelInput.Value}); err != nil {
```
PetarTodorovv, line 392: The `updateScenariosLabel` seems to work well for the `Update` method but I'm not sure about the `SetLabel` method. What is my concern:
From the old code, we have LabelInput with the new label value that is added to the `currentRuntimeLabels` resulting in `newRuntimeLabels`, and then `s.upsertScenariosLabelIfShould` was invoked with it(along with other parameters). Then `s.upsertScenariosLabelIfShould ` merge the provided `newRuntimeLabels` with those from the ASA and Upsert/Update them.

Now with the new changes, if I understand it correctly, we call `updateScenariosLabel` if the input label key is `scenarios` and provide the label value to the method. `updateScenariosLabel ` then get this input and merge it with the "scenarios from the ASA", called `mergedScenariosSlice` that will be used to invoke `AssignFormation` and for the other scenarios(`currentScenarios`), `UnassignFormation` will be invoked.

To me, it seems to have two different behaviors? Example:
Consider the following setup,
runtime with two scenarios(`scenario1` and `scenario2`) + ASA that will add another scenario - `asa-scenario`. And label value input - `input-label-val` as argument of the `SetLabel` method.
The "old SetLabel behavior" will get the `scenario1` and `scenario2` will add `input-label-val`, then call `upsertScenariosLabelIfShould` which will add the scenarios from the ASA - `asa-scenario` and finally will execute upsert, resulting in "scenarios": ["scenario1", "scenario2", "input-label-val", "asa-scenario"]

The "new SetLabel behavior" will call `updateScenariosLabel` with the label input value, provided to the `SetLabel` method - `input-label-val`, and from the explanation above for the `updateScenariosLabel ` to me it seems to be result in: "scenarios": ["input-label-val", "asa-scenario"]? The other two scenarios will be part of the `currentScenariosSlice` and call `UnassignFormation` for them? 

Am I missing something? I could be totally wrong, that's why I describe it in details. If needed we could have a quick conversation/sync

PetarTodorovv (changes requested):
Shouldn't we also modify `AddDefaultScenarioIfEnabled` to use the new AssignFormation API?

components/director/internal/domain/runtime/service.go
```diff
@@ -525,39 +490,97 @@ func (s *service) ensureRuntimeExists(ctx context.Context, tnt string, runtimeID
...
+	if !areCurrentScenariosInLabels {
+		currentScenarios = []interface{}{}
+	}
+
+	currentScenariosSlice, err := label.ValueToStringsSlice(currentScenarios)
+	if err != nil {
+		return errors.Wrapf(err, "while converting current runtime scenarios: %+v to slice of strings", currentScenarios)
+	}
+
+	// Calculate the diff between current scenarios and merged scenarios.
```
StanislavStefanov, line 556: Maybe it will be easier to follow the method if you create two maps:
currentScenarios map[string]struct{} and mergedScenarios map[string]struct{}
then you can range currentScenarios and call unassignFormation for each scenario that is not found in mergedScenarios. After that you can range mergedScenarios and call assignFormation for each scenario that is not foun in currentScenarios.
This will reduce the complexity and nesting a lot and will be more straightforward approach.

components/director/internal/domain/runtime/service.go
StanislavStefanov, line 392: you could just return s.updateScenariosLabel(...). Then 'if err' and the else clause wont be needed

components/director/internal/domain/runtime/service.go
```diff
@@ -261,6 +258,14 @@ func (s *service) CreateWithMandatoryLabels(ctx context.Context, in model.Runtim
 		in.Labels[key] = value
 	}
 
+	if scenarios, areScenariosInLabels := in.Labels[model.ScenariosKey]; areScenariosInLabels {
+		if err := s.assignRuntimeScenarios(ctx, rtmTenant, id, scenarios); err != nil {
+			return err
+		}
+
+		delete(in.Labels, model.ScenariosKey)
+	}
```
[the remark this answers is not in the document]
la4ezar: I assign them here because on the next line the input labels are upserted and we want to use the Formation API for the "scenarios" label modifications. Maybe we can assign them below(on line 296) with the labels that are coming from ASAs, but at least we need to store the scenarios that are coming from the input in a slice and delete them from the input labels so they will not end up being upserted.

StanislavStefanov (changes requested):

components/director/internal/domain/runtime/service.go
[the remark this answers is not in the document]
la4ezar: It is bad documentation. I think it will be better if I delete the whole comment because it is obvious what the code below is doing.

la4ezar: At first, this was my approach, but I came up with this second approach with the maps which I think is a little bit smarter, but if the code is not readable I will revert it.

components/director/internal/domain/runtime/service.go
```diff
@@ -261,6 +258,14 @@ func (s *service) CreateWithMandatoryLabels(ctx context.Context, in model.Runtim
 		in.Labels[key] = value
 	}
 
+	if scenarios, areScenariosInLabels := in.Labels[model.ScenariosKey]; areScenariosInLabels {
```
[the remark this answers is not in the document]
la4ezar: Okay, this comment is similar to Petar's comment, so I will do it this way.

components/director/internal/domain/runtime/service.go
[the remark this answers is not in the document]
la4ezar: Stani has a similar comment, so I will merge this code with the code below.

components/director/internal/domain/runtime/service.go
la4ezar: We've discussed it and also tested the example above with the 'old SetLabel behavior' and the result was `scenarios": ["input-label-val", "asa-scenario"]`.

components/director/internal/domain/runtime/service.go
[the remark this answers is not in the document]
la4ezar: After we discussed it, we can't assign all of the formations at once. The input formations must be assigned in the subaccount and the ASAs formations must be assigned in the global account.

components/director/internal/domain/runtime/service.go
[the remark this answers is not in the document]
la4ezar: After we discussed it, we can't assign all of the formations at once. The input formations must be assigned in the subaccount and the ASAs formations must be assigned in the global account.

StanislavStefanov (approved):

PetarTodorovv (approved):

PR #2393, opened 2022-06-01
Allow generating OTT for applications registered on subaccount level from Pairing Adapter

la4ezar:
**Description**
Allow generating OTT for applications registered on subaccount level from Pairing Adapter

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] [N/A] Integration tests


components/director/internal/domain/onetimetoken/service.go
```diff
@@ -236,13 +239,25 @@ func (s *service) getTokenFromAdapter(ctx context.Context, adapterURL string, ap
...
 	if len(extTenant) == 0 {
-		extTenant, err = s.extTenantsSvc.GetExternalTenant(ctx, tntCtx.InternalID)
-		if err != nil {
-			return nil, errors.Wrapf(err, "while getting external tenant for internal tenant [%s]", tntCtx.InternalID)
+		if tnt, err = s.extTenantsSvc.GetTenantByID(ctx, tntCtx.InternalID); err != nil {
+			return nil, errors.Wrapf(err, "while getting tenant with internal ID %q", tntCtx.InternalID)
+		}
+	} else {
+		if tnt, err = s.extTenantsSvc.GetTenantByExternalID(ctx, tntCtx.ExternalID); err != nil {
+			return nil, errors.Wrapf(err, "while getting tenant with external ID %q", tntCtx.ExternalID)
```
ivantenevvasilev, line 249: Is this if/else needed? We'll have the internal ID regardless and it'll be associated with an external tenant ID, so this branching of the code might not be needed and it'll make the mocking in the tests a bit easier, but in the end it's ok as is

components/director/internal/domain/onetimetoken/service.go
```diff
@@ -236,13 +239,25 @@ func (s *service) getTokenFromAdapter(ctx context.Context, adapterURL string, ap
...
 		}
 	}
 
+	if tnt.Type == tenantpkg.Subaccount {
+		if extTenant, err = s.extTenantsSvc.GetExternalTenant(ctx, tnt.Parent); err != nil {
+			return nil, errors.Wrapf(err, "while getting parent external tenant for internal tenant %q", tnt.Parent)
+		}
+	} else {
+		extTenant = tnt.ExternalTenant
```
ivantenevvasilev, line 258: Maybe set it by default to `tnt.ExternalTenant` above the check for subaccount? I think this might be a bit easier for readability.

StanislavStefanov (dismissed):

PetarTodorovv (changes requested):

PetarTodorovv (approved):

StanislavStefanov (approved):

PR #2396, opened 2022-06-03
Fix self reg flow for app templates

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


dzahariev (approved):

components/director/internal/domain/apptemplate/resolver.go
```diff
@@ -25,8 +25,8 @@ import (
 // ApplicationTemplateService missing godoc
 //go:generate mockery --name=ApplicationTemplateService --output=automock --outpkg=automock --case=underscore --disable-version-string
 type ApplicationTemplateService interface {
-	Create(ctx context.Context, in model.ApplicationTemplateInput) (string, error)
-	CreateWithLabels(ctx context.Context, in model.ApplicationTemplateInput, labels map[string]interface{}) (string, error)
+	Create(ctx context.Context, in model.ApplicationTemplateInput, selfRegID *string) (string, error)
```
NickyMateev, line 28: It kind of looks weird having this new custom ID parameter, especially with this cryptic name "selfRegID" and then passing nil whenever it is not needed (ns adapter).

How about adapting the `model.ApplicationTemplateInput` type to have a `ID *string` and assigning it to the selfRegID before calling the `ApplicationTemplateService`. Then when you reach the following inner logic:

```
appTemplateID := s.uidService.Generate()
	if len(str.PtrStrToStr(selfRegID)) > 0 {
		appTemplateID = *selfRegID
	}
```

Just adjust it to:
```
appTemplateID := s.uidService.Generate()
	if len(str.PtrStrToStr(in.ID)) > 0 {
		appTemplateID = *in.ID
	}
```

NickyMateev (changes requested):

NickyMateev (approved):

PR #2400, opened 2022-06-03
Fix director local setup

StefanCholakov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/domain/formation/resolver.go
```diff
@@ -26,22 +26,16 @@ type Converter interface {
 	ToGraphQL(i *model.Formation) *graphql.Formation
 }
 
-// TenantFetcher calls an API which fetches details for the given tenant from an external tenancy service, stores the tenant in the Compass DB and returns 200 OK if the tenant was successfully created.
-//go:generate mockery --name=TenantFetcher --output=automock --outpkg=automock --case=underscore --disable-version-string
-type TenantFetcher interface {
```
desislavaa, line 31: The tests were using the interfaces and if someone regenerates the mocks via the regenerate script, they will start failing. Or refactor the tests to not use them.

I think we follow the "define interface where used" almost everywhere in our project and this might be considered a deviation from our common practices, however it is a simple interface and we will save ourselves a little boilerplate.

I'm fine with both approaches, it's up to you :)

desislavaa (approved):

PR #2420, opened 2022-06-15
Introduce application namespace to AppTemplate

DimitarPetrov:
**Description**

- Introduce a new field in ApplicationTemplate entity for application namespace
- Expose the new field through graphQL API
- Add `region: default` label to all S/4HANA Cloud applications
- Rename ORD Labels field in Application entity to minimize confusion with normal compass lables.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/pkg/graphql/schema.graphql
```diff
@@ -314,6 +314,7 @@ input ApplicationTemplateInput {
 	applicationInput: ApplicationRegisterInput!
 	placeholders: [PlaceholderDefinitionInput!]
 	accessLevel: ApplicationTemplateAccessLevel!
+	applicationNamespace: String
```
NickyMateev, line 317: Seems like the new `applicationNamespace` property is added only to the graphql `input` types. Does the use case also require us to expose it for reading? This would require adding it to `type ApplicationTemplate` as well.

components/director/internal/model/application.go
```diff
@@ -20,7 +20,7 @@ type Application struct {
 	ApplicationTemplateID *string
 	SystemNumber          *string
 	BaseURL               *string         `json:"baseUrl"`
-	Labels                json.RawMessage `json:"labels"`
+	OrdLabels             json.RawMessage `json:"labels"`
```
NickyMateev, line 23: Shouldn't we also update the `json` tag?

NickyMateev: On second thought, this comment would also be addressed for the DB entity... but there I guess the tag is tightly coupled to the DB column, so we can't rename the tag there without renaming the column. So it would be strange if the tag here is changed whereas the tag in the db entity stays the same.

I guess either rename the DB column, or just leave the tags as is.

components/director/pkg/graphql/schema.graphql
DimitarPetrov: It is added in the type ApplicationTemplate as well. Maybe you missed it?

NickyMateev: ah, yes - my bad! 😄

NickyMateev (approved):

PR #2412, opened 2022-06-09
Introduce Formation Templates

ivantenevvasilev:
**Description**

Changes proposed in this pull request:
- introduce formation template table in database
- introduce CRUD operations for formation templates
- implement CRUD e2e tests for newly added GraphQL operations
- cover new code with unit tests
- introduce `formation_template:read` and `formation_template:write` scopes
- generate examples for new GraphQL operations

**Related issue(s)**
- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests
- [x] `chart/compass/values.yaml` is updated 
- [x] Mocks are regenerated, using the automated script

PR #2423, opened 2022-06-16
Introduce formation entity

PetarTodorovv:
**Description**
Introduce new Formation entity that will be used to persist/manage the formations we have. Currently they are modelled as labels that comes with their drawbacks.
Also, introducing new optional(for time being) parameter of the `CreateFormation` mutation containing the formation template name. There will be default value to preserve the backward compatibility in case no template name is provided

- introduce new DB entity - Formations
- CRUD API for the repo layer
- new optional(for time being) parameter of the `CreateFormation` mutation containing formation template name
- unit tests adaptation + introducing new ones

**Related issue(s)**
- https://github.com/kyma-incubator/compass/pull/2412

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/formation/service.go
```diff
@@ -128,23 +147,46 @@ func (s *service) GetFormationsForObject(ctx context.Context, tnt string, objTyp
...
+			return nil, err
 		}
+		if err = s.labelDefService.CreateWithFormations(ctx, tnt, []string{formationName}); err != nil {
+			return nil, err
+		}
+		// TODO:: Currently we need to support both mechanisms of formation creation/deletion(through label definitions and Formations entity) for backwards compatibility
+		if err = s.createFormation(ctx, tnt, *templateName, formationName); err != nil {
+			return nil, err
+		}
+		return &model.Formation{Name: formationName}, nil
```
ivantenevvasilev, line 164: What about the other new `ID`, `TenantID`, `FormationTemplateID` fields? Should they be added to the GQL schema and/or resolver and service?

components/director/internal/domain/formation/service.go
```diff
@@ -610,3 +652,25 @@ func (s *service) getAvailableScenarios(ctx context.Context, tenantID string) ([
...
+		FormationTemplateID: fTmpl.ID,
+		Name:                formationName,
+	}
+	log.C(ctx).Debugf("Creating formation with name: %q and template ID: %q...", formationName, fTmpl.ID)
+	if err = s.formationRepository.Create(ctx, formation); err != nil {
+		log.C(ctx).Errorf("An error occurred while creating formation with name: %q and template ID: %q", formationName, fTmpl.ID)
+		return errors.Wrapf(err, "An error occurred while reating formation with name: %q and template ID: %q", formationName, fTmpl.ID)
+	}
+
+	return nil
```
ivantenevvasilev, line 675: If the previous comment is applicable, it could return `(*Formation, error)`

components/director/internal/domain/labeldef/resolver.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+var defaultTemplateName = "Side-by-side extensibility with Kyma"
```
ivantenevvasilev, line 17: Could be moved to `internal/model/formation.go` in a similar way to the ScenarioKey in `scenarios.go`, but it's ok here for now.

components/director/internal/domain/formation/service.go
PetarTodorovv: I'm not sure it's in the scope of this task? We could have a quick discussion about that

components/director/internal/domain/formation/service.go
PetarTodorovv: answered above

PetarTodorovv: answered above

ivantenevvasilev (approved):

la4ezar (changes requested):

desislavaa (approved):

la4ezar (approved):

dzahariev (approved):

NickyMateev (approved):

PR #2433, opened 2022-06-22
Remove optional parameter from createFormation mutation and introduce unique constraint

PetarTodorovv:
**Description**

- Add unique constraint between name and tenant_id in the formation table

**Related issue(s)**
- https://github.com/kyma-incubator/compass/pull/2423

**Pull Request status**
- [x] Implementation
- [x] [N/A] Unit tests
- [x] [N/A] Integration tests

### DIFF
--- a/components/director/internal/domain/formation/service.go
+++ b/components/director/internal/domain/formation/service.go
@@ -158,11 +158,6 @@ func (s *service) CreateFormation(ctx context.Context, tnt string, formation mod
 		if err = s.labelDefService.CreateWithFormations(ctx, tnt, []string{formationName}); err != nil {
 			return nil, err
 		}
-		// TODO:: Currently we need to support both mechanisms of formation creation/deletion(through label definitions and Formations entity) for backwards compatibility
-		if err = s.createFormation(ctx, tnt, templateName, formationName); err != nil {
-			return nil, err
-		}
-		return &model.Formation{Name: formationName}, nil
 	}
 
 	// TODO:: Currently we need to support both mechanisms of formation creation/deletion(through label definitions and Formations entity) for backwards compatibility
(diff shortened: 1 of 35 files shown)


la4ezar (dismissed):

components/director/internal/domain/formation/service.go
```diff
@@ -159,14 +159,14 @@ func (s *service) CreateFormation(ctx context.Context, tnt string, formation mod
 			return nil, err
 		}
 		// TODO:: Currently we need to support both mechanisms of formation creation/deletion(through label definitions and Formations entity) for backwards compatibility
-		if err = s.createFormation(ctx, tnt, *templateName, formationName); err != nil {
+		if err = s.createFormation(ctx, tnt, templateName, formationName); err != nil {
```
StanislavStefanov, line 162: Can we remove lines 162-165? That way if everithing is fine with the labelDeffinition we will proceed with the code after the if statement. That way we will reduce the code duplication and the flow will be more straightforward.

PetarTodorovv: Currently the inner if(duplicated code) is needed because it cover the case when we have error that is `not found` and a few lines above we have `CreateWithFormations` method that creates the "old" formation and we need to double maintain(for time being) a formation entity as well. The outer if is when there is no errors from the `modifyFormations` method.

StanislavStefanov: If we do not return on line 165 we will proceed with the code on line 169 which is the same as line 162. If we remove lines 162-165 the flow will be - try to modify formations -> if notfound call create with formations and then create formation on line 169

PetarTodorovv: ah, to remove the return statement, yep, that will simplify the function. Initially I understand you to remove only the `s.createFormation` function

StanislavStefanov (approved):

PR #2438, opened 2022-06-22
Introduce Read Formations API

la4ezar:
**Description**
Introduce a dedicated API to read Formations.

- Add ID and FormationTemplateID in the Formation response
- Introduce a List with paging API
- Introduce a GetByID API
- Introduce new formation:read scope
- Adapt the unit tests and write new ones
- E2E tests for the Read API

**Related issue(s)**
- https://github.com/kyma-incubator/compass/pull/2423

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


tests/director/tests/formation_api_test.go
```diff
@@ -18,6 +18,72 @@ import (
 const assignFormationCategory = "assign formation"
 const unassignFormationCategory = "unassign formation"
 
+func TestGetFormation(t *testing.T) {
+	ctx := context.Background()
+	formationName := "formation1"
+
+	t.Logf("Should create formation: %q", formationName)
+	formation := fixtures.CreateFormation(t, ctx, certSecuredGraphQLClient, formationName)
+	require.Equal(t, formationName, formation.Name)
```
nyordanoff, line 27: This comparison is duplicated because the same one is made in the `fixtures.CreateFormation()`. I don't know if it is on purpose but probably we can leave it at one place only - doesn't matter to me where.

tests/director/tests/formation_api_test.go
```diff
@@ -18,6 +18,72 @@ import (
...
+	t.Logf("Should create formation: %q", secondFormationName)
+	secondFormation := fixtures.CreateFormation(t, ctx, certSecuredGraphQLClient, secondFormationName)
+	defer fixtures.DeleteFormation(t, ctx, certSecuredGraphQLClient, secondFormationName)
+
+	expectedFormations = 2
+	t.Logf("List should return %d formations", expectedFormations)
+	formationPage2 := fixtures.ListFormations(t, ctx, certSecuredGraphQLClient, first, expectedFormations)
+	require.NotNil(t, formationPage2)
+	assert.Equal(t, expectedFormations, formationPage2.TotalCount)
+	assert.Subset(t, formationPage2.Data, []*graphql.Formation{
```
nyordanoff, line 81: I think it would be better to use `assert.ElementsMatch(t, expected, actual)` instead of `assert.Subset()` because the `Subset()` doesn't check that the number of elements in both arrays is equal while `ElementsMatch()` does:

`ElementsMatch asserts that the specified listA(array, slice...) is equal to specified listB(array, slice...)`

For example, if we use `Subset()` and we expect only `firstFormation` and `secondFormation` to be present in `formationPage2.Data` but in `formationPage2.Data` there are X more formations this assert will pass because the two formations will be indeed present but we want  to be sure that only the two are returned.

nyordanoff (changes requested):

tests/director/tests/formation_api_test.go
la4ezar: Actually, on the line above(80) we check if their count is the same. And if their count is the same and one of the sets is a subset of the other one => they are equal, but I will change it because it's a little bit misleading.

components/director/internal/domain/formation/service.go
```diff
@@ -163,18 +194,18 @@ func (s *service) CreateFormation(ctx context.Context, tnt string, formation mod
...
 // DeleteFormation removes the provided formation from the scenario label definitions of the given tenant.
 // Also, removes the Formation entity from the DB
 func (s *service) DeleteFormation(ctx context.Context, tnt string, formation model.Formation) (*model.Formation, error) {
 	formationName := formation.Name
-	err := s.modifyFormations(ctx, tnt, formationName, deleteFormation)
+	if err := s.modifyFormations(ctx, tnt, formationName, deleteFormation); err != nil {
+		return nil, err
+	}
+
+	f, err := s.getFormationByName(ctx, formation.Name, tnt)
```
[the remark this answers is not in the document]
la4ezar: Because only the name is available. You can see that in the schema, `FormationInput` don't have the formation ID. And we want all of the functions for formations(create, delete, assign, unassign) to return the whole formation as a result, so we need to get the formation from the DB and the only property that we have available is the formation name.

NickyMateev (dismissed):

components/director/examples/merge-applications-with-self-register-distinguish-label-key/merge-applications-with-self-register-distinguish-label-key.graphql
```diff
@@ -0,0 +1,463 @@
+# Code generated by Compass integration tests, DO NOT EDIT.
```
NickyMateev, line 1: This mutation example is a duplicate of the existing `merge-applications` one. Why is it added?

nkanchev (dismissed):

components/director/examples/merge-applications-with-self-register-distinguish-label-key/merge-applications-with-self-register-distinguish-label-key.graphql
la4ezar: It was generated from generate_examples.sh

tests/director/tests/formation_api_test.go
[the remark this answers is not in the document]
la4ezar: outdated

tests/director/tests/formation_api_test.go
```diff
@@ -18,6 +18,72 @@ import (
...
+	firstFormationName := "formation1"
+	secondFormationName := "formation2"
+
+	first := 100
+
+	expectedFormations := 0
+	t.Logf("List should return %d formations", expectedFormations)
+	formationPage1 := fixtures.ListFormations(t, ctx, certSecuredGraphQLClient, first, expectedFormations)
+	require.NotNil(t, formationPage1)
+	require.Equal(t, expectedFormations, formationPage1.TotalCount)
```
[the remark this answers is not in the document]
la4ezar: The total count must be 0, so I don't know if we need to check the Data too, because they are related.

PetarTodorovv (approved):

nyordanoff (approved):

nkanchev (approved):

NickyMateev (approved):

PR #2445, opened 2022-06-24
SubscribeTenantToApplication should be Idempotent

dzahariev:
**Description**
- Create Test to verify idempotence of the call

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


alextargov (approved):

components/director/internal/domain/subscription/service.go
```diff
@@ -277,6 +277,18 @@ func (s *service) SubscribeTenantToApplication(ctx context.Context, providerID,
 		return false, errors.Wrapf(err, "while getting application template with filter labels %q and %q", providerID, region)
 	}
 
+	applications, err := s.appSvc.ListAll(ctx)
```
DimitarPetrov, line 280: Isn't this list in the provider tenant and the application that is result from the subscription should be in the consumer tenant?

PR #2450, opened 2022-06-27
Allow Application Providers to query individual SaaS consumer systems

alextargov:
**Description**

- enhance Applications query to support certificate flow (certificate + "user_context" header). If Applications query is executed in a certificate flow
    - It fetches model.ApplicationTemplate by labels: scenarioassignment.SubaccountIDKey, region and  selfRegisterDistinguishLabelKey. 
    - Lists all tenant header scoped []model.Application and finds the one created by the previously fetched App Template

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


NickyMateev (changes requested):
The more I looked at the newly introduced query the more it got me thinking that it won't do anything special than what the existing `applications` query already does with the only detail that it will take into account two tenants (provider & consumer) instead of one.

From this point of view I have a suggestion which will probably simplify things and I'd like to have your PoV on it.. **how about us not introducing a new query and instead just modify the existing `applications` query to consider when the flow is certificate based (i.e. has provider subaccount) and also a `Tenant` header is present which holds the consumer subaccount - even though the `applications` query usually returns multiple items, this will be a special case where it will always return just one application**. I think this would greatly simplify things for us and for consumers which in terms of keeping with a consistent API interface instead of introducing new standalone APIs for specific use-cases.

Haven't looked into the code implementation much and the implications, but I assume it will be less effort than having to introduce a new query and dedicated tests for it.

I assume that the `applications` query ought to fetch all applications in the context of the consumer subaccount first and then leave only the which is for the correct app template (the app template labeled with the provider subaccount).

What do you think?

components/director/internal/domain/application/resolver.go
```diff
@@ -238,6 +254,62 @@ func (r *Resolver) Application(ctx context.Context, id string) (*graphql.Applica
 	return r.appConverter.ToGraphQL(app), nil
 }
 
+// ApplicationForTenant could be executed only in a certificate flow. It fetches model.ApplicationTemplate by label scenarioassignment.SubaccountIDKey
+// with value equal to the consumer ID. Lists all tenant header scoped []model.Application and finds the one created by the previously fetched App Template
+func (r *Resolver) ApplicationForTenant(ctx context.Context) (*graphql.Application, error) {
```
NickyMateev, line 259: As discussed w/t the team - we want the logic of this function to be actually embedded in the existing `applications` query so that we don't introduce new standalone queries unnecessarily.

components/director/internal/domain/application/resolver.go
```diff
@@ -238,6 +254,62 @@ func (r *Resolver) Application(ctx context.Context, id string) (*graphql.Applica
...
+	consumerInfo, err := consumer.LoadFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if !consumerInfo.Flow.IsCertFlow() {
```
NickyMateev, line 265: In the spirit of the previous comment regarding moving this whole logic to the `applications` query - a similar check should be used to guard this logic from executing in the `applications` query resolver handler. Just one modification needed - don't just rely on checking that the flow is cert one (`consumerInfo.Flow.IsCertFlow()`) - the best way to guard this logic and be sure that this is a provider-consumer flow (i.e. two tenants are part of the tenant context) is to actually base the `if` check on `consumerInfo.OnBehalfOf != nil` - only when OnBehalfOf is existing can we be sure that there are two tenants at play (provider & consumer).

components/director/internal/domain/application/resolver.go
```diff
@@ -238,6 +254,62 @@ func (r *Resolver) Application(ctx context.Context, id string) (*graphql.Applica
...
+	tx, err := r.transact.Begin()
+	if err != nil {
+		return nil, err
+	}
+	defer r.transact.RollbackUnlessCommitted(ctx, tx)
+
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	filters := []*labelfilter.LabelFilter{
+		labelfilter.NewForKeyWithQuery(scenarioassignment.SubaccountIDKey, fmt.Sprintf("\"%s\"", consumerInfo.ConsumerID)),
```
NickyMateev, line 278: Also add `region` and `xsappname` as label filters here. To do that:
1. Add two more fields - Region and TokenClientID (it contains the xsappname) to the [Consumer type](https://github.com/kyma-incubator/compass/blob/896cfbcca39fdf402cb0d82118be2dbc841379f4/components/director/pkg/consumer/consumer.go#L28):
```
	Region        string                `json:"region"`
	TokenClientID string                `json:"tokenClientID"`
```
2. Make sure to assign it to the consumerInfo context before saving it to the context [HERE](https://github.com/kyma-incubator/compass/blob/8a386b80908a4b0d1c1c2d4384a837a4e455d9b2/components/director/internal/authenticator/claims/claims.go#L78)
3. Make use of it here (you have access to the consumerInfo object)

> Note: The `consumerInfo.TokenClientID` field is not exactly equal to the xsappname. The xsappname is part of the TokenClientID string - more specifically it's "<some-prefix>-<xsappname>". There is a configuration property for what this prefix is (it is known to us), so check [HERE](https://github.com/kyma-incubator/compass/blob/896cfbcca39fdf402cb0d82118be2dbc841379f4/components/director/internal/authenticator/claims/validator.go#L106) for reference. Only after you trim the prefix can you use the resulting string as the xsappname label filter value.

NickyMateev (changes requested):

NickyMateev (approved):

PR #2451, opened 2022-06-27
Subscription fixes

alextargov:
**Description**

- Use consumer tenant instead of the provider
**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


tests/ord-service/tests/subscription_flow_test.go
```diff
@@ -276,6 +276,9 @@ func TestConsumerProviderFlow(stdT *testing.T) {
 	})
 }
 
+// for localEnv
```
dzahariev, line 279: Look like TODO left in code?

tests/director/tests/application_templates_subscription_test.go
```diff
@@ -0,0 +1,110 @@
...
+		subscriptionProviderSubaccountID := conf.TestProviderSubaccountID
+		subscriptionConsumerSubaccountID := conf.TestConsumerSubaccountID
+		subscriptionConsumerTenantID := conf.TestConsumerTenantID
+
+		// Prepare provider external client certificate and secret and Build graphql director client configured with certificate
+		providerClientKey, providerRawCertChain := certprovider.NewExternalCertFromConfig(t, ctx, conf.ExternalCertProviderConfig)
+		directorCertSecuredClient := gql.NewCertAuthorizedGraphQLClientWithCustomURL(conf.DirectorExternalCertSecuredURL, providerClientKey, providerRawCertChain, conf.SkipSSLValidation)
+
+		// Create Application Template
+		appTemplateName := createAppTemplateName("app-template-name-subscription")
```
dzahariev, line 38: With e2e test the issue described above can be tested if you create more than one AppTemplate in the same region and verify that subscription is removed on unsubscribe, the expectation would be that the test will be unstable - will pass half of the times. To make it more unstable, you may create 10 or 20 templates in the same region as part of the test instead 2.

components/director/internal/domain/subscription/service.go
```diff
@@ -260,14 +260,10 @@ func (s *service) UnsubscribeTenantFromRuntime(ctx context.Context, providerID,
 }
 
 // SubscribeTenantToApplication fetches model.ApplicationTemplate by region and provider and registers an Application from that template
-func (s *service) SubscribeTenantToApplication(ctx context.Context, providerID, subscribedSubaccountID, providerSubaccountID, region, subscribedAppName string) (bool, error) {
-	providerInternalTenant, err := s.tenantSvc.GetInternalTenant(ctx, providerSubaccountID)
-	if err != nil {
-		return false, errors.Wrapf(err, "while getting provider subaccount internal ID: %q", providerSubaccountID)
+func (s *service) SubscribeTenantToApplication(ctx context.Context, providerID, subscribedSubaccountID, region, subscribedAppName string) (bool, error) {
+	filters := []*labelfilter.LabelFilter{
```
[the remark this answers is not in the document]
alextargov: I think there should be only one template registered per region. @NickyMateev can you give a feedback here?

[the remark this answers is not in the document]
NickyMateev: There's no use case where 1 SaaS app provider has multiple app templates registered in Compass for the same region.

However there's the use case where 2 SaaS app providers have registered their own single app template for one and the same region with Compass (they've onboarded with Compass essentially).

By the looks of the current label filter it looks to me that in such a use case the filter will return the templates for both SaaS app providers in that region, when we're only interested in one specifically (the one bound to the provider subaccount of the current request).

dzahariev (dismissed):

tests/director/tests/application_templates_subscription_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/kyma-incubator/compass/tests/pkg/gql"
+	"github.com/kyma-incubator/compass/tests/pkg/subscription"
+	"github.com/kyma-incubator/compass/tests/pkg/tenant"
+	"github.com/kyma-incubator/compass/tests/pkg/testctx"
+	testingx "github.com/kyma-incubator/compass/tests/pkg/testing"
+	"github.com/kyma-incubator/compass/tests/pkg/token"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSubscriptionApplicationTemplateFlow(stdT *testing.T) {
```
PetarTodorovv, line 25: That test only the "subscribe" scenario. Enhance a little bit the test at the end to make "unsubscribe" and to validate there is no application of the application template in the given tenant

PetarTodorovv (changes requested):

PetarTodorovv:
Also, a gentle reminder to update the description of the PR. And maybe it's not a good idea to link our internal repository to the open-source project?

tests/director/tests/application_templates_subscription_test.go
```diff
@@ -24,47 +24,93 @@ import (
...
+	defer func() {
+		if err := response.Body.Close(); err != nil {
+			t.Logf("Could not close response body %s", err)
+		}
+	}()
+	require.Equal(t, http.StatusOK, response.StatusCode)
 
-		appTmpl, err := fixtures.CreateApplicationTemplateFromInput(t, ctx, directorCertSecuredClient, tenant.TestTenants.GetDefaultTenantID(), appTemplateInput)
-		defer fixtures.CleanupApplicationTemplate(t, ctx, directorCertSecuredClient, tenant.TestTenants.GetDefaultTenantID(), &appTmpl)
+	t.Run("Subscribe tenant to Application flow", func(t *testing.T) {
```
PetarTodorovv, line 69: With the new approach, we made two subscriptions and have code duplication. We could either extract the subscribe logic as common and in the sub-tests to add the assertions or we could use the old setup just add the invoke the "unsubscribe" method - `BuildAndExecuteUnsubscribeRequest` and add assert that there is no application available

dzahariev (approved):

PetarTodorovv (approved):

PR #2456, opened 2022-06-29
Make DB dumps environment/DB specific

PetarTodorovv:
**Description**
Currently, the DB dump is one for all environments, and it's used in the `pre-main-compass-integration` prow job(and other places as well). But in the case of hotfix/fix when we have different DB states on the different environments this job fails.

- Make DB dump environment-specific

**Pull Request status**
- [x] Implementation
- [x] [N/A] Unit tests
- [x] [N/A] Integration tests


components/schema-migrator/run.sh
```diff
@@ -104,12 +104,13 @@ fi
...
-    echo "Will reuse existing dump in seeds/dump.sql"
-    PGPASSWORD="${DB_PASSWORD}" psql --username="${DB_USER}" --host="${DB_HOST}" --port="${DB_PORT}" --dbname="${DB_NAME}" --single-transaction -f seeds/dump.sql --set ON_ERROR_STOP=on --set "${SSL_OPTION}"
+DUMP_FILE=$(find seeds -type f -name "dump-*.sql")
+if [[ $DUMP_FILE ]] && [[ "${DIRECTION}" == "up" ]]; then
+    echo "Will reuse existing dump - $DUMP_FILE"
+    PGPASSWORD="${DB_PASSWORD}" psql --username="${DB_USER}" --host="${DB_HOST}" --port="${DB_PORT}" --dbname="${DB_NAME}" --single-transaction -f "$DUMP_FILE" --set ON_ERROR_STOP=on --set "${SSL_OPTION}"
 
     REMOTE_MIGRATION_VERSION=$(PGPASSWORD="${DB_PASSWORD}" psql -qtAX -U "${DB_USER}" -h "${DB_HOST}" -p "${DB_PORT}" -d "${DB_NAME}" -c "SELECT version FROM schema_migrations")
-    LOCAL_MIGRATION_VERSION=$(echo $(ls migrations/director | tail -n 1) | grep -o -E '[0-9]+' | head -1 | sed -e 's/^0\+//')
+    LOCAL_MIGRATION_VERSION=$(echo $(ls migrations/director | tail -n 1) | grep -o -E '^[0-9]+' | head -1 | sed -e 's/^0\+//')
```
nyordanoff, line 113: Could you please elaborate in what context this change was needed?

components/schema-migrator/update-expected-schema-version.sh
```diff
@@ -36,7 +36,7 @@ do
     esac
 done
 
-LATEST_SCHEMA=$(ls -lr migrations/${MIGRATION_PATH} | head -n 2 | tail -n 1 | tr -s ' ' | cut -d ' ' -f9 | cut -d '_' -f1)
+LATEST_SCHEMA=$(ls migrations/${MIGRATION_PATH} | tail -n 1 | grep -o -E '^[0-9]+' | sed -e 's/^0\+//')
```
nyordanoff, line 39: Could you please elaborate in what context this change was needed?

installation/cmd/run.sh
```diff
@@ -195,12 +196,35 @@ if [[ ${DUMP_DB} ]]; then
         sed -i 's/image\:.*compass-schema-migrator.*/image\: k3d-kyma-registry\:5001\/compass-schema-migrator\:'$DUMP_IMAGE_TAG'/' ${ROOT_PATH}/chart/compass/templates/update-expected-schema-version-job.yaml
     fi
 
+    REMOTE_VERSIONS=($(gsutil ls -R gs://sap-cp-cmp-dev-db-dump/ | grep -o -E '[0-9]+' | sed -e 's/^0\+//' | sort -r))
+    LOCAL_VERSIONS=($(ls migrations/director | grep -o -E '^[0-9]+' | sed -e 's/^0\+//' | sort -ru))
```
nyordanoff, line 200: I think we will have to pass the full path to the migrations dir because currently the `integration` [job](https://status.build.kyma-project.io/view/gs/kyma-prow-logs/pr-logs/pull/kyma-incubator_compass/2456/pre-main-compass-integration/1542053141107707904) fails with (line 908) : 
`DB dump will be used to prepopulate installation
ls: cannot access 'migrations/director': No such file or directory`

components/schema-migrator/update-expected-schema-version.sh
PetarTodorovv: If you trace what both command do they extract the latest schema version but the "old" one did it with a lot unnecessary "pipes" that didn't apply any changes and it was difficult to understand what every pipe do

components/schema-migrator/run.sh
PetarTodorovv: It's an improvement in the method of getting schema version from the local files. More specifically regex improvement that expects the file name to start with number

installation/cmd/run.sh
```diff
@@ -195,12 +196,36 @@ if [[ ${DUMP_DB} ]]; then
...
```
nyordanoff, line 200: By mistake this var was not deleted/edited and the job keeps failing with the same error because this var is processed first. As the correct syntax is in the var below, this one can be deleted.

nyordanoff (approved):

PR #2461, opened 2022-06-30
Fix subscription test 

alextargov:
**Description**
This E2e should be skipped on dev-main env but it is not. It is probably due to using the `T` param in the test suites.
- Wrap tests with a "dummy" test run so the T param can be properly used
**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/tests/director/tests/application_templates_subscription_test.go
+++ b/tests/director/tests/application_templates_subscription_test.go
@@ -24,7 +24,7 @@ import (
 
 func TestSubscriptionApplicationTemplateFlow(stdT *testing.T) {
 	t := testingx.NewT(stdT)
-	t.Run("Success", func(t *testing.T) {
+	t.Run("When creating app template with a certificate", func(t *testing.T) {
 		// GIVEN
 		ctx := context.Background()
 
--- a/tests/director/tests/application_templates_subscription_test.go
+++ b/tests/director/tests/application_templates_subscription_test.go
@@ -24,87 +24,89 @@ import (
 
 func TestSubscriptionApplicationTemplateFlow(stdT *testing.T) {
 	t := testingx.NewT(stdT)
-	ctx := context.Background()
-
-	// GIVEN
-	subscriptionProviderSubaccountID := conf.TestProviderSubaccountID
-	subscriptionConsumerSubaccountID := conf.TestConsumerSubaccountID
-	subscriptionConsumerTenantID := conf.TestConsumerTenantID
-
-	// Prepare provider external client certificate and secret and Build graphql director client configured with certificate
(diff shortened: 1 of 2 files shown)


tests/director/tests/application_templates_subscription_test.go
```diff
@@ -24,87 +24,89 @@ import (
...
-	subscriptionConsumerSubaccountID := conf.TestConsumerSubaccountID
-	subscriptionConsumerTenantID := conf.TestConsumerTenantID
-
-	// Prepare provider external client certificate and secret and Build graphql director client configured with certificate
-	providerClientKey, providerRawCertChain := certprovider.NewExternalCertFromConfig(stdT, ctx, conf.ExternalCertProviderConfig)
-	directorCertSecuredClient := gql.NewCertAuthorizedGraphQLClientWithCustomURL(conf.DirectorExternalCertSecuredURL, providerClientKey, providerRawCertChain, conf.SkipSSLValidation)
-
-	apiPath := fmt.Sprintf("/saas-manager/v1/application/tenants/%s/subscriptions", subscriptionConsumerTenantID)
-	subscriptionToken := token.GetClientCredentialsToken(stdT, ctx, conf.SubscriptionConfig.TokenURL+conf.TokenPath, conf.SubscriptionConfig.ClientID, conf.SubscriptionConfig.ClientSecret, "tenantFetcherClaims")
+	t.Run("Success", func(t *testing.T) {
```
PetarTodorovv, line 27: It has to be not a common name here because other tests that have skip patterns could match and skip them unintentionally. Also, change the regex pattern in our internal repo as well

PetarTodorovv (changes requested):

tests/director/tests/application_templates_subscription_test.go
alextargov: Why unique? If someone would create a regex for his test, he'd probably use a pattern like `MyRootTest_Success` or something like that.
Also the general regex in the internal repo would catch the whole test suite - why changing it?

PetarTodorovv: About the first thing, it's hard to predict what someone will do or will not do. And creating it with a common name we risk skipping other tests which I think is not what we want, also, it will be very hard for us to notice some tests are skipped. However, the "skip mechanism" that we have just does not work that way. It aims to match and skip a "test run" not a "test suite" you're creating and the name of the test suite should not be included in the skip regex pattern. So if you do not change the regex in our internal repo I think the result will be the same and that change will be obsolete. You could drill down to double-check that.

alextargov: image
A local test run with a hardcoded pattern. Seems like the test suite is skipped after all.
I can change the sub-test name, tho.

PetarTodorovv (approved):

PR #2465, opened 2022-07-01
Check for provider runtime existence before creating a new one

PetarTodorovv:
**Description**
Now we can self-register more than one provider runtime with the same labels whereas we shouldn't be able to.

- Check if provider runtime with specified labels already exists before creating it

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


PetarTodorovv (commented):
Looks okay to me ✅
While I reviewing it I noticed a leftover here. If you do not want to wait the tests we could remove it in separate PR.
https://github.com/kyma-incubator/compass/blob/d0bd5a4a7fdedb191bc9382e5e5123ea0a9bfeb9/components/director/internal/domain/runtime/service_test.go#L47-L48

PetarTodorovv:
Found one more place - https://github.com/kyma-incubator/compass/blob/973582f57c84d1ece38bc7fa7befaf4133f57412/components/director/internal/domain/runtime/resolver.go#L873

PR #2474, opened 2022-07-06
Allow one self registration for app templates

alextargov:
**Description**
At the moment, multiple self-registration for app templates for the same region and distinguished label is allowed. 
- Restrict self-registration for application templates to only one for the pair: region-distinguished label

**Related issue(s)**

- https://github.com/kyma-incubator/compass/pull/2465

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


tests/director/tests/application_templates_api_test.go
```diff
@@ -24,42 +24,80 @@ import (
...
+		require.NoError(t, err)
+		require.NotEmpty(t, output1.ID)
+
+		// WHEN
+		t.Log("Create second application template")
+		err = testctx.Tc.RunOperationNoTenant(ctx, certSecuredGraphQLClient, createApplicationTemplateRequest2, &output2)
+		defer fixtures.CleanupApplicationTemplate(t, ctx, certSecuredGraphQLClient, tenant.TestTenants.GetDefaultTenantID(), &output2)
+
+		//THEN
+		require.Error(t, err)
```
PetarTodorovv, line 98: Assert the error contains the message from the productive code(with the labels) to ensure it's not some other error

components/director/internal/domain/apptemplate/resolver.go
```diff
@@ -564,3 +569,31 @@ func validateAppTemplateForSelfReg(name string, placeholders []model.Application
...
+		filters := []*labelfilter.LabelFilter{
+			labelfilter.NewForKeyWithQuery(distinguishLabelKey, fmt.Sprintf("\"%s\"", distinguishLabelValue)),
+			labelfilter.NewForKeyWithQuery(regionLabelKey, fmt.Sprintf("\"%s\"", region)),
+		}
+
+		log.C(ctx).Infof("Getting application template for labels %q: %q and %q: %q", regionLabelKey, region, distinguishLabelKey, distinguishLabelValue)
+		appTemplate, err := r.appTemplateSvc.GetByFilters(ctx, filters)
+		if err != nil && !apperrors.IsNotFoundError(err) {
+			return errors.Wrap(err, fmt.Sprintf("Failed to get application template for labels %q: %q and %q: %q", regionLabelKey, region, distinguishLabelKey, distinguishLabelValue))
+		}
```
PetarTodorovv, line 590: Add unit test to cover the scenario when we have an error and it's different from `not found`

PetarTodorovv (dismissed):

NickyMateev (commented):
LGTM, can you just rename `fixAppTemplateInput` to `fixAppTemplateInputWithDefaultRegionAndDistinguishLabel` in `tests/director/tests/application_templates_api_test.go`?

I know it's longer, but it's more descriptive of what the function does. I always seem to forget that all e2e tests now run w/t certificates which makes causes the complication where all app template e2e tests have to basically do a self reg app templates (i.e. one which has region + distinguished label). So the `fixAppTemplateInput` function doesn't imply that in any way and you're left with the impression that regular global template is just created.

NickyMateev (approved):

PetarTodorovv (dismissed):

PetarTodorovv (approved):

PR #2475, opened 2022-07-06
Respect subscriptions in Application Provider Metadata Push APIs

dzahariev:
**Description**

- Respect subscriptions in Application Provider Metadata Push APIs by extending the existing Claims Validator to also look for Application Provider subscriptions during double authentication requests.

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/cmd/director/main.go
```diff
@@ -660,26 +661,104 @@ func runtimeCtxSvc(cfg config) claims.RuntimeCtxService {
...
+	labelConverter := label.NewConverter()
+	labelRepo := label.NewRepository(labelConverter)
+	labelDefConverter := labeldef.NewConverter()
+	labelDefRepo := labeldef.NewRepository(labelDefConverter)
+	labelSvc := label.NewLabelService(labelRepo, labelDefRepo, uidSvc)
+
+	return apptemplate.NewService(appTemplateRepo, webhookRepo, uidSvc, labelSvc, labelRepo)
+}
+
+func applicaitonSvc(cfg config, httpClient *http.Client, certCache certloader.Cache) claims.ApplicationService {
```
NickyMateev, line 691: fix typo `applicaitonSvc` -> `applicationSvc`

components/director/internal/authenticator/claims/validator.go
```diff
@@ -153,6 +177,77 @@ func (v *validator) validateRuntimeConsumer(ctx context.Context, claims Claims)
...
+		log.C(ctx).WithError(err).Errorf("Error while getting application template in provider tenant %s for labels %s: %s and %s: %s: %v", providerInternalTenantID, tenant.RegionLabelKey, claims.Region, v.subscriptionProviderLabelKey, tokenClientID, err)
+		return errors.Wrapf(err, "failed to get application template in tenant %s for labels %s: %s and %s: %s", providerInternalTenantID, tenant.RegionLabelKey, claims.Region, v.subscriptionProviderLabelKey, tokenClientID)
+	}
+	log.C(ctx).Infof("Found application template with ID %q in provider tenant %s for labels %s: %s and %s: %s", applicationTemplate.ID, providerInternalTenantID, tenant.RegionLabelKey, claims.Region, v.subscriptionProviderLabelKey, tokenClientID)
+
+	consumerInternalTenantID := claims.Tenant[tenantmapping.ConsumerTenantKey]
+	consumerExternalTenantID := claims.Tenant[tenantmapping.ExternalTenantKey]
+	ctxWithConsumerTenant := tenant.SaveToContext(ctx, consumerInternalTenantID, consumerExternalTenantID)
+
+	applicatinsFilter := []*labelfilter.LabelFilter{
```
NickyMateev, line 221: I think this filter is unnecessary

components/director/internal/authenticator/claims/validator.go
```diff
@@ -153,6 +177,77 @@ func (v *validator) validateRuntimeConsumer(ctx context.Context, claims Claims)
...
+		labelfilter.NewForKeyWithQuery(v.consumerSubaccountLabelKey, fmt.Sprintf("\"%s\"", consumerExternalTenantID)),
+	}
+
+	found := false
+
+	log.C(ctx).Infof("Listing applications in the consumer tenant %q for application template with ID: %q and label with key: %q and value: %q", consumerExternalTenantID, applicationTemplate.ID, v.consumerSubaccountLabelKey, consumerExternalTenantID)
+	applicationsPage, err := v.applicationSvc.List(ctxWithConsumerTenant, applicatinsFilter, 100, "")
```
NickyMateev, line 228: Can you increase the page size to 200 here, so that we postpone running into the issue of missing out some application due to us not going over all pages

NickyMateev: btw, you can do the same for the `validateRuntimeConsumer` function

components/director/internal/authenticator/claims/validator.go
dzahariev: Without it will look for apps in all subaccounts instead only in subscriber subaccount only. 
The idea is to get all apps in this subaccount and go over them to look for app that is based on the template from the provider.

NickyMateev:
Let's not merge this until the release has passed successfully.

alextargov (dismissed):

desislavaa (approved):

components/director/internal/authenticator/claims/validator.go
```diff
@@ -145,6 +169,73 @@ func (v *validator) validateRuntimeConsumer(ctx context.Context, claims Claims)
...
+	providerInternalTenantID := claims.Tenant[tenantmapping.ProviderTenantKey]
+	providerExternalTenantID := claims.Tenant[tenantmapping.ProviderExternalTenantKey]
+	ctxWithProviderTenant := tenant.SaveToContext(ctx, providerInternalTenantID, providerExternalTenantID)
+
+	log.C(ctx).Infof("Get application template in provider tenant %s for labels %s: %s and %s: %s", providerInternalTenantID, tenant.RegionLabelKey, claims.Region, v.subscriptionProviderLabelKey, tokenClientID)
+	applicationTemplate, err := v.appTemplateSvc.GetByFilters(ctxWithProviderTenant, filters)
+	if err != nil {
+		log.C(ctx).WithError(err).Errorf("Error while getting application template in provider tenant %s for labels %s: %s and %s: %s: %v", providerInternalTenantID, tenant.RegionLabelKey, claims.Region, v.subscriptionProviderLabelKey, tokenClientID, err)
+		return errors.Wrapf(err, "failed to get application template in tenant %s for labels %s: %s and %s: %s", providerInternalTenantID, tenant.RegionLabelKey, claims.Region, v.subscriptionProviderLabelKey, tokenClientID)
+	}
```
PetarTodorovv, line 206: Something very small - this `if` seems not covered with test. Also, the if below when we call `v.applicationSvc.ListAll`

tests/director/tests/application_templates_subscription_test.go
```diff
@@ -88,51 +88,37 @@ func TestSubscriptionApplicationTemplateFlow(stdT *testing.T) {
...
 		})
 
-		t.Run("Subscribe tenant to Application flow and add bundle", func(t *testing.T) {
+		t.Run("Application is deleted successfully in consumer subaccount as a result of unsubscription", func(t *testing.T) {
 			//GIVEN
 			subscriptionToken := token.GetClientCredentialsToken(t, ctx, conf.SubscriptionConfig.TokenURL+conf.TokenPath, conf.SubscriptionConfig.ClientID, conf.SubscriptionConfig.ClientSecret, "tenantFetcherClaims")
-
-			// Subscribe
 			createSubscription(t, ctx, httpClient, appTmpl, apiPath, subscriptionToken, subscriptionConsumerTenantID, subscriptionConsumerSubaccountID, subscriptionProviderSubaccountID)
-			defer subscription.BuildAndExecuteUnsubscribeRequest(t, appTmpl.ID, appTmpl.Name, httpClient, conf.SubscriptionConfig.URL, apiPath, subscriptionToken, conf.SubscriptionConfig.PropagatedProviderSubaccountHeader, subscriptionConsumerSubaccountID, subscriptionConsumerTenantID, subscriptionProviderSubaccountID)
```
PetarTodorovv: Most probably you've removed this defer because you explicitly execute unsubscribe?
But in case something fails after the subscription and before the explicit unsubscribe request(get apps) we'll have leftover resources as part of the subscription.
So maybe it's safer to have a defer for such cases and it should not fails because currently `subscription.BuildAndExecuteUnsubscribeRequest` is implemented not to fail if the subscription is not found for such cases and cleanup

NickyMateev: Mistake 😄

PetarTodorovv (approved):

NickyMateev (approved):

desislavaa (approved):

PR #2478, opened 2022-07-07
Respect runtime types in ASAs

la4ezar:
**Description**
When processing ASAs we need to respect the runtime types in the formation template and assign to the formation only the runtimes (or runtime contexts of that runtimes) that are of the type provided in the formation template.
If we have a runtime(even if the runtime has owner=true) with runtime contexts, only runtime contexts must be assigned to the formation.

Changes proposed in this pull request:
- Respect runtime types in ASAs
- Unit tests
- E2E test

**Related issue(s)**

- https://github.com/kyma-incubator/compass/pull/2434

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests
- [x] `chart/compass/values.yaml` is updated 
- [x] Mocks are regenerated, using the automated script

PR #2505, opened 2022-07-27
Fix runtime/runtime contexts ASA processing

PetarTodorovv:
- Fix regarding runtime context ASA matching
- Additional logs

**Related issue(s)**
- https://github.com/kyma-incubator/compass/pull/2478

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


tests/director/tests/formation_api_test.go
```diff
@@ -703,6 +703,131 @@ func TestRuntimeContextsFormationProcessingFromASA(stdT *testing.T) {
...
+		kymaFormationName := "kyma-formation-name"
+		fixtures.CreateFormationWithinTenant(t, ctx, certSecuredGraphQLClient, subscriptionConsumerAccountID, kymaFormationName, &kymaFormationTmplName)
+		defer fixtures.DeleteFormationWithinTenant(t, ctx, certSecuredGraphQLClient, subscriptionConsumerAccountID, kymaFormationName)
+
+		// Create provider formation
+		providerFormationName := "provider-formation-name"
+		fixtures.CreateFormationWithinTenant(t, ctx, certSecuredGraphQLClient, subscriptionConsumerAccountID, providerFormationName, &providerFormationTmplName)
+		defer fixtures.DeleteFormationWithinTenant(t, ctx, certSecuredGraphQLClient, subscriptionConsumerAccountID, providerFormationName)
+
+		t.Run("Register kyma runtime and validate scenarios label", func(t *testing.T) {
```
DimitarPetrov, line 740: Register also the provider runtime and check that runtime context will not be assigned in this case.

That way you will catch the problem we have with the runtime context matching right now.

tests/director/tests/formation_api_test.go
```diff
@@ -703,6 +703,131 @@ func TestRuntimeContextsFormationProcessingFromASA(stdT *testing.T) {
...
+			// Register kyma runtime
+			kymaRtmInput := fixtures.FixRuntimeRegisterInput("kyma-runtime")
+			kymaRuntime := registerKymaRuntime(t, ctx, subscriptionConsumerSubaccountID, kymaRtmInput)
+			defer fixtures.CleanupRuntime(t, ctx, certSecuredGraphQLClient, subscriptionConsumerSubaccountID, &kymaRuntime)
+
+			t.Logf("Assert scenarios label exists after assigning tenant with ID: %q to formation", subscriptionConsumerSubaccountID)
+			checkRuntimeFormationLabelsExists(t, ctx, subscriptionConsumerAccountID, kymaRuntime.ID, "scenarios", []string{kymaFormationName})
+		})
+
+		t.Run("Register provider runtime and validate scenarios label", func(t *testing.T) {
```
DimitarPetrov, line 754: Register also Kyma and validate it is not assigned in this case

components/director/internal/domain/formation/service.go
```diff
@@ -561,29 +564,7 @@ func (s *service) isASAMatchingRuntime(ctx context.Context, asa *model.Automatic
...
-		if err != nil {
-			return false, errors.Wrapf(err, "while getting runtime contexts for runtime with id %q", r.ID)
-		}
-		if hasRuntimeContext {
-			return true, nil
-		}
-	}
-
-	return false, nil
+	return s.runtimeContextRepo.Exists(ctx, asa.TargetTenantID, runtimeContextID)
```
DimitarPetrov, line 567: This should be type dependant. Already discussed it offline.

DimitarPetrov (changes requested):

tests/director/tests/formation_api_test.go
DimitarPetrov: You can structure the test differently to avoid so much dumplication.

First assign tenant to both formations then register both runtimes and verify that kyma has formation1 and do not have formation2 and provider runtime has formation2 and do not have formation1

DimitarPetrov (approved):

PR #2518, opened 2022-08-02
Fix ord aggregator e2e test.

la4ezar:
**Description**
Fix ord aggregator e2e test.

**Related issue(s)**
- [HOTFIX] https://github.com/kyma-incubator/compass/pull/2519

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/tests/ord-aggregator/tests/handler_test.go
+++ b/tests/ord-aggregator/tests/handler_test.go
@@ -619,6 +619,14 @@ func TestORDAggregator(stdT *testing.T) {
 		oauthCredentialData, ok := intSystemCredentials.Auth.Credential.(*directorSchema.OAuthCredentialData)
 		require.True(t, ok)
 
+		subJobStatusPath := resp.Header.Get(subscription.LocationHeader)
+		require.NotEmpty(t, subJobStatusPath)
+		subJobStatusURL := testConfig.SubscriptionConfig.URL + subJobStatusPath
+		require.Eventually(t, func() bool {
+			return subscription.GetSubscriptionJobStatus(t, httpClient, subJobStatusURL, subscriptionToken) == subscription.JobSucceededStatus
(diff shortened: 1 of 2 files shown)


tests/ord-aggregator/tests/handler_test.go
```diff
@@ -619,6 +619,14 @@ func TestORDAggregator(stdT *testing.T) {
...
 		require.True(t, ok)
 
+		subJobStatusPath := resp.Header.Get(subscription.LocationHeader)
+		require.NotEmpty(t, subJobStatusPath)
+		subJobStatusURL := testConfig.SubscriptionConfig.URL + subJobStatusPath
+		require.Eventually(t, func() bool {
+			return subscription.GetSubscriptionJobStatus(t, httpClient, subJobStatusURL, subscriptionToken) == subscription.JobSucceededStatus
+		}, subscription.EventuallyTimeout, subscription.EventuallyTick)
+		t.Logf("Successfully created subscription between consumer with subaccount id: %q and tenant id: %q, and provider with name: %q, id: %q and subaccount id: %q", subscriptionConsumerSubaccountID, subscriptionConsumerTenantID, appTemplate.Name, appTemplate.ID, subscriptionProviderSubaccountID)
+
```
PetarTodorovv, line 629: It seems to me that this subscription job status check needs to be even earlier - after the subscription call. Because now between the subscription call and the job status validation we have some other operations, like integration system creation?

Also, this test could be revisited again because a few lines above we have subscription request executed and defer before the `Accepted` status - https://github.com/kyma-incubator/compass/pull/2518/files#diff-82649b741f4d472b7d81e1975fc00697ac52b7a66e5f19594c85bd5d3b5db976R599

kaloyanspiridonov (approved):

PR #2521, opened 2022-08-03
Consumer context provider

PetarTodorovv:
**Description**
Introduce a new object context provider that is similar to the `authenticator_context_provider` but instead of authenticator and JWT token, the necessary information will be provided within `user_context` header

- Introduce a new object context provider in the hydrator component
- Unit and E2E tests adjustments

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/hydrator/internal/tenantmapping/consumer_context_provider_test.go
```diff
@@ -0,0 +1,399 @@
...
+			objectCtx, err := provider.GetObjectContext(ctx, testCase.ReqDataInput, authDetails)
+
+			// THEN
+			if testCase.ExpectedErrMsg == "" {
+				require.NoError(t, err)
+				require.Equal(t, consumer.User, objectCtx.ConsumerType)
+				require.Equal(t, authID, objectCtx.ConsumerID)
+				require.Equal(t, clientID, objectCtx.OauthClientID)
+				require.Equal(t, oathkeeper.ConsumerProviderFlow, objectCtx.AuthFlow)
+				//require.Equal(t, consumerInternalTenantID, objectCtx.TenantContext.TenantID)
```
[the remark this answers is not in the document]
PetarTodorovv: Yep, I forgot to adapt this one.

tests/director/tests/application_templates_subscription_test.go
```diff
@@ -125,7 +138,8 @@ func TestSubscriptionApplicationTemplateFlow(baseT *testing.T) {
 
 			// THEN
 			consumerToken := token.GetUserToken(t, ctx, conf.ConsumerTokenURL+conf.TokenPath, conf.ProviderClientID, conf.ProviderClientSecret, conf.BasicUsername, conf.BasicPassword, "subscriptionClaims")
-			headers := map[string][]string{subscription.AuthorizationHeader: {fmt.Sprintf("Bearer %s", consumerToken)}}
+			consumerClaims := token.FlattenTokenClaims(stdT, consumerToken)
+			headers := map[string][]string{subscription.UserContextHeader: {consumerClaims}}
```
nyordanoff, line 142: Currently, all app_template tests are adapted to send the new header instead of the whole token? Shouldn't we test both calls (with header & with token) separately or we decided to adapt and test only the header calls?

tests/pkg/token/token_utils.go
```diff
@@ -202,3 +204,48 @@ func GetTokenWithClient(t *testing.T, ctx context.Context, client *http.Client,
...
+	require.Equal(t, 3, len(tokenParts), "invalid token format")
+	payload := tokenParts[1]
+
+	b, err := base64.RawURLEncoding.DecodeString(payload)
+	require.NoError(t, err)
+
+	return string(b)
+}
+
+func flatten(m map[string]interface{}) map[string]interface{} {
```
nyordanoff, line 237: Just double check question - have you validated if that func works as expected with some test/dummy input or it is tested properly across the tests?

components/hydrator/pkg/oathkeeper/model.go
```diff
@@ -46,6 +48,8 @@ const (
 	ClientIDTokenKey = "client-id-from-token"
 	// ExternalTenantKey missing godoc
 	ExternalTenantKey = "tenant"
+	// UserContextKey is a header key containing consumer data
+	UserContextKey = "User_context"
```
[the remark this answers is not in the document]
PetarTodorovv: Initially, it was with lower `u` but I guess the oathkeeper makes the first letter uppercase so to successfully get the header I have to change it.

components/director/pkg/graphql/schema.graphql
```diff
@@ -1204,6 +1204,7 @@ type Tenant {
 	parentID: ID
 	initialized: Boolean
 	labels(key: String): Labels
+	provider: String!
```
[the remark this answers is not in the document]
PetarTodorovv: Currently, in the hydrator component, the new object context provider will check if the tenant has a subdomain label and if it does not exist will update it through `writeTenants` mutation, which will update the tenant with whatever we provide as parameters. So we need to be sure the `getTenants` query returns everything that `writeTenants` accepts as parameters to avoid overriding the tenant fields with wrong information.

tests/director/tests/application_templates_subscription_test.go
PetarTodorovv: The applicatio template will be using the new functionality with the `user_context` header and that's why we're testing it only with the header. On the other hand in ord-service tests, we have a "subscription" test with cert + token, and that way we cover both approaches with the e2e tests. But to be more robust/confident in our tests, there is one more test case that uses cert + user_context header to be sure we cover both approaches in "our" tests and we're not relying only on app template tests

tests/pkg/token/token_utils.go
PetarTodorovv: It is used in the newly introduced tests(or the adapted ones) and the ext-svc-mock returns a token that has nested fields that should be validated with this function.

nyordanoff (approved):

nyordanoff (approved):

PR #2525, opened 2022-08-04
Support more than one locality in cert-rotation job

la4ezar:
**Description**
We want to allow more than one locality to be configured as expected in the certificate-rotation job and fail only if the locality of the issuer is neither of them.

- Allow more than one locality to be configured as expected in the job and fail only if the locality of the issuer is neither of them.

**Pull Request status**

- [x] Implementation
- [x] [N/A] Unit tests
- [x] [N/A] Integration tests
- [x] [N/A] `chart/compass/values.yaml` is updated 


chart/compass/charts/director/templates/external-certificate-rotation-job.yaml
```diff
@@ -79,16 +79,22 @@ spec:
...
-                        sleep 0.2 # Sleep for 200ms before next retry
-                      else
-                        echo -e "${GREEN}Issuer locality of the client certificate is valid. Proceeding with the next steps... ${NC}"
-                        IS_CERT_ISSUER_LOCALITY_VALID=true
-                        break
-                      fi
+                      
+                      ARRAY_OF_LOCALITIES=($(echo "$EXPECTED_ISSUER_LOCALITY" | tr ',' '\n'))
+
+                      for LOCALITY in ${ARRAY_OF_LOCALITIES[@]}
```
nyordanoff, line 85: Functionally the current approach will work but it is a bit ineffective.

Let's say we have two possible valid locality values: l1, l2. And let's say that the cert is with locality = l2. 

The script will start looping through the possible localities (l1, l2) and the first iteration will be with locality = l1. It will see that l2 != l1 and will make call to get a new cert. This call is unnecessary because on the second iteration of the loop, the cert locality (l2) would have  matched the second possible value from the array - meaning that the original cert would have been with valid locality.

My suggestion is right after we initialise `ARRAY_OF_LOCALITIES` to traverse it and check if the `ISSUER_LOCALITY` matches any of its values and set the bool result to a var. Then use that var to determine whether a retry is needed (if the cert locality didn't match anything from the array). That way we will call `getClientCert` only if it's really necessary.

nyordanoff (changes requested):

nyordanoff (approved):

PR #2535, opened 2022-08-08
[Tenants Aggregator] Introduce Regional API Configuration

desislavaa:
**Description**
Sometimes we might have tenants that are only available on a certain region in a landscape. 
One TF job maintains tenants of a given type from a given landscape.
We maintain all tenants from that lanscape via an Event Service - that event service has one central client, that is able to retrieve all centrally available tenants, and might have multiple regional clients, that can retrieve only regional tenants.

- Rework Tenants Resync configuration
- Introduce Regional Clients
- Move ex-TF logic in a separate package
- Split TF service.go and simplify fetching logic

Documentation: https://github.com/kyma-incubator/compass/blob/b0745c769cccb080d19f75dbb3f4f43c2f5b479e/docs/tenants-aggregator/01-overview.md#region-based-tenancy

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/tenantfetchersvc/resync/config.go
```diff
@@ -0,0 +1,290 @@
...
+		ctx:             ctx,
+		name:            name,
+		environmentVars: environmentVars,
+		jobConfig:       nil,
+	}
+}
+
+func (jc *JobConfig) Validate() error {
+	switch {
+	case len(jc.JobName) == 0:
```
StefanCholakov, line 135: If statement would be more suitable here and also some meaningful error

installation/resources/compass-overrides-local.yaml
```diff
@@ -90,37 +90,43 @@ global:
         enabled: false
       job:
         interval: "10s"
-      accountRegion: "local"
+      tenantType: " account"
```
StefanCholakov, line 93: redundant whitespace

components/director/internal/tenantfetchersvc/handler.go
```diff
@@ -134,7 +110,7 @@ func (h *handler) FetchTenantOnDemand(writer http.ResponseWriter, request *http.
 
 	log.C(ctx).Infof("Fetching create event for tenant with ID %s", tenantID)
 
-	err := h.fetcher.FetchTenantOnDemand(ctx, tenantID, parentTenantID)
+	err := h.fetcher.SynchronizeTenant(ctx, tenantID, parentTenantID)
 	if err != nil {
```
StefanCholakov, line 114: if err := h.fetcher.SynchronizeTenant(ctx, tenantID, parentTenantID); err!=nil

StefanCholakov (changes requested):

components/director/internal/tenantfetchersvc/resync/synchronizer.go
```diff
@@ -0,0 +1,386 @@
...
+	defer ts.transact.RollbackUnlessCommitted(ctx, tx)
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	log.C(ctx).Infof("Checking if tenant with external tenant ID %s already exists...", tenantID)
+	tnt, err := ts.tenantStorageService.GetTenantByExternalID(ctx, tenantID)
+	if err != nil && !apperrors.IsNotFoundError(err) {
+		return nil, errors.Wrapf(err, "while checking if tenant with external ID %s already exists", tenantID)
+	}
+
+	return tnt, tx.Commit()
```
StefanCholakov, line 250: Isn't it better if we first commit the transaction, because we will return either (nil, err) or (tnt,nil)?
Now there is a chance to return both error and tenant.

components/director/internal/tenantfetchersvc/resync/synchronizer.go
```diff
@@ -0,0 +1,386 @@
...
+//go:generate mockery --name=TenantMover --output=automock --outpkg=automock --case=underscore --disable-version-string
+type TenantMover interface {
+	TenantsToMove(ctx context.Context, region, fromTimestamp string) ([]model.MovedSubaccountMappingInput, error)
+	MoveTenants(ctx context.Context, movedSubaccountMappings []model.MovedSubaccountMappingInput) error
+}
+
+// TenantsSynchronizer takes care of synchronizing tenants with external tenants registry.
+// It creates, updates, deletes and moves tenants that were created, updated, deleted or moved in that external registry.
+type TenantsSynchronizer struct {
+	supportedRegions []string // k=region, v=prefix
```
StefanCholakov, line 55: leftover

components/director/internal/tenantfetchersvc/resync/tenant_manager.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+
+	log.C(ctx).Infof("Tenant not found from central region, checking regional APIst")
+
+	tenantChan := make(chan *model.BusinessTenantMappingInput, len(tm.regionalClients))
+	for region, regionalClient := range tm.regionalClients {
+		go func(ctx context.Context, region string, regionalClient EventAPIClient, ch chan *model.BusinessTenantMappingInput) {
+			createdRegionalTenants, err := fetchCreatedTenantsWithRetries(regionalClient, tm.config.RetryAttempts, tm.supportedEventTypes, configProvider)
+			if err != nil {
+				log.C(ctx).WithError(err).Errorf("Failed to fetch created tenants from")
```
StefanCholakov, line 134: Can you add the name of the region here

desislavaa: AT least I had that intention :D

StefanCholakov (approved):

PR #2543, opened 2022-08-09
Make subscribeTenantToRuntime idempotent

la4ezar:
**Description**
Currently, multiple subscription calls may lead to multiple records in our DB.

- Make SubscribeTenantToRuntime check if a subscription is already registered and should reuse it instead of creating a new one.

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


PetarTodorovv (commented):
Maybe it's nice to have an e2e test for this checking two same consecutive subscriptions are successful and the runtime context in the DB is only one?

PetarTodorovv (approved):

PetarTodorovv (approved):

PR #2559, opened 2022-08-15
Check application type when assigning to formation

ivantenevvasilev:
**Description**
- when assigning to formation check `applicationType` to make sure it matches the `applicationType` in the application template for the application that it's being assigned to
- adapt existing e2e tests that register applications to register applications with valid `applicationType`
- add `applicationType` label before assigning to formation
- add new e2e test to check that assigning to a formation with unallowed application type fails

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/formation/fixtures_test.go
```diff
@@ -217,6 +218,12 @@ func fixApplicationModelWithoutTemplate(applicationID string) *model.Application
 	return appModel
 }
 
+func fixApplicationWithType(applicationType string) *model.Application {
```
alextargov, line 221: Don't see the param applicationType being used here

components/director/internal/domain/formation/service.go
```diff
@@ -362,6 +364,12 @@ func (s *service) assign(ctx context.Context, tnt, objectID string, objectType g
 	if err != nil {
 		return nil, errors.Wrapf(err, "while getting formation %q", formation.Name)
 	}
+	// TODO Remove default scenario check after removing default scenario
```
alextargov, line 367: Is this comment referring to the current PR or a future improvement?

tests/director/tests/application_api_test.go
```diff
@@ -597,21 +597,11 @@ func TestDeleteApplication(t *testing.T) {
...
-			"items": map[string]interface{}{
-				"type": "string",
-				"enum": scenarios,
-			},
-		}
-		var schema interface{} = jsonSchema
-
-		fixtures.CreateLabelDefinitionWithinTenant(t, ctx, certSecuredGraphQLClient, ScenariosLabel, schema, tenantID)
+		fixtures.UpsertScenariosLabelDefinitionWithinTenant(t, ctx, certSecuredGraphQLClient, tenantID, scenarios)
+		defer fixtures.UpdateScenariosLabelDefinitionWithinTenant(t, ctx, certSecuredGraphQLClient, tenantID, []string{defaultValue})
```
alextargov, line 601: Wouldn't it be better to use the "conf.DefaultScenario" value instead of the "defaultValue" variable?

tests/director/tests/application_api_test.go
```diff
@@ -705,23 +699,25 @@ func TestUnpairApplication(t *testing.T) {
...
-		applicationInput.Labels = graphql.Labels{ScenariosLabel: scenarios}
+		applicationInput.Labels = graphql.Labels{ScenariosLabel: scenarios, conf.ApplicationTypeLabelKey: createAppTemplateName("Cloud for Customer")}
 		appInputGQL, err := testctx.Tc.Graphqlizer.ApplicationRegisterInputToGQL(applicationInput)
 		require.NoError(t, err)
 
 		createApplicationReq := fixtures.FixRegisterApplicationRequest(appInputGQL)
 		application := graphql.ApplicationExt{}
 
 		err = testctx.Tc.RunOperationWithCustomTenant(ctx, certSecuredGraphQLClient, tenantID, createApplicationReq, &application)
+		require.NoError(t, err)
```
alextargov, line 714: Think it's better first to execute the defers and then check for errors so we would be sure that the defers will be executed

tests/director/tests/formation_api_test.go
```diff
@@ -1095,6 +1095,32 @@ func TestFormationNotifications(stdT *testing.T) {
...
+
+	tenantId := tenant.TestTenants.GetDefaultTenantID()
+
+	formation := fixtures.CreateFormation(t, ctx, certSecuredGraphQLClient, formationName)
+	defer fixtures.DeleteFormation(t, ctx, certSecuredGraphQLClient, formation.Name)
+
+	formationTemplate := fixtures.QueryFormationTemplate(t, ctx, certSecuredGraphQLClient, formation.FormationTemplateID)
+
+	actualApplication, err := fixtures.RegisterApplicationWithApplicationType(t, ctx, certSecuredGraphQLClient, applicationName, conf.ApplicationTypeLabelKey, invalidApplicationType, tenantId)
+	defer fixtures.CleanupApplication(t, ctx, certSecuredGraphQLClient, tenantId, &actualApplication)
```
alextargov, line 1113: We can assert the error is nil after the defer here - `require.NoError(t, err)`

tests/director/tests/main_test.go
```diff
@@ -47,6 +47,7 @@ type DirectorConfig struct {
 	ConsumerSubaccountLabelKey             string
 	SubscriptionLabelKey                   string
 	RuntimeTypeLabelKey                    string
+	ApplicationTypeLabelKey                string `envconfig:"APP_APPLICATION_TYPE_LABEL_KEY,default=applicationType"`
```
alextargov, line 50: Are these explicit tags necessary? I see them in a couple of places within the pr.

components/director/internal/domain/formation/service.go
```diff
@@ -362,6 +364,12 @@ func (s *service) assign(ctx context.Context, tnt, objectID string, objectType g
...
+	if formation.Name != model.DefaultScenario && objectType == graphql.FormationObjectTypeApplication {
```
la4ezar, line 368: Change these 3 if's(lines 368, 373, 378) into a switch statement where the cases will be the different object types. And after we remove the `DefaultScenario` the code `switch` will look even more readable. Also, this `switch` can be extracted into a function.

la4ezar (changes requested):

tests/director/tests/application_api_test.go
ivantenevvasilev: If we defer before checking for errors the evaluation of `formation.FormationTemplateID` will cause a nil ptr and we'll lose the error message on the following line.

tests/director/tests/main_test.go
ivantenevvasilev: Not really necessary, but it's easier for me to check where the value comes from when searching, I'll change it to implicit if it's better that way.

components/director/internal/domain/formation/service.go
ivantenevvasilev: The comment is for when the default scenario is removed in the future.

components/director/internal/domain/application/service.go
```diff
@@ -1014,15 +1014,14 @@ func (s *service) genericCreate(ctx context.Context, in model.ApplicationRegiste
 	}
 	in.Labels[nameKey] = normalizedName
 
+	var scenariosToAssign []string = nil
```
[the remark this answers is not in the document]
ivantenevvasilev: I was just being cautious by being explicit, but I'll remove it.

components/director/internal/domain/formation/fixtures_test.go
ivantenevvasilev: Forgot to remove and adapt it, it's only used in one test anyway. I will remove it.

la4ezar (approved):

alextargov (approved):

desislavaa (approved):

PR #2570, opened 2022-08-18
Application templates are managed by the System Fetcher

kaloyanspiridonov:
**Description**

- create application templates and related integration systems on system-fetcher start up

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/systemfetcher/loader.go
```diff
@@ -0,0 +1,212 @@
...
+			return nil, errors.Wrapf(err, "while unmarshalling application templates from file %s", applicationTemplatesDirectoryPath+f.Name())
+		}
+		log.C(ctx).Infof("Successfully loaded application templates from file: %s", f.Name())
+		appTemplateInputs = append(appTemplateInputs, templatesFromFile...)
+	}
+
+	return appTemplateInputs, nil
+}
+
+func (d *DataLoader) upsertIntegrationSystems(ctx context.Context, intSystems []model.IntegrationSystem) error {
```
NickyMateev, line 146: This upsetting func and the one for app templates only create the respective entities if they are missing. From this I think it would be more accurate to name the function'insertIntegrationsSystemsIfMissing' as there won't be any updating of the entity if it already exists.

NickyMateev: Would it be much effort to also update the entity if it already exists with the provided configuration? In this way we would truly have one configuration which controls completely the state of predefined app templates in our db.

chart/compass/templates/system-fetcher-configmap.yaml
```diff
@@ -0,0 +1,190 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: compass-system-fetcher-templates-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Chart.Name }}
+    release: {{ .Release.Name }}
+data:
+  integration-systems.json: |-
```
NickyMateev, line 11: This goes for the Int systems and app templates - can we only define the keys 'integration-systems.json' and 'templates.json' as they are here, but have their values (the big chunks of raw json) as placeholder configuration values?

The point is that the current values won't make sense for the open source. They are SAP-specific, hence it would be best to have them defined only in our internal mgt-plane-config repo as overrides.

chart/compass/templates/system-fetcher-configmap.yaml
```diff
@@ -0,0 +1,190 @@
...
+    [
+      {
+        "id":"{{ (index .Values.global.systemFetcher.intSystems 0).id }}",
```
NickyMateev, line 14: Can we scrap the `global.systemFetcher.intSystems` configuration array altogether and just check in the `upsertIntegrationSystems` function in `loader.go` if an integration system with the given name exists (as we do for the templates)? You could list all integration systems (they should not be many) and iterate by name.

NickyMateev: It's just that those intSystems ids look tricky to maintain and also why not just simplify things where we can..?

NickyMateev: A thing to consider is how then to provide the integration system id to those templates that need it in the applicationInputJSON

chart/compass/templates/system-fetcher-configmap.yaml
```diff
@@ -0,0 +1,190 @@
...
+      },
+      {
+        "id":"{{ (index .Values.global.systemFetcher.intSystems 1).id }}",
+        "name":"sap-successfactors",
+        "description":"Integration system for SAP SuccessFactors"
+      }
+    ]
+  templates.json: |-
+    [
+      {
```
NickyMateev, line 26: There are a lot of duplications and also default properties that we always set to app templates like the following:
- name and description are always the same
- all app templates have the same 2 placeholders - name and display-name
- all of these app templates have the label managed_app_provisioning:false
- all app templates have the legacy field accessLevel: GLOBAL

So it appears that the only dynamic property (which also has very predictable and similar structure and values) is the `applicationInputJSON`.
From this POV would it make sense to you to simplify the `templates.json` to a simpler structure of the like:

```
[
  {
    "name": "SAP ..."
    "applicationInputJSON: "..."
  }
]
```

After that we can programatically add the description, placeholders, labels and accessLevel. Of course it's possible that we may want to be able to specify different values than the defaults that we'll put programatically but we can extend this json array structure as needed

Another benefit of a simpler structure and one that is decoupled from the actual ApplicationTemplate programatic Go structure (as it currently stands this json can be automatically unmarsheled into a Go structure) is that we have freedom in adding additional fields like `integrationSystemName`. This way we can for example solve the issue of not knowing the Integration system ID at design time here, but by knowing the name of the int system we can programatically find it and add it to the applicationInputJSON later at runtime.

components/director/internal/systemfetcher/loader.go
NickyMateev: On second thoughts:
- we don't have much use case of updating int systems as they have no meaningful data associated with them
- as far as app templates are concerned - if you look at my comment below about restructuring the templates.json array where the only two properties that would make sense are `name` and `applicationInputJSON` then updating and existing apptemplate here would mean only to update it's applicationInputJSON property as is provided in the configuration.

chart/compass/templates/system-fetcher-configmap.yaml
kaloyanspiridonov: in my opinion this is not a simplification but a complication. In order to skip a few lines of configuration we will introduce a complex "magic" code.  If we decouple our configuration json from the actual ApplicationTemplate programatic Go structure then we will have: unmarshal to map[string]interface, if exists:=map[key] {<do-smth>}... or we will have to introduce another configuration go struct

components/director/internal/systemfetcher/loader.go
```diff
@@ -175,25 +205,135 @@ func (d *DataLoader) upsertAppTemplates(ctx context.Context, appTemplateInputs [
...
+}
+
+func enrichWithIntegrationSystemIDLabel(applicationInputJSON, intSystemID string) (string, error) {
+	var appInput map[string]interface{}
+
+	if err := json.Unmarshal([]byte(applicationInputJSON), &appInput); err != nil {
+		return "", errors.Wrapf(err, "while unmarshaling application input json")
+	}
+
+	labels, ok := appInput[labelsJSONKey]
```
NickyMateev, line 293: Integration system ID is a top level property of the app, not a label, maybe this is a mistake here?

NickyMateev (dismissed):

NickyMateev (approved):

PR #2572, opened 2022-08-18
Unify the names of the docker images

PetarTodorovv:
**Description**
Unify the docker images names

- Change `compass-tests` image name to `compass-e2e-tests`. Also, `pairing-adapter` image to `compass-pairing-adapter`
- Adapt images names where they referenced
- Add one new property of the images containing the name. We need this in order to sign docker images

**Pull Request status**
- [x] Implementation
- [x] [N/A] Unit tests
- [x] [N/A] Integration tests


tests/Makefile
```diff
@@ -1,4 +1,5 @@
 APP_NAME = compass-tests
```
DimitarPetrov, line 1: Remove

DimitarPetrov:
I am not sure If you rely on the name convention that this is stable enough.
Maybe it would be better to extract the name of the image under the component in values .yaml, something like this:

```yaml
    director:
      dir:
      name: compass-director
      version: "PR-2521"
```

Then whenever you need those images you can get them from the values.yaml.

alextargov (approved):

PR #2577, opened 2022-08-22
Trigger ORD Aggregator job manually in ORD Aggregator e2e tests

dragobt:
**Description**

- trigger ORD Aggregator job manually in ORD Aggregator e2e tests
- add role and role binding for ORD Aggregator test
- add `suspend` field in `ord-aggregator-job.yaml` which is set to `true` - that guarantees the ORD Aggregator job will be suspended during testing. So if you want to trigger the ORD Aggregator job in your tests you have to do it manually OR to set       `suspend: false`

**Pull Request status**

- [X] Implementation
- [ ] Unit tests
- [X] Integration tests


chart/compass/values.yaml
```diff
@@ -636,6 +636,7 @@ global:
       maxIdleConnections: 2
     globalRegistryUrl: http://compass-external-services-mock.compass-system.svc.cluster.local:8087/.well-known/open-resource-discovery
     maxParallelApplicationProcessors: 4
+    containerName: "ord-aggregator"
```
NickyMateev, line 639: I see you've added this config similarly how it is done in the System Fetcher test. Make sure to use this config value in the actual `ord-aggregator-job.yaml` as the name of the container (similarly to `system-fetcher-job.yaml` in the sys fetcher PR that you're looking from) otherwise it's practically useless to define it

chart/compass/templates/tests/ord-aggregator/ord-aggregator-test.yaml
```diff
@@ -22,14 +22,16 @@ spec:
     spec:
       serviceAccountName: {{ $.Chart.Name }}-e2e-ord-aggregator
       containers:
-        - name: ord-aggregator-tests
+        - name: {{ .Values.global.ordAggregator.containerName }}
```
NickyMateev, line 25: Going from `ord-aggregator-tests` to the parametarized `{{ .Values.global.ordAggregator.containerName }}` breaks the sandbox tests (as we noticed yesterday).

Make this `name` property value `{{ .Values.global.ordAggregator.containerName }}-tests` instead so that the sandbox make targets keep working

NickyMateev (changes requested):

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -327,6 +330,18 @@ func TestORDAggregator(stdT *testing.T) {
 		defer fixtures.CleanupApplication(t, ctx, certSecuredGraphQLClient, testConfig.DefaultTestTenant, &seventhApp)
 		require.NoError(t, err)
 
+		k8sClient, err := clients.NewK8SClientSet(ctx, time.Second, time.Minute, time.Minute)
+		require.NoError(t, err)
+		jobName := "ord-aggregator-test"
+		namespace := "compass-system"
+		k8s.CreateJobByCronJob(t, ctx, k8sClient, "compass-ord-aggregator", jobName, namespace)
```
NickyMateev, line 337: Here you spawn the ord aggregator job on-demand, but if there is already an existing ord aggregator job running then we don't know how both of them might collide. Additionally even if there was no other parallel job already running, you might hit the case where this e2e tests finishes and tries to cleanup it's created app resources but just at that time a new ord aggregator regular job spawns naturally and locks the app resources that you want to delete which will block/slow down the test until the ord aggregator job finishes.

That's why you need to ensure that for the duration of this e2e test that the regular ord aggregator job doesn't interfere.

Here's a suggestion, but if you have a better idea, we can discuss:
- add a new configuration in values.yaml called `ordAggregator.suspend`:
image
- Then use it in the ord aggregator job yaml:
image
- The `suspend` value will be `true` by default as it is in the picture, but we'll make sure to override it to `false` in the mgt-plane-config repo for our real environments. This way we can disable the ord aggregator for local scenarios and for our dev-e2e-tests, but have it running regularly (not suspended) for our real envs

NickyMateev (changes requested):

tests/ord-aggregator/tests/handler_test.go
NickyMateev: The only down-side to this approach is that if someone develops sth about the ord aggregator on his local machine, he needs to be aware that the ord aggregator will be suspended by default. He'll have to either enable it manually, or just trigger the cronjob manually (as you do in the tests) whenever he wants to check something with the aggregator.

I think this is a smaller price to pay than the alternative that I can think of - the alternative is that we programatically suspend the ord aggregator job from the e2e test code by patching the ord aggregator. But this has the bigger (in my opinion) downside of the ord aggregator potentially runnning at the same time - we'll have to wait for it to finish before the test can suspend the job and then start running the test cases.

NickyMateev (approved):

PR #2607, opened 2022-09-05
Fix endspoints of destination fetcher and destination service

petar-iv:
**Description**

- Fixes in destination fetcher configuration


components/director/internal/destinationfetchersvc/client.go
```diff
@@ -186,7 +186,7 @@ func (c *Client) FetchTenantDestinationsPage(ctx context.Context, page string) (
 
 	pageCount := res.Header.Get(c.apiConfig.PagingCountHeader)
 	if pageCount == "" {
-		return nil, errors.Errorf("failed to extract header '%s' from destinations response", c.apiConfig.PagingCountParam)
+		return nil, errors.Errorf("failed to extract header '%s' from destinations response", c.apiConfig.PagingCountHeader)
```
PetarTodorovv, line 189: Because there is a change in the director component you need to bump/update the director component version in the root values.yaml as well. 
Here - https://github.com/kyma-incubator/compass/blob/915f2b7b1c39dfd6dca29c275754ea52dab8c854/chart/compass/values.yaml#L126 needs to be `PR-2607`

PetarTodorovv (changes requested):

PetarTodorovv (approved):

PR #2611, opened 2022-09-08
Destination fetcher fix extension service defaults

georgifarashev:
**Description**

- In order for destination fetcher to work out of the box with S/4 destinations created from extension service - proper defaults need to be set.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


NickyMateev (commented):
Overall I don't see a technical problem with the PR, my only concern is the explicit hardcoding logic surrounding one specific type (S4) instead of thinking of something more generic which could make the code more maintainable but also better suited for the open source.

NickyMateev (approved):

PR #2616, opened 2022-09-09
Move DB data

dzahariev:
Move DB data to be in accessible (not RO) file system place.


dzahariev:
Local installation works fine with and without DB, should proceed and merge so we can continue investigation of benchmark tests issue.

PetarTodorovv (approved):

PR #2617, opened 2022-09-09
Remove Default Scenario

la4ezar:
**Description**
Currently, we have a feature to add a DEFAULT scenario by default on each application and runtime when registered when the feature is enabled. We want to get rid of it as it is disabled on real environments and only contributes to the difference between local and productive envs. Also, this feature is poorly designed and implemented and has a negative impact on our codebase and new features.

- Drop the 'DEFAULT' scenario
- Adapt e2e tests
- Remove automatically creation of label defs
- Remove cleanup of tenants in e2e tests

**Related issue(s)**

- https://github.com/kyma-incubator/compass-console/pull/69

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


tests/director/tests/eventing_test.go
```diff
@@ -31,29 +31,36 @@ func TestGetDefaultRuntimeForEventingForApplication_DefaultBehaviourWhenNoEventi
 	runtime1Eventing := "eventing.runtime1.local"
 	runtime1EventingURL := fmt.Sprintf(runtimeEventURLFormat, runtime1Eventing)
 	runtime2EventingURL := "https://eventing.runtime2.local"
-	defaultScenarios := []string{"DEFAULT"}
 
 	appName := "app-test-eventing"
-	application, err := fixtures.RegisterApplication(t, ctx, certSecuredGraphQLClient, appName, tenantId)
+	applicationInput := fixtures.FixSampleApplicationRegisterInput(appName)
```
StanislavStefanov, line 36: Extract the applicationInput setup as it is done the same way several times

tests/pkg/fixtures/application_queries.go
```diff
@@ -244,23 +244,12 @@ func GenerateOneTimeTokenForApplicationWithSuggestedToken(t require.TestingT, ct
...
-	if defaultScenarioEnabled {
-		defaultValue := "DEFAULT"
 
-		scenarios := []string{defaultValue}
-		var labelValue interface{} = scenarios
-
-		setLabelRequest := FixSetApplicationLabelRequest(applicationID, labelKey, labelValue)
-		label := graphql.Label{}
-		err := testctx.Tc.RunOperationWithCustomTenant(ctx, gqlClient, tenantID, setLabelRequest, &label)
+	deleteLabelRequest := FixDeleteApplicationLabelRequest(applicationID, labelKey)
```
StanislavStefanov, line 250: Why do we delete the labels instead of calling unassign from formation?

tests/pkg/fixtures/formation_queries.go
```diff
@@ -31,7 +31,18 @@ func CreateFormation(t require.TestingT, ctx context.Context, gqlClient *gcli.Cl
 	return formation
 }
 
-func CreateFormationWithinTenant(t *testing.T, ctx context.Context, gqlClient *gcli.Client, tenantID, formationName string, formationTemplateName *string) graphql.Formation {
+func CreateFormationWithinTenant(t *testing.T, ctx context.Context, gqlClient *gcli.Client, tenantId, formationName string) graphql.Formation {
+	t.Logf("Creating formation with name: %q", formationName)
+	var formation graphql.Formation
+	createFirstFormationReq := FixCreateFormationRequest(formationName)
```
StanislavStefanov, line 37: Maybe the name should be createFormaionReq

StanislavStefanov (changes requested):

components/director/internal/domain/application/service_test.go
```diff
@@ -1431,122 +1021,25 @@ func TestService_CreateFromTemplate(t *testing.T) {
...
-	modelInputWithInvalidPpmsProductVersion := model.ApplicationRegisterInput{
-		Name: "foo.bar-not",
-		Labels: map[string]interface{}{
-			"label":                "value",
-			"applicationType":      "test-app-with-ppms",
-			"ppmsProductVersionId": "2",
-		},
-		IntegrationSystemID: &intSysID,
-		BaseURL:             str.Ptr("123://localhost"),
+			mock.AssertExpectationsForObjects(t, appRepo, intSysRepo, uidSvc, bndlSvc, formationSvc)
```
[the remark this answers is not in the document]
la4ezar: `appNameNormalizer` which is `*normalizer.DefaultNormalizator` is not `mock.assertExpectationser` and can't pass it to `mock.AssertExpectationsForObjects` because the `AssertExpectations` method is missing.

tests/pkg/fixtures/application_queries.go
la4ezar: Because the deletion of the application label(if the label is == `scenarios`) leads to unassignment of all formations. Reference - https://github.com/kyma-incubator/compass/blob/060b16d3fecf6315ffcf04a833b81360701b3283/components/director/internal/domain/application/service.go#L799.

StanislavStefanov (dismissed):

StanislavStefanov (approved):

alextargov (approved):

tests/pkg/tenant/tenant_setup.go
```diff
@@ -272,47 +249,3 @@ func (mgr TestTenantsManager) List() []Tenant {
 
 	return toReturn
 }
-
-func (mgr TestTenantsManager) cleanup(ids []string) {
```
DimitarPetrov, line 276: Now as we have removed DB access logic from tests, we can remove their DB access entirely by removing:

- The cloudSQL proxy from the testDefinitions from charts.
- The DB credentials and connection details from their env.
- Their service accounts from the proxy-user workload identity in our infra projects.

PR #2618, opened 2022-09-09
Add destination fetcher e2e test

stefan-sto:
**Description**

- Add e2e test which involves ORD service, destination fetcher and destination service

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [x] Integration tests


components/director/internal/destinationfetchersvc/client.go
```diff
@@ -43,6 +43,7 @@ type DestinationServiceAPIConfig struct {
 	PagingCountParam              string        `envconfig:"APP_DESTINATIONS_PAGE_COUNT_PARAM,default=$pageCount"`
 	PagingCountHeader             string        `envconfig:"APP_DESTINATIONS_PAGE_COUNT_HEADER,default=Page-Count"`
 	SkipSSLVerify                 bool          `envconfig:"APP_DESTINATIONS_SKIP_SSL_VERIFY,default=false"`
+	OAuthTokenPath                string        `envconfig:"APP_DESTINATION_OAUTH_TOKEN_PATH,default=/oauth/token"`
```
PetarTodorovv, line 46: I saw that in the charts we also use `/oauth/token` but as far as I remember in the external-services-mock we do not have such endpoint? We have `/secured/oauth/token` for the unsecured server and `/cert/token` for the secured one. 

The interesting part is all of the tests are passing so I might be missing something. Do you know the exact URL that is called in the local setup? or which endpoint/handler of the external services mock is invoked in this case?

tests/director/tests/application_api_test.go
```diff
@@ -493,7 +493,23 @@ func TestUpdateApplicationWithLocalTenantIDShouldBeAllowedOnlyForIntegrationSyst
 	updatedApp := graphql.ApplicationExt{}
 
 	t.Run("should fail for non-integration system", func(t *testing.T) {
-		err = testctx.Tc.RunOperation(ctx, certSecuredGraphQLClient, request, &updatedApp)
+		runtime, err := fixtures.RegisterRuntime(t, ctx, certSecuredGraphQLClient, "test-runtime", tenant.TestTenants.GetDefaultTenantID())
```
PetarTodorovv, line 496: Can you elaborate on why this change is necessary? From the test name, I assume the old code with the test run should fail as well? because it uses certificate creds instead of an integration system?

tests/ord-service/tests/subscription_flow_test.go
```diff
@@ -276,15 +308,37 @@ func TestConsumerProviderFlow(stdT *testing.T) {
...
+		// With destinations
+		respBody = makeRequestWithHeaders(stdT, certHttpClient, conf.ORDExternalCertSecuredServiceURL+
+			"/systemInstances?$expand=consumptionBundles($expand=destinations)&$format=json&reload=true", headers)
+		require.Equal(stdT, 1, len(gjson.Get(respBody, "value").Array()))
+		require.Equal(stdT, consumerApp.Name, gjson.Get(respBody, "value.0.title").String())
+		require.NotEmpty(stdT, gjson.Get(respBody, "value.0.consumptionBundles.0.destinations").Raw)
+		destinationsFromResponse := gjson.Get(respBody, "value.0.consumptionBundles.0.destinations").Array()
+		require.Len(stdT, destinationsFromResponse, 1)
+		require.Equal(stdT, destination.Name, destinationsFromResponse[0].Get("sensitiveData.destinationConfiguration.Name").String())
+		stdT.Log("Successfully fetched system with bundles and destinations")
```
PetarTodorovv, line 335: I assume that's the validation when subscribe request is executed - to create a destination? What's the case when someone has triggered an unsubscribe request - the destination should be deleted? If there is a logic on unsubscribe requests we could add additional checks/validation for the destinations that cover the unsubscribe flow as well? 

On line 337 we explicitly execute an unsubscribe request and below that is the validation regarding it.

tests/director/tests/application_api_test.go
```diff
@@ -493,7 +493,23 @@ func TestUpdateApplicationWithLocalTenantIDShouldBeAllowedOnlyForIntegrationSyst
...
+		defer fixtures.CleanupRuntime(t, ctx, certSecuredGraphQLClient, tenant.TestTenants.GetDefaultTenantID(), &runtime)
+		require.NoError(t, err)
+		require.NotEmpty(t, runtime.ID)
+
+		runtimeAuth := fixtures.RequestClientCredentialsForRuntime(t, ctx, certSecuredGraphQLClient, tenant.TestTenants.GetDefaultTenantID(), runtime.ID)
+		require.NotEmpty(t, runtimeAuth)
```
StanislavStefanov, line 502: Shouldn't we be checking if runtimeAuth.ID is populated?

StanislavStefanov (changes requested):

PetarTodorovv:
Also, one additional comment about the E2E test. In the `subscription_flow_test.go` file, there is a `TestConsumerProviderFlow` test with two tests run - `ConsumerProvider flow` and `Consumer provider with user context header`. Currently, the new dest svc test is in the former one but I think it's a good idea to have it also in the latter one. The difference between both of them is we're using different approaches to send the "provider" data. In the first one, we're sending it within a token whereas in the second one we sending it in the `User_context` header.

components/director/internal/destinationfetchersvc/client.go
PetarTodorovv: ah, I overlook the local overrides, sorry. I guess we're using this approach to override only the local setup because the productive one is with `/oauth/token` and that way avoiding "productive" override?

tests/director/tests/application_api_test.go
PetarTodorovv: Thanks for the clarification!

StanislavStefanov (approved):

PetarTodorovv (approved):

PR #2640, opened 2022-09-20
Enrich self register labels for runtimes

PetarTodorovv:
**Description**
Enrich the labels that we return during self-registration of runtimes

- return the SaaS application name as part of the self-registered runtime's labels
- Make the new label immutable
- Adjust unit and E2E tests

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


tests/director/tests/runtime_api_test.go
```diff
@@ -798,53 +798,65 @@ func TestSelfRegMoreThanOneProviderRuntime(t *testing.T) {
 
 func TestRuntimeTypeImmutability(t *testing.T) {
```
DimitarPetrov, line 799: Add assert that the new label is added to the self reg test as well.

components/director/pkg/config/selfreg_config.go
```diff
@@ -68,3 +86,24 @@ func (c *SelfRegConfig) MapInstanceConfigs() error {
...
+	if err != nil {
+		return errors.Wrapf(err, "while getting SaaS application names secret")
+	}
+
+	m, err := ParseConfigToJSONMap(secretData)
+	if err != nil {
+		return err
+	}
+
+	c.RegionToSaaSAppName = make(map[string]string)
```
la4ezar, line 102: You can create this map with capacity = len(m).

la4ezar (changes requested):

la4ezar (approved):

la4ezar (approved):

PR #2641, opened 2022-09-20
Soften certain ORD aggregation validation rules temporarily for S/4 systems

emmakarova:
**Description**
Soften certain validation rules:
 - bundle: `shortDescription` should be between 1 & 256
- api: `successors`: cannot be blank
- event: `extensible` field must be provided when `policyLevel` is either `sap:core:v1` or `sap:partner:v1`;
- event: `resourceDefinitions:` when event resource visibility is public or internal, resource definitions must be provided;
- event: `title`: cannot be blank

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/open_resource_discovery/validation.go
```diff
@@ -681,14 +680,15 @@ func validateEventResourceDefinition(value interface{}, event model.EventDefinit
...
 	if !ok {
 		return errors.New("error while casting to EventResourceDefinition")
 	}
 
-	if len(eventResourceDef) == 0 {
-		return errors.New("when event resource visibility is public or internal, resource definitions must be provided")
-	}
+	// soften validation temporarily
+	// if len(eventResourceDef) == 0 {
+	//	return errors.New("when event resource visibility is public or internal, resource definitions must be provided")
```
dzahariev, line 690: I think the better approach for softening is not to remove the code snippet (comment out) but to log the error instead returning it here. 
With this the errors will still be visible in logs and we will know when to harden the rule again (when no more errors are logged)

dzahariev (changes requested):

components/schema-migrator/migrations/director/20220914135000_ord_increase_short_description_size.down.sql
```diff
@@ -0,0 +1,46 @@
+BEGIN;
```
alextargov, line 1: Make sure to update the file name as the current latest migration has a timestamp of `20220920145100`.
When we introduce UP and DOWN migrations, they should have the latest timestamp.

components/director/internal/open_resource_discovery/ord_document.go
```diff
@@ -179,17 +179,35 @@ func (docs Documents) Validate(calculatedBaseURL string, apisFromDB map[string]*
...
+
+		eventsIndicesWithEmptyName := make([]int, 0)
+
+		for i, event := range doc.EventResources {
+			if event.Name == "" {
+				eventsIndicesWithEmptyName = append(eventsIndicesWithEmptyName, i)
+				continue
+			}
+
 			if err := validateEventInput(event, packagePolicyLevels, eventsFromDB, resourceHashes); err != nil {
```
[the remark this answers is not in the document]
dzahariev: continue on line 188 will address this.

components/director/internal/open_resource_discovery/validation.go
emmakarova: Added logging of the error with default logger.

components/schema-migrator/migrations/director/20220914135000_ord_increase_short_description_size.down.sql
emmakarova: Renamed the file names with the latest timestamp.

dzahariev (approved):

PR #2644, opened 2022-09-21
Allow Technical Client authentication

StefanCholakov:
**Description**
In Atom, when a technical client is created, a secret, containing client certificate and key is created as well. We would like to use this certificate for authentication when executing queries and mutations for tenants. Since technical client certificates do not contain OU specifying the region, a few modifications of the logic for cert subject validation are made.

**Changes proposed in this PR**
- add a new consumer type(Technical Client). For now, each Technical Client has tenant:read and tenant:write scopes, but based on the required access, new consumers can be added with the corresponding scopes
- adjust logic for validating external certificate subject
- add e2e test for the new scenario

**Related PRs**
- https://github.tools.sap/cmp/management-plane-config/pull/563

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/pkg/cert/certutils_test.go
```diff
@@ -99,13 +102,23 @@ func TestSubjectExtraction(t *testing.T) {
...
-			assert.Equal(t, testCase.remainingOrgUnit, cert.GetRemainingOrganizationalUnit(testCase.orgUnitPattern)(testCase.subject))
+			assert.Equal(t, testCase.remainingOrgUnit, cert.GetRemainingOrganizationalUnit(testCase.orgUnitPattern, testCase.orgRegionPattern)(testCase.subject))
 			assert.Equal(t, testCase.commonName, cert.GetCommonName(testCase.subject))
-			assert.Equal(t, testCase.possibleOrgUnitMatches, cert.GetPossibleRegexTopLevelMatches(testCase.orgUnitPattern))
+			assert.Equal(t, testCase.possibleOrgUnitMatches, cert.GetPossibleRegexTopLevelMatches(constructOURegex(testCase.orgUnitPattern, testCase.orgRegionPattern)))
 		})
 	}
 }
 
+func constructOURegex(patterns ...string) string {
```
krasish, line 112: The same function is present in `cerutills.go`. Make it an exported function there and use `cert.ConstructOURegex` instead of duplicating it here.

krasish (commented):
Even after the discussion we've had, I am not sure that splitting the OU pattern in two parts and then constructing it in the code afterward is absolutely necessary.In practice, we use the same pattern for the regex checking.

We do need to remove/adapt the part which checks whether there only is a single remaining OU that does not match the regex (notice that if we only remove the `expectedOrgUnits-matchedOrgUnits == 0` part from the if statement everything will work just fine even in the scenario in which the region is missing) but that still is just a half-a-line adjustment which we will result in less code & configuration changes.

 Other reviewers can give it a thought too. In case we leave it as it is (split into OU pattern & OU regional pattern), we must carefully adapt configurations for all installations.

desislavaa (approved):

la4ezar (approved):

PR #2646, opened 2022-09-26
replace buildx with img

Ressetkk:
With Prow we aim to reduce the number of privileged containers, so we have to tackle the problem of building images without docker. `img` should be a drop-in replace for buildx that doesn't require privileged mode.

In the future I'll be working on removing this workaround and introduce a new centrally managed build infrastructure


dzahariev:
@Ressetkk, can you please confirm if multi architecture build is triggered with "img", as we need ARM and AMD architecture based images to be produced on each build.

PR #2662, opened 2022-09-30
Fix unpair process for extensitions svc

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/pkg/graphql/schema.graphql
+++ b/components/director/pkg/graphql/schema.graphql
@@ -185,7 +185,6 @@ enum WebhookType {
 	REGISTER_APPLICATION
 	UNREGISTER_APPLICATION
 	OPEN_RESOURCE_DISCOVERY
-	UNPAIR_APPLICATION
 }
 
 interface OneTimeToken {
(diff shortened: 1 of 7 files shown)


la4ezar (approved):

components/director/pkg/graphql/schema.graphql
```diff
@@ -185,7 +185,6 @@ enum WebhookType {
 	REGISTER_APPLICATION
 	UNREGISTER_APPLICATION
 	OPEN_RESOURCE_DISCOVERY
-	UNPAIR_APPLICATION
```
PetarTodorovv, line 188: If we remove the `UNPAIR_APPLICATION` enum type shouldn't we also remove the other "unpair" statuses in the code, for example [here](https://github.com/kyma-incubator/compass/blob/fix-unpair-process-for-extensitions-svc/components/director/pkg/graphql/schema.graphql#L67-L68) or [here](https://github.com/kyma-incubator/compass/blob/fix-unpair-process-for-extensitions-svc/components/director/pkg/graphql/models_gen.go#L827-L828)? I saw these two but there could be more usages as well.

P.S. I do not have context for the change and the comment could be irrelevant

PetarTodorovv (changes requested):

components/director/pkg/graphql/schema.graphql
alextargov: No, we shouldn't. The statuses are essential for the cockpit UI. There will still be an unpair flow, but the extensions will use the unregister webhook.

PetarTodorovv (approved):

dragobt (approved):

PR #2680, opened 2022-10-10
Make certificate subject trust more strict

DimitarPetrov:
**Description**

- Expect exactly one OU left after pattern matching
- Match based on whole words / phrases instead of allowing submatches

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- N/A Integration tests
- N/A Mocks are regenerated, using the automated script


la4ezar (approved):

components/hydrator/internal/subject/cert_subject.go
```diff
@@ -140,17 +139,7 @@ func ExternalCertIssuerSubjectMatcher(externalSubjectConsts ExternalIssuerSubjec
 		if cert.GetCountry(subject) != externalSubjectConsts.Country || cert.GetOrganization(subject) != externalSubjectConsts.Organization {
 			return false
 		}
-		orgUnitRegex := regexp.MustCompile(externalSubjectConsts.OrganizationalUnitPattern)
```
PetarTodorovv, line 143: How was this working before when the `GetPossibleRegexTopLevelMatches` did not exist?

components/director/pkg/cert/certutils.go
```diff
@@ -68,7 +76,7 @@ func ConstructOURegex(patterns ...string) string {
 	nonEmptyStr := make([]string, 0)
 	for _, pattern := range patterns {
 		if len(pattern) > 0 {
-			nonEmptyStr = append(nonEmptyStr, pattern)
+			nonEmptyStr = append(nonEmptyStr, fmt.Sprintf("\\b%s\\b", pattern))
```
PetarTodorovv, line 79: what is the idea of `\\b` prefix/suffix?

PetarTodorovv (approved):

components/hydrator/internal/subject/cert_subject.go
DimitarPetrov: Old reference in go mod from a commit part of the previous PR history.

components/director/pkg/cert/certutils.go
DimitarPetrov: Match only a whole word / phrase instead of just match.

For example without it the regex "Region" will match the string "thisIsAFakeRegion", now it will match only "Region"

PR #2689, opened 2022-10-12
Rework System Fetcher to use application templates labels

alextargov:
**Description**

- Parametrize filter criteria
- Stop using mappings for system type determination. Instead, use a label on the app template entity

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


chart/compass/values.yaml
```diff
@@ -742,8 +742,8 @@ global:
     systemsAPIEndpoint: ""
     # systemsAPIFilterCriteria - criteria for fetching systems
     systemsAPIFilterCriteria: ""
-    # systemToTemplateMappings - how to map system properties to an existing application template
-    systemToTemplateMappings: '{}'
+    appTemplatesLabelFilter: "systemRole"
```
NickyMateev, line 745: Can we rename `appTemplatesLabelFilter` to `appTemplatesProductLabel`? Because the value of this property is practically the label value which we put to app templates, not an actual filter. It just so happens that we later filter based on that.

NickyMateev (approved):

PR #2692, opened 2022-10-14
Refactor metrics pusher

desislavaa:
**Description**
As we have several aggregators, the metrics pusher used by the Tenants Aggregator (TA) can be reused in them as well. 
- Refactor Metrics pusher to be generic instead of specific for the Tenants Aggregator
- Remove the Events Service request status pusher - it was useless combined with the metric for aggregation failure, as they will either fire both, or none.

**Pull Request status**

- [x] Implementation
- [x] Unit tests


components/director/internal/tenantfetchersvc/resync/client.go
```diff
@@ -187,13 +174,8 @@ func NewClient(oAuth2Config OAuth2Config, authMode oauth.AuthMode, clientConfig
...
 }
 
-// SetMetricsPusher missing godoc
-func (c *Client) SetMetricsPusher(metricsPusher MetricsPusher) {
-	c.metricsPusher = metricsPusher
-}
-
 // FetchTenantEventsPage missing godoc
-func (c *Client) FetchTenantEventsPage(eventsType EventsType, additionalQueryParams QueryParams) (*EventsPage, error) {
+func (c *Client) FetchTenantEventsPage(ctx context.Context, eventsType EventsType, additionalQueryParams QueryParams) (*EventsPage, error) {
```
StefanCholakov, line 178: why is context needed here?

desislavaa: because of the logs in the metrics pusher - they went from `log.D()` to `log.C(ctx)`

StefanCholakov (approved):

PR #2695, opened 2022-10-17
Placeholders for apps in subscription

alextargov:
**Description**

- Subscriptions to use subdomain and region in the placeholders so the baseURL would be populated correctly

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/domain/subscription/service.go
```diff
@@ -28,6 +29,13 @@ type Config struct {
 	RuntimeTypeLabelKey        string `envconfig:"APP_RUNTIME_TYPE_LABEL_KEY,default=runtimeType"`
 }
 
+const (
+	// SubdomainLabelKey is the key of the tenant label for subdomain.
+	SubdomainLabelKey = "subdomain"
+	// RegionPrefix a prefix to be trimmed when creating an app from template
```
NickyMateev, line 35: RegionPrefix a prefix to be trimmed **from the region placeholder value** when creating an app from template

tests/director/tests/application_templates_subscription_test.go
```diff
@@ -91,6 +97,7 @@ func TestSubscriptionApplicationTemplateFlow(baseT *testing.T) {
 		t.Run("Application is created successfully in consumer subaccount as a result of subscription", func(t *testing.T) {
 			//GIVEN
 			subscriptionToken := token.GetClientCredentialsToken(t, ctx, conf.SubscriptionConfig.TokenURL+conf.TokenPath, conf.SubscriptionConfig.ClientID, conf.SubscriptionConfig.ClientSecret, "tenantFetcherClaims")
+			expectedBaseURL := fmt.Sprintf(baseURLTemplate, DefaultSubdomain, RegionValue)
```
NickyMateev, line 100: Providing subdomain and region placeholders is an optional thing. The common case of stakeholders might be not to provide it.

Right now you've modified the existing test **"Application is created successfully in consumer subaccount as a result of subscription"** to work as if the two optional parameters are provided in the app template. Now we don't have such a test when they are not provided.

So it would be better to rather leave the original test as is and add a new one below it (with a clarification in the test name that it will include subdomain and region placeholders in the app template), which will test the scenario when the optional subdomain + region placeholders are part of the app template.

NickyMateev (changes requested):

tests/director/tests/application_templates_subscription_test.go
```diff
@@ -26,6 +26,12 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	RegionValue      = "eu-1"
```
NickyMateev, line 30: I think I've seen this `eu-1` string as part of some of our values.yaml configurations. Maybe it's the `global.defaultTenantRegion` value? In this case it would be better to not hard code this here but rather use an env variable for accessing this value.

tests/director/tests/application_templates_subscription_test.go
```diff
@@ -26,6 +26,12 @@ import (
...
+	DefaultSubdomain = "compass-external-services-mock-sap-mtls"
```
NickyMateev, line 31: Same as above comment for `eu-1`... maybe you can get this value from `global.externalServicesMock.certSecuredHost` instead of hardcoding it

NickyMateev (changes requested):

NickyMateev (dismissed):

NickyMateev (approved):

PR #2712, opened 2022-10-27
Introduce Formation Mapping Async API Skeleton

PetarTodorovv:
**Description**
Introduce a formation mapping async API skeleton that will be used by the formation participants to call us asynchronously to update the exchanged configuration. Also, custom authentication/authorization logic is introduced as HTTP middleware.

- HTTP middleware with custom authentication/authorization logic
- New formation mapping async API handler as part of the director component
- Unit and E2E tests, (re)generated mocks
- Updated inter dependencies

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


dragobt (dismissed):

tests/director/tests/formation_mapping_api_test.go
```diff
@@ -0,0 +1,558 @@
...
+		asyncFormationTmplName := "async-formation-template-name"
+		t.Logf("Creating formation template with name: %q", asyncFormationTmplName)
+		ft := createFormationTemplate(t, ctx, asyncFormationTmplName, conf.KymaRuntimeTypeLabelValue, []string{appType}, graphql.ArtifactTypeEnvironmentInstance)
+		defer fixtures.CleanupFormationTemplate(t, ctx, certSecuredGraphQLClient, ft.ID)
+
+		asyncFormationName := "async-formation-name"
+		t.Logf("Creating formation with name: %q from template with name: %q", asyncFormationName, asyncFormationTmplName)
+		formation := fixtures.CreateFormationFromTemplateWithinTenant(t, ctx, certSecuredGraphQLClient, parentTenantID, asyncFormationName, &asyncFormationTmplName)
+		defer fixtures.DeleteFormationWithinTenant(t, ctx, certSecuredGraphQLClient, parentTenantID, asyncFormationName)
+		formationID := formation.ID
```
ivantenevvasilev, line 85: `require.NotNil(formation)` just in case, if you decide to add them in the previous two places.

components/director/internal/formationmapping/handler.go
```diff
@@ -0,0 +1,171 @@
...
+	if b.Error != "" {
+		fieldRules = append(fieldRules, validation.Field(&b.State, validation.In(ConfigurationStateCreateError)))
+		fieldRules = append(fieldRules, validation.Field(&b.Configuration, validation.Empty))
+		return validation.ValidateStruct(&b, fieldRules...)
+	} else if len(b.Configuration) > 0 {
+		fieldRules = append(fieldRules, validation.Field(&b.State, validation.In(ConfigurationStateReady, ConfigurationStateConfigPending)))
+		fieldRules = append(fieldRules, validation.Field(&b.Error, validation.Empty))
+		return validation.ValidateStruct(&b, fieldRules...)
+	} else {
+		return errors.New("The Request Body cannot contains only State")
```
ivantenevvasilev, line 108: missing test case and typo

components/director/internal/formationmapping/handler.go
```diff
@@ -0,0 +1,171 @@
...
+		return
+	}
+
+	routeVars := mux.Vars(r)
+	formationID := routeVars[FormationIDParam]
+	formationAssignmentID := routeVars[FormationAssignmentIDParam]
+
+	if formationID == "" || formationAssignmentID == "" {
+		log.C(ctx).Errorf("Missing required parameters: %q or/and %q", FormationIDParam, FormationAssignmentIDParam)
+		respondWithError(ctx, w, http.StatusBadRequest, errors.New("Not all of the required parameters are provided"))
```
ivantenevvasilev, line 85: missing return

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -0,0 +1,317 @@
...
+		if consumerType == consumer.IntegrationSystem && app.IntegrationSystemID != nil && *app.IntegrationSystemID == consumerID {
+			log.C(ctx).Infof("The caller with ID: %q and type: %q has owner access to the formation assignment with target: %q and target type: %q that is being updated", consumerID, consumerType, fa.Target, fa.TargetType)
+			return true, http.StatusOK, nil
+		}
+
+		// Verify if the caller has owner access to the formation assignment with type application that is being updated
+		log.C(ctx).Infof("Getting parent tenant of the caller with ID: %q and type: %q", consumerID, consumerType)
+		tnt, err := a.tenantRepo.Get(ctx, consumerInternalTenantID)
+		if err != nil {
+			log.C(ctx).Warningf("an error occurred while getting tenant from ID: %q", consumerInternalTenantID)
```
ivantenevvasilev, line 200: missing return

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+
+		if tnt != nil && tnt.Parent != "" {
+			exists, err := a.appRepo.OwnerExists(ctx, tnt.Parent, fa.Target)
+			if err != nil {
+				log.C(ctx).Warningf("an error occurred while verifying caller with ID: %q and type: %q has owner access to formation assignment with type: %q and target ID: %q", consumerInternalTenantID, consumerType, fa.TargetType, fa.Target)
```
ivantenevvasilev, line 206: missing return in case of db error

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -0,0 +1,317 @@
...
+	consumerTenantPair, err := tenant.LoadTenantPairFromContext(ctx)
+	if err != nil {
+		return false, http.StatusInternalServerError, errors.Wrap(err, "while loading tenant pair from context")
+	}
+	consumerInternalTenantID := consumerTenantPair.InternalID
+	consumerExternalTenantID := consumerTenantPair.ExternalID
+
+	log.C(ctx).Infof("Consumer with internal ID: %q and external ID: %q with type: %q is trying to update formation assignment with ID: %q for formation with ID: %q about source: %q and source type: %q, and target: %q and target type: %q", consumerInternalTenantID, consumerExternalTenantID, consumerType, fa.ID, fa.FormationID, fa.Source, fa.SourceType, fa.Target, fa.TargetType)
+	if fa.TargetType == model.FormationAssignmentTypeApplication {
+		log.C(ctx).Infof("The formation assignment that is being update has type: %s and ID: %q", model.FormationAssignmentTypeApplication, fa.Target)
```
ivantenevvasilev, line 182: updated

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+		}
+		log.C(ctx).Warningf("The caller with ID: %q and type: %q has NOT direct owner access to formation assignment with type: %q and target ID: %q. Checking if the application is made through subscription...", consumerInternalTenantID, consumerType, fa.TargetType, fa.Target)
+
+		// Validate if the application is registered through subscription and the caller has owner access to that application
+		return a.validateSubscriptionProvider(ctx, app.ApplicationTemplateID, consumerExternalTenantID, string(consumerType), fa.Target, string(fa.TargetType))
+	}
+
+	if fa.TargetType == model.FormationAssignmentTypeRuntime && (consumerType == consumer.Runtime || consumerType == consumer.ExternalCertificate || consumerType == consumer.SuperAdmin) { // consumer.SuperAdmin is needed for the local testing setup
+		log.C(ctx).Infof("The formation assignment that is being update has type: %s and ID: %q", model.FormationAssignmentTypeRuntime, fa.Target)
```
ivantenevvasilev, line 221: updated

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -0,0 +1,317 @@
...
+		if exists {
+			log.C(ctx).Infof("The caller with ID: %q and type: %q has owner access to the formation assignment with target: %q and target type: %q that is being updated", consumerInternalTenantID, consumerType, fa.Target, fa.TargetType)
+			return true, http.StatusOK, nil
+		}
+
+		return false, http.StatusUnauthorized, nil
+	}
+
+	if fa.TargetType == model.FormationAssignmentTypeRuntimeContext && (consumerType == consumer.Runtime || consumerType == consumer.ExternalCertificate || consumerType == consumer.SuperAdmin) { // consumer.SuperAdmin is needed for the local testing setup
+		log.C(ctx).Infof("The formation assignment that is being update has type: %s and ID: %q", model.FormationAssignmentTypeRuntimeContext, fa.Target)
```
ivantenevvasilev, line 237: updated

components/director/internal/formationmapping/auth_middleware.go
PetarTodorovv: This is also intentional. The whole idea here is to try to validate the consumer among different scenarios and if one of these scenarios fail we don't want to interrupt the validation and return but rather abstain and continue to the other checks.

tests/director/tests/formation_mapping_api_test.go
PetarTodorovv: I'm not sure we need this because if you look at the `CreateFormationFromTemplateWithinTenant` function there is `require.Equal(t, formationName, formation.Name)` check and at the mentioned point I believe we cannot receive nil value. Even if there is a problem and nil value it will fail in the `require.Equal(t, formationName, formation.Name)`. But I think that's the reason of `require.NoError(t, err)` check, to validate there is no error hence formation is not nil?

tests/director/tests/formation_api_test.go
```diff
@@ -103,8 +103,6 @@ func TestListFormations(t *testing.T) {
 	expectedFormations = 2
 	t.Logf("List should return %d formations", expectedFormations)
 	formationPage2 := fixtures.ListFormations(t, ctx, certSecuredGraphQLClient, listFormationsReq, expectedFormations)
-	require.NotNil(t, formationPage2)
```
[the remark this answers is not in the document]
PetarTodorovv: same comment as above 😄

tests/director/tests/formation_api_test.go
```diff
@@ -88,8 +90,6 @@ func TestListFormations(t *testing.T) {
 	listFormationsReq := fixtures.FixListFormationsRequestWithPageSize(first)
 	saveExample(t, listFormationsReq.Query(), "query formations")
 	formationPage1 := fixtures.ListFormations(t, ctx, certSecuredGraphQLClient, listFormationsReq, expectedFormations)
-	require.NotNil(t, formationPage1)
```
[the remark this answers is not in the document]
PetarTodorovv: This check is unnecessary because if you look at the `fixtures.ListFormations` function we have `require.NotEmpty(t, formationPage)` and `require.Equal(t, expectedCount, formationPage.TotalCount)` where not empty check validates not nil as well, from the go doc `NotEmpty asserts that the specified object is NOT empty.  I.e. not nil, "", false, 0 or ...`
So this will be handled in the function as well as the count and here I think we do not need this?

components/director/internal/formationmapping/auth_middleware.go
ivantenevvasilev: I understand, but in the case of an error with the connection to database, isn't it correct to return the error? Logging in the case of non found is ok.

components/director/internal/formationmapping/auth_middleware.go
ivantenevvasilev: It's more a distinction between potential error types rather than just returning it outright. If the tenant doesn't exist we don't want to return an error, but if the database call itself fails due to a networking issue I think it's correct to return the error.

PetarTodorovv: I take your point. Yep, I've adapted the two `if`s to return erros.

components/director/internal/formationmapping/auth_middleware.go
PetarTodorovv: see the comment above

dragobt (approved):

ivantenevvasilev (approved):

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -0,0 +1,317 @@
...
+	consumerType := consumerInfo.ConsumerType
+
+	tx, err := a.transact.Begin()
+	if err != nil {
+		return false, http.StatusInternalServerError, errors.Wrap(err, "Unable to establish connection with database")
+	}
+	defer a.transact.RollbackUnlessCommitted(ctx, tx)
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	fa, err := a.faService.GetGlobalByID(ctx, formationAssignmentID)
```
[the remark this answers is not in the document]
PetarTodorovv: It's already adapted in https://github.com/kyma-incubator/compass/pull/2722

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -0,0 +1,317 @@
...
+
+		app, err := a.appRepo.GetGlobalByID(ctx, fa.Target)
+		if err != nil {
+			return false, http.StatusInternalServerError, errors.Wrapf(err, "while getting application with ID: %q globally", fa.Target)
+		}
+		log.C(ctx).Infof("Successfully got application with ID: %q globally", fa.Target)
+
+		// If the consumer is integration system validate the formation assignment type is application that can be managed by the integration system caller
+		if consumerType == consumer.IntegrationSystem && app.IntegrationSystemID != nil && *app.IntegrationSystemID == consumerID {
+			log.C(ctx).Infof("The caller with ID: %q and type: %q has owner access to the formation assignment with target: %q and target type: %q that is being updated", consumerID, consumerType, fa.Target, fa.TargetType)
```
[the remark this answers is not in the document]
PetarTodorovv: I was wondering what will be the best way to log these operations because as you mentioned in the other comments the caller could be diffent, that's why on all of them additionally to the caller ID we have the type of it.

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -0,0 +1,317 @@
...
+				return true, http.StatusOK, nil
+			}
+		}
+		log.C(ctx).Warningf("The caller with ID: %q and type: %q has NOT direct owner access to formation assignment with type: %q and target ID: %q. Checking if the application is made through subscription...", consumerInternalTenantID, consumerType, fa.TargetType, fa.Target)
+
+		// Validate if the application is registered through subscription and the caller has owner access to that application
+		return a.validateSubscriptionProvider(ctx, app.ApplicationTemplateID, consumerExternalTenantID, string(consumerType), fa.Target, string(fa.TargetType))
+	}
+
+	if fa.TargetType == model.FormationAssignmentTypeRuntime && (consumerType == consumer.Runtime || consumerType == consumer.ExternalCertificate || consumerType == consumer.SuperAdmin) { // consumer.SuperAdmin is needed for the local testing setup
```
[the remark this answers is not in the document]
PetarTodorovv: why do we have two `if`s? I'm not sure about the context of the question but here we check the if the target type is runtime and do some validations/checks and below we have another validations/checks for the runtime context

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -0,0 +1,317 @@
...
+		return false, http.StatusUnauthorized, errors.Errorf("both %q and %q labels should be provided as part of the provider's application template", a.selfRegDistinguishLabelKey, a.consumerSubaccountLabelKey)
+	}
+
+	consumerSubaccountLblValue, ok := consumerSubaccountLbl.Value.(string)
+	if !ok {
+		return false, http.StatusUnauthorized, errors.Errorf("unexpected type of %q label, expect: string, got: %T", a.consumerSubaccountLabelKey, consumerSubaccountLbl.Value)
+	}
+
+	if consumerExternalTenantID == consumerSubaccountLblValue {
+		log.C(ctx).Infof("The caller with external ID: %q and type: %q has owner access to the formation assignment with target: %q and target type: %q that is being updated", consumerExternalTenantID, consumerType, faTarget, faTargetType)
```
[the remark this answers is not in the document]
PetarTodorovv: discuss offline and the logs are adapted

PR #2723, opened 2022-11-03
Send Slack notification when the Aggregation of ORD documents fails

emmakarova:
**Description**
Send new notifications to slack channels when aggregation of ORD documents fails

- new prometheus rule introduced

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/metrics/const.go
```diff
@@ -11,4 +11,12 @@ const (
 	TenantFetcherJobName = TenantFetcherSubsystem
 	// InstanceIDKeyName missing godoc
 	InstanceIDKeyName = "instance"
+	// OrdAggregatorSubsystem missing godoc
+	OrdAggregatorSubsystem = "ordaggregator"
```
alextargov, line 15: Can you add meaningful go docs for all new constants?

emmakarova: Added go doc for each constant

alextargov (approved):

krasish (approved):

PR #2735, opened 2022-11-14
user_context header encode/decode processing

PetarTodorovv:
**Description**
In the hydrator component and more specifically in the `consumre_context_provider` we could receive non-ascii characters in the `user_context` header. They will be escaped/encoded and we need to unescape/decode before using the data.

- Unescape the` user_context` header before data processing
- adapt unit and e2e tests

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/hydrator/internal/tenantmapping/consumer_context_provider.go
```diff
@@ -96,12 +107,30 @@ func (c *consumerContextProvider) Match(_ context.Context, data oathkeeper.ReqDa
 }
 
 func (c *consumerContextProvider) getUserContextData(userContextHeader string) (*userContextData, error) {
-	clientID := gjson.Get(userContextHeader, c.consumerClaimsKeysConfig.ClientIDKey)
+	decodedUserContextHeader, err := url.QueryUnescape(userContextHeader)
+	if err != nil {
```
DimitarPetrov, line 111: There is a fair amount of code duplication.

The same can be achieved with

```go
if decodedUserContextHeader, err := url.QueryUnescape(userContextHeader); err == nil {
    userContextHeader = decodedUserContextHeader
}
```

And everything below stays the same.

PR #2744, opened 2022-11-16
Cleanup image cache in update compass script

ivantenevvasilev:
**Description**

- remove cached images in the k3d instance nodes as it doesn't detect changes after a second execution of PR jobs

**Pull Request status**

- [x] Implementation
- [x] [N/A] Unit tests
- [x] [N/A] Integration tests
- [x]  [N/A] `chart/compass/values.yaml` is updated 


installation/scripts/prow/provision.sh
```diff
@@ -41,3 +41,22 @@ else
     sudo ${INSTALLATION_DIR}/cmd/run.sh --k3d-memory 12288MB
 fi
 
+# Patch deployment imagePullPolicy to always pull image on subsequent executions of update-compass.sh
```
DimitarPetrov, line 44: It would be faster from prow job PoV to do that patching on the charts level with some cli text editor like sed on the charts.

installation/scripts/prow/provision.sh
```diff
@@ -41,3 +41,22 @@ else
...
+       	kubectl patch -n $namespace "deployment/$DEPLOYMENT" -p "{\"spec\":{\"template\":{\"spec\":{\"containers\":[{\"name\":\"pairing-adapter\",\"imagePullPolicy\":\"Always\"}]}}}}"
+	    kubectl rollout restart -n $namespace "deployment/$DEPLOYMENT"
+        continue
+    fi
+    CONTAINER_NAME=$(echo "$DEPLOYMENT" | cut -d '-' -f2-)
+   	kubectl patch -n $namespace "deployment/$DEPLOYMENT" -p "{\"spec\":{\"template\":{\"spec\":{\"containers\":[{\"name\":\"$CONTAINER_NAME\",\"imagePullPolicy\":\"Always\"}]}}}}"
+	kubectl rollout restart -n $namespace "deployment/$DEPLOYMENT"
+done
+
+kubectl patch -n $namespace cronjob/compass-ord-aggregator -p '{"spec":{"jobTemplate":{"spec":{"template":{"spec":{"containers":[{"name":"ord-aggregator","imagePullPolicy":"Always"}]}}}}}}'
```
DimitarPetrov, line 59: Can we try to add some kind of a generic mechanism for jobs so that we can make sure that when someone introduces a job / deployment / statefulset or anything that uses compass docker images, he don't have to remember to put it here in order for prow jobs to work correctly.

installation/scripts/prow/update-compass.sh
```diff
@@ -9,4 +9,7 @@ set -o errexit
 CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 INSTALLATION_DIR=${CURRENT_DIR}/../../
 
+docker exec k3d-kyma-agent-0  sh -c "ctr images rm \$(ctr image list -q name~=eu.gcr.io/kyma-project/incubator)"
```
PetarTodorovv: From the man page - ```ctr is an unsupported debug and administrative client for interacting with the containerd daemon. Because it is unsupported, the commands, options, and operations are not guaranteed to be backward compatible or stable from release to release of the containerd project.```

Could be because of that in the future releases to encounter the same problem that this PR fix? the drawback of the problem is it's tricky to be found

PetarTodorovv (changes requested):

installation/scripts/prow/update-compass.sh
ivantenevvasilev: Wouldn't we have to explicitly bump the k3d version? If it updates and the command fails, then it's better to have the job fail than to give false positives/negatives. Also from what I understand is that they do not gurantee the backwards compatibility, but it's not very likely to happen. Plus in the logs we can see when and which images it has removed.

PetarTodorovv: That's valid, my point was if the "breaking change" is not a breaking but rather the command not working as expected or misbehaves and we encounter the same issue which is hard to notice in first glance. If the command directly fail that okay because we'll find out something is wrong and as you mentioned most probably there will be another option to fix it/delete the image.

I just saw that the `ctr` is unsupported and was wondering if will be an impact for us. But if you think there is no such probability or the impact is low and we can fast determine what's happening we can use it.

DimitarPetrov (approved):

PetarTodorovv (approved):

PR #2755, opened 2022-11-22
Expose aggregated formation status

la4ezar:
**Description**
Exposing aggregated status for our formation creation/deletion, which takes into account the status of the formation notifications sent as part of it.

- Expose a sub-resolver in the formation object that aggregates the status of all the assignments and returns it in a more compact way.
- Adapt the existing e2e tests to assert the formation condition

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/formation/resolver.go
```diff
@@ -352,3 +352,86 @@ func (r *Resolver) FormationAssignment(ctx context.Context, obj *graphql.Formati
...
+
+	ctx := keys[0].Ctx
+	formationIDs := make([]string, 0, len(keys))
+	for _, key := range keys {
+		formationIDs = append(formationIDs, key.ID)
+	}
+
+	var ( // These are the default values for listing
+		cursor = ""
+		first  = 200
```
nyordanoff, line 376: Setting a maximum amount of FAs that can be retrieved per formation is incorrect and may lead to invalid behaviour.

With the current implementation, it is assumed that each formation must have maximum 200 FAs which is not guaranteed at all. Assume that given formation has 201 FAs for the sake of the example -> 200 are in status READY and 1 is in status ERROR -> we will get the 200 READY and will assume formation status READY which is invalid.

We should change how we retrieve the FAs. I assume you kept the `cursor` and `first` because they are required for the `unionLister` that supports paging, but in this case we won't need any paging.

So, in the FA repository a new func has to added that retrieves all FAs for the given formationIDs with `IN` clause instead of a `UNION` query:

1. `select all FAs where formation_id IN (<formation IDs slice>)`
2. then map them accordingly (FAs per Formation) and return them
3. then in the resolver the status is calculated based on all FAs

tests/director/tests/formation_api_test.go
```diff
@@ -1021,6 +1021,7 @@ func TestFormationNotifications(stdT *testing.T) {
 		require.Equal(t, providerFormationName, assignedFormation.Name)
 		//list assignments - expect 0
 		assertFormationAssignments(t, ctx, subscriptionConsumerAccountID, formation.ID, 0, nil)
+		assertFormationStatus(t, ctx, subscriptionConsumerAccountID, formation.ID, graphql.FormationStatus{Condition: graphql.FormationStatusConditionReady, Errors: nil})
```
nyordanoff, line 1024: In all tests only two (`READY`, `ERROR`) of the possible three Status Conditions are asserted. I think it is worth adding a test if no such is present where we assert that the Status is `IN_PROGRESS`.

nyordanoff (changes requested):

nyordanoff (approved):

nyordanoff (approved):

PR #2757, opened 2022-11-22
Validate Application Template name based on provider

alextargov:
**Description**

- validate that Application Template name starts with SAP only when the provider is SAP

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/selfregmanager/selfreg_manager.go
```diff
@@ -82,83 +82,80 @@ func (s *selfRegisterManager) PrepareForSelfRegistration(ctx context.Context, re
 		return nil, errors.Wrapf(err, "while loading consumer")
 	}
 
-	if consumerInfo.Flow.IsCertFlow() {
```
PetarTodorovv, line 85: Is this an application template specific or used for a specific use-case, there is no provided description of the PR? I think for the runtime flow we still need this check because a certificate is used for the self registration but I might be missing something?

kaloyanspiridonov: actually the logic is the same, it is just a little bit softened. The region and the consumer id are retrieved from the certificate with the difference that they can also be extracted from the resource

PetarTodorovv: Yep, I see but I think we need that check for the runtime and this should not be softened for the runtime, maybe only for the application template. Offline follow-up discussion had been conducted regarding this case.

kaloyanspiridonov: we've decided to revert the change in selfreg_manager

kaloyanspiridonov:
reset-vm

PetarTodorovv (approved):

PR #2768, opened 2022-11-28
Soften app template name and displayName requirements

NickyMateev:
# Soften app template name and displayName requirements

**Description**

Previously when onboarding with app template we mandated two placeholders to be provided - `name` and `display-name` - both respectively used for the appInputJSON `name` property value and `labels['displayName']` value.

As of today the values of the `name` and `display-name` placeholders are provided by our team in an opinionated fashion as part of the saas callback handlers. We've already had a circumstance where a stakeholder would like to use different values for the `name` and `labels['displayName']` values. This forces us to soften the requirement in the sense that clients no longer need to necessarily provide the `name` and `display-name` placeholder, but they still do need to define the `name` and `labels['displayName']` properties - with whichever value they decide (non of our business).

- Only require the definition of `name` and `labels['displayName']` properties as part of appInputJSON of app template and don't concern ourselves with the actual value specfied.

**Pull Request status**

- [x] Implementation
- [x] Unit tests


chart/compass/values.yaml
```diff
@@ -127,7 +127,7 @@ global:
       name: compass-pairing-adapter
     director:
       dir:
-      version: "PR-2770"
+      version: "PR-2768"
```
la4ezar, line 130: Bump also e2e test version

la4ezar (changes requested):

chart/compass/values.yaml
NickyMateev: good catch 👍

la4ezar (approved):

components/director/internal/domain/apptemplate/resolver.go
```diff
@@ -449,7 +448,7 @@ func (r *Resolver) UpdateApplicationTemplate(ctx context.Context, id string, in
 		return nil, err
 	}
 	if isSelfRegFlow {
-		if err := validateAppTemplateForSelfReg(convertedIn.Placeholders); err != nil {
+		if err := validateAppTemplateForSelfReg(convertedIn.ApplicationInputJSON); err != nil {
```
kaloyanspiridonov, line 451: use "in.ApplicationInput" in order to avoid casting like - appName.(string), labels.(map[string]interface{})...

kaloyanspiridonov (changes requested):

components/director/internal/domain/apptemplate/resolver.go
NickyMateev: that's a good idea, yep 👍

la4ezar (approved):

kaloyanspiridonov (approved):

PR #2786, opened 2022-12-05
Add TenantsToSync env var to system fetcher

NickyMateev:
...


dzahariev:
replaced by https://github.com/kyma-incubator/compass/pull/2997

PR #2791, opened 2022-12-07
SaaS subscription refactoring

dzahariev:
**Description**
- SaaS subscription refactoring to use full subscription payload parsing it with placeholder's JSONPath property of application templates.

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/subscription/service.go
```diff
@@ -400,13 +409,31 @@ func (s *service) DetermineSubscriptionFlow(ctx context.Context, providerID, reg
...
 
-func (s *service) createApplicationFromTemplate(ctx context.Context, appTemplate *model.ApplicationTemplate, subscribedSubaccountID, consumerTenantID, subscribedAppName, subdomain, region string) error {
+func (s *service) createApplicationFromTemplate(ctx context.Context, appTemplate *model.ApplicationTemplate, subscribedSubaccountID, consumerTenantID, subscribedAppName, subdomain, region string, subscriptionPayload *string) error {
 	values := []*model.ApplicationTemplateValueInput{
-		{Placeholder: "name", Value: subscribedAppName},
-		{Placeholder: "display-name", Value: subscribedAppName},
 		{Placeholder: "subdomain", Value: subdomain},
 		{Placeholder: "region", Value: strings.TrimPrefix(region, RegionPrefix)},
 	}
+	if subscriptionPayload != nil && len(*subscriptionPayload) > 0 {
```
desislavaa, line 417: can we extract this whole part in a function like `additionalValues()`?
and then also reduce the if-elses

components/director/internal/tenantfetchersvc/handler_test.go
```diff
@@ -52,6 +52,17 @@ func TestService_SubscriptionFlows(t *testing.T) {
 	})
 	assert.NoError(t, err)
 
+	validRequestBodyWithMatchingParentID, err := json.Marshal(regionalTenantCreationRequest{
```
desislavaa, line 55: isn't this the same as `bodyWithMatchingAccountAndSubaccountIDs ` (with missing new prop, ocf)? we can rename the variable to this one, but the test case should be already covered

desislavaa (commented):
some minor comments and a question from my side - will E2E tests be added?

StanislavStefanov (approved):

desislavaa (dismissed):

components/director/pkg/graphql/schema.graphql
```diff
@@ -1738,7 +1738,7 @@ type Mutation {
 	writeTenant(in: BusinessTenantMappingInput!): String! @hasScopes(path: "graphql.mutation.writeTenants")
 	deleteTenants(in: [String!]): Int! @hasScopes(path: "graphql.mutation.deleteTenants")
 	updateTenant(id: ID!, in: BusinessTenantMappingInput!): Tenant! @hasScopes(path: "graphql.mutation.updateTenant")
-	subscribeTenant(providerID: String!, subaccountID: String!, providerSubaccountID: String!, consumerTenantID: String!, region: String!, subscriptionAppName: String!): Boolean! @hasScopes(path: "graphql.mutation.subscribeTenant")
+	subscribeTenant(providerID: String!, subaccountID: String!, providerSubaccountID: String!, consumerTenantID: String!, region: String!, subscriptionAppName: String!, subscriptionPayload: String): Boolean! @hasScopes(path: "graphql.mutation.subscribeTenant")
```
NickyMateev, line 1741: This is very minor - maybe it's a good idea to make the `subscriptionPayload` argument mandatory? (by using `String!`)

As far as I understand the only caller of this `subscribeTenant` mutation is the tenant fetcher in the context of SaaS  onSubscribe callbacks - these callbacks will always have a payload and in the code of the tenant fetcher if made it so that it is always sent along this mutation (as it should be). In this sense this property will always be provided?

NickyMateev: My only concern is someone in the future looking at this code and wondering why is the `subscriptionPayload` parameter not mandatory and wondering if there is some other use case where it is not provided :)

dzahariev:
reset-vm
/test pre-main-compass-integration-with-dump

tests/director/tests/main_test.go
```diff
@@ -61,6 +61,18 @@ type DirectorConfig struct {
...
+}
+
+type TenantProviderConfig struct {
+	TenantIDProperty                    string `envconfig:"APP_TENANT_PROVIDER_TENANT_ID_PROPERTY"`
+	SubaccountTenantIDProperty          string `envconfig:"APP_TENANT_PROVIDER_SUBACCOUNT_TENANT_ID_PROPERTY"`
+	CustomerIDProperty                  string `envconfig:"APP_TENANT_PROVIDER_CUSTOMER_ID_PROPERTY"`
+	SubdomainProperty                   string `envconfig:"APP_TENANT_PROVIDER_SUBDOMAIN_PROPERTY"`
+	SubscriptionProviderIDProperty      string `envconfig:"APP_TENANT_PROVIDER_SUBSCRIPTION_PROVIDER_ID_PROPERTY"`
+	ProviderSubaccountIDProperty        string `envconfig:"APP_TENANT_PROVIDER_PROVIDER_SUBACCOUNT_ID_PROPERTY"`
+	ConsumerTenantIDProperty            string `envconfig:"APP_TENANT_PROVIDER_CONSUMER_TENANT_ID_PROPERTY"`
```
PetarTodorovv, line 74: Where all of these are used? I found only the `SubscriptionProviderAppNameProperty` is used in the `formation_mapping_api_test.go` and `application_templates_subscription_test.go`? If they aren't used we could remove this structure + the environment variables in the chart?

tests/director/tests/formation_mapping_api_test.go
```diff
@@ -345,6 +351,20 @@ func Test_UpdateStatus(baseT *testing.T) {
...
+			{
+				Name:        "name",
+				Description: &placeholderName,
+				JSONPath:    &conf.TenantProviderConfig.SubscriptionProviderAppNameProperty,
+			},
+			{
+				Name:        "display-name",
+				Description: &placeholderDisplayName,
+				JSONPath:    &conf.TenantProviderConfig.SubscriptionProviderAppNameProperty,
+			},
```
PetarTodorovv, line 366: Why this is needed? We already have the `Name`, `Description` and `JSONPath` properties set in the fixture above - `fixtures.FixApplicationTemplateWithoutWebhooks`?

components/director/internal/domain/subscription/service.go
```diff
@@ -401,13 +417,43 @@ func (s *service) DetermineSubscriptionFlow(ctx context.Context, providerID, reg
...
-func (s *service) createApplicationFromTemplate(ctx context.Context, appTemplate *model.ApplicationTemplate, subscribedSubaccountID, consumerTenantID, subscribedAppName, subdomain, region string) error {
+func (s *service) createApplicationFromTemplate(ctx context.Context, appTemplate *model.ApplicationTemplate, subscribedSubaccountID, consumerTenantID, subscribedAppName, subdomain, region string, subscriptionPayload string) error {
 	values := []*model.ApplicationTemplateValueInput{
-		{Placeholder: "name", Value: subscribedAppName},
-		{Placeholder: "display-name", Value: subscribedAppName},
 		{Placeholder: "subdomain", Value: subdomain},
 		{Placeholder: "region", Value: strings.TrimPrefix(region, RegionPrefix)},
 	}
+
+	oldPlaceholders := appTemplate.Placeholders
```
NickyMateev, line 426: Can we extract the code from line 421 to 455 into a separate function - something like `preparePlaceholderValues` as currently it's a little hard to follow the parent `createApplicationFromTemplate` function flow because there are a lot of low level constructs like loops and re-assignments going on?

tests/director/tests/formation_mapping_api_test.go
dzahariev: I thought will be clear to replace the whole array, will replace only the JSONPath property here, as it is different than the one created with fix-function.

components/director/internal/domain/subscription/service.go
```diff
@@ -401,13 +417,43 @@ func (s *service) DetermineSubscriptionFlow(ctx context.Context, providerID, reg
...
+	if err != nil {
+		return errors.Wrapf(err, "while parsing the callback payload with the Application template %q", appTemplate.Name)
+	}
+
+	appTemplate.Placeholders = oldPlaceholders
+
+	if len(subscriptionPayload) > 0 {
+		values = append(appFromTemplateInput.Values, values...)
+	} else {
+		additionalValues := []*model.ApplicationTemplateValueInput{
```
NickyMateev, line 450: I think this code in the `else` statement is never going to be reachable in an e2e flow, because in theory there should always be a subscriptionPayload + there's a GraphQL requirement that a `subscriptionPayload: String!` is provided

dzahariev: Yes, I add this for compatibility when it was not mandatory, but will remove it now.

NickyMateev (dismissed):

NickyMateev (approved):

PetarTodorovv (dismissed):

StanislavStefanov (approved):

NickyMateev (approved):

StanislavStefanov (approved):

PetarTodorovv (approved):

desislavaa (approved):

PR #2792, opened 2022-12-08
Use LocalTenantId during ORD Aggregation

dragobt:
**Description**
- Pass `local tenant id` value to `Tenant_Id` header during ORD Aggregation
- Create `LocalTenantIDContextKey` under which the local tenant id is saved in a given context
- fix `webhook` typo 
- update dependencies
- adapt e2e tests

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [x] Integration tests


components/director/internal/domain/tenant/tenant.go
```diff
@@ -53,3 +57,23 @@ func SaveToContext(ctx context.Context, internalID, externalID string) context.C
...
+
+// LoadLocalTenantIDFromContext retrieves the local tenant ID from the provided context. It returns error if such ID cannot be found.
+func LoadLocalTenantIDFromContext(ctx context.Context) (string, error) {
+	localTenantID, ok := ctx.Value(LocalTenantIDContextKey).(string)
+
+	if !ok {
+		return "", apperrors.NewInternalError("cannot read local tenant id from context")
+	}
+
+	if localTenantID == "" {
```
dzahariev, line 69: Trim for empty characters before check `strings.TrimSpace(localTenantID)`

dzahariev (commented):
Also as we discuss you missed to commit a main.go file where the new operation is consumed.

dzahariev (dismissed):

components/director/pkg/accessstrategy/cmp_mtls_access_strategy.go
```diff
@@ -19,7 +19,7 @@ type HTTPRoundTripper interface {
 	GetTransport() *http.Transport
 }
 
-const tenantHeader = "tenant"
+const tenantHeader = "Tenant_Id"
```
PetarTodorovv, line 22: I approved the PR but I was wondering what is the motivation behind this change? Because until now we were using the  `tenant` key and the other side(whoever is it) has known to extract the value from it but now we're changing the header key, so I guess the other participant should adapt that from his side as well?

PetarTodorovv (approved):

components/director/pkg/accessstrategy/cmp_mtls_access_strategy.go
dragobt: the other side is aware of that change

PetarTodorovv (approved):

dzahariev (approved):

PR #2796, opened 2022-12-09
Introduce tenant scoped formation templates

nyordanoff:
**Description**

This PR introduces global account scoped formation types. These are formation templates that are visible only in a single GA. In a given GA, all global formation templates will be visible as well as the formation templates registered in that GA.

As a client one will be able to modify only FTs that are scoped for their GA. As an admin, one will be able to see and modify all global FTs without passing a tenant and respectively see and modify tenant scoped FTs when passsing a tenant header in the request.

- add `tenant_id` column to `formation_templates` table which will be populated for tenant scoped FTs and will be `null` for global ones
- FTs are now unique by their `(name, tenant_id)`
- introduce tenancy across the layers for the necessary FT operations
- adapt unit tests and add e2e test for the tenant scoped FTs

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


nyordanoff:
reset-vm

nyordanoff:
reset-vm

components/director/internal/domain/formationtemplate/repository.go
```diff
@@ -81,12 +93,37 @@ func (r *repository) Get(ctx context.Context, id string) (*model.FormationTempla
...
+	// If the call is with tenant but the query (select * from FT where name = ? and tenant_id = ?) returns NOT FOUND that means that there is no such tenant scoped FT and the call should get the global FT with that name.
+	//
+	// With this approach we allow the client to create a tenant scoped FT with the same name as a global FT - so when we get the FT first we will try to match it by name and tenant
+	// and if there is no such FT, we will get the global one
+	if tenantID == "" {
+		if err := r.singleGetterGlobal.GetGlobal(ctx, conditionsEqualNameAndNullTenant, repo.NoOrderBy, &entity); err != nil {
+			return nil, err
+		}
+	} else {
+		err := r.singleGetterWithEmbeddedTenant.Get(ctx, resource.FormationTemplate, tenantID, conditionsEqualName, repo.NoOrderBy, &entity)
```
la4ezar, line 116: You can make this `err..` one-liner. But if you think it will be less readable you can leave it this way.

components/director/internal/domain/formationtemplate/service.go
```diff
@@ -93,9 +120,71 @@ func (s *service) Update(ctx context.Context, id string, in *model.FormationTemp
...
+//		2. If it's GA -> return the GA id
+//		3. If it's a SA -> return its parent GA id
+func (s *service) extractTenantIDForTenantScopedFormationTemplates(ctx context.Context) (string, error) {
+	internalTenantID, err := s.getTenantFromContext(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	if internalTenantID == "" {
+		return internalTenantID, nil
```
la4ezar, line 168: Add unit test for this case.

la4ezar (changes requested):

tests/director/tests/formation_templates_test.go
```diff
@@ -196,3 +201,66 @@ func TestQueryFormationTemplates(t *testing.T) {
...
+
+func TestTenantScopedFormationTemplates(t *testing.T) {
+	ctx := context.Background()
+	first := 100
+
+	// Prepare provider external client certificate and secret and Build graphql director client configured with certificate
+	providerClientKey, providerRawCertChain := certprovider.NewExternalCertFromConfig(t, ctx, conf.ExternalCertProviderConfig, true)
+	directorCertSecuredClient := gql.NewCertAuthorizedGraphQLClientWithCustomURL(conf.DirectorExternalCertSecuredURL, providerClientKey, providerRawCertChain, conf.SkipSSLValidation)
+
+	scopedFormationTemplateName := "tenant-scoped-formation-template-test"
```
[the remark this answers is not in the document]
nyordanoff: Regarding the first point to create a FT with the same name as a global FT - we will be testing that the postgres UNIQUE constraint works which from my PoV is an overhead in the e2e tests.

Regarding the querying of that FT - currently, our API allows you to query a single FT by its ID or list them all. There is not exposed API for querying a FT by its name, thus checking which one will be returned so no such e2e test can be added.

la4ezar (approved):

components/director/internal/domain/formationtemplate/service.go
```diff
@@ -93,9 +120,71 @@ func (s *service) Update(ctx context.Context, id string, in *model.FormationTemp
...
+
+	if tenantObject.Type != tnt.Account && tenantObject.Type != tnt.Subaccount {
+		return "", errors.New("tenant used for tenant scoped Formation Templates must be of type account or subaccount")
+	}
+
+	if tenantObject.Type == tnt.Account {
+		return tenantObject.ID, nil
+	}
+
+	gaTenantObject, err := s.tenantSvc.GetTenantByID(ctx, tenantObject.Parent)
```
DimitarPetrov, line 184: Why don't you directly return tenantObject.Parent?

components/director/internal/domain/formationtemplate/service.go
DimitarPetrov, line 168: `return "", nil` is more explicit

components/director/internal/domain/formationtemplate/service.go
```diff
@@ -93,9 +120,71 @@ func (s *service) Update(ctx context.Context, id string, in *model.FormationTemp
...
 	}
 
 	return nil
 }
+
+// getTenantFromContext validates and returns the tenant present in the context:
+// 		1. if only one ID is present -> throw TenantNotFoundError
+// 		2. if both internalID and externalID are present -> proceed with tenant scoped formation templates (return the internalID from ctx)
+// 		3. if both internalID and externalID are not present -> proceed with global formation templates (return empty id)
+func (s *service) getTenantFromContext(ctx context.Context) (string, error) {
```
DimitarPetrov, line 139: This function would be easier to digest if you do it the other way around:
```go
       tntCtx, err := tenant.LoadTenantPairFromContextNoChecks(ctx)
 	if err != nil {
 		return "", err
 	}

 	if tntCtx.InternalID != "" && tntCtx.ExternalID != "" {
 		return tntCtx.InternalID, nil
 	}

 	if tntCtx.InternalID == "" && tntCtx.ExternalID == "" {
 		return "", nil
 	}
 	return "", apperrors.NewTenantNotFoundError(tntCtx.ExternalID)
```

components/director/internal/repo/delete.go
```diff
@@ -130,12 +130,14 @@ func (g *universalDeleter) delete(ctx context.Context, resourceType resource.Typ
 		return err
 	}
 
+	isTenantScopedUpdate := g.tenantColumn != nil
```
DimitarPetrov, line 133: Did you men `isTenantScopedDelete`?

components/director/internal/repo/testdb/generic_repo_get_tests.go
```diff
@@ -15,16 +15,20 @@ import (
...
+	ExpectedModelEntity                   interface{}
+	ExpectedDBEntity                      interface{}
+	MethodArgs                            []interface{}
+	AdditionalConverterArgs               []interface{}
+	DisableConverterErrorTest             bool
+	MethodName                            string
+	ExpectNotFoundError                   bool
+	AfterNotFoundErrorSQLQueryDetails     []SQLQueryDetails
+	AfterNotFoundErrorExpectedModelEntity interface{}
+	AfterNotFoundErrorExpectedDBEntity    interface{}
```
DimitarPetrov, line 31: Doesn't seems to be very generic feature for the generic suite. The assumption that if "NotFoundError" is expected then a new query will be executed is hard to imagine to be reusable. Even if it is, it would be in so few places that the cognitive overhead to grasp those new long and specific arguments won't be worth it every-time the suite is used (and most of the time not needed).

It could be introduced a bit less static as now it is specific for just one test and pollutes the suite.
You can introduce only the `ExpectNotFoundError bool` and configure it, if it is true, then the `returns not found error when no rows` will be branched to not expect not found error.

Then in the `SQLQueryDetails` you can add one more query for the NotFoundCase and rely on the generic success and error cases that are generated to handle that for you.

If that does not work out for whatever reason (since I am not much into the details, but it seems doable), it is better to me to move the tests as additional tests in the specific repo. Otherwise with a few more such overly specific arguments the suites may become too complex to be used and their purpose would be defeated.

components/schema-migrator/migrations/director/20221206114754_add-tenant-to-formation-templates.up.sql
```diff
@@ -0,0 +1,15 @@
+BEGIN;
+
+ALTER TABLE formation_templates ADD COLUMN tenant_id uuid;
+
+ALTER TABLE formation_templates
+    ADD CONSTRAINT formation_templates_tenant_fk
+        FOREIGN KEY (tenant_id) REFERENCES business_tenant_mappings (id) ON DELETE CASCADE;
```
DimitarPetrov, line 7: Currently, tenant scoped formation templates has ON DELETE CASCADE when a tenant is deleted.
Formations also has ON DELETE CASCADE when a tenant is deleted.
However, formations also has a FK to formation templates which does NOT have ON DELETE CASCADE.

This creates and ambiguous trigger execution path during our tenant orphan data deletion.

Let's imagine that a tenant is deleted, if this first deletes cascadely the formations and then the formation templates everything is fine. 

However, if it first deletes the formation templates, and there are formations of these types, this will fail and the whole tenant deprovision will fail.

PR #2798, opened 2022-12-12
Fix UnassignFormation in Async API

ivantenevvasilev:
**Description**
Due to the visibility and isolation between the transactions opened in the async API and the UnassignFormation call, whenever the UnassignFormation call is reached, it still sees the now-deleted formation assignment and does not Unassign the participant from the formation. The UnassignFormation call itself is successful, as this is the expected behaviour, so no error is returned.

Also unassigning just the `lastOperatorInitiator` can potentially leave applications/runtimes/runtime contexts in the formation that were called to be unassigned, but remained in the formation due to the leftover formation assignment. A check for when there are no formation assignments left for both the `Source` and `Target` is necessary when deleting the last formation assignment.
With the implementation of the `Source` and `Target` logic for unassigning in the API, we need to introduce a way to know which participants should be removed. One such way is to add formation assignments for each participant with `Source` and `Target` equal to the ID of the participant. This way we can be certain that `Unassign` has been called for the participant with the newly introduced formation assignment, as it will be directly deleted without having to be processed.

When unassigning from a formation some formation assignments are deleted which causes the mapping between notifications and formation assignments to leave the `ReverseAssignment` to `nil`, leaving notifications to not get sent, as they expect the `Assignment` and `ReverseAssignment` fields to be set.

With the introduction of the formation assignments with equal `Source` and `Target` we can now remove `last_operation_initiator`, `last_operation_initiator_type` and `last_operation` columns as the checks can be made based on the presence of the self assignments and state of the formation assignment that is stored in the database.

- commit transaction before calling `UnassignFormation`
- open a second transaction for `UnassignFormation`
- adapt unit tests to expect successful transaction
- check both `Source` and `Target` for unassigning when deleting last formation assignment
- add default values to `ApplicationTenantMappingInput` and `FormationConfigurationChangeInput`
- add missing checks for empty configuration in `ApplicationTenantMappingInput` and Async API conversion
- introduce checks in TestFormationAssignments e2e test after deleting formation assignment that the formation objects are indeed unassigned afterwards
- add a formation assignment with matching `Source` and `Target`
- adapt e2e tests to factor in the new formation assignments
- skip sending notifications for formation assignments with matchin `Source` and `Target`
- introduce a migration to add the formation assignments for existing formations
- drop `last_operation_initiator`, `last_operation_initiator_type` and `last_operation` columns
- adapt unit tests, repo layer and service layer to not use the dropped columns
- adapt change from checking `lastOperation` to check for the presence of formation assignments with matching `Source` and `Target` when checking if notifications should be generated for objects that are marked to be unassigned

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/formationmapping/handler.go
```diff
@@ -127,6 +131,33 @@ func (h *Handler) UpdateStatus(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		if isFADeleted {
```
PetarTodorovv, line 134: what is the idea of this `if`? It just extract part of the `h.processFormationAssignmentAsynchronousUnassign` into the handler function without any changes? Or I'm missing something? And the transaction seems to be the same

PetarTodorovv (changes requested):

PetarTodorovv (changes requested):

components/director/internal/formationmapping/handler.go
```diff
@@ -127,6 +131,46 @@ func (h *Handler) UpdateStatus(w http.ResponseWriter, r *http.Request) {
...
+					return
+				}
+				log.C(unassignCtx).Infof("Object with type: %q and ID: %q was successfully unassigned from formation with name: %q", lastOpInitiatorType, lastOpInitiatorID, f.Name)
+			}
+			if err = unassignTx.Commit(); err != nil {
+				log.C(ctx).WithError(err).Error("An error occurred while closing database transaction")
+				respondWithError(ctx, w, http.StatusInternalServerError, errResp)
+				return
+			}
+		}
```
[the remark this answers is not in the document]
PetarTodorovv: why we would commit twice? there is commit in the current situation but we'll remove it/transfer it to the `h.processFormationAssignmentAsynchronousUnassign `? Or I'm missing something?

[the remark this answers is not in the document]
ivantenevvasilev: Having the `.Commit()` done in a separate function when in all other cases it's done in `UpdateStatus` might be misleading. I think it's clear enough when the logic is now split into different functions.

components/schema-migrator/migrations/director/20221219120419_add-self-fa-and-drop-last-operation.up.sql
```diff
@@ -0,0 +1,19 @@
...
+    DROP COLUMN last_operation;
+
+INSERT INTO formation_assignments (id, formation_id, tenant_id, source, source_type, target, target_type, state)
+SELECT uuid_generate_v4(), fa.formation_id, fa.tenant_id, fa.source, fa.source_type, fa.source, fa.source_type, 'READY'
+FROM (
+         (
+             SELECT DISTINCT fa_source.formation_id, fa_source.tenant_id, fa_source.source, fa_source.source_type
+             FROM formation_assignments AS fa_source
+                      INNER JOIN formation_assignments AS fa_target
+                                 ON fa_source.source = fa_target.target AND fa_source.source_type = fa_target.target_type
```
DimitarPetrov, line 15: We should cross check the targets as well, otherwise imagine that we have the following formation - X, Y and Z.
X is being unassigned and everyone has async webhooks.

Then we have the following assignments: (X,Y), (Y,X), (X,Z), (Z,X), (Y,Z), (Z,Y)

Z has processed that X is being removed, however, X also processed that it is being removed from Y, however, Y still didn't processed that X is removed and X didn't process that is being removed from Z.

This will result in the following assignments:
(X,Y), (Z,X), (Y,Z), (Z,Y)

The first two will satisfy your condition and a (X,X) assignment will be created, which is not correct.

We should do full cross check `ON fa_source.source = fa_target.target AND fa_source.source_type = fa_target.target_type AND fa_source.target = fa_target.source AND fa_source.target_type = fa_target.source_type`

components/director/internal/formationmapping/handler.go
```diff
@@ -186,7 +200,7 @@ func (h *Handler) UpdateStatus(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	log.C(ctx).Infof("Generating reverse formation assignment notifications for ID: %q and formation ID: %q about last initiator ID: %q and type: %q", fa.ID, fa.FormationID, fa.LastOperationInitiator, fa.LastOperationInitiatorType)
+	log.C(ctx).Infof("Generating reverse formation assignment notifications for ID: %q and formation ID: %q about last initiator ID", fa.ID, fa.FormationID)
```
DimitarPetrov, line 203: Remove the `about last initiator ID` part as well.

components/director/internal/formationmapping/handler.go
```diff
@@ -226,6 +240,47 @@ func (h *Handler) UpdateStatus(w http.ResponseWriter, r *http.Request) {
...
+	return nil
+}
+
+func (h *Handler) unassignObjectFromFormationWhenThereAreNoFormationAssignments(ctx context.Context, fa *model.FormationAssignment, objectID string, objectType model.FormationAssignmentType) error {
+	formationAssignmentsForObject, err := h.faService.ListFormationAssignmentsForObjectID(ctx, fa.FormationID, objectID)
+	if err != nil {
+		return errors.Wrapf(err, "while listing formation assignments for object with type: %q and ID: %q", objectType, objectID)
+	}
+
+	if len(formationAssignmentsForObject) == 0 { // if there are no formation assignments left after the deletion, execute formation unassign for the last operation initiator
```
DimitarPetrov, line 268: Remove reference to last initiator from the comment.

tests/director/tests/formation_api_test.go
```diff
@@ -1978,102 +2083,286 @@ func TestFormationAssignments(stdT *testing.T) {
...
-		}
-		require.True(t, unassignNotificationFound)
+			require.True(t, unassignNotificationFound)
+
+			// rtCtx -> App notifications
+			assertNotificationsCountForTenant(t, body, localTenantID, 3)
+			notificationsForConsumerTenant = gjson.GetBytes(body, localTenantID)
+			assertSeveralFormationNotifications(t, notificationsForConsumerTenant, rtCtx, formation.ID, regionLbl, unassignOperation, 1)
+
+			//list assignments - expect 0
```
DimitarPetrov, line 2186: expect 1, but I would propose to remove those comments. Those are the kind of comments that exactly replicate the semantics of the code and it is inevitable to become outadate and therefore making more harm than good.

DimitarPetrov (changes requested):

DimitarPetrov (approved):

PR #2807, opened 2022-12-20
Adapt E2E tests to use new SaaS Registry subscription API

kaloyanspiridonov:
**Description**

-
- Adapt the new Subscription API for making subscriptions during e2e tests on dev (+ ext. svc. mock facade subscription APIs to look like this as well):
_**POST /saas-manager/v1/applications/{appName}/subscription**_
Currently we use in our e2e tests:
**_POST /saas-manager/v1/application/tenants/{tenantId}/subscriptions_**

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/external-services-mock/internal/subscription/handler_test.go
```diff
@@ -40,8 +40,8 @@ func NewTestClient(fn RoundTripFunc) *http.Client {
 
 func TestHandler_SubscribeAndUnsubscribe(t *testing.T) {
 	// GIVEN
-	consumerTenantID := "94764028-8cf8-11ec-9ffc-acde48001122"
-	apiPath := fmt.Sprintf("/saas-manager/v1/application/tenants/%s/subscriptions", consumerTenantID)
+	appName := "94764028-8cf8-11ec-9ffc-acde48001122"
```
la4ezar, line 43: Maybe change this to some name(for example, "test-app-name") and not UUID.

components/external-services-mock/internal/subscription/handler.go
```diff
@@ -23,8 +23,7 @@ type handler struct {
 }
 
 type JobStatus struct {
-	ID    string `json:"id"`
```
la4ezar, line 26: Why are we removing the `ID`?

kaloyanspiridonov: There is no ID in the response anymore:
Sample response:
{
	"status": "COMPLETED"
}

la4ezar (dismissed):

dragobt (approved):

tests/ord-aggregator/tests/handler_test.go
```diff
@@ -605,8 +605,7 @@ func TestORDAggregator(stdT *testing.T) {
 		subscriptionProviderSubaccountID := testConfig.TestProviderSubaccountID
 		subscriptionConsumerSubaccountID := testConfig.TestConsumerSubaccountID
 		subscriptionConsumerTenantID := testConfig.TestConsumerTenantID
```
PetarTodorovv, line 607: With the new APIs seems the `TestConsumerTenantID` is not used for the subscription call but rather the "application name"? If that's the case, could we remove that property from the tests configurations and adapt the logs as well? And the the yaml configuration too(locally and productively because it's a little bit misleading to have all of the setup/configuration + the logs and not to use them.

https://github.com/kyma-incubator/compass/blob/saas-url/chart/compass/values.yaml#L967

PetarTodorovv (changes requested):

PetarTodorovv (approved):

dragobt (approved):

PR #2808, opened 2022-12-21
Log subscribe payload

NickyMateev:


components/director/internal/tenantfetchersvc/subscriber.go
```diff
@@ -44,6 +46,7 @@ func (s *subscriber) applySubscriptionChange(ctx context.Context, subscriptionPr
 	var err error
 
 	if subscribe {
+		log.C(ctx).Infof("Subscribe payload:\n %s", subscriptionPayload)
```
PetarTodorovv, line 49: Is there any sensitive information as part of the payload? From the documentation I see there is `userInfo` property with `userID/name`, `email`, and other fields? Also there is a `additionalInformation` property that contains: `clientid`, `clientsecret` and `tokenurl` and this `additionalInformation` is present `ONLY for applications using XSUAA with "tenant-mode: external"`. I'm not sure which tenant mode we're using.

PR #2815, opened 2022-12-29
Ignore ORD resources with validation errors

emmakarova:
**Description**

When there are validation errors for ORD documents, the flow should continue and save in db only the valid resources. There should be concerns about the parent-child relations of the resources:
 - If a bundle is invalid -> the many-to-many relation with apis and events should not be created
 - If a package is invalid -> the related apis and events should not be stored as well

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/internal/open_resource_discovery/ord_document.go
+++ b/components/director/internal/open_resource_discovery/ord_document.go
@@ -2,7 +2,9 @@ package ord
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/url"
+	"reflect"
 	"regexp"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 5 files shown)


components/director/internal/open_resource_discovery/ord_document.go
```diff
@@ -138,19 +138,33 @@ func (docs Documents) Validate(calculatedBaseURL string, apisFromDB map[string]*
 		}
 	}
 
+	invalidApisIndices := make([]int, 0)
+	invalidEventIndices := make([]int, 0)
```
dragobt, line 142: change variable name to `invalidEventsIndices`. Do the same to the others bellow.

components/director/internal/open_resource_discovery/ord_document.go
```diff
@@ -201,27 +216,36 @@ func (docs Documents) Validate(calculatedBaseURL string, apisFromDB map[string]*
...
 			}
 			vendorIDs[vendor.OrdID] = true
 		}
-		for _, tombstone := range doc.Tombstones {
+
+		for i, tombstone := range doc.Tombstones {
 			if err := validateTombstoneInput(tombstone); err != nil {
 				errs = multierror.Append(errs, errors.Wrapf(err, "error validating tombstone with ord id %q", tombstone.OrdID))
+				invalidTombstoneIndices = append(invalidTombstoneIndices, i)
+				continue
```
dragobt, line 235: do we really need this `continue` here?

components/director/internal/open_resource_discovery/ord_document.go
```diff
@@ -280,15 +307,18 @@ func (docs Documents) Validate(calculatedBaseURL string, apisFromDB map[string]*
 				}
 			}
 		}
+
+		doc.APIResources = deleteInvalidInputObjects(invalidApisIndices, doc.APIResources)
+		doc.EventResources = deleteInvalidInputObjects(invalidEventIndices, doc.EventResources)
 	}
```
dragobt, line 313: should we do `invalidApisIndices = nil` and  `invalidEventIndices = nil` here as well?

components/director/internal/open_resource_discovery/ord_document.go
```diff
@@ -589,3 +619,14 @@ func stringPtrToString(p *string) string {
...
 	return ""
 }
+
+func deleteInvalidInputObjects[T any](invalidObjectsIndices []int, objects []T) []T {
+	decreaseIndexForDeleting := 0
+	for _, eventIndex := range invalidObjectsIndices {
```
dragobt, line 625: maybe consider renaming `eventIndex` to `invalidObjectIdx` ?

components/director/internal/open_resource_discovery/ord_document.go
emmakarova: Actually we don't need it, yes, because it will go to the next iteration even without it.

components/director/internal/open_resource_discovery/ord_document.go
emmakarova: Yes, we should, because if they aren't there, the indices from previous iterations will be stored in the next ones. Thanks for the comment  I added the clearing of the slices.

components/director/internal/open_resource_discovery/ord_document.go
emmakarova: That was a leftover from previous implementation, thanks for the observation I changed the name.

dragobt (approved):

PR #2817, opened 2023-01-05
Add local-tenant-id:write scope to external certificate

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -1060,6 +1060,7 @@ global:
         - "application_template:write"
         - "formation_template:read"
         - "formation_template:write"
+        - "application.local_tenant_id:write"
       application:
         - "application:read"
         - "application:write"


chart/compass/values.yaml
```diff
@@ -1060,6 +1060,7 @@ global:
         - "application_template:write"
         - "formation_template:read"
         - "formation_template:write"
+        - "application.local_tenant_id:write"
```
DimitarPetrov, line 1063: This would allow anyone to provide local tenant id to any system which means that one can register a system and steal another customer's system. A major security issue from my PoV which was the reason for this scope.

DimitarPetrov (changes requested):

chart/compass/values.yaml
alextargov: Yeah, we figured that out as well. We'll now merge it.

PR #2819, opened 2023-01-06
System fetcher does not update app templates

alextargov:
**Description**

- Historically, SF took into consideration only the ApplicationInputJSON object when it compares the two templates. However, when there are differences in the labels, placeholders, etc, but not in the ApplicationInputJSON, it does not update the template. This PR fixes that.

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/internal/systemfetcher/loader.go
+++ b/components/director/internal/systemfetcher/loader.go
@@ -341,10 +341,9 @@ func areAppTemplatesEqual(appTemplate *model.ApplicationTemplate, appTemplateInp
 
 	isAppInputJSONEqual := appTemplate.ApplicationInputJSON == appTemplateInput.ApplicationInputJSON
 	isLabelEqual := reflect.DeepEqual(appTemplate.Labels, appTemplateInput.Labels)
-	isWebhookEqual := (appTemplate.Webhooks == nil && appTemplateInput.Webhooks == nil) || reflect.DeepEqual(appTemplate.Webhooks, appTemplateInput.Webhooks)
 	isPlaceholderEqual := reflect.DeepEqual(appTemplate.Placeholders, appTemplateInput.Placeholders)
 
-	if isAppInputJSONEqual && isLabelEqual && isWebhookEqual && isPlaceholderEqual {
+	if isAppInputJSONEqual && isLabelEqual && isPlaceholderEqual {
 		return true
 	}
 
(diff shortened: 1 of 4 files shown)


components/director/internal/systemfetcher/loader.go
```diff
@@ -230,7 +230,7 @@ func (d *DataLoader) upsertAppTemplates(ctx context.Context, appTemplateInputs [
 			return errors.Wrapf(err, "while unmarshaling application input json")
 		}
```
kaloyanspiridonov, line 231: delete appInput1 and appInput2

components/director/internal/systemfetcher/loader.go
```diff
@@ -341,3 +342,20 @@ func retrieveRegion(labels map[string]interface{}) (string, error) {
 	}
 	return regionValue, nil
 }
+
+func areAppTemplatesEqual(appTemplate *model.ApplicationTemplate, appTemplateInput model.ApplicationTemplateInput) bool {
+	if appTemplate == nil {
+		return false
+	}
+
+	isAppInputJSONEqual := reflect.DeepEqual(appTemplate.ApplicationInputJSON, appTemplateInput.ApplicationInputJSON)
```
kaloyanspiridonov, line 351: do you really need DeepEqual for strings?

components/director/internal/systemfetcher/loader.go
```diff
@@ -341,3 +342,20 @@ func retrieveRegion(labels map[string]interface{}) (string, error) {
...
+	isLabelEqual := reflect.DeepEqual(appTemplate.Labels, appTemplateInput.Labels)
+	isWebhookEqual := (appTemplate.Webhooks == nil && appTemplateInput.Webhooks == nil) || reflect.DeepEqual(appTemplate.Webhooks, appTemplateInput.Webhooks)
```
kaloyanspiridonov, line 353: reflect.DeepEqual(nil, nil) returns true, so i think that (appTemplate.Webhooks == nil && appTemplateInput.Webhooks == nil) is redundant

kaloyanspiridonov (changes requested):

components/director/internal/systemfetcher/loader.go
alextargov: Not really. I've tested it previously. 
image

components/director/internal/systemfetcher/loader.go
alextargov: Yeah, you are right. Before this change, they were deepEqually checked and that's why I left it that way. But it seems that it is not necessary

components/director/internal/systemfetcher/loader.go
kaloyanspiridonov: Thats because you are comparing two objects of different types. []model.Webhook and []*model.WebhookInput

kaloyanspiridonov (approved):

PR #2825, opened 2023-01-12
Introduce tenant hierarchy to formation notifications payload

nyordanoff:
**Description**

This PR enriches formation notification payload with the tenant hierarchy of the customer creating the formation. A new field is added to the `FormationConfigurationChangeInput` and `ApplicationTenantMappingInput` for the tenant hierarchy of the customer creating the formation and is populated  with their external IDs when sending formation notifications.

- new recursive function that finds the customer parent of a given tenant
- new e2e test where tenants are selected so that the one creating a formation has a customer parent

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


nyordanoff:
reset-vm

nyordanoff:
reset-vm

components/director/internal/domain/formation/notifications.go
```diff
@@ -778,3 +799,29 @@ func (ns *notificationsService) createWebhookRequest(ctx context.Context, webhoo
 		CorrelationID: correlation.CorrelationIDFromContext(ctx),
 	}, nil
 }
+
+func (ns *notificationsService) extractCustomerTenantContext(ctx context.Context, internalTenantID string) (*webhookdir.CustomerTenantContext, error) {
```
dragobt, line 803: i am not very familiar with the structure of the code about the formations, but i noticed that this function has a "copy" of itself in the `/formationassignment/notifications.go`. Is it possible to define the function once and somehow both files use it?

components/director/internal/domain/formation/notifications.go
```diff
@@ -778,3 +799,29 @@ func (ns *notificationsService) createWebhookRequest(ctx context.Context, webhoo
...
+	tenantObject, err := ns.tenantRepository.Get(ctx, internalTenantID)
+	if err != nil {
+		return nil, err
+	}
+
+	var accountID *string
+	var path *string
+	if tenantObject.Type == tenant.Account {
+		accountID = &tenantObject.ExternalTenant
+	} else if tenantObject.Type == tenant.ResourceGroup {
```
dragobt, line 813: maybe unit test for this case

components/director/internal/domain/formationassignment/notifications.go
```diff
@@ -344,3 +358,29 @@ func convertFormationAssignmentFromModel(formationAssignment *model.FormationAss
...
+	customerID, err := fan.tenantRepository.GetCustomerIDParentRecursively(ctx, internalTenantID)
+	if err != nil {
+		return nil, err
+	}
+
+	var accountID *string
+	var path *string
+	if tenantObject.Type == tenant.Account {
+		accountID = &tenantObject.ExternalTenant
+	} else if tenantObject.Type == tenant.ResourceGroup {
```
dragobt, line 377: maybe unit test here as well

components/director/internal/domain/formationassignment/notifications.go
```diff
@@ -344,3 +358,29 @@ func convertFormationAssignmentFromModel(formationAssignment *model.FormationAss
...
 	}
 }
+
+func (fan *formationAssignmentNotificationService) extractCustomerTenantContext(ctx context.Context, internalTenantID string) (*webhook.CustomerTenantContext, error) {
+	tenantObject, err := fan.tenantRepository.Get(ctx, internalTenantID)
+	if err != nil {
+		return nil, err
+	}
+
+	customerID, err := fan.tenantRepository.GetCustomerIDParentRecursively(ctx, internalTenantID)
```
StanislavStefanov, line 368: The customerID may be empty string. Should we check for that and return error if empty?

components/director/internal/domain/tenant/repository.go
```diff
@@ -370,6 +370,45 @@ func (r *pgRepository) GetLowestOwnerForResource(ctx context.Context, resourceTy
...
+		ExternalCustomerTenant string `db:"external_tenant"`
+		Type                   string `db:"type"`
+	}{}
+
+	if err := persist.GetContext(ctx, &dest, recursiveQuery, tenantID); err != nil {
+		return "", persistence.MapSQLError(ctx, err, resource.Tenant, resource.Get, "while getting parent external customer ID for internal tenant: %q", tenantID)
+	}
+
+	if dest.Type != tenant.TypeToStr(tenant.Customer) {
+		return "", nil
```
StanislavStefanov, line 402: The funtion looks for CustomerID and such is not found. Maybe an error should be returned herer? Otherwise there should be checks for empty string whenever the function is invoked which is errorprone.

components/director/internal/domain/formationassignment/notifications.go
```diff
@@ -344,3 +358,29 @@ func convertFormationAssignmentFromModel(formationAssignment *model.FormationAss
...
+	if err != nil {
+		return nil, err
+	}
+
+	var accountID *string
+	var path *string
+	if tenantObject.Type == tenant.Account {
```
StanislavStefanov, line 375: Is it possible for the tenant to be sth different from Account or ResourceGroup?

components/director/internal/domain/formation/notifications.go
nyordanoff: The main thing is that this function has two different receivers in both cases (`notificationsService` and `formationAssignmentNotificationService`) that use their own repository layer functions.

Currently, the used repo layer functions coincide but we don't have any guarantee that in the future they will always be the same/with the same signature depending on the flow. That's why I separated them.

components/director/internal/domain/formationassignment/notifications.go
nyordanoff: No, this is the desired state. If `customerID` is empty, that means that the parent of `internalTenantID ` is not of type customer. In the notification payload we want to populate only the customer parent, so if there isn't such, we leave it empty.

components/director/internal/domain/tenant/repository.go
nyordanoff: If no customer tenant is found, we are OK with it - just return empty string and use in the notification payload.

components/director/internal/domain/formationassignment/notifications.go
nyordanoff: It shouldn't be, because both Formation/FormationAssignment are scoped to Account/ResourceGroup and we store that Account/ResourceGroup as tenant field in them.

For example, one should not be able to CRUD a Formation from a subaccount.

dragobt (approved):

desislavaa (approved):

StanislavStefanov (approved):

PR #2836, opened 2023-01-23
Application tenancy synchroization

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


chart/compass/templates/tests/director/director-test.yaml
```diff
@@ -43,6 +43,8 @@ spec:
               value: "https://{{ .Values.global.gateway.mtls.external.host }}.{{ .Values.global.ingress.domainName }}{{ .Values.global.director.prefix }}/graphql"
             - name: APP_DIRECTOR_EXTERNAL_CERT_FORMATION_MAPPING_ASYNC_URL
               value: "https://{{ .Values.global.gateway.mtls.external.host }}.{{ .Values.global.ingress.domainName }}{{ .Values.global.director.formationAsyncApi.pathPrefix }}{{ .Values.global.director.formationAsyncApi.path }}"
+            - name: APP_DIRECTOR_INTERNAL_URL
+              value: "https://{{ $.Values.global.gateway.tls.secure.internal.host }}.{{ $.Values.global.ingress.domainName }}{{ $.Values.global.director.prefix }}/graphql"
```
PetarTodorovv, line 47: Could we rename it `APP_DIRECTOR_INTERNAL_GATEWAY_URL` or something similar because with that name it seems to be "internal director" and it's an "external" through the gateway. There are two types of internal communication, so to say, internal(inside the cluster) and internal but the call is from the "outside" - istio, oathkeeper rules, etc. This seems to be the second one?

PetarTodorovv (changes requested):

components/director/internal/domain/scenarioassignment/resolver.go
```diff
@@ -29,7 +29,6 @@ type asaService interface {
 //go:generate mockery --exported --name=tenantService --output=automock --outpkg=automock --case=underscore --disable-version-string
 type tenantService interface {
 	GetExternalTenant(ctx context.Context, id string) (string, error)
-	CreateManyIfNotExists(ctx context.Context, tenantInputs ...model.BusinessTenantMappingInput) error
```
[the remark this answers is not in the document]
alextargov: Yeah, everything compiled without it.

chart/compass/values.yaml
```diff
@@ -110,6 +110,18 @@ global:
...
+      type: organization
+      parent: 2c4f4a25-ba9a-4dbc-be68-e0beb77a7eb0
+    - name: TestAtomFolder
+      id: 4c31b7c7-2bea-4bd5-9ea5-e9a8d704f900
+      type: folder
+      parent: f2724f8e-1a58-4f32-bfd0-8b831de34e71
+    - name: TestAtomResourceGroup
+      id: ff30da87-7685-4462-869a-baae6441898b
+      type: resource-group
+      parent: 4c31b7c7-2bea-4bd5-9ea5-e9a8d704f900
```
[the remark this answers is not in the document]
alextargov: Wouldn't the tenant loader job insert these as well when an update is happening?

chart/compass/templates/tests/director/director-test.yaml
alextargov: I use this one for a new client in the tests using the service account token instead of the cert client. I've copied the env variable from another place in the tests. But if it makes more sense, it wouldn't be a problem to rename it.

PetarTodorovv: Maybe the other place that you've copied is not correctly named or perhaps it's a different case, nevertheless, such name differentiation between the two types of internal clients/URLs would be a nice thing to have.

chart/compass/values.yaml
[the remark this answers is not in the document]
PetarTodorovv: Yep, that's my assumption from the above comment but I've seen cases when the tenant loader doesn't insert them. So I left the comment as a heads-up/note to be double-checked and if that's the case to be inserted manually.

PetarTodorovv (dismissed):

components/director/pkg/applicationtenancy/directive.go
```diff
@@ -0,0 +1,234 @@
...
+
+		tntModel, err := d.tenantService.GetTenantByID(ctx, tntIDFromContext)
+		if err != nil {
+			log.C(ctx).WithError(err).Errorf("An error occurred while getting tenant with ID %s", tntIDFromContext)
+			return nil, errors.Wrapf(err, "while fetching tenant with id: %s", tntIDFromContext)
+		}
+
+		log.C(ctx).Debugf("Found a matching tenant in the database: %s", tntModel.ID)
+
+		if !(tntModel.Type == tenantpkg.ResourceGroup || tntModel.Type == tenantpkg.Folder || tntModel.Type == tenantpkg.Organization) {
```
kaloyanspiridonov, line 86: tntModel.Type != tenantpkg.ResourceGroup && tntModel.Type != tenantpkg.Folder && tntModel.Type != tenantpkg.Organization) is more readable for me

kaloyanspiridonov, line 86: Or extract to method isAtomTenant()

components/director/pkg/applicationtenancy/directive.go
```diff
@@ -0,0 +1,234 @@
...
+			log.C(ctx).Infof("Tenant type is %s. Will not continue with tanancy synchronization", tntModel.Type)
+			return resp, nil
```
kaloyanspiridonov, line 88: Have in mind that if you return a successful response without Commit, a transaction rollback error will appear in the logs.

components/director/pkg/applicationtenancy/directive.go
```diff
@@ -0,0 +1,234 @@
...
+			return nil, err
+		}
+	} else if eventType == graphql.EventTypeNewMultipleTenants {
+		tenantIDs, ok := resp.([]string)
+		if !ok {
+			log.C(ctx).WithError(err).Errorf("An error occurred while casting the response entity: %v", err)
+			return nil, apperrors.NewInvalidDataError("An error occurred while casting the response entity: %v", resp)
+		}
+
+		filteredTenants := make([]*model.BusinessTenantMapping, 0)
```
kaloyanspiridonov, line 126: [Code duplication] Extract the code for eventType == graphql.EventTypeNewSingleTenant into a method processSingleTenantEvent(tenantID) and then
for _, tenantID := range tenantIDs{
   processSingleTenantEvent(tenantID) 
}

components/director/pkg/applicationtenancy/directive.go
```diff
@@ -0,0 +1,234 @@
...
+		log.C(ctx).WithError(err).Errorf("An error occurred while opening database transaction")
+		return nil, apperrors.NewInternalError("Unable to initialize database transaction: %s", err.Error())
+	}
+	defer d.transact.RollbackUnlessCommitted(ctx, tx)
+
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	log.C(ctx).Debugf("Preparing tenant access creation for event: %s", eventType)
+
+	if eventType == graphql.EventTypeNewApplication {
```
kaloyanspiridonov, line 71: you can extract every case in a separate method

kaloyanspiridonov (changes requested):

components/director/internal/domain/root_resolver.go
```diff
@@ -882,7 +882,7 @@ func (r *mutationResolver) CreateAutomaticScenarioAssignment(ctx context.Context
 }
 
 // WriteTenants creates tenants of type customer, account, subaccount, organization, folder, or resource-group
-func (r *mutationResolver) WriteTenants(ctx context.Context, in []*graphql.BusinessTenantMappingInput) (int, error) {
```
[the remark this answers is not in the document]
alextargov: I've traced the usage of it and I think it is safe. https://github.com/kyma-incubator/compass/blob/4e0bd66b0684992b1b6f3a26637942373014c2ba/components/director/pkg/graphql_client/director_client.go#L43 This snippet is being used by the tenant fetcher service. And it does not care about the result of the request

components/director/internal/domain/tenant/service.go
```diff
@@ -169,15 +174,42 @@ func (s *service) Update(ctx context.Context, id string, tenantInput model.Busin
 	return nil
 }
 
+func (s *service) GetCustomerIDParentRecursively(ctx context.Context, tenantID string) (string, error) {
```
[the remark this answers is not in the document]
alextargov: I'm just extracting the repo layer func to the service. So I guess it would be better to leave the naming as it is. But if you have strong feelings about this, I could rename the svc function.

components/director/internal/domain/root_resolver.go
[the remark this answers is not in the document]
kaloyanspiridonov: i suppose that it is not possible someone outside ucl to use this mutation?

components/director/internal/domain/tenant/service.go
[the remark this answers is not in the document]
kaloyanspiridonov: My point was that i don't care if it is "Recursively"  or not, but i am okey to leave it as it is.

kaloyanspiridonov (approved):

krasish (approved):

PetarTodorovv (approved):

PR #2854, opened 2023-01-30
Skip sending duplicate tenant mapping notifications

DimitarPetrov:
**Description**

Currently, when a notification receiver (forward assignment) returns config synchronously, we recursively process the reverse assignment. We should not process it a second time, as it will result in out-of-order notifications (first with config then without) in the sync case.

This is unavoidable in the async case and the receiver should take care of the wrong order. However, in the sync case, we can do it on their behalf.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


DimitarPetrov:
reset-vm

PetarTodorovv (approved):

PR #2872, opened 2023-02-07
ORD Aggregation of C4C CNS fails with local tenant id is required

dragobt:
**Description**

- send tenant header only when webhook is associated with app template
- update dependencies

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


chart/compass/values.yaml
```diff
@@ -139,7 +139,7 @@ global:
       name: compass-pairing-adapter
     director:
       dir:
-      version: "PR-2868"
```
la4ezar, line 142: Update the versions of all other components, because you've changed the go.sum/mod files.

la4ezar (changes requested):

la4ezar (dismissed):

alextargov (approved):

la4ezar (approved):

PR #2886, opened 2023-02-16
Introduce Formation Async Status API Skeleton

PetarTodorovv:
**Description**
Introducing async formation status API skeleton. The idea of which is when there is `formation_lifecycle` webhook attached to the formation and notification(s) are sent to the interested party, they can return a status asynchronously to that API.

- Adapt the current async status API to handle both - formation assignment and formation status update requests
- New auth middleware handler with the necessary authn/authz check for the formation status updates
- Adding new unit tests + adaptation of the old ones
- Negative authorization E2E test

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


PetarTodorovv:
reset-vm

tests/director/tests/formation_mapping_api_test.go
```diff
@@ -546,8 +595,22 @@ func getFormationAssignmentIDByTargetTypeAndSourceID(t *testing.T, assignmentsPa
...
-	formationMappingEndpoint := strings.Replace(conf.DirectorExternalCertFormationMappingURL, fmt.Sprintf("{%s}", formationIDPathParam), formationID, 1)
-	formationMappingEndpoint = strings.Replace(formationMappingEndpoint, fmt.Sprintf("{%s}", formationAssignmentIDPathParam), formationAssignmentID, 1)
-	return formationMappingEndpoint
+func resolveFAAsyncStatusAPIURL(formationID, formationAssignmentID string) string {
+	faAsyncStatusAPIURL := strings.Replace(conf.DirectorExternalCertFAAsyncStatusURL, fmt.Sprintf("{%s}", formationIDPathParam), formationID, 1)
+	faAsyncStatusAPIURL = strings.Replace(faAsyncStatusAPIURL, fmt.Sprintf("{%s}", formationAssignmentIDPathParam), formationAssignmentID, 1)
+	return faAsyncStatusAPIURL
+}
+
+func createFormationTemplateWithLeadingProductIDs(t *testing.T, ctx context.Context, formationTemplateName, runtimeType string, applicationTypes []string, runtimeArtifactKind graphql.ArtifactType, leadingProductIDs []*string) graphql.FormationTemplate {
```
StanislavStefanov, line 604: Should be in tests/pkg/fixtures/formation_template_queries.go

StanislavStefanov (changes requested):

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -172,8 +190,92 @@ func (a *Authenticator) Handler() func(next http.Handler) http.Handler {
...
+				w.WriteHeader(http.StatusMethodNotAllowed)
+				return
+			}
+
+			routeVars := mux.Vars(r)
+			formationID := routeVars[FormationIDParam]
+
+			if formationID == "" {
+				log.C(ctx).Errorf("Missing required parameters: %q", FormationIDParam)
+				respondWithError(ctx, w, http.StatusBadRequest, errors.New("Not all of the required parameters are provided"))
```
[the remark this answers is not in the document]
PetarTodorovv: I haven't included it because the parameter in the code(`ucl-formation-id`) is slightly different from the one in our swagger documentation(`uclFormationId `). And if we include it in the error message then we need to take care when the swagger is updated to update our code too. We could forget, and the error message could bring more confusion than help.

StanislavStefanov (approved):

NickyMateev (approved):

components/director/internal/formationmapping/handler.go
```diff
@@ -300,8 +362,20 @@ func (b RequestBody) Validate() error {
 	return nil
 }
 
+// Validate validates the formation's request body input
+func (b FormationRequestBody) Validate() error {
+	var fieldRules []*validation.FieldRules
+	fieldRules = append(fieldRules, validation.Field(&b.State, validation.In(model.ReadyFormationState, model.CreateErrorFormationState, model.DeleteErrorFormationState)))
+
+	if b.Error != "" {
+		fieldRules = append(fieldRules, validation.Field(&b.State, validation.In(model.CreateErrorFormationState, model.DeleteErrorFormationState)))
```
DimitarPetrov, line 371: You are adding those states to the field rules when an error is present, this means that if error is present you allow all 5 states.

I believe the correct way to implement this is:
```go
func (b FormationRequestBody) Validate() error {
	return validation.ValidateStruct(&b,
		validation.Field(&b.State,
			validation.When(len(b.Error) > 0, validation.In(model.CreateErrorFormationState, model.DeleteErrorFormationState)),
			validation.When(len(b.Error) == 0, validation.In(model.ReadyFormationState, model.CreateErrorFormationState, model.DeleteErrorFormationState))))
}
```

Maybe this is valid for the FA validation as well.

PR #2899, opened 2023-02-21
Switch helm driver to sql on local env

kaloyanspiridonov:
**Description**

For every helm chart, helm 3 generates a new secret.
The purpose of this auto generated secret is for recording release information and its size is correlated with the size of the helm chart.
The default release information storage for Helm 3 is Secret (in the namespace of the release).
In k8s the individual secrets are limited to 1MiB in size. Which means that if our chart become bigger than 1MB we will hit that limit.

In order to avoid hitting this limit we have to change the default helm storage backend.
The supported storage backends for helm are - secret, configmap, sql
The secrets and configmaps are limited to 1 MiB in size. So the only available option for us is to use SQL as a storage backend.

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/docs/compass/04-01-installation.md
+++ b/docs/compass/04-01-installation.md
@@ -124,12 +124,12 @@ global:
 Start the Database installation by using the following command:
 
 ```bash
-<script from ../../installation/scripts/install-db.sh> --overrides-file <file from ../../installation/resources/compass-overrides-local.yaml> --overrides-file <file from above step - e.g. additionalCompassOverrides.yaml> --timeout <e.g: 30m0s> --sql-helm-backend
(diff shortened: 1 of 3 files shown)


docs/compass/04-01-installation.md
```diff
@@ -124,7 +124,7 @@ global:
 Start the Database installation by using the following command:
 
 ```bash
-<script from ../../installation/scripts/install-db.sh> --overrides-file <file from ../../installation/resources/compass-overrides-local.yaml> --overrides-file <file from above step - e.g. additionalCompassOverrides.yaml> --timeout <e.g: 30m0s>
+<script from ../../installation/scripts/install-db.sh> --overrides-file <file from ../../installation/resources/compass-overrides-local.yaml> --overrides-file <file from above step - e.g. additionalCompassOverrides.yaml> --timeout <e.g: 30m0s> --sql-helm-backend
```
DimitarPetrov, line 127: Looks like the wrong line `install-db.sh`

nkanchev (approved):

PR #2901, opened 2023-02-22
Filter ORD Service response based on formations for SaaS application consumers

nyordanoff:
**Description**

Currently, ORD Service provides filtering only by tenant. This means that we can discover all the systems that are in a given tenant (Customer, global account or subaccount). However, for the subaccount we had a custom logic that checks all the systems that are in a formation with the subaccount and enrich the subaccount response with those systems. (Systems that are registered in the GlobalAccount, which owns the Subaccount, and that have a configured Formation between the system and the given Subaccount.)

This was enough when formations were not typed and anything could be in any formation. However, with formation types we introduced various restrictions over different formations.

The expected behaviour is, as a acapability provider for a formation type to receive only systems that are in those kind of formations, which will handle all the restrictions applied to that formation type.

- in the `apps_subaccounts view` we need to return not only the ID of the system and the tenant in formation with it, but also the formation_id
- in the other views using `apps_subaccounts` we need to not have only the resource + the tenant_id, but also the formation_id
- remove JOIN from `tenants_destinations` because it was obsolete from the beginning but was present by mistake
- adapt consumer provider E2E tests in order to validate that the formation based discovery (filtering) is in place. 

**Related issue(s)**

- https://github.com/kyma-incubator/ord-service/pull/87

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [x] Integration tests


nyordanoff:
reset-vm

components/schema-migrator/migrations/director/20230223160000_ord_filter_by_formations.up.sql
```diff
@@ -0,0 +1,438 @@
...
+
+DROP VIEW IF EXISTS apps_subaccounts;
+
+-- enhance the apps_subaccounts view to return not only the ID of the application and the tenant ID (subaccount) in formation with it but also the formation ID
+
+CREATE VIEW apps_subaccounts(id, tenant_id, formation_id)
+AS
+SELECT l.app_id                 AS id,
+       asa.target_tenant_id AS tenant_id,
+       (SELECT id FROM formations f WHERE name = asa.scenario AND (f.tenant_id = asa.tenant_id OR f.tenant_id = asa.target_tenant_id)) as formation_id
```
PetarTodorovv, line 24: `f.tenant_id = asa.tenant_id OR f.tenant_id = asa.target_tenant_id` why we're checking both `tenant_id` and `target_tenant_id`. Shouldn't the `f.tenant_id = asa.tenant_id` check be enough?

tests/ord-service/tests/subscription_flow_test.go
```diff
@@ -181,16 +187,18 @@ func TestConsumerProviderFlow(stdT *testing.T) {
...
-		createFormationReq := fixtures.FixCreateFormationRequest(consumerFormationName)
-		executeGQLRequest(stdT, ctx, createFormationReq, consumerFormationName, secondaryTenant)
-		defer func() {
-			stdT.Logf("Deleting formation with name: %s...", consumerFormationName)
-			deleteRequest := fixtures.FixDeleteFormationRequest(consumerFormationName)
-			executeGQLRequest(stdT, ctx, deleteRequest, consumerFormationName, secondaryTenant)
-			stdT.Logf("Successfully deleted formation with name: %s...", consumerFormationName)
-		}()
+		stdT.Logf("Creating formation with name: %q from template with name: %q", consumerFormationName, formationTmplName)
+		formation := fixtures.CreateFormationFromTemplateWithinTenant(stdT, ctx, certSecuredGraphQLClient, secondaryTenant, consumerFormationName, &formationTmplName)
```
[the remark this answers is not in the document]
nyordanoff: It didn't work with the old setup (using the default formation template) because after the `subscriptionConsumerSubaccountID` tenant was assigned to formation and then a subscription was created (resulting in a runtimeContext), there was no record in the `labels` table indicating that the runtimeContext is in that formation (`runtime_context_id is not null and key='scenarios'`). This resulted in error in the ORD Service because there it entered the consumer-provider flow -> checked that there is a runtimeContext present -> tries to get all the formations and set them as claims it is part of (none due to the issue) -> and then there is validation to fail if no formations/claims were found.

The reason why no labels with `key='scenario'` were created for the runtimeContext was that there was a mismatch of the runtimeTypes of the default formation Template and the actual one from the test (`conf.SubscriptionProviderAppNameValue`) and then the ASA processing logic was broken.

That part with the default formation template was just not adapted like the other similar places across the tests because with the old ORD flow, all that was needed  was an ASA but now we need the formations as well.

components/schema-migrator/migrations/director/20230223160000_ord_filter_by_formations.up.sql
nyordanoff: On paper it should be enough, but I checked that some formations have subaccount as their tenant on all landscapes:

`select distinct type from business_tenant_mappings where id in (select tenant_id from formations)` returns `account & subaccount`

So, I decided to add both tenants in the query to cover all cases.

PetarTodorovv: Hm, most probably these formations are for testing purposes? We should not have productive formations in subaccount? And if the data is for testing purposes maybe we shouldn't adapt the productive code based on that? We could sync it offline as well.

tests/ord-service/tests/subscription_flow_test.go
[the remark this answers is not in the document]
PetarTodorovv: I get it, it was a hidden trap :D I was wondering is it a good idea to change the default `conf.SubscriptionProviderAppNameValue` to be equal to the default formation template's runtimeType?

components/schema-migrator/migrations/director/20230223160000_ord_filter_by_formations.up.sql
nyordanoff: Discussed offline - will remove the `f.tenant_id = asa.target_tenant_id` part of the query because it shouldn't happen in a productive flow.

tests/ord-service/tests/subscription_flow_test.go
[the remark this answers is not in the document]
nyordanoff: Discussed offline - for the time being no adaption is needed, we will create FT and formations in the tests with the proper configuration.

PetarTodorovv (approved):

PR #2907, opened 2023-02-23
Support optional application template placeholders

emmakarova:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


dzahariev (approved):

components/director/internal/domain/apptemplate/service.go
```diff
@@ -321,7 +321,14 @@ func (s *service) PrepareApplicationCreateInputJSON(appTemplate *model.Applicati
...
 		newValue, err := values.FindPlaceholderValue(placeholder.Name)
-		if err != nil {
+		var optional bool
+		if placeholder.Optional == nil {
+			optional = false
+		} else {
+			optional = *placeholder.Optional
+		}
+
+		if err != nil && !optional {
```
alextargov, line 331: This can be simplified:
```suggestion
		isOptional := false
		if placeholder.Optional != nil {
			isOptional = *placeholder.Optional
		}

		if err != nil && !isOptional {
```

What do you think?

emmakarova: Yes, it looks better when simplified. I will refactor it in another PR so that this one can be released.

PR #2918, opened 2023-02-28
Allow Formation Template registration in ResourceGroup tenants

DimitarPetrov:
**Description**

- Allow Formation Template registration in ResourceGroup tenants

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/formationtemplate/service.go
```diff
@@ -192,9 +192,8 @@ func (s *service) getTenantFromContext(ctx context.Context) (string, error) {
 }
 
 // extractTenantIDForTenantScopedFormationTemplates returns the tenant ID based on its type:
-//		1. If it's not SA or GA -> return error
```
nyordanoff, line 195: The same adaptation has to be applied [here](https://github.com/kyma-incubator/compass/blob/main/components/director/internal/domain/webhook/service.go#L257) because that func is present in the webhook service as well.

At the time when we had to use it for the webhooks as well, we (me and the reviewer) didn't come up with a 'clean' approach how to extract it and that's why it is double maintained.

As far as I remember, our concern was that if we extracted the func to the pkg, we had to pass the tenantSvc as param, but it will have to be different for both flows and we do not have any guarantee that in the future the exact same func from tenantSvc will be used/needed in both flows.

nyordanoff (changes requested):

components/director/internal/domain/formationtemplate/service.go
PetarTodorovv: Couldn't we extract that function in the "tenant service" layer? that way in both - "formation template service" and "webhook service" we have "tenant service" and we could reuse it? Also, it will be easier for testing because I saw it's used in different functions in the "formation template service" - `Create`, `List`, `Update`, `Delete` and `ListWebhooksForFormationTemplate`

PetarTodorovv (approved):

nyordanoff (approved):

PR #2929, opened 2023-03-06
Extend registerApplicationFromTemplate mutation with labels

dragobt:
**Description**

- extend the registerApplicationFromTemplate mutation to have labels
- extend the Application type to have Application Template
- adapt unit tests
- generate the graphql schema
- update dependencies

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/application/resolver.go
```diff
@@ -785,6 +809,37 @@ func (r *Resolver) Bundle(ctx context.Context, obj *graphql.Application, id stri
 	return gqlBundle, nil
 }
 
+// ApplicationTemplate missing godoc
+func (r *Resolver) ApplicationTemplate(ctx context.Context, obj *graphql.Application) (*graphql.ApplicationTemplate, error) {
+	if obj == nil {
+		return nil, apperrors.NewInternalError("Application cannot be empty")
+	}
+	if obj.ApplicationTemplateID == nil {
+		return nil, apperrors.NewInternalError("Application does not contain app template id")
```
alextargov, line 818: IMO, we should not throw an error here but rather return an empty ApplicationTemplate object or return nil - depends on the implementation. But why we shouldn't return an error is the more essential question - if somebody has used the `registerApplication` mutation and not the `registerApplicationFromTemplate` mutation, then the created App will not have AppTemplateID. When I want to execute the `applications` graphQL query for my tenant, I'll get an error because of this line here. 
So in order not to experience such behavior, the error should be omitted and should be `nil` instead.
Here's an example: 
image
I had 2 applications in the GA - one created from a template and one created from scratch. Because of this line here, I got both an error for the app from scratch and an object for the app from template

PetarTodorovv (dismissed):
From the team-raptor perspective and components, the PR looks okay.

alextargov (changes requested):

nkanchev (approved):

PetarTodorovv (approved):

dragobt:
retest

nkanchev (approved):

PR #2947, opened 2023-03-10
Allow BI to create formations in INITIAL state even if webhook is not present

DimitarPetrov:
**Description**

- Allow BI to create formations in INITIAL state even if webhook is not present

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/formation/service_test.go
```diff
@@ -247,6 +247,14 @@ func TestServiceCreateFormation(t *testing.T) {
 		State:               model.ReadyFormationState,
 	}
 
+	expectedFormationInInitialState := &model.Formation{
```
PetarTodorovv, line 250: **_minor:_** Could we use a fixture here - `expectedFormationInInitialState := fixFormationModelWithState(model.InitialFormationState)`. ?

PetarTodorovv (approved):

PR #2948, opened 2023-03-13
Add IAS Adapter

ZdravkoGyurov:
**Description**

- Add IAS Adapter

**Related issue(s)**

-

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/ias-adapter/internal/service/ias/service.go
```diff
@@ -0,0 +1,181 @@
...
+				Value:     consumedAPIs,
+			},
+		},
+	}
+	appUpdateBytes, err := json.Marshal(appUpdate)
+	if err != nil {
+		return errors.Newf("failed to marshal body: %w", err)
+	}
+	url := buildPatchApplicationURL(iasHost, applicationID)
+	timeoutCtx, cancel := context.WithTimeout(ctx, s.cfg.RequestTimeout)
```
petar-iv, line 150: Would be nice to encapsulate the timeout in a method - in this way will be written once and we will not forget adding it. E.g., have a dedicated _Client_ struct with and `http.Client` as a field which handles the timeout.

components/ias-adapter/internal/service/tenant_mapping.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (s TenantMappingsService) ProcessTenantMapping(ctx context.Context, tenantMapping1 types.TenantMapping) error {
+	log := logger.FromContext(ctx)
+
+	formationID := tenantMapping1.FormationID
+	tenantMappings, err := s.Storage.ListTenantMappings(ctx, formationID)
+	if err != nil {
+		log.Err(err).Send()
+		return errors.Newf("failed to get tenant mappings for formation '%s': %w", formationID, err)
+	}
```
petar-iv, line 37: Usually an error is either logged or returned. In this case the returned error has more context (what operation has failed + the reason it failed) so logging it (possibly in the layer above (where it is called)) seems beneficial.

Same below (lines 51, and 61).

components/ias-adapter/internal/storage/postgres/tenant_mapping.go
```diff
@@ -0,0 +1,85 @@
...
+const tenantMappingsTable = "tenant_mappings"
+
+var (
+	upsertTenantMappingQuery  = fmt.Sprintf("INSERT INTO %s (formation_id, ucl_application_id, value) VALUES ($1, $2, $3) ON CONFLICT ON CONSTRAINT pk DO UPDATE SET value=$3", tenantMappingsTable)
+	selectTenantMappingsQuery = fmt.Sprintf("SELECT value FROM %s WHERE formation_id=$1", tenantMappingsTable)
+	deleteTenantMappingQuery  = fmt.Sprintf("DELETE FROM %s WHERE formation_id=$1 AND ucl_application_id=$2", tenantMappingsTable)
+)
+
+func (c Connection) UpsertTenantMapping(ctx context.Context, tenantMapping types.TenantMapping) error {
+	timeoutCtx, cancel := context.WithTimeout(ctx, c.cfg.RequestTimeout)
```
petar-iv, line 21: The timeout logic could be encapsulated in a struct implementing the `pgxPool` interface, could be the `Connection` struct itself (in private methods). In this way we will have a single place (still, several functions) in which we deal with the timeout (and with a quick look at it we can check that it is used in every db operation) and will not introduce additional lines here (which slightly distracts the reader from the main logic of the method).

components/schema-migrator/migrations/ias-adapter/20230315104000_initialize_schema.up.sql
NickyMateev, line 1: The migrations are put in a dedicated `components/schema-migrator/migrations/ias-adapter` folder instead of the already used one `components/schema-migrator/migrations/director`.

While at first I thought that that might be a nice separation two thoughts came to mind:


1. The `tenant_mappings` table might be added with this IAS Adapter PR but as it's name suggests it could be used for other adapter-like scenarios, where UCL has to simulate the tenant mapping call to a specific application. So this table is really general purpose and I'm not sure that it should be in a dedicated migrations folder with the name `ias-adapter`.
2. It will be a lot more difficult for people to track the timestamp of the latest migration when adding new migrations as they will have to double check in two folders. We do have a script for creating new migration which uses the current timestamp, but I've seen cases where people just add new migrations by hand (in other words, manually create the file and put the timestamp themselves)

DimitarPetrov: Correct me if I am wrong, but I think that IAS Adapter uses a separate DB schema from other compass components.

components/ias-adapter/go.mod
```diff
@@ -0,0 +1,114 @@
+module github.com/kyma-incubator/compass/components/ias-adapter
+
+go 1.20
```
PetarTodorovv, line 3: The go version is 1.20 here but 1.18 in the Makefile. I'm not sure if there will be a problem when some of the make targets are used? Did you try to execute `deploy-on-k3d` for example to see if everything is working as expected?

components/ias-adapter/tests/ias-adapter/Makefile
```diff
@@ -0,0 +1,34 @@
+APP_NAME = ias-adapter-tests
+COMPONENT = ias-adapter
+APP_PATH = tests/ias-adapter
+BUILDPACK = eu.gcr.io/kyma-project/test-infra/buildpack-golang:go1.20
```
PetarTodorovv, line 4: Is there a specific reason for the e2e tests to be in the components themselves instead of in the "root" tests directory where other e2e tests are? Also, I didn't find `TestDefinition` CRD(from octopus) similar to the other components, so I'm wondering how the e2e tests will be executed?
Currently, we create a cluster test suite(also a CRD from octopus) that manages the mentioned test definitions and execute them in the cluster. For part of the tests we need external-services-mock and we temporarily deploy it on the cluster during the e2e tests execution.
Will the new component's e2e tests work "outside" this approach that we have

chart/compass/charts/ias-adapter/templates/rbac.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  {{- if eq .Values.global.database.embedded.enabled false }}
+  annotations:
+    iam.gke.io/gcp-service-account: {{ $.Values.global.database.sqlProxyServiceAccount | quote }}
```
PetarTodorovv, line 8: When we have this gcp annotation that means we're using the workload identity functionality from GCP. And the other half of it is to maintain/update the `proxy-user` service account in GCP. For that, a PR in the infra repo is needed. Valid for all environments and also for all k8s SA that use such annotation.

chart/compass/templates/ias-adapter-migrate-down-job.yaml
```diff
@@ -0,0 +1,103 @@
...
+            claimName: {{ .Values.global.migratorJob.ias_adapter.pvc.name }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ $.Chart.Name }}-ias-adapter-migration-down-job
+  namespace: {{ $.Release.Namespace }}
+  {{- if eq .Values.global.database.embedded.enabled false }}
+  annotations:
+    iam.gke.io/gcp-service-account: {{ $.Values.global.database.sqlProxyServiceAccount | quote }}
```
PetarTodorovv, line 96: See the other comment about the gcp SA annotation

chart/compass/templates/ias-adapter-migrate-up-job.yaml
```diff
@@ -0,0 +1,99 @@
...
+              claimName: {{ .Values.global.migratorJob.ias_adapter.pvc.name }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ $.Chart.Name }}-ias-adapter-migration-up-job
+  namespace: {{ $.Release.Namespace }}
+  {{- if eq .Values.global.database.embedded.enabled false }}
+  annotations:
+    iam.gke.io/gcp-service-account: {{ $.Values.global.database.sqlProxyServiceAccount | quote }}
```
PetarTodorovv, line 92: here as well

PetarTodorovv (changes requested):

components/ias-adapter/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM golang:1.20-alpine3.17 as builder
```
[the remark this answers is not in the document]
georgifarashev: Dockerfile aligned with https://github.com/kyma-incubator/compass/pull/2948/commits/1d8ae0813cd545f55e33fc21e42c5777fbf335f0

components/schema-migrator/migrations/ias-adapter/20230315104000_initialize_schema.up.sql
PetarTodorovv: If that's the case, what is the situation with the infrastructure/terraform scripts? Are they adapted as well or is the new DB initially created manually(through the GCP console for example)? The main question: is there a risk if a new change in the infra repo is made, the terraform "actual" and "desired" states to differ and to "rollback"/delete the DB?

Also, what will be the behavior on the validation clusters? I guess they are not "urgent" but we need to maintain them as well. As far as I saw there is no such DB there. The setup around the validation clusters is a little bit different(we recreate the SQL instance every time a cluster is created or deleted).

yordanovagabriela: @PetarTodorovv we already created the db manually on all landscapes, because we didn't know about the terraform scripts

NickyMateev (dismissed):

components/ias-adapter/internal/service/ias/service.go
```diff
@@ -0,0 +1,188 @@
...
+	applications := types.Applications{}
+	if err := json.NewDecoder(resp.Body).Decode(&applications); err != nil {
+		return types.Application{}, err
+	}
+
+	if len(applications.Applications) == 0 {
+		return types.Application{}, errors.Newf("no applications found with clientID '%s'", clientID)
+	}
+	if len(applications.Applications) == 1 {
+		return applications.Applications[0], nil // TODO do we leave this?
```
georgifarashev, line 99: Leave for testing in dev-main. Later to be removed!

PR #2973, opened 2023-03-24
[Constraint] Adding multiple systems of a given type to a formation should be forbidden

la4ezar:
**Description**
We want to forbid adding multiple systems of a given type to a formation. In this PR I'm adding a new Constraint which forbids that.

- Add new constraint
- Unit tests
- E2e tests

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


chart/compass/values.yaml
dragobt, line 1: update the dependencies as you have changes in the pkg dir and after that do not forget to bump the components versions in the values.yaml

components/director/internal/domain/formationconstraint/constraint_operators.go
```diff
@@ -85,6 +93,46 @@ func (e *ConstraintEngine) IsNotAssignedToAnyFormationOfType(ctx context.Context
...
+	i, ok := input.(*formationconstraint.DoesNotContainResourceOfSubtypeInput)
+	if !ok {
+		return false, errors.New(fmt.Sprintf("Incompatible input for operator %q", DoesNotContainResourceOfSubtypeOperator))
+	}
+
+	log.C(ctx).Infof("Enforcing constraint on resource of type: %q, subtype: %q and ID: %q", i.ResourceType, i.ResourceSubtype, i.ResourceID)
+
+	switch i.ResourceType {
+	case model.ApplicationResourceType:
+		inputAppType, err := e.labelService.GetByKey(ctx, i.Tenant, model.ApplicationLabelableObject, i.ResourceID, e.applicationTypeLabelKey)
```
nyordanoff, line 109: This call to get the input App type is most probably redundant because you already have the input App type calculated in the joinPointDetails and then propagated to the operatorInput, so you can just use it from the operator.

nyordanoff (changes requested):

components/director/internal/domain/formationconstraint/constraint_engine.go
```diff
@@ -36,28 +36,50 @@ type labelRepository interface {
...
 	constraintSvc             formationConstraintSvc
 	tenantSvc                 tenantService
 	asaSvc                    automaticScenarioAssignmentService
 	formationRepo             formationRepository
 	labelRepo                 labelRepository
+	labelService              labelService
+	applicationRepository     applicationRepository
 	operators                 map[OperatorName]OperatorFunc
 	operatorInputConstructors map[OperatorName]OperatorInputConstructor
+	applicationTypeLabelKey   string
```
nyordanoff, line 60: We discussed offline whether `applicationTypeLabelKey` to be passed in the constructor or to be added in the constraint inputTemplate -> joinPointDetails -> operatorInput.

As I gave it a further thought I lean more towards the second option for the following 3 reasons:
- if we can calculate a field that will be used by the engine in the joinPointDetails, why not to do it there but separate the places from where 'input' information is passed; if we move the key to the joinPointDetails we will encapsulate the logic in one place;
- if we add the `applicationTypeLabelKey` to joinPointDetails (in our case `AssignFormationOperationDetails`) we will know more granually in which flow this field is needed/will be used whereas if we add it directly in the constructor, initially we won't have an idea why and where this field will be used. In the future, if more such fields are added, from my PoV it will become messy.
- if we add it in the joinPointDetails, the resource type key will be determined dynamically based on the object type that is passed, meaning that we will have it out-of-the-box for all supported object types by `AssignFormation`. If we leave it in the constructor and in the future we will have to use the runtimeType key, we will have to add it separately as another argument and so on for every other type.

What do you think, I am open for a discussion.

dragobt (approved):

nyordanoff (dismissed):

dragobt (approved):

nyordanoff (approved):

dragobt (dismissed):

nyordanoff (approved):

dragobt (approved):

PR #2980, opened 2023-03-28
Forbid deletion of application templates if there are existing associated applications

dragobt:
**Description**
Currently application templates can be deleted even if there are existing applications created from it. With this PR we forbid this action.

- introduce schema migrations for the new constraint
- add e2e test

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [x] Integration tests

### DIFF
--- a/components/schema-migrator/migrations/director/20230328144000_forbid_deleting_app_template_if_it_has_associated_application.down.sql
+++ b/components/schema-migrator/migrations/director/20230328144000_forbid_deleting_app_template_if_it_has_associated_application.down.sql
@@ -4,4 +4,3 @@ ALTER TABLE applications
     DROP CONSTRAINT IF EXISTS applications_app_template_id_fkey;
 
 COMMIT;
-
--- a/components/schema-migrator/migrations/director/20230328144000_forbid_deleting_app_template_if_it_has_associated_application.down.sql
+++ b/components/schema-migrator/migrations/director/20230328144000_forbid_deleting_app_template_if_it_has_associated_application.down.sql
@@ -2,5 +2,8 @@ BEGIN;
 
 ALTER TABLE applications
     DROP CONSTRAINT IF EXISTS applications_app_template_id_fkey;
+ALTER TABLE applications
+    ADD CONSTRAINT applications_app_template_id_fkey
+        FOREIGN KEY (app_template_id) REFERENCES app_templates(id) ON DELETE SET NULL;
 
 COMMIT;
--- a/components/schema-migrator/migrations/director/20230328144000_forbid_deleting_app_template_if_it_has_associated_application.down.sql
+++ b/components/schema-migrator/migrations/director/20230328144000_forbid_deleting_app_template_if_it_has_associated_application.down.sql
@@ -1,0 +1,7 @@
+BEGIN;
+
+ALTER TABLE applications
+    DROP CONSTRAINT IF EXISTS applications_app_template_id_fkey;
+
+COMMIT;
+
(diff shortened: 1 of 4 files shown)


tests/director/tests/application_templates_api_test.go
```diff
@@ -584,6 +584,74 @@ func TestDeleteApplicationTemplate(t *testing.T) {
...
+		},
+	}
+
+	tenantId := tenant.TestTenants.GetDefaultTenantID()
+
+	t.Log("Creating application template")
+	appTemplate, err := fixtures.CreateApplicationTemplateFromInput(t, ctx, certSecuredGraphQLClient, tenantId, appTemplateInput)
+	defer fixtures.CleanupApplicationTemplate(t, ctx, certSecuredGraphQLClient, tenantId, appTemplate)
+	require.NoError(t, err)
+	require.Equal(t, conf.SubscriptionConfig.SelfRegRegion, appTemplate.Labels[tenantfetcher.RegionKey])
```
alextargov, line 614: This would not be needed if we drop the creation of the app template in the subscription flow scenario.

tests/director/tests/application_templates_api_test.go
```diff
@@ -584,6 +584,74 @@ func TestDeleteApplicationTemplate(t *testing.T) {
...
+			Placeholder: "display-name",
+			Value:       "new-display-name",
+		}}}
+	appFromTemplateGQL, err := testctx.Tc.Graphqlizer.ApplicationFromTemplateInputToGQL(appFromTemplate)
+	require.NoError(t, err)
+	createAppFromTemplateRequest := fixtures.FixRegisterApplicationFromTemplate(appFromTemplateGQL)
+	outputApp := graphql.ApplicationExt{}
+
+	t.Logf("Creating application from application template with id %s", appTemplate.ID)
+	err = testctx.Tc.RunOperation(ctx, certSecuredGraphQLClient, createAppFromTemplateRequest, &outputApp)
```
alextargov, line 631: I think we should have a defer here to unregister the app. Because if something fails between line 631 and line 649, then the app will stay in the db as garbage, as well as the app template because the differ [here](https://github.com/kyma-incubator/compass/pull/2980/files#diff-cd91fb82642cfc9f5b55639748e76cc678d32a6a96513051cc75efae61540ce7R612) will fail due to the FK restriction.

alextargov: We might think about adapting other tests that create app from template like "TestRegisterApplicationFromTemplate" for example, in order to guarantee that other tests will not produce garbage.

alextargov (changes requested):

components/schema-migrator/migrations/director/20230328144000_forbid_deleting_app_template_if_it_has_associated_application.down.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE applications
+    DROP CONSTRAINT IF EXISTS applications_app_template_id_fkey;
```
dzahariev, line 4: The old constraint need to be added in down migration:
```
ALTER TABLE applications
    ADD CONSTRAINT applications_app_template_id_fkey FOREIGN KEY (app_template_id) REFERENCES app_templates(id) ON DELETE SET NULL;
```

dzahariev (approved):

alextargov:
I see there are about 20 tests that register app from template. Have we checked them out for possible garbage generation if something errors between the defers in them?

dragobt:
as far as i can i checked all the places where `fixtures.FixRegisterApplicationFromTemplate` is being called and all of them looked good to me except the ones that i have changed in this PR

alextargov (approved):

PR #2981, opened 2023-03-28
Derive region for auto-discovered systems

dzahariev:
**Description**
Derive region for auto-discovered systems

- Fix a bug where optional properties that were not supplied, were processed as empty strings. They should be removed from payload.
- 

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


dzahariev:
reset-vm
/test pre-main-compass-integration-no-dump

chart/compass/values.yaml
dragobt, line 1: bump the `e2e_test` PR version as well as you have changes in the e2e tests.

components/director/internal/domain/apptemplate/service.go
```diff
@@ -338,9 +338,61 @@ func (s *service) PrepareApplicationCreateInputJSON(appTemplate *model.Applicati
...
+
+	for _, placeholder := range appTemplate.Placeholders {
+		newValue, err := values.FindPlaceholderValue(placeholder.Name)
+		isOptional := false
+		if placeholder.Optional != nil {
+			isOptional = *placeholder.Optional
+		}
+
+		if err != nil && !isOptional {
+			return "", errors.Wrap(err, "required placeholder not provided")
```
dragobt, line 350: add unit test for this case

components/director/internal/domain/apptemplate/service.go
```diff
@@ -338,9 +338,61 @@ func (s *service) PrepareApplicationCreateInputJSON(appTemplate *model.Applicati
...
+		}
```
dragobt, line 351: I see that this logic is the same as the one from l323 to l332. Can we somehow do the whole thing in one loop? If not so, perhaps consider taking it out in a separate function.

components/director/internal/domain/apptemplate/service.go
```diff
@@ -338,9 +338,61 @@ func (s *service) PrepareApplicationCreateInputJSON(appTemplate *model.Applicati
...
+
+		if strings.TrimSpace(newValue) == "" && isOptional {
+			appCreateInputJSON, err = removeEmptyKeyFromLabels(appCreateInputJSON, placeholder.Name)
+			if err != nil {
+				return "", errors.Wrap(err, "error while clear optional empty value")
```
dragobt, line 356: add unit test for this case as well

components/director/internal/domain/apptemplate/service.go
dzahariev: This statement is never reached, as the same check was done several lines above. 
Will look for a way to remove this redundant check.

components/director/internal/domain/apptemplate/service.go
dzahariev: Not possible to be done on one pass.
The idea of first pass is to process entries and the second pass require the appCreateInputJSON to be already filled and now we just clear the empty labels from it.

dzahariev: Now I clear the output now on each property process. Slower than before, but penalty can be ignored, as data that is processed is usually les than 15 fields.

components/director/internal/domain/apptemplate/service.go
```diff
@@ -337,10 +337,47 @@ func (s *service) PrepareApplicationCreateInputJSON(appTemplate *model.Applicati
...
+		if _, ok := value.(string); ok {
+			// String value
+			if value == "" && key == keyName {
+				if !rootObject {
+					delete(*input, key)
+				}
+			}
+		} else if mapValue, ok := value.(map[string]interface{}); ok {
+			// Object value - process only labels object
+			if key == "labels" {
```
dragobt, line 374: we can take `labels` out as const

dzahariev:
Validated!

dragobt (approved):

PR #2986, opened 2023-03-30
Expose formation constraints subresolver for formation template

la4ezar:
**Description**
Expose formation constraints subresolver for formation template. Also, add one more formation constraint type('UI') for the Cockpit constraints.

- Expose formation constraints subresolver for formation template
- Add 'UI' constraint type
- Unit tests
- E2e tests

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/pkg/graphql/schema.graphql
```diff
@@ -1241,6 +1242,7 @@ type FormationTemplate {
 	runtimeArtifactKind: ArtifactType
 	webhooks: [Webhook!] @sanitize(path: "graphql.field.formation_template.webhooks")
 	leadingProductIDs: [String!]
+	formationConstraints: [FormationConstraint!]!
```
StanislavStefanov, line 1245: Should be formationConstraints: [FormationConstraint!] as the constraints are not mandatory for the template

components/director/pkg/graphql/schema.graphql
```diff
@@ -108,6 +108,7 @@ enum ConstraintScope {
 enum ConstraintType {
 	PRE
 	POST
```
StanislavStefanov, line 110: Can you add:
LOAD_FORMATIONS
SELECT_SYSTEMS_FOR_FORMATION 
keys in enum TargetOperation

tests/director/tests/formation_constraint_api_test.go
```diff
@@ -287,9 +287,27 @@ func TestListFormationConstraintsForFormationTemplate(t *testing.T) {
...
 	fixtures.AttachConstraintToFormationTemplate(t, ctx, certSecuredGraphQLClient, constraint.ID, formationTemplate.ID)
 
+	// Assert the constraint is attached only to the first formation template
+	t.Logf("Get formation template with name %q and id %q, and assert there are is one constraint attached to it", formationTemplate.Name, formationTemplate.ID)
+	formationTemplateOutput = fixtures.QueryFormationTemplateWithConstraints(t, ctx, certSecuredGraphQLClient, formationTemplate.ID)
+	assert.Equal(t, formationTemplateOutput.FormationConstraints, []graphql.FormationConstraint{*constraint})
+
+	t.Logf("Get formation template with name %q and id %q, and assert there are no constraints attached to it", secondFormationTemplate.Name, secondFormationTemplate.ID)
+	secondFormationTemplateOutput = fixtures.QueryFormationTemplateWithConstraints(t, ctx, certSecuredGraphQLClient, secondFormationTemplate.ID)
+	assert.Empty(t, secondFormationTemplateOutput.FormationConstraints)
```
PetarTodorovv, line 309: Do we need this? As we already check that a few lines above and the "attach" operation before this assert is related to the "first" formation template, we do not expect the "second" one to be changed?

tests/director/tests/formation_constraint_api_test.go
```diff
@@ -309,6 +327,15 @@ func TestListFormationConstraintsForFormationTemplate(t *testing.T) {
 	t.Logf("Attaching second constraint to formation template")
 	fixtures.AttachConstraintToFormationTemplate(t, ctx, certSecuredGraphQLClient, constraintSecond.ID, formationTemplate.ID)
 
+	// Assert the two constraints are attached only to the first formation template
+	t.Logf("Get formation template with name %q and id %q, and assert there are is one constraint attached to it", formationTemplate.Name, formationTemplate.ID)
+	formationTemplateOutput = fixtures.QueryFormationTemplateWithConstraints(t, ctx, certSecuredGraphQLClient, formationTemplate.ID)
+	assert.Equal(t, formationTemplateOutput.FormationConstraints, []graphql.FormationConstraint{*constraint, *constraintSecond})
```
PetarTodorovv, line 333: Mismatch between the log message and the assert, the log says one constraint but the assert check for two

tests/director/tests/formation_constraint_api_test.go
```diff
@@ -309,6 +327,15 @@ func TestListFormationConstraintsForFormationTemplate(t *testing.T) {
...
+
+	t.Logf("Get formation template with name %q and id %q, and assert there are no constraints attached to it", secondFormationTemplate.Name, secondFormationTemplate.ID)
+	secondFormationTemplateOutput = fixtures.QueryFormationTemplateWithConstraints(t, ctx, certSecuredGraphQLClient, secondFormationTemplate.ID)
+	assert.Empty(t, secondFormationTemplateOutput.FormationConstraints)
```
PetarTodorovv, line 337: I think we don't need this, the reason is the same as the comment above?

tests/director/tests/formation_constraint_api_test.go
```diff
@@ -328,6 +355,15 @@ func TestListFormationConstraintsForFormationTemplate(t *testing.T) {
 	t.Logf("Attaching constraintForOtherTemplate to formation template other")
 	fixtures.AttachConstraintToFormationTemplate(t, ctx, certSecuredGraphQLClient, constraintForOtherTemplate.ID, secondFormationTemplate.ID)
 
+	// Assert the two constraints are attached to the first formation template and one to the second formation template
+	t.Logf("Get formation template with name %q and id %q, and assert there are is one constraint attached to it", formationTemplate.Name, formationTemplate.ID)
+	formationTemplateOutput = fixtures.QueryFormationTemplateWithConstraints(t, ctx, certSecuredGraphQLClient, formationTemplate.ID)
+	assert.Equal(t, formationTemplateOutput.FormationConstraints, []graphql.FormationConstraint{*constraint, *constraintSecond})
```
PetarTodorovv, line 361: log and assert mismatch, check the comment above

tests/director/tests/formation_constraint_api_test.go
```diff
@@ -328,6 +355,15 @@ func TestListFormationConstraintsForFormationTemplate(t *testing.T) {
...
+
+	t.Logf("Get formation template with name %q and id %q, and assert there are no constraints attached to it", secondFormationTemplate.Name, secondFormationTemplate.ID)
```
PetarTodorovv, line 363: The log is wrong here as well.

PetarTodorovv (changes requested):

tests/director/tests/formation_constraint_api_test.go
la4ezar: It can be removed but with this, we are checking that the sub-resolver is working correctly(that it returns the correct constraints for the right formation templates).

tests/director/tests/formation_constraint_api_test.go
la4ezar: As mentioned above, it can be removed, but I don't think it interferes in any way. Also, we are checking if the sub-resolver is working correctly.

components/director/internal/domain/formationtemplate/resolver.go
```diff
@@ -280,3 +298,47 @@ func (r *Resolver) Webhooks(ctx context.Context, obj *graphql.FormationTemplate)
...
 }
+
+// FormationConstraint retrieves a FormationConstraint for the specified FormationTemplate
+func (r *Resolver) FormationConstraint(ctx context.Context, obj *graphql.FormationTemplate) ([]*graphql.FormationConstraint, error) {
+	params := dataloader.ParamFormationConstraint{ID: obj.ID, Ctx: ctx}
+	return dataloader.FormationTemplateFor(ctx).FormationConstraintByID.Load(params)
+}
+
+// FormationConstraintsDataLoader retrieves Formation Constraints for each FormationTemplate ID in the keys
+func (r *Resolver) FormationConstraintsDataLoader(keys []dataloader.ParamFormationConstraint) ([][]*graphql.FormationConstraint, []error) {
```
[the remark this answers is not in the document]
la4ezar: As we discussed, if we agree to add Page structures it will be done in a follow-up PR.

To sum up the discussion: 
Currently, we don't need the Page structures for the pagination, because the Formation Constraints are just a few. But the Page structures could make the code more readable because we will not work with an array of arrays([][]...) but with an array of page structures. If that is the case, it will be better to add pages.

components/director/internal/domain/formationconstraint/service.go
```diff
@@ -99,6 +101,54 @@ func (s *service) ListByFormationTemplateID(ctx context.Context, formationTempla
...
+
+	for _, constraint := range formationConstraints {
+		if constraint.ConstraintScope == model.GlobalFormationConstraintScope {
+			globalConstraints = append(globalConstraints, constraint)
+		} else {
+			attachedConstraintsMap[constraint.ID] = constraint
+		}
+	}
+
+	formationConstraintsPerFormationTemplate := make([][]*model.FormationConstraint, len(formationTemplateIDs))
```
PetarTodorovv, line 136: To me it seems the logic in this function(`ListByFormationTemplateIDs`) is unnecessarily complex. 

E.g. here you break up into smaller units the `formationConstraints` -  into "global" and "non-global"(which are stored into a new map[constrainID][constraintValue]. And in the for cycle below you combine them again. What's the idea behind this? In the end we combine them again?

Moreover, the final structure is `[][]*model.FormationConstraint` but in the resolver we're interested only of the second part(`[]*model.FormationConstraint`) and the indexes are skipped and not used - `for _, formationConstraints := range formationConstraintsPerFormation {}`? Why we need [][].. when we don't use it

Could we simplify the whole logic to:
```
func (s *service) ListByFormationTemplateIDs(ctx context.Context, formationTemplateIDs []string) ([]*model.FormationConstraint, error) {
	constraintRefs, err := s.formationTemplateConstraintReferenceRepo.ListByFormationTemplateIDs(ctx, formationTemplateIDs)
	if err != nil {
		return nil, errors.Wrapf(err, "while listing Formation Constraint References for FormationTemplates with IDs: %q", formationTemplateIDs)
	}

	constraintIDs := make([]string, 0, len(constraintRefs))
	for _, cr := range constraintRefs {
		constraintIDs = append(constraintIDs, cr.ConstraintID)
	}

	formationConstraints, err := s.repo.ListByIDsAndGlobal(ctx, constraintIDs)
	if err != nil {
		return nil, errors.Wrapf(err, "while listing Formation Constraints by IDs %q and the global ones", formationTemplateIDs)
	}

	return formationConstraints, nil
}
```

components/director/internal/domain/formationtemplate/resolver.go
```diff
@@ -280,3 +298,47 @@ func (r *Resolver) Webhooks(ctx context.Context, obj *graphql.FormationTemplate)
...
+	ctx = persistence.SaveToContext(ctx, tx)
+
+	formationConstraintsPerFormation, err := r.formationConstraintSvc.ListByFormationTemplateIDs(ctx, formationTemplateIDs)
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	gqlFormationConstraints := make([][]*graphql.FormationConstraint, 0, len(formationConstraintsPerFormation))
+
+	for _, formationConstraints := range formationConstraintsPerFormation {
```
PetarTodorovv, line 335: Additionally to the comment in the `formationconstraint/service.go`, this will be simplified as well and most probably will have a similar look:
```
func (r *Resolver) FormationConstraintsDataLoader(keys []dataloader.ParamFormationConstraint) ([]*graphql.FormationConstraint, []error) {
	if len(keys) == 0 {
		return nil, []error{apperrors.NewInternalError("No Formation Templates found")}
	}

	ctx := keys[0].Ctx
	formationTemplateIDs := make([]string, 0, len(keys))
	for _, key := range keys {
		formationTemplateIDs = append(formationTemplateIDs, key.ID)
	}

	tx, err := r.transact.Begin()
	if err != nil {
		return nil, []error{err}
	}
	defer r.transact.RollbackUnlessCommitted(ctx, tx)

	formationConstraints, err := r.formationConstraintSvc.ListByFormationTemplateIDs(ctx, formationTemplateIDs)
	if err != nil {
		return nil, []error{err}
	}

	gqlFormationConstraints := r.formationConstraintConverter.MultipleToGraphQL(formationConstraints)

	if err = tx.Commit(); err != nil {
		return nil, []error{err}
	}

	return gqlFormationConstraints, nil
}
```

Also, that way in the root_resolver we'll have `func (r *RootResolver) FormationConstraintsDataLoader(ids []dataloader.ParamFormationConstraint) ([]*graphql.FormationConstraint, []error) {}` which is similar to the other data loaders that we have, instead of `([][]*graphql.FormationConstraint, []error)`

image


I may be missing something but at the first glance, I'm not sure we need this matrix structure in the service and resolver files?

PetarTodorovv (changes requested):

components/director/internal/domain/formationconstraint/service.go
la4ezar: We discussed it offline. To sum it up: the structure is an array of arrays because of the dataloader(that's how it works). 
And in the result, at the index 0 are the formation constraints for the formation template ID at index 0 from the input formationTemplateIDs; at index 1 are the FCs for the FT at index 1, etc. That's how the dataloaders aggregate which constraints are for which formation template.

dragobt (approved):

components/director/internal/domain/formationtemplate/resolver_test.go
```diff
@@ -943,3 +946,114 @@ func TestResolver_Webhooks(t *testing.T) {
...
+			ExpectedConstraints: nil,
+			ExpectedErrors:      []error{testErr},
+		},
+		{
+			Name:  "Returns error when listing the formation templates by ids fail",
+			TxFn:  txGen.ThatDoesntExpectCommit,
+			Input: []dataloader.ParamFormationConstraint{{ID: formationConstraintIDs[0], Ctx: ctx}, {ID: formationConstraintIDs[1], Ctx: ctx}},
+			FormationConstraintSvc: func() *automock.FormationConstraintService {
+				svc := &automock.FormationConstraintService{}
+				svc.On("ListByFormationTemplateIDs", txtest.CtxWithDBMatcher(), formationConstraintIDs).Return(nil, testErr)
```
StanislavStefanov, line 1011: Can we specify how many invocations we expect

tests/director/tests/formation_constraint_api_test.go
```diff
@@ -287,9 +287,27 @@ func TestListFormationConstraintsForFormationTemplate(t *testing.T) {
 	defer fixtures.CleanupFormationConstraint(t, ctx, certSecuredGraphQLClient, constraint.ID)
 	require.NotEmpty(t, constraint.ID)
 
+	// Assert no constraints attached
+	t.Logf("Get formation template with name %q and id %q, and assert there are no constraints attached to it", formationTemplate.Name, formationTemplate.ID)
+	formationTemplateOutput := fixtures.QueryFormationTemplateWithConstraints(t, ctx, certSecuredGraphQLClient, formationTemplate.ID)
```
StanislavStefanov, line 292: We have to regenerate the examples as the query has changed. Running the scrip should be sufficient as the graphqlizer is already adapted

StanislavStefanov (changes requested):

dragobt (approved):

nkanchev (approved):

PetarTodorovv (approved):

PR #2987, opened 2023-03-30
Allow cld app providers to onboard with cert

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/cmd/director/main.go
+++ b/components/director/cmd/director/main.go
@@ -181,6 +181,8 @@ type config struct {
 
 	ExternalClientCertSecretName string `envconfig:"APP_EXTERNAL_CLIENT_CERT_SECRET_NAME"`
 	ExtSvcClientCertSecretName   string `envconfig:"APP_EXT_SVC_CLIENT_CERT_SECRET_NAME"`
+
+	ApplicationTemplateProductLabel string `envconfig:"APP_APPLICATION_TEMPLATE_PRODUCT_LABEL"`
 }
 
 func main() {
@@ -294,6 +296,7 @@ func main() {
 		ordWebhookMapping,
 		tenantMappingConfig,
 		cfg.TenantMappingCallbackURL,
+		cfg.ApplicationTemplateProductLabel,
 	)
 	exitOnError(err, "Failed to initialize root resolver")
 
--- a/components/director/internal/domain/apptemplate/resolver.go
+++ b/components/director/internal/domain/apptemplate/resolver.go
@@ -121,10 +121,11 @@ type Resolver struct {
 	uidService               UIDService
 	tenantMappingConfig      map[string]interface{}
 	tenantMappingCallbackURL string
+	appTemplateProductLabel  string
 }
 
 // NewResolver missing godoc
-func NewResolver(transact persistence.Transactioner, appSvc ApplicationService, appConverter ApplicationConverter, appTemplateSvc ApplicationTemplateService, appTemplateConverter ApplicationTemplateConverter, webhookService WebhookService, webhookConverter WebhookConverter, selfRegisterManager SelfRegisterManager, uidService UIDService, tenantMappingConfig map[string]interface{}, tenantMappingCallbackURL string) *Resolver {
+func NewResolver(transact persistence.Transactioner, appSvc ApplicationService, appConverter ApplicationConverter, appTemplateSvc ApplicationTemplateService, appTemplateConverter ApplicationTemplateConverter, webhookService WebhookService, webhookConverter WebhookConverter, selfRegisterManager SelfRegisterManager, uidService UIDService, tenantMappingConfig map[string]interface{}, tenantMappingCallbackURL string, appTemplateProductLabel string) *Resolver {
 	return &Resolver{
 		transact:                 transact,
 		appSvc:                   appSvc,
@@ -137,6 +138,7 @@ func NewResolver(transact persistence.Transactioner, appSvc ApplicationService, 
 		uidService:               uidService,
 		tenantMappingConfig:      tenantMappingConfig,
 		tenantMappingCallbackURL: tenantMappingCallbackURL,
+		appTemplateProductLabel:  appTemplateProductLabel,
 	}
 }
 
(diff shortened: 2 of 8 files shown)


alextargov:
reset-vm

alextargov:
reset-vm

components/director/cmd/director/main.go
```diff
@@ -181,6 +181,8 @@ type config struct {
 
 	ExternalClientCertSecretName string `envconfig:"APP_EXTERNAL_CLIENT_CERT_SECRET_NAME"`
 	ExtSvcClientCertSecretName   string `envconfig:"APP_EXT_SVC_CLIENT_CERT_SECRET_NAME"`
+
+	ApplicationTemplateProductLabel string `envconfig:"APP_APPLICATION_TEMPLATE_PRODUCT_LABEL"`
```
dragobt, line 185: As far as I see this label is used only in the SelfReg flow so why don't we put it into the `SelfRegConfig`? This will prevent the need from passing the label as an argument to many functions. And as i know this label is also some kind of "distinguish label", so we can keep the convention of naming. Maybe something like `SelfRegisterAppTemplateProductLabelKey`.

components/director/internal/domain/apptemplate/resolver.go
```diff
@@ -745,28 +747,41 @@ func validateAppTemplateNameBasedOnProvider(name string, appInput *graphql.Appli
...
+		}
+
+		if regionExists {
+			filters = append(filters, labelfilter.NewForKeyWithQuery(regionLabelKey, fmt.Sprintf("\"%s\"", region)))
+			msg += fmt.Sprintf(" and %q: %q", regionLabelKey, region)
 		}
 
-		log.C(ctx).Infof("Getting application template for labels %q: %q and %q: %q", regionLabelKey, region, distinguishLabelKey, distinguishLabelValue)
+		log.C(ctx).Infof("Getting application template for labels %s", msg)
 		appTemplate, err := r.appTemplateSvc.GetByFilters(ctx, filters)
```
dragobt, line 776: If the filters are applied with an "AND" in the query will it even be possible to find an app template which has all these 2 or 3 labels? Can we check that?

components/director/internal/selfregmanager/selfreg_manager.go
```diff
@@ -85,12 +86,26 @@ func (s *selfRegisterManager) PrepareForSelfRegistration(ctx context.Context, re
...
+			labels[scenarioassignment.SubaccountIDKey] = consumerInfo.ConsumerID
+		}
+
+		distinguishLabel, distinguishLabelExists := labels[s.cfg.SelfRegisterDistinguishLabelKey]
+		_, productLabelExists := labels[s.appTemplateProductLabel]
+
+		switch {
+		case !distinguishLabelExists && !productLabelExists:
 			if resourceType == resource.Runtime {
 				return labels, nil
```
dragobt, line 99: If we are runtime do we want to have `Subaccount` label returned?

components/director/internal/selfregmanager/selfreg_manager.go
```diff
@@ -85,12 +86,26 @@ func (s *selfRegisterManager) PrepareForSelfRegistration(ctx context.Context, re
...
 			}
-			return nil, errors.Errorf("missing %q label", s.cfg.SelfRegisterDistinguishLabelKey)
+			return nil, errors.Errorf("missing %q or %q label", s.cfg.SelfRegisterDistinguishLabelKey, s.appTemplateProductLabel)
+		case distinguishLabelExists && productLabelExists:
+			if resourceType == resource.Runtime {
+				return labels, nil
+			}
+			return nil, errors.Errorf("should provide either %q or %q label - providing both at the same time is not allowed", s.cfg.SelfRegisterDistinguishLabelKey, s.appTemplateProductLabel)
+		case productLabelExists:
+			return labels, nil
```
dragobt, line 108: Do we want to return the current labels or to continue the execution of the function?

tests/director/tests/application_templates_api_test.go
```diff
@@ -66,6 +66,49 @@ func TestCreateApplicationTemplate(t *testing.T) {
...
+		require.NoError(t, err)
+
+		createApplicationTemplateRequest := fixtures.FixCreateApplicationTemplateRequest(appTemplate)
+		output := graphql.ApplicationTemplate{}
+
+		// WHEN
+		t.Log("Create application template")
+		err = testctx.Tc.RunOperation(ctx, certSecuredGraphQLClient, createApplicationTemplateRequest, &output)
+		defer fixtures.CleanupApplicationTemplate(t, ctx, certSecuredGraphQLClient, tenant.TestTenants.GetDefaultTenantID(), output)
+
```
dragobt, line 106: maybe check if the label is present in the created app template

components/director/cmd/director/main.go
alextargov: Because this label has nothing to do with the self reg flow. We use it in the SelfReg manager, yes, but that's only because the `if`s are there that check the flows. The label is part of the CLD flow and that's why it is not in the SelfReg Config

components/director/internal/domain/apptemplate/resolver.go
alextargov: Of course. There currently we are looking for a maximum of 2 labels at once - region + self reg OR region + cld. Why wouldn't we find App Templates by this criteria?

components/director/internal/selfregmanager/selfreg_manager.go
alextargov: I do not understand that. On L89 we set the subaccount only if we are in AppTemplate flow. That was the previous behavior as well: 
image

components/director/internal/selfregmanager/selfreg_manager.go
alextargov: If we continue, we will proceed with the self reg flow. As far as I understand, this is undesirable.

tests/director/tests/application_templates_api_test.go
alextargov: Which label? I'm creating a negative test here with no labels.

dragobt: I put the comment on the wrong line, but you got the point

components/director/internal/domain/apptemplate/resolver.go
dragobt: when i double check it - all is good

components/director/internal/domain/apptemplate/resolver.go
```diff
@@ -246,13 +251,30 @@ func (r *Resolver) CreateApplicationTemplate(ctx context.Context, in graphql.App
...
+
+	labels := convertedIn.Labels
+	if _, err := tenant.LoadFromContext(ctx); err == nil && consumerInfo.Flow.IsCertFlow() {
+		isSelfReg, selfRegFlowErr := r.isSelfRegFlow(labels)
+		if selfRegFlowErr != nil {
+			return nil, selfRegFlowErr
+		}
+
+		if isSelfReg {
+			labels[scenarioassignment.SubaccountIDKey] = consumerInfo.ConsumerID
```
dragobt, line 268: I think we want this label when we are not only in self reg flow, but also in cld flow.

components/director/internal/domain/apptemplate/resolver.go
```diff
@@ -246,15 +251,32 @@ func (r *Resolver) CreateApplicationTemplate(ctx context.Context, in graphql.App
...
 	convertedIn.ID = &selfRegID
-	validate := func() error {
-		return validateAppTemplateForSelfReg(in.ApplicationInput)
-	}
 
-	selfRegLabels, err := r.selfRegManager.PrepareForSelfRegistration(ctx, resource.ApplicationTemplate, convertedIn.Labels, selfRegID, validate)
+	consumerInfo, err := consumer.LoadFromContext(ctx)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "while loading consumer")
```
dragobt, line 257: unit test for this case

dragobt (approved):

alextargov:
reset-vm

dragobt (dismissed):

alextargov:
reset-vm

alextargov:
reset-vm

alextargov:
retest

dragobt (approved):

PR #2988, opened 2023-03-30
Return not found error when deleting Tenant Access that does not exist

StanislavStefanov:
**Description**

- return not found error during delete if Tenant Access record does not exist

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


StanislavStefanov:
reset vm

StanislavStefanov:
reset-vm

PR #3001, opened 2023-04-06
Adapt latest ORD specification - from 1.2.0 to 1.2.2

dragobt:
**Description**
Adopt latest ORD specification - from 1.2.0 to 1.2.2

- add version, resource hash and local tenant id properties to bundles
- add local tenant id property to apis
- add local tenant id property to events
- add validation rules for the new properties
- add and adapt unit tests
- introduce new schema migration for the new properties
- regenerate mocks
- store ord spec local id as application local tenant id if application does not have local tenant id

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/application/service.go
```diff
@@ -52,6 +52,7 @@ type repoCreatorFunc func(ctx context.Context, tenant string, application *model
 type repoUpserterFunc func(ctx context.Context, tenant string, application *model.Application) (string, error)
 
 // ApplicationRepository missing godoc
+//
```
dzahariev, line 55: Only formatting changes. I notice that for some files I also get different formatting lately, even without meaningful changes in the same file. Live it like this.

components/director/internal/domain/bundle/fixtures_test.go
```diff
@@ -173,7 +173,10 @@ const (
 	tenantID         = "b91b59f7-2563-40b2-aba9-fef726037aa3"
 	externalTenantID = "eeeeeeeee-eeee-eeee-eeee-eeeeeeeeeeee"
 	ordID            = "com.compass.v1"
+	localTenantID    = "localTenantID"
 	correlationIDs   = `["id1", "id2"]`
+	version          = "v1.0.5"
```
dzahariev, line 178: This is not a valid version according ORD spec!
Cannot start with letter and 0 that is not followed by dot!

components/director/internal/open_resource_discovery/fixtures_test.go
```diff
@@ -298,7 +303,9 @@ func fixORDDocumentWithBaseURL(providedBaseURL string) *ord.Document {
 			{
 				Name:                         "BUNDLE TITLE",
 				Description:                  str.Ptr("lorem ipsum dolor nsq sme"),
+				Version:                      str.Ptr("1.1.2"),
```
dzahariev, line 306: Again not valid version according ORD

components/director/internal/open_resource_discovery/fixtures_test.go
```diff
@@ -725,7 +736,9 @@ func fixBundles() []*model.Bundle {
 			ApplicationID:                appID,
 			Name:                         "BUNDLE TITLE",
 			Description:                  str.Ptr("lorem ipsum dolor nsq sme"),
+			Version:                      str.Ptr("1.1.2"),
```
dzahariev, line 739: Same.
It definitely can be used in tests, but is strange that we anyway allow invalid values to be persisted when API is used and perform  input validation only when we are scrapping the ORD data. 
Not something that should be fixed in this PR, but something that can be improved in the future.

components/director/internal/open_resource_discovery/validation_test.go
```diff
@@ -21,6 +21,7 @@ const (
 	invalidOrdID                  = "invalidOrdId"
 	invalidShortDescriptionLength = 257 // max allowed: 256
 	invalidTitleLength            = 256 // max allowed: 255
+	invalidLocalTenantIDLength    = 256 //max allowed: 255
```
dzahariev, line 24: Validation for version is not tested. Or at least I cannot see it here.

components/director/internal/open_resource_discovery/fixtures_test.go
dragobt: i think this is valid version

dzahariev: My mistake!

dzahariev (approved):

PR #3002, opened 2023-04-07
Reduce amount of logs from destination fetcher

georgifarashev:
**Description**

- Reduce amount of logs from destination fetcher

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


petar-iv (dismissed):

NickyMateev (approved):

components/director/internal/destinationfetchersvc/client.go
```diff
@@ -50,9 +49,16 @@ type DestinationServiceAPIConfig struct {
...
+	HTTPClient        *http.Client
+	apiConfig         DestinationServiceAPIConfig
+	authConfig        config.InstanceConfig
+	authToken         string
+	authTokenValidity time.Time
+}
+
+// Close closes idle connections
+func (c *Client) Close() {
+	c.HTTPClient.CloseIdleConnections()
```
alexpenev-s, line 61: By default the connection are kept to be reused if possible. 
Would the setting: MaxIdleConnsPerHost work better to prevent leaks

alexpenev-s: The setting is per host. I suspect that we have a relatively small number of hosts that we communicate with.
I don't think it should be 1 either.

alexpenev-s (approved):

PR #3003, opened 2023-04-07
Enrich runtime metadata with all the necessary properties for tenant mapping

nyordanoff:
**Description**

This PR enriches the runtime by storing a new property to it - `application_namespace` and for kyma runtimes the `region` is set as label.

- add optional `application_namespace` field to the runtime and expose it to graphql
- for kyma runtimes, add logic to store the region based on the region of the subaccount passed as `global_subaccount_label` on runtime creation
- for kyma runtimes, automatically set the `application_namespace` to `sap.kyma`
- unit & e2e tests
- manual migration of the two new fields for existing kyma runtimes

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/runtime/resolver_test.go
```diff
@@ -1,5 +1,6 @@
 package runtime_test
 
+/*
```
PetarTodorovv, line 3: I guess it's leftover? and the test fixtures should be adapted?

tests/director/tests/runtime_api_test.go
```diff
@@ -745,6 +748,17 @@ func TestRuntimeTypeLabels(t *testing.T) {
...
+		t.Log("Validate that the Application Namespace is available...")
+		appNamespace := runtime.ApplicationNamespace
+		require.NotNil(t, appNamespace)
+		require.Equal(t, conf.KymaApplicationNamespaceValue, *appNamespace)
+
+		t.Log("Validate that the region label of the runtime is available...")
+		rtRegionLabelValue, ok := runtime.Labels[tenantfetcher.RegionKey].(string)
+		require.True(t, ok)
+		require.Equal(t, conf.DefaultTenantRegion, rtRegionLabelValue)
+
```
PetarTodorovv, line 761: Could we also add "negative" assertions in the sub-test below that is registering the runtime without integration system credentials?

tests/director/tests/runtime_api_test.go
PetarTodorovv, line 1: Also, I'm wondering if we should update the test(+ the sub-tests) names as well because currently, it's related only to runtimeType. Example name of the test - `TestRuntimeTypeAndRegionLabels`, and for the sub-tests - `fmt.Sprintf("Validate %q and %q labels - they are added when runtime is registered with integration system credentials", conf.RuntimeTypeLabelKey, tenantfetcher.RegionKey)` and for the second sub-test(the negative one) - `fmt.Sprintf("Validate %q and %q labels - they are missing when runtime is NOT registered with integration system credentials", conf.RuntimeTypeLabelKey, tenantfetcher.RegionKey)`

PetarTodorovv (changes requested):

components/director/internal/domain/runtime/resolver_test.go
nyordanoff: My bad, will adapt them.

chart/compass/values.yaml
```diff
@@ -298,6 +298,7 @@ global:
     runtimeTypeLabelKey: "runtimeType"
     applicationTypeLabelKey: "applicationType"
     kymaRuntimeTypeLabelValue: "kyma"
+    kymaApplicationNamespaceValue: "sap.kyma"
```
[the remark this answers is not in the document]
nyordanoff: I was considering it but decided against this approach because to me it looked that it's not worth it maintaining an override for such a 'narrow scoped'/'single use case' variable.

In addition, currently we don't have an override for the similar `kymaRuntimeTypeLabelValue` variable and decided to keep it consistent in that aspect too.

What's your opinion? I am open for a discussion.

alextargov (dismissed):

PetarTodorovv (approved):

alextargov (approved):

PR #3017, opened 2023-04-13
Allow for returning state in the response body for sync TM Notifications instead of relying on status codes only

la4ezar:
**Description**
Currently, we are relying only on status codes in synchronous tenant mapping notifications. We want to allow returning state in the response body too.

- Allow returning state in the response body in sync TM notifications.
- Unit tests
- E2e test
- Adapt external-svc-mock handlers

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/external-services-mock/internal/notification/handler.go
```diff
@@ -116,56 +116,85 @@ func NewHandler(notificationConfiguration NotificationsConfiguration) *Handler {
...
+					Key2 struct {
+						Key string `json:"key"`
+					} `json:"key2"`
+				}{
+					Key: "value",
+					Key2: struct {
+						Key string `json:"key"`
+					}{Key: "value2"},
+				},
+			}
```
PetarTodorovv, line 194: Isn't there a better approach instead of this `if`-`else` approach?(Valid for the other handlers as well). For example - to use a separate handler instead of modifying the existing one with this `shouldHaveStateInBody` property. Similar to the async handlers we could extract the main logic into a function and define a "response function" that in the "with-state" handler will be configured to send `state` but in the "normal handler"(patch/delete) will send only status code?

Also, could we define a typed stucture(similar to the ones for the request body - `FormationAssignmentRequestBody` and `FormationRequestBody`) instead of using an anonymous one? Even before when we had only one response it was hard/ugly to read it. But now we duplicate the same thing with an additional property.

tests/director/tests/formation_api_test.go
```diff
@@ -1519,6 +1519,168 @@ func TestFormationNotificationsWithApplicationOnlyParticipants(t *testing.T) {
...
+		urlTemplate := "{\\\"path\\\":\\\"" + conf.ExternalServicesMockMtlsSecuredURL + "/formation-callback/with-state/{{.TargetApplication.ID}}{{if eq .Operation \\\"unassign\\\"}}/{{.SourceApplication.ID}}{{end}}\\\",\\\"method\\\":\\\"{{if eq .Operation \\\"assign\\\"}}PATCH{{else}}DELETE{{end}}\\\"}"
+		inputTemplate := "{\\\"ucl-formation-id\\\":\\\"{{.FormationID}}\\\",\\\"globalAccountId\\\":\\\"{{.CustomerTenantContext.AccountID}}\\\",\\\"crmId\\\":\\\"{{.CustomerTenantContext.CustomerID}}\\\",\\\"items\\\":[{\\\"region\\\":\\\"{{ if .SourceApplication.Labels.region }}{{.SourceApplication.Labels.region}}{{ else }}{{.SourceApplicationTemplate.Labels.region}}{{ end }}\\\",\\\"application-namespace\\\":\\\"{{.SourceApplicationTemplate.ApplicationNamespace}}\\\",\\\"tenant-id\\\":\\\"{{.SourceApplication.LocalTenantID}}\\\",\\\"ucl-system-tenant-id\\\":\\\"{{.SourceApplication.ID}}\\\"}]}"
+		outputTemplate := "{\\\"config\\\":\\\"{{.Body.Config}}\\\", \\\"state\\\":\\\"{{.Body.state}}\\\", \\\"location\\\":\\\"{{.Headers.Location}}\\\",\\\"error\\\": \\\"{{.Body.error}}\\\",\\\"success_status_code\\\": 200, \\\"incomplete_status_code\\\": 204}"
```
PetarTodorovv, line 1530: I think we don't need the `incomplete_status_code` here, do we?

PetarTodorovv (changes requested):

components/director/internal/domain/formationassignment/service.go
```diff
@@ -674,6 +682,29 @@ func (s *service) processFormationAssignmentsWithReverseNotification(ctx context
...
+	if newState != string(model.InitialAssignmentState) &&
+		newState != string(model.ReadyAssignmentState) &&
+		newState != string(model.ConfigPendingAssignmentState) &&
+		newState != string(model.CreateErrorAssignmentState) &&
+		newState != string(model.DeletingAssignmentState) &&
+		newState != string(model.DeleteErrorAssignmentState) {
+		return false
+	}
+
+	if previousState == string(model.DeletingAssignmentState) &&
```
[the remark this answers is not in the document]
la4ezar: Currently, this function `validateResponseState` is used only in the synchronous case, when there is a state in the body.

components/director/internal/domain/formationassignment/service.go
```diff
@@ -610,13 +610,21 @@ func (s *service) processFormationAssignmentsWithReverseNotification(ctx context
 		return nil
 	}
 
-	if *response.ActualStatusCode == *response.SuccessStatusCode {
-		assignment.State = string(model.ReadyAssignmentState)
-		assignment.Value = nil
-	}
+	if response.State != nil { // if there is a state in the response
```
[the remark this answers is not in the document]
la4ezar: The task was to adapt this function to handle when there is a state in the body, but maybe the unassign should be adjusted too. I'll double-check it and adjust it if needed.

tests/director/tests/formation_api_test.go
la4ezar: We don't need it but I left it there to make sure that even when there is an incomplete status code, the business logic gets the state from the body(the flow goes through the `if response.State != nil {...}` in `processFormationAssignmentsWithReverseNotification` function).

components/director/internal/domain/formationassignment/service.go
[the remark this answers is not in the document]
PetarTodorovv: hm, if it's related only to the "assign" why do we have the first `if` - with "deleting/delete error" states? The formation assignment could be in them only during unassign?

tests/director/tests/formation_api_test.go
PetarTodorovv: Yep, my question is regarding that because that way there is a risk if someone for some reason(not intentionally) changes the status code in the ext-svc-mock from 200 OK to 204 no content by mistake we cannot be sure config_pending state where is coming from - is it from the `response.state` logic or from the "old" logic. So, maybe it's safer to remove it from the output template? What's your thoughts?

components/external-services-mock/internal/notification/handler.go
```diff
@@ -64,6 +64,25 @@ type FormationRequestBody struct {
 	Error string         `json:"error,omitempty"`
 }
 
+// FormationResponseBody contains the synchronous formation assignment notification response body
+type FormationResponseBody struct {
```
PetarTodorovv, line 68: `FormationAssignmentResponseBody`

components/external-services-mock/internal/notification/handler.go
```diff
@@ -281,66 +232,28 @@ func (h *Handler) GetResponses(writer http.ResponseWriter, r *http.Request) {
...
-		if r.Method == http.MethodDelete {
-			applicationId, ok := mux.Vars(r)["applicationId"]
-			if !ok {
-				httphelpers.WriteError(writer, errors.New("missing applicationId in url"), http.StatusBadRequest)
-				return
+		responseFunc := func([]byte) {
+			response := struct {
+				Error string `json:"error"`
+			}{
+				Error: "failed to parse request",
```
PetarTodorovv, line 239: Couldn't we reuse the FAResponseBody with Error field in it and json tag `json:"error,omitempty"` and reuse it here?

components/external-services-mock/internal/notification/handler.go
```diff
@@ -281,66 +232,28 @@ func (h *Handler) GetResponses(writer http.ResponseWriter, r *http.Request) {
...
 	if r.Method == http.MethodPatch {
-		h.Patch(writer, r)
+		patchHandler := h.Patch
+		patchHandler(writer, r)
 	}
 
 	if r.Method == http.MethodDelete {
-		h.Delete(writer, r)
+		deleteHandler := h.Delete
+		deleteHandler(writer, r)
```
PetarTodorovv, line 256: Why this is needed? Couldn't we use just `h.Patch(writer, r)` and `h.Delete(writer, r)`

components/external-services-mock/internal/notification/handler_test.go
```diff
@@ -63,7 +63,66 @@ func TestHandler_Patch(t *testing.T) {
 			r := httptest.NewRecorder()
 
 			//WHEN
-			h.Patch(r, req)
+			patchHandler := h.Patch
+			patchHandler(r, req)
```
PetarTodorovv, line 67: why do we need this? couldn't we use `h.Patch(r, req)`?
P.S. Valid for the other places as well

tests/director/tests/formation_api_test.go
```diff
@@ -1519,6 +1519,168 @@ func TestFormationNotificationsWithApplicationOnlyParticipants(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, formationName, assignedFormation.Name)
+
+		expectedAssignments = map[string]map[string]fixtures.AssignmentState{
+			app1.ID: {
+				app1.ID: fixtures.AssignmentState{State: "READY", Config: nil},
+				app2.ID: fixtures.AssignmentState{State: "READY", Config: nil},
+			},
+			app2.ID: {
+				app1.ID: fixtures.AssignmentState{State: "READY", Config: str.Ptr("{\"key\":\"value\",\"key2\":{\"key\":\"value2\"}}")},
```
PetarTodorovv, line 1573: Why in this [commit](https://github.com/kyma-incubator/compass/pull/3017/commits/f9f1edddc80feaaa26a4090981d2a20d8057bf3c) in the `tests/director/tests/formation_api_test.go` file you changed the `config_pending` state to `ready`? Shouldn't that be `config_pending` and test the new functionality where the state is provided in the response body? In this test, we call `.../formation-callback/with-state/..` which should return `config_pending`? Am I missing something?

That's valid for the other assertions in this commit as well - lines 1578, 1627, and 1633

PetarTodorovv (changes requested):

tests/director/tests/formation_api_test.go
PetarTodorovv: Hmm, I think I found the reason for that - you changed the state in the ext-svc-mock in the Patch(and other handlers) from `config_pending` to `ready`? [source link](https://github.com/kyma-incubator/compass/pull/3017/commits/f9f1edddc80feaaa26a4090981d2a20d8057bf3c)
image

Is there a specific reason for that? I think it's better to use a different state than `ready` because we already return "success status code" 200 and the "default behavior" for that is to set the assignment in a ready state? that way we cannot be sure from where the `ready` state comes from, is it from the state field in the response or either from the "old logic" that had?

components/director/internal/domain/formationassignment/service.go
```diff
@@ -738,6 +762,12 @@ func (s *service) CleanupFormationAssignment(ctx context.Context, mappingPair *A
 		return false, nil
 	}
 
+	if response.State != nil && *response.State == string(model.DeleteErrorAssignmentState) {
```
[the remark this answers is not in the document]
la4ezar: This `if` handles the corner case when there is an error state in the body, but no error in the response. If we don't have this `if` and there is `DELETE_ERROR` in the `response.State`, but nothing in the `response.Error` then we will not set the assignment in an error state. But we want to set it in an error state because there is a state(which is an error) in the response body and we want to respect it. Also, the message is empty, because we don't have message to put there.

PS: this was discussed with another colleague too.

tests/director/tests/formation_api_test.go
la4ezar: No, there is no reason. I switched it to READY because there were problems with the unassign method and I forgot to return it to READY.

components/external-services-mock/internal/notification/handler.go
PetarTodorovv: seems the delete is not adapted?

PetarTodorovv (approved):

PR #3021, opened 2023-04-18
Add customer adoption metric

nyordanoff:
**Description**
This PR introduces a chart in grafana that displays our customer adoption and track its growth overtime. It tracks the number of tenants that have at least one non auto-discovered Application (the tenant is owner and its parent is null) or Global Account tenants that have at least 1 formation different from the default one.

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/chart/compass/charts/prometheus-postgres-exporter/values.yaml
+++ b/chart/compass/charts/prometheus-postgres-exporter/values.yaml
@@ -319,7 +319,7 @@ config:
             description: "Number of tenants which are assigned to at least one runtime or application"
 
     compass_director_customer_adoption:
-      query: "SELECT COUNT(DISTINCT (id)) FROM business_tenant_mappings WHERE (id IN ( SELECT b.id FROM applications a JOIN tenant_applications ta ON a.id = ta.id JOIN business_tenant_mappings b ON ta.tenant_id = b.id WHERE ta.owner IS true AND ta.id IN (SELECT app_id FROM labels WHERE key='managed' AND value ? 'false') AND b.parent IS null) OR id IN ( SELECT b2.id FROM formations JOIN business_tenant_mappings b2 ON formations.tenant_id = b2.id WHERE name <> 'DEFAULT' AND (parent IS null OR parent IN (SELECT id FROM business_tenant_mappings b3 WHERE type = 'customer'))));"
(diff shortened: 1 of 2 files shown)


nyordanoff:
reset-vm

chart/compass/charts/prometheus-postgres-exporter/values.yaml
```diff
@@ -318,6 +318,14 @@ config:
             usage: "COUNTER"
             description: "Number of tenants which are assigned to at least one runtime or application"
 
+    compass_director_customer_adoption:
+      query: "SELECT COUNT(DISTINCT (id)) as quantity FROM business_tenant_mappings WHERE (id IN ( SELECT b.id FROM applications a JOIN tenant_applications ta ON a.id = ta.id JOIN business_tenant_mappings b ON ta.tenant_id = b.id WHERE ta.owner IS true AND ta.id IN (SELECT app_id FROM labels WHERE key='managed' AND value ? 'false') AND b.parent IS null) OR id IN ( SELECT b2.id FROM formations JOIN business_tenant_mappings b2 ON formations.tenant_id = b2.id WHERE name <> 'DEFAULT' AND (parent IS null OR parent IN (SELECT id FROM business_tenant_mappings b3 WHERE type = 'customer'))));"
+      master: true
+      metrics:
+        - quantity:
+            usage: "COUNTER"
+            description: "Number of tenants that have at least one non auto-discovered Application (the tenant is owner and its parent is null) or Global Account tenants that have at least 1 formation different from the default one"
```
PetarTodorovv, line 327: `.. or Global Account tenants that ..` from the queary above it seems not only GA tennats are included in that result? There could be `customer` tenants(they may also not have parent), also atom tenats - organizations? Are they expected in this customer adoption calculation?

nyordanoff: Yes, there are not only GAs. But the second part of the query related to the formations gives us GAs - that's why the `or` in the description.

Maybe I can change it a bit for more clarity: `Number of customer tenants <bla bla> or GA tenants that <bla bla>`?

PetarTodorovv (approved):

PR #3033, opened 2023-04-25
Do not send formation assignment error in notifications

ivantenevvasilev:
**Description**

- do not send error as part of configuration when resynchronizing formation
- add e2e test that checks that no error is sent
- add reverse assignment and checks for received configuration in e2e tests
- add `fail` endpoint that always returns errors in external service mock

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


ivantenevvasilev:
reset-vm

ivantenevvasilev:
reset-vm

ivantenevvasilev:
reset-vm

components/director/internal/domain/formationassignment/notifications.go
```diff
@@ -401,7 +401,7 @@ func convertFormationAssignmentFromModel(formationAssignment *model.FormationAss
 		return &webhook.FormationAssignment{Value: "\"\""}
 	}
 	config := string(formationAssignment.Value)
-	if config == "" {
+	if config == "" || formationAssignment.State == string(model.CreateErrorAssignmentState) || formationAssignment.State == string(model.DeleteErrorAssignmentState) {
```
PetarTodorovv, line 404: Shouldn't we also adapt the state as well? In case of "create_error" to set the state in "initial" and when we have "delete_error" --> "deleting"?

P.S. Valid for the other changes as well

components/director/pkg/webhook/application_tenant_mapping.go
PetarTodorovv, line 1: Most probably our e2e tests or/and tenant mappings mocks in the external-service-mock should be adapted so we could verify the correct state and configuration is set during the resync of the notifications? Maybe the ext-svc-mock is the logical/better place since it's a mock implementation of the TM API and another benefit would be that we'll do that assert/check only in one place(not in every e2e tests)?
What's your thoughts on that?

PetarTodorovv (changes requested):

components/director/internal/domain/formationassignment/notifications.go
ivantenevvasilev: This is during the sending of the notifications. The transition is already done after it is sent in `CleanupFormationAssignment` and `processFormationAssignmentsWithReverseNotification`. 
The flow is as follows:
1. Get all assignments
2. Build notifications payload
3. Send notification 
4. Update the Formation Assignment based on the response (and in the case from CREATE_ERROR/DELETE_ERROR to INITIAL/DELETING it is cleared)

components/director/pkg/webhook/application_tenant_mapping.go
ivantenevvasilev: It would be better for the testing to be in the e2e tests in my opinion. If we send a wrong payload any one of the operations could fail assign/unassign/create/delete/resync and we wouldn't know why the notification failed in the e2e tests. The adaptation could be in `assertFormationAssignmentsNotification` and it would have the same logic as when we check the other properties.

components/director/internal/domain/formationassignment/notifications.go
ivantenevvasilev: There are even e2e tests that currently check the cleanup of errors when changing to `DELETING` and `INITIAL`, but there are missing asserts for the notifications that are received by external service mock.

ivantenevvasilev:
reset-vm

ivantenevvasilev:
reset-vm

PetarTodorovv (approved):

PR #3035, opened 2023-04-26
Remove full Kyma installation

la4ezar:
**Description**
The 'full' Kyma installation is outdated and we are not using it, that's why we are removing everything related to the full Kyma installation that we have around our scripts and configuration files.

- Remove full Kyma installation from scripts & docs.

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


ekaterina-mitova (approved):
All looks good to me.

docs/compass/04-01-installation.md
```diff
@@ -245,11 +245,6 @@ The Kyma version is read from the [`KYMA_VERSION`](../../installation/resources/
 ```bash
 ./installation/cmd/run.sh --kyma-release {KYMA_VERSION} --oidc-host {URL_TO_OIDC_SERVER} --oidc-client-id {OIDC_CLIENT_ID} --oidc-admin-group {OIDC_ADMIN_GROUP}
 ```
-You can also specify if you want the Kyma installation to contain only `minimal` components or whether you want `full` Kyma
```
StanislavStefanov, line 248: there are few places where we are talking about minimal Kyma instalation. As It is the only type of installation that we support and there is no flag minimal that is passed to the script we could remove the minimal from the documentation and the file names

StanislavStefanov: NOTE! If we decide to rename the files that contain minimal in their names they should be updated in the config repo also

StanislavStefanov (approved):

PR #3040, opened 2023-04-27
Change the image registry

PetarTodorovv:
**Description**
Changing the kyma image registry from `eu.gcr.io` to `europe-docker.pkg.dev`. For the time being the directory of the images will be `dev/incubator/` but when we clarify the process it most probably will be changed to `prod/incubator`

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] [N/A] Unit tests
- [x] [N/A] Integration tests

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -124,7 +124,7 @@ global:
       parent: 4c31b7c7-2bea-4bd5-9ea5-e9a8d704f900
   images:
     containerRegistry:
-      path: eu.gcr.io/kyma-project/incubator
+      path: europe-docker.pkg.dev/kyma-project/dev/incubator
     connector:
       dir:
       version: "PR-3022"
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -124,49 +124,49 @@ global:
       parent: 4c31b7c7-2bea-4bd5-9ea5-e9a8d704f900
   images:
     containerRegistry:
-      path: europe-docker.pkg.dev/kyma-project/dev/incubator
+      path: europe-docker.pkg.dev/kyma-project
     connector:
-      dir:
+      dir: dev/incubator/
       version: "PR-3022"
       name: compass-connector
     connectivity_adapter:
-      dir:
+      dir: dev/incubator/
       version: "PR-3022"
       name: compass-connectivity-adapter
     pairing_adapter:
-      dir:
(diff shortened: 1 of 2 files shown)


chart/compass/values.yaml
```diff
@@ -124,49 +124,49 @@ global:
...
-      dir:
+      dir: dev/incubator/
       version: "PR-3022"
       name: compass-schema-migrator
     system_broker:
-      dir:
+      dir: dev/incubator/
       version: "PR-3022"
       name: compass-system-broker
     certs_setup_job:
```
dzahariev, line 172: This is a corner case. 
The image is missing on the new repo. We need to clarify what will happen with old repo. If it is decommissioned, then we may need to transport this image to the new one.

dzahariev (commented):
What about buildpack-golang, it is available at https://console.cloud.google.com/artifacts/docker/kyma-project/europe/prod/test-infra%2Fbuildpack-golang
Probably we should switch it in Makefile-s.

The following charts also need to be adapted:
- chart/compass/charts/testing/values.yaml
- chart/compass/templates/agent-configuration-job.yaml
- chart/compass/templates/registration-job.yaml
- docs/compass/assets/order-service.yaml (cannot find the same image in prod repo)
- installation/scripts/prom-mtls-patch.sh

PetarTodorovv:
> What about buildpack-golang, it is available at https://console.cloud.google.com/artifacts/docker/kyma-project/europe/prod/test-infra%2Fbuildpack-golang Probably we should switch it in Makefile-s.
> 
> The following charts also need to be adapted:
> 
> * chart/compass/charts/testing/values.yaml
> * chart/compass/templates/agent-configuration-job.yaml
> * chart/compass/templates/registration-job.yaml
> * docs/compass/assets/order-service.yaml (cannot find the same image in prod repo)
> * installation/scripts/prom-mtls-patch.sh

The buildpack images, cert-setup-job image, and the others will be available in the "old" registry. Only the "PR images" will be removed. About the first ones maybe in the near future, we should consider migrating them.

chart/compass/values.yaml
PetarTodorovv: Check the other comment as well but it will be available in the old registry

dzahariev (dismissed):

dzahariev (approved):

PR #3041, opened 2023-04-28
Switch between pointer values

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/pkg/auth/converter.go
dragobt, line 1: as you have changes in pkg dir, you should update the dependencies

dragobt: we discussed it offline.

dragobt (approved):

PR #3054, opened 2023-05-04
Fix runtime update

nyordanoff:
**Description**

Before this PR, when we updated a Runtime, if the update input did not include some pointer fields of the runtime (`description`, `applicationNamespace`) they were automatically set to `null` after the update.

This PR fixes that issue and updates the pointer values only if they are present in the updated input. Otherwise, the current value is kept.

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -139,7 +139,7 @@ global:
       name: compass-pairing-adapter
     director:
       dir: dev/incubator/
-      version: "PR-3051"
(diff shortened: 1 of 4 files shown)


nyordanoff:
reset-vm

PetarTodorovv (approved):

PR #3061, opened 2023-05-12
[Formation constraint operator] Do not send notifications to formation participants of a given type

DimitarPetrov:
**Description**

- Add constraints operator that disables formation assignment notifications
- Fixes and adjustments around the `GenerateFormationAssignmentNotificationOperationDetails` joinpoint

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests
- [N/A] Mocks are regenerated, using the automated script


components/director/internal/domain/formationconstraint/constraint_operators_test.go
```diff
@@ -386,3 +385,249 @@ func TestConstraintOperators_DoesNotContainResourceOfSubtype(t *testing.T) {
...
+		{
+			Name:  "Success for a system when notifications should be skipped",
+			Input: in,
+			LabelSvc: func() *automock.LabelService {
+				svc := &automock.LabelService{}
+				svc.On("GetByKey", ctx, testTenantID, model.ApplicationLabelableObject, appID, applicationTypeLabel).Return(&model.Label{Value: inputAppType}, nil).Once()
+				return svc
+			},
+			ExpectedResult:   false,
+			ExpectedErrorMsg: "",
```
la4ezar, line 464: In `success` tests this `ExpectedErrorMsg` can be skipped, it will use the default string value which is `""`.

components/director/internal/domain/formationconstraint/constraint_operators_test.go
```diff
@@ -386,3 +385,249 @@ func TestConstraintOperators_DoesNotContainResourceOfSubtype(t *testing.T) {
...
+				repo.On("GetByID", ctx, testTenantID, runtimeCtxID).Return(&model.RuntimeContext{RuntimeID: runtimeID}, nil).Once()
+				return repo
+			},
+			ExpectedResult:   false,
+			ExpectedErrorMsg: "",
+		},
+		{
+			Name:  "Error for runtime context when get rt ctx fails",
+			Input: runtimeContextIn,
+			LabelSvc: func() *automock.LabelService {
```
la4ezar, line 573: Change to `LabelSvc: UnusedLabelService`. This applies to the other tests too.

tests/director/tests/formation_constraint_api_test.go
```diff
@@ -16,6 +16,7 @@ import (
 const (
 	IsNotAssignedToAnyFormationOfTypeOperator = "IsNotAssignedToAnyFormationOfType"
 	DoesNotContainResourceOfSubtypeOperator   = "DoesNotContainResourceOfSubtype"
+	DoNotSendNotificationOperator = "DoNotSendNotification"
```
la4ezar, line 19: Wouldn't it be better if we update the dependencies in the tests and use `graphql.DoNotSendNotificationOperator` instead of double-maintaining it? This applies to other operator names too.

la4ezar (changes requested):

tests/director/tests/formation_api_test.go
```diff
@@ -2415,6 +2419,449 @@ func TestFormationNotificationsWithApplicationOnlyParticipants(t *testing.T) {
...
+		t.Logf("exceptionTypeApp ID: %q", exceptionTypeApp.ID)
+
+		t.Logf("Assign application 3 (exception one) to formation %s", formationName)
+		defer fixtures.CleanupFormation(t, ctx, certSecuredGraphQLClient, graphql.FormationInput{Name: formationName}, exceptionTypeApp.ID, graphql.FormationObjectTypeApplication, tnt)
+		assignReq = fixtures.FixAssignFormationRequest(exceptionTypeApp.ID, string(graphql.FormationObjectTypeApplication), formationName)
+		err = testctx.Tc.RunOperationWithCustomTenant(ctx, certSecuredGraphQLClient, tnt, assignReq, &assignedFormation)
+		require.NoError(t, err)
+		require.Equal(t, formationName, assignedFormation.Name)
+
+		expectedAssignments = map[string]map[string]fixtures.AssignmentState{
```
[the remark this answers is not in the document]
DimitarPetrov: This map contains the assignments, they are generated and READY even if there are no notifications.

Below (on line 2551) you can see that there is an assert that only one notification is received in the external services mock - the excepted one.

tests/director/tests/formation_constraint_api_test.go
DimitarPetrov: We don't have such value `graphql.DoNotSendNotificationOperator`. Operators are just strings for the graphQL schema. The const in the director is in the internal package. I also tried it, but it looks like we need to make the operators enum in the graphQL schema, which is quite limiting having in mind the UI constraint and the operators there which does not involve a change in compass.

la4ezar: We have the strings in the `components/director/pkg/graphql/formation_constraint_validation.go` which are double-maintained to be == from the internal package.

la4ezar (approved):

alextargov (approved):

PR #3071, opened 2023-05-16
Allow CLD-based application providers to override UCL defaults for individual properties when mapping SLIS systems

dragobt:
**Description**
As a CLD-based application provider I want to be able to override UCL's defaults for system property mappings with SLIS/CLD so that I can have the right properties associated with my system records in UCL.

- System Fetcher respects jsonPath defined in placeholders of an application template created by a customer when he onboard their CLD-based product to UCL
- Change jsonPaths values to look like this `$.<jsonPath>` and adapt the code
- Build `$select=` filter once during runtime
- Drop `systemBase` struct and make System Fetcher to work directly with the payload from SLIS
- Drop temporary logic for systems which `productId == S4_PC`
- Add and adapt unit tests

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/systemfetcher/client.go
```diff
@@ -179,15 +180,14 @@ func (c *Client) buildFilter() map[string]string {
...
-
 	result := map[string]string{"fetchAcrossZones": "true"}
 
 	if len(c.apiConfig.FilterCriteria) > 0 {
 		result["$filter"] = fmt.Sprintf(c.apiConfig.FilterCriteria, filterBuilder.buildFilterQuery())
 	}
 
 	if len(c.apiConfig.SelectCriteria) > 0 {
-		result["$select"] = c.apiConfig.SelectCriteria
+		result["$select"] = fmt.Sprintf(c.apiConfig.SelectCriteria, strings.Join(SelectFilter, ","))
```
kaloyanspiridonov, line 190: Do we need c.apiConfig.SelectCriteria?

kaloyanspiridonov, line 190: Do we need c.apiConfig.SelectCriteria?

chart/compass/values.yaml
```diff
@@ -914,7 +914,7 @@ global:
...
+    templatePlaceholderToSystemKeyMappings: '[ { "placeholder_name": "name", "system_key": "$.displayName" }, { "placeholder_name": "display-name", "system_key": "$.displayName" }, { "placeholder_name": "systemNumber", "system_key": "$.systemNumber" }, { "placeholder_name": "productId", "system_key": "$.productId" }, { "placeholder_name": "tenant-id", "system_key": "$.systemNumber" }, { "placeholder_name": "ppmsProductVersionId", "system_key": "$.ppmsProductVersionId", "optional": true }, { "placeholder_name": "region", "system_key": "$.additionalAttributes.systemSCPLandscapeID", "optional": true }, { "placeholder_name": "description", "system_key": "$.productDescription", "optional": true }, { "placeholder_name": "baseUrl", "system_key": "$.additionalUrls.mainUrl", "optional": true }, { "placeholder_name": "providerName", "system_key": "$.infrastructureProvider", "optional": true } ]'
```
NickyMateev, line 917: How much effort would it be to not define the `tenant-id` placeholder here and rather define in the internal config with `jsonPath`? Reason being is that having it here indicates that in general `$.systemNumber` is a reasonable default for the local tenant id, which is not the case. There isn't a well-established value which System Fetcher can rely on being "in most cases" the right local tenant id. So putting it here is very S4-biased.

Also this makes the S4 app template (in the internal config) very coupled wit this `templatePlaceholderToSystemKeyMappings`, so at least from a readability and troubleshooting standpoint it would be better to define the placeholder with it's jsonPath in the internal config as part of the S4 app template.

At the end of the day this is not something that's a MUST fix, but I think it would overall result in a better setup/configuration. So pls if you agree, just consider what's the effort doing this and if it would risk the timely delivery of this item, I'd postpone this tweak for a later item/task.

kaloyanspiridonov (changes requested):

kaloyanspiridonov (approved):

ekaterina-mitova (approved):
Looks good to me!

PR #3097, opened 2023-06-01
Fix Resync API with generate FA constraint attached

ivantenevvasilev:
**Description**
When there is a constraint with type `GENERATE_FORMATION_ASSIGNMENT_NOTIFICATION` attached to a participant in the formation and the resync API is called, the constraint throws an error rather than not generating a notification. It should just skip the notification, not throw an error. 

- change the error to a non-fatal one and just skip the generation of a FA notification

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] [N/A] Integration tests


ivantenevvasilev:
reset-vm

StanislavStefanov (approved):

PR #3101, opened 2023-06-05
[HOTFIX] Fix get Application template Webhook for Formation Notifications

StanislavStefanov:
**Description**

When selecting Application templates Webhooks for Formation Notifications the webhook type is not specified. It is not deterministic which webhook will be returned if the template has more than one webhook.

- select only webhooks of specific type


StanislavStefanov:
reset-vm

PR #3105, opened 2023-06-08
Delete formation assignment at the start of unassign from formation

ivantenevvasilev:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] [N/A] Integration tests

### DIFF
--- a/components/director/internal/domain/formation/service.go
+++ b/components/director/internal/domain/formation/service.go
@@ -792,6 +792,33 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
 	}
 
 	formationFromDB := ft.formation
+	if objectType != graphql.FormationObjectTypeTenant {
+		selfFATx, err := s.transact.Begin()
+		if err != nil {
+			return nil, err
+		}
+		selfFATransactionCtx := persistence.SaveToContext(ctx, selfFATx)
+		defer s.transact.RollbackUnlessCommitted(selfFATransactionCtx, selfFATx)
+
+		fa, err := s.formationAssignmentService.GetReverseBySourceAndTarget(selfFATransactionCtx, formationFromDB.ID, objectID, objectID)
+		if err != nil {
+			commitErr := selfFATx.Commit()
+			if commitErr != nil {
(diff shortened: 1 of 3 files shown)


ivantenevvasilev:
reset-vm

components/director/internal/domain/formation/resolver.go
```diff
@@ -490,6 +498,32 @@ func (r *Resolver) ResynchronizeFormationNotifications(ctx context.Context, form
 	return r.conv.ToGraphQL(updatedFormation)
 }
 
+func (r *Resolver) deleteSelfFormationAssignment(ctx context.Context, tnt, formationName, objectID string) error {
```
PetarTodorovv, line 501: **_minor:_** what do you thing about `deleteSelfReferencedFormationAssignment`?

components/director/internal/domain/formation/resolver.go
```diff
@@ -490,6 +498,32 @@ func (r *Resolver) ResynchronizeFormationNotifications(ctx context.Context, form
...
+	selfFATransactionCtx := persistence.SaveToContext(ctx, selfFATx)
+	defer r.transact.RollbackUnlessCommitted(selfFATransactionCtx, selfFATx)
+
+	formationFromDB, err := r.service.GetFormationByName(selfFATransactionCtx, formationName, tnt)
+	if err != nil {
+		log.C(ctx).Errorf("An error occurred while getting formation by name: %q: %v", formationName, err)
+		return errors.Wrapf(err, "An error occurred while getting formation by name: %q", formationName)
+	}
+
+	fa, err := r.formationAssignmentSvc.GetReverseBySourceAndTarget(selfFATransactionCtx, formationFromDB.ID, objectID, objectID)
```
PetarTodorovv, line 515: **_minor:_** This works because the source and target are the same but I wonder if it would be more readable to use `r.formationAssignmentSvc.GetBySourceAndTarget` when we get "self-referenced FA"?

ivantenevvasilev: I think so too, but there is no such method in the service for formation assignments and it's already overloaded with methods.

PR #3108, opened 2023-06-09
Fix matching certificate subjects

StanislavStefanov:
**Description**
When CMP is called with certificate as part of the request processing the hydarator component is matching the certificate subject against whitelisted certificate subject mappings. The OU parts of the subject can be separated by `,` or by `+`. If the subject from the request uses `,` for separator and the whitelisted subject uses `+` as separator the hydrator fails to match them.

- Use cert utils methods to compare the subject parts

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests

### DIFF
--- a/components/director/pkg/cert/certutils_test.go
+++ b/components/director/pkg/cert/certutils_test.go
@@ -10,7 +10,7 @@ import (
 )
 
 func TestSubjectExtraction(t *testing.T) {
-	for _, testCase := range []struct {
+	testCases := []struct {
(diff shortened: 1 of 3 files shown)


PetarTodorovv (approved):

PetarTodorovv (approved):

PetarTodorovv:
reset-vm

PetarTodorovv (approved):

PR #3112, opened 2023-06-13
Fix formation assignments stuck in initial state when object is without any webhooks

ivantenevvasilev:
**Description**
When there are formation lifecycle notifications all formation assignments are created in `INITIAL` state and when the formation is ready they get resynchronized. After #3097 the formation assignments that do not have webhooks get stuck in `INITIAL` state because they are skipped.

- in `resynchronizeFormationAssignments` update assignments that are in `INITIAL` state with no webhook to `READY`

**Related issue(s)**
- #3097

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] [N/A] Integration tests


ivantenevvasilev:
reset-vm

PR #3092, opened 2023-05-30
Introduce formation destination operator

PetarTodorovv:
**Description**

Changes proposed in this pull request:
- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests
- [ ] `chart/compass/values.yaml` is updated 
- [ ] Mocks are regenerated, using the automated script

PR #3152, opened 2023-06-29
Fix error handling in the destination creator operator

PetarTodorovv:
**Description**
Currently, in the destination creator formation "constraint" operator we have special error handling if something fails during destination creation to update the formation assignment in an error state and to return no errors. But with that behaviour, we do not stop the formation notifications propagation whereas we should.

- remove the "special" error handling in the destination creator operator and return the error, the rest of the code will handle it independently
- During "delete with constraints" before deleting the formation assignment we need to update it to "READY" state

**Related issue(s)**

- https://github.com/kyma-incubator/compass/pull/3092

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -139,7 +139,7 @@ global:
       name: compass-pairing-adapter
     director:
       dir: dev/incubator/
-      version: "PR-3135"
+      version: "PR-3152"
       name: compass-director
     hydrator:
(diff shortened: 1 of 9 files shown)


PetarTodorovv:
reset-vm

nyordanoff (approved):

PR #3163, opened 2023-07-11
Split Formation assignment error and configuration

StanislavStefanov:
**Description**
Keeping the configuration and errors in the same column for formation assignments is errorprone. Also we can not keep old configurations in case of error(the error overrides the configuration).

- separate configuration and error in different columns
- expose two new properties for the formation assignments - configuration and error
- calculate the value of the value property of formation assignments - If error != nil -> take that with priority, otherwise return the configuration
- enrich e2e tests to validate the new properties 

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


tests/director/tests/formation_api_test.go
```diff
@@ -5491,10 +5516,12 @@ func assertFormationAssignmentsAsynchronously(t *testing.T, ctx context.Context,
 
 		assignmentExpectation, ok := targetAssignmentsExpectations[assignment.Target]
 		require.Truef(t, ok, "Could not find expectations for assignment with ID: %q, source %q and target %q", assignment.ID, assignment.Source, assignment.Target)
-
+		fmt.Printf("assignment with ID: %q, source %q and target %q", assignment.ID, assignment.Source, assignment.Target)
```
ivantenevvasilev, line 5519: Maybe change to `t.Logf`?

StanislavStefanov: It was a debugging message

ivantenevvasilev (dismissed):

ivantenevvasilev (approved):

PR #3176, opened 2023-07-24
Forbid creation of cert subject mappings with the same subject

la4ezar:
**Description**
Currently, it is possible to create cert subject mappings with the same subjects if their delimiter is different. For example, instead of comma(',') you use plus sign('+') for delimiter. This should be forbidden and this PR does that.

- Forbid creation of cert subject mappings with the same subject

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -138,8 +138,8 @@ global:
       version: "v20230714-506a64e8"
       name: compass-pairing-adapter
     director:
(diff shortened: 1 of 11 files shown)


PetarTodorovv:
closed in favor of https://github.com/kyma-incubator/compass/pull/3177

PR #3177, opened 2023-07-24
Forbid creation of cert subject mappings with the same subject

la4ezar:
**Description**
Currently, it is possible to create cert subject mappings with the same subjects if their delimiter is different. For example, instead of comma(',') you use plus sign('+') for delimiter. This should be forbidden and this PR does that.

- Forbid creation of cert subject mappings with the same subject

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


StanislavStefanov (dismissed):

StanislavStefanov (dismissed):

chart/compass/values.yaml
```diff
@@ -139,7 +139,7 @@ global:
       name: compass-pairing-adapter
     director:
       dir: dev/incubator/
-      version: "PR-3170"
+      version: "3177"
```
dzahariev, line 142: Please Add "PR-" here.

la4ezar: Good catch!

StanislavStefanov (approved):

PR #3187, opened 2023-08-01
Use service layer to handle CSM on AT Deletion

dzahariev:
**Description**

- Use service layer to handle CSM on AT Deletion
- Increase PostgreSQL container version used for validation

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/schema-migrator/migrations/director/20230801103000_add_trigger_for serv_sub_mappings_deletion.up.sql
```diff
@@ -0,0 +1,19 @@
...
+BEGIN
+    DELETE FROM cert_subject_mapping
+    WHERE internal_consumer_id = OLD.id::varchar(256);
+    RETURN OLD;
+END;
+$$
+LANGUAGE plpgsql;
+
+
+CREATE TRIGGER trigger_delete_cert_subject_mapping_on_app_template_delete
```
StanislavStefanov, line 15: Wouldnt it be better if the logic for deleting the certificate subject mappings is implemented in the service layer of the application template. If this approach is not suitable for the MAP operator another approach would be to list the certificate subject mappings for the consumer ID(new query could be introduced) and then delete the respective resources. Using APIs or implementing in the service layer will result in better traceability(we will have logs for the deleted resources) and will be easier to follow the logic of the flow. By using the trriger the business logic is kind of hidden and one should remember that there is a trigger for deleting the resource.

StanislavStefanov (changes requested):

components/schema-migrator/migrations/director/20230801103000_add_trigger_for serv_sub_mappings_deletion.up.sql
dzahariev: It was implemented in business operation inside MAP operator, but we decide to move it lower in layers. 
Relation between CSM and AT is hard, and we should ensure all such hard links to be strictly kept. In short, Certificate Subject Mapping without Application template is a leak that cannot be traced and cleared. Such leak can happen in case business logic is not fully executed due to connectivity glitch, crash or bug in code. Having a trigger that handles this ensures such leaks will not occur. 
Similar is situation with all resources that are connected (have foreign keys).

StanislavStefanov: If the deletion of the CSM is processed in the service layer of the Application template the deletion of both of the resources will be performed in one db transaction. If any of the resources fails to be deleted the transaction will be rolled back. If that is not enough we can ensure that tere is no CSM resource left by deleting it before deleting the Application template

dzahariev: I did not understand correctly the proposition. We will move the logic from Trigger to Service layer. No need of additional discussion.

dzahariev:
Do not merge for now. Changes are coming...

components/schema-migrator/validate.sh
```diff
@@ -19,7 +19,7 @@ DATA_DIR="${COMPONENT_PATH}/seeds"
 IMG_NAME="compass-schema-migrator"
 NETWORK="migration-test-network"
 POSTGRES_CONTAINER="test-postgres"
-POSTGRES_VERSION="12"
+POSTGRES_VERSION="15"
```
StanislavStefanov, line 22: After removing the migration the postgres version bumping should not be needed. Also the postgres version on real env is 9.6 so there may be potential problems if the version is bumped

components/director/internal/domain/certsubjectmapping/service.go
```diff
@@ -103,3 +104,14 @@ func (s *service) List(ctx context.Context, pageSize int, cursor string) (*model
 
 	return csmPage, nil
 }
+
+// ListByConsumerID retrieves all certificate subject mappings for a specific consumer id
+func (s *service) ListByConsumerID(ctx context.Context, consumerID string) ([]*model.CertSubjectMapping, error) {
```
StanislavStefanov, line 109: Can we just delete the records instead of fetching them and deleting them one by one? We could use `r.deleterGlobal.DeleteManyGlobal(ctx, repo.Conditions{repo.NewEqualCondition("internal_consumer_id", consumerID)})`
Then we could just invoke the method in the Application template resolver and this will save us some db requests

kaloyanspiridonov: Of course it will be better ....maybe its misunderstanding...i thought that you want first to list them and then delete them one by one ....

components/schema-migrator/validate.sh
dzahariev: The script is broken, probably from more than a month. Validate script seem to not be used in prow jobs and it fails without this change.
The reason is that provisioner image that is used to make DB export on deployment is using newest version of pg_dump utility, that produces the export that cannot be imported with older tools.

tests/director/tests/application_templates_api_test.go
```diff
@@ -832,6 +832,56 @@ func TestDeleteApplicationTemplate(t *testing.T) {
...
 }
 
+func TestDeleteApplicationTemplateWithCertSubjMapping(t *testing.T) {
+	// GIVEN
+	ctx := context.Background()
+	appTemplateName := createAppTemplateName("app-template")
+
+	tenantId := tenant.TestTenants.GetDefaultTenantID()
+
+	t.Log("Create application template")
```
la4ezar, line 842: Add the application template name in the log. For example, `t.Logf("Create application template with name: %q", appTemplateName)`

tests/director/tests/application_templates_api_test.go
```diff
@@ -832,6 +832,56 @@ func TestDeleteApplicationTemplate(t *testing.T) {
 	saveExample(t, deleteApplicationTemplateRequest.Query(), "delete application template")
 }
 
+func TestDeleteApplicationTemplateWithCertSubjMapping(t *testing.T) {
```
la4ezar, line 835: I think it will be better if the test creates 2 cert subject mappings and at the end checks that both of them are deleted. That way we will check the mass deletion.

la4ezar (changes requested):

la4ezar (dismissed):

la4ezar (approved):

PR #3188, opened 2023-08-01
Small improvement in the installation script

PetarTodorovv:
**Description**
Invoke `patchJWKS` after compass DB installation to minimise the unnecessary logs. Also, increased sleep interval in the same function.

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/installation/scripts/install-db.sh
+++ b/installation/scripts/install-db.sh
@@ -51,7 +51,7 @@ do
 done
 set -- "${POSITIONAL[@]}" # restore positional parameters
 
-echo "Install DB"
+echo "Installing DB"
 helm upgrade --install --atomic --debug --timeout "${TIMEOUT}" -f ./mergedOverrides.yaml --create-namespace --namespace compass-system localdb "${DB_CHARTS}"
 
 echo "Look for DB dump in progress ..."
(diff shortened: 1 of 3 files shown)


installation/scripts/install-db.sh
```diff
@@ -51,7 +51,7 @@ do
 done
 set -- "${POSITIONAL[@]}" # restore positional parameters
 
-echo "Install DB"
+echo "Installing DB"
 helm upgrade --install --atomic --debug --timeout "${TIMEOUT}" -f ./mergedOverrides.yaml --create-namespace --namespace compass-system localdb "${DB_CHARTS}"
```
ognyvrac, line 55: As you are already removing the spam from the logs, maybe remove the `--debug` flag to exclude verbose output

PetarTodorovv: not sure what was the initial motivation to use this flag but when I compared the logs most of the debug logs are about the helm charts/values.yaml. there are couple with resource creation info. Nevertheless, I remove the debug flag.

ognyvrac (approved):

la4ezar (approved):

PR #3234, opened 2023-08-17
Add missing transaction commit in aggregator

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -139,7 +139,7 @@ global:
       name: compass-pairing-adapter
     director:
       dir: dev/incubator/
-      version: "PR-3227"
+      version: "PR-3234"
       name: compass-director
     hydrator:
       dir: prod/incubator/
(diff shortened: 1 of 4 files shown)


dzahariev:
I look at the code and notice that there is a place where the transaction is closed twice, and there are some entities that are recreated (deleted and created) outside of the transaction. 
The operation that needs revision is resyncTenantMappingWebhooksInTx.
Can you please check it also.

alextargov:
@dzahariev I see a transaction opening at [L1028](https://github.com/kyma-incubator/compass/blob/6b1b354359526312c4c900132bd1f45fb53d0e29/components/director/internal/open_resource_discovery/service.go#L1028) and committing 40 lines below at [L1067](https://github.com/kyma-incubator/compass/blob/6b1b354359526312c4c900132bd1f45fb53d0e29/components/director/internal/open_resource_discovery/service.go#L1067).

There is a special case at [L1054](https://github.com/kyma-incubator/compass/blob/6b1b354359526312c4c900132bd1f45fb53d0e29/components/director/internal/open_resource_discovery/service.go#L1054) where we return the result of the transaction so that the create and delete functions a few lines below would not be executed.
So as a conclusion it does not seem to me that there is a double transaction committing. Could you confirm?

dzahariev (approved):

PR #3239, opened 2023-08-21
FA not present when updating status

StanislavStefanov:
**Description**

If we have async TM notification receiver, when we notify them, they may be faster to report status than the transaction that created the respective assignment.

This will result in a status report before the assignments are persisted and commited in the DB. Then the status API will not see such assignment and will fail with Not Found.

- persist FAs before processing the notifications
- delete the persisted resources in case the forrmation assignment has failed after ther persisting

**Pull Request status**

- [X] Implementation
- [X] Unit tests


components/director/internal/domain/formation/service.go
```diff
@@ -514,34 +512,91 @@ func (s *service) AssignFormation(ctx context.Context, tnt, objectID string, obj
...
+		if terr != nil {
+			return nil, terr
+		}
+
+		// We need to persist the FAs before we proceed to notification processing as for scenarios where there are both
+		// participants with ASYNC notifications and SYNC notifications it is possible that a FA status update request is
+		// received before the FA is persisted in the database.
+		// Example P1 has ASYNC webhook - W1 , P2 has SYNC webhook - W2. Assign both P1 and P2 to a formation.
+		// Formation assignments are generated.
+		// Execute W1, then execute W2. W2 takes for example 10 seconds. Before the W2 processing finishes a FA status update request for W1 is received.
```
PetarTodorovv, line 535:
```
// Example: Participant 1(P1) has ASYNC webhook(W1) and Participant2(P2) has SYNC webhook(W2). Assign both P1 and P2 to a formation.
// Formation assignments are generated.
// Execute W1, then execute W2. W2 takes, for example, 10 seconds. Before the W2 processing finishes, an FA status update request for W1 is received.
// When fetching the corresponding FA from the DB we get an object not found error, as the status update is performed in new transaction and the transaction in which the FA were generated is still running.
```
?

components/director/internal/domain/formation/service.go
```diff
@@ -514,34 +512,91 @@ func (s *service) AssignFormation(ctx context.Context, tnt, objectID string, obj
...
+			return nil, terr
 		}
 
+		if terr = tx.Commit(); terr != nil {
+			return nil, terr
+		}
+
+		// If the assigning of the object fails the transaction opened in the resolver fill be rolled back.
+		// The FA records and the labels for the object will not be reverted as they we re persisted as part of another
+		// transaction. The leftover resources should be deleted separately.
```
PetarTodorovv, line 555:
```
// If the assigning of the object fails, the transaction opened in the resolver will be rolled back.
// The FA records and the labels for the object will not be reverted as they were persisted as part of another
// transaction. The leftover resources should be deleted separately.
```

components/director/internal/domain/formation/service.go
PetarTodorovv, line 1: One general question - do we need an E2E test for this scenario? E.g. "async" handler in the ext-svc-mock that doesn't have a delay but rather immediately executes "FA status update request" to the status API, opposite to what we have now in the async handler with the fixed time delay.

PetarTodorovv (changes requested):

components/director/internal/domain/formation/service.go
```diff
@@ -514,34 +512,91 @@ func (s *service) AssignFormation(ctx context.Context, tnt, objectID string, obj
...
+		// is created. As the defer statement is declared in the scope of the case fragment of the switch it will be bound to the err variable in the same scope
+		// which is the new one. Then the deffer will not execute its logic in case of error in the outer scope.
+		assignmentInputs, terr := s.formationAssignmentService.GenerateAssignments(ctx, tnt, objectID, objectType, formationFromDB)
+		if terr != nil {
+			return nil, terr
+		}
+
+		// We need to persist the FAs before we proceed to notification processing as for scenarios where there are both
+		// participants with ASYNC notifications and SYNC notifications it is possible that a FA status update request is
+		// received before the FA is persisted in the database.
```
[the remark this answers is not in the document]
StanislavStefanov: It is unlikely as the async implementations respond with accepted and then perform their logic. The issue is the big delay when processing the syncronous notification that provides the ASYNC participant to process its notification and to respond

components/director/internal/domain/formation/service.go
StanislavStefanov: I dont think that we need as this is kind of a corner case

PetarTodorovv (dismissed):

PetarTodorovv (approved):

PR #3241, opened 2023-08-21
Initial skeleton of instace-creator component and SM instances client

nyordanoff:
**Description**
This PR introduces a 'skeleton' for a new component that will be responsible for instance creation. A client for communicating with remote service for managing the instances is implemented as well.

- charts for the new component
- initial code-level structure of the new component including tenant and token validations
- preparation of regional creds that will be used by the client to call the remote service
- client implementation and unit tests

**Related issue(s)**

- https://github.com/kyma-project/test-infra/pull/8732

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


chart/compass/charts/external-services-mock/templates/regional-sm-instances-credentials.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if eq .Values.global.externalServicesMock.regionSMInstancesCredentials.manage true }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Values.global.instanceCreator.secrets.name }}
+  namespace: compass-system
```
ognyvrac, line 6: `{{ .Release.Namespace }}` ?

chart/compass/charts/instance-creator/templates/service.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  type: ClusterIP
+  ports:
+    - port: {{ .Values.global.instanceCreator.port }}
+      protocol: TCP
+      name: http
+    - port: {{ .Values.global.istio.proxy.port }}
+      protocol: TCP
+      name: proxy-status
+  selector:
```
ognyvrac, line 18: You can consider extracting these labels in a `_helpers.tpl` once again. Afterwards you can use the new helpers definition here, in the `.spec.matchLabels` and `.spec.template.metadata.labels` of the Deployment. This way you avoid issues if you change the labels, but might be an overkill if they are not going to change as often.

chart/compass/charts/instance-creator/templates/deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+              value: "{{ .Values.global.externalCertConfiguration.secrets.externalClientCertSecret.name }}"
+            - name: APP_SM_INSTANCE_OAUTH_TOKEN_PATH
+              value: {{ .Values.global.instanceCreator.oauthTokenPath }}
+            - name: APP_SM_ASYNC_API_TICKER
+              value: {{ .Values.global.instanceCreator.ticker }}
+            - name: APP_SM_ASYNC_API_TIMEOUT
+              value: {{ .Values.global.instanceCreator.timeout }}
+          livenessProbe:
+            httpGet:
+              port: {{.Values.deployment.args.containerPort }}
```
ognyvrac, line 101: Could we use { .Values.global.instanceCreator.port }}? Same for the readinessProbe

Saw it above the code again, maybe the `{{.Values.deployment.args.containerPort }}` is not needed

chart/compass/charts/instance-creator/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+  image:
+    pullPolicy: IfNotPresent
+  resources: {}
+  resourcesIstioProxy:
+    requests: {}
+    limits: {}
+  args:
+    containerPort: 8080
+  securityContext: # Set on container level
+    runAsUser: 2000
```
ognyvrac, line 16: Do we need specifically user 2000? Or this is more or less not to run as root?

chart/compass/charts/instance-creator/templates/deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+  selector:
+    matchLabels:
+      app: {{ .Chart.Name }}
+      release: {{ .Release.Name }}
+  strategy:
+    {{- toYaml .Values.deployment.strategy | nindent 4 }}
+  template:
+    metadata:
+      annotations:
+        restartOnNewInstallationMarker: {{ randAlphaNum 5 | quote }} # Restarts the deployment on a new Helm installation. (https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments)
```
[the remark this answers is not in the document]
nyordanoff: I saw that it is used in other deployments and applied it here but haven't dug into it.  By memory, it was related to able to restart components when new installation is triggered so that the secrets can be refreshed. I don't remember the semantics of the annotation, though.

However, I also can't find info about it.

We can investigate it further in the team to double check if this functionality still works and if needed search/validate a new approach and apply it across all the resources in the project. 

What do you think?

chart/compass/charts/instance-creator/templates/deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+        cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
```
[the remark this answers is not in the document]
nyordanoff: I put it for consistency purposes as it is present in most of the other deployments.

chart/compass/charts/instance-creator/templates/service.yaml
```diff
@@ -0,0 +1,20 @@
...
+  labels:
+    app: {{ .Chart.Name }}
+    release: {{ .Release.Name }}
+spec:
+  type: ClusterIP
+  ports:
+    - port: {{ .Values.global.instanceCreator.port }}
+      protocol: TCP
+      name: http
+    - port: {{ .Values.global.istio.proxy.port }}
```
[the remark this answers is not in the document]
nyordanoff: I added it for consistency purposes.

chart/compass/charts/instance-creator/templates/service.yaml
nyordanoff: I lean towards the latter because I don't think there will be many (if any) changes on a regular basis.

chart/compass/charts/instance-creator/templates/deployment.yaml
nyordanoff: We could but from my PoV it is a bit safer to use the 'local' chart `values.yaml` for these kind of deployment related values because it's scoped only for the deployment. And for the `global` value, as it can be used in more places (since it's global), if it changes in the future for whatever reason the deployment will be affected.

chart/compass/charts/instance-creator/values.yaml
nyordanoff: I followed the pattern of the other charts. TBH, I don't remember why we came up with that user initially.

ognyvrac (approved):

ZdravkoGyurov (approved):

components/instance-creator/internal/client/client.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.Errorf("failed to get object(s), status: %d, body: %s", resp.StatusCode, body)
+	}
+
+	return body, nil
+}
+
+func (c *client) executeAsyncRequest(ctx context.Context, resp *http.Response, caller ExternalSvcCaller, subaccountID string, isCreateRequest bool) (string, error) {
```
StanislavStefanov, line 341: Why did we choose the polling approach, was it not possible to be notified via callback? If we wait for the async resource processing we risk reaching the timeout of the client that is calling us or the server timeout. Can we push the async operations in some queue and return something that notifies that. Then when all of the operations are processed we could aggregate the resources and then notify the caller asynchronously with the information

StanislavStefanov (changes requested):

components/instance-creator/internal/client/types/service_binding.go
```diff
@@ -0,0 +1,87 @@
...
+// Match matches a ServiceKey based on some criteria
+func (sk ServiceKeys) Match(args resources.ResourceArguments) (string, error) {
+	return "", nil // implement me when needed
+}
+
+// MatchMultiple matches several ServiceKeys based on some criteria
+func (sk ServiceKeys) MatchMultiple(args resources.ResourceArguments) []string {
+	serviceKeyIDs := make([]string, 0, sk.NumItems)
+	for _, item := range sk.Items {
+		if item.ServiceInstanceID == args.(ServiceKeyArguments).ServiceInstanceID {
```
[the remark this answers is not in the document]
nyordanoff: Nice catch, thanks

components/instance-creator/internal/client/client.go
```diff
@@ -0,0 +1,446 @@
...
+		return errors.Errorf("error while getting caller for region: %s", region)
+	}
+
+	resourceIDs, err := c.retrieveMultipleResources(ctx, caller, subaccountID, resources, resourceArgs)
+	if err != nil {
+		return err
+	}
+
+	for _, resourceID := range resourceIDs {
+		resourcePath := resourceArgs.GetURLPath() + fmt.Sprintf("/%s", resourceID)
```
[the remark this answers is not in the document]
nyordanoff: I don't think it will work because we have to call `DeleteResource()` with specific implementation of the `ResourceMatchParameters` and `Resource` interfaces but the `DeleteMultipleResources()` is generic so we don't have the knowledge what resource is being deleted.

ZdravkoGyurov (dismissed):

StanislavStefanov (dismissed):

StanislavStefanov (approved):

ZdravkoGyurov (approved):

chart/compass/charts/instance-creator/values.yaml
```diff
@@ -0,0 +1,23 @@
+deployment:
+  minReplicas: 1
```
DimitarPetrov, line 2: As per product standards we should have at least 2 replicas in order to ensure high availability

components/instance-creator/internal/client/caller.go
```diff
@@ -0,0 +1,43 @@
...
+// CallerProvider is used to provide ExternalSvcCaller to call external services with given authentication
+type CallerProvider struct{}
+
+// NewCallerProvider creates new CallerProvider
+func NewCallerProvider() *CallerProvider {
+	return &CallerProvider{}
+}
+
+// GetCaller provides ExternalSvcCaller to call external services with given authentication
+func (c *CallerProvider) GetCaller(config config.Config, region string) (ExternalSvcCaller, error) {
```
DimitarPetrov, line 19: We should always return concrete structs and avoid returning interfaces to avoid need of refactoring in case more methods are added.

https://github.com/golang/go/wiki/CodeReviewComments#interfaces

PR #3243, opened 2023-08-21
Fix concurrent write iteration

alextargov:
**Description**

- When fetching the Specs concurrently an error occurs about concurrent unsafe loop/write to the http Headers. This PR introduces sync maps, which behind the scene handle concurrent reading/writing.
- Validated on dev-main and also I've extracted the problematic code in an isolated environment. Ran the "go run" command with -race flag before the fix - it stated that there is a concurrency problem. After the approach was implemented, I ran the command with flag again multiple times and no concurrency error was reported by the command.

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


NickyMateev (commented):
What did the validation procedure of this fix involve?

components/director/pkg/accessstrategy/open_access_strategy.go
```diff
@@ -13,14 +16,17 @@ func NewOpenAccessStrategyExecutor() *openAccessStrategyExecutor {
...
 	if err != nil {
 		return nil, err
 	}
 
 	if additionalHeaders != nil {
-		req.Header = additionalHeaders
+		additionalHeaders.Range(func(key, value any) bool {
+			req.Header.Set(str.CastOrEmpty(key), str.CastOrEmpty(value))
+			return true
+		})
```
dragobt, line 29: This part is not covered with unit test. Do you think it is relevant to do it?

dragobt:
Add description to the PR as well.

alextargov:
@NickyMateev  it included multiple aggregator runs for three systems on dev-main. All three had their specs fetched (this is the place that was concurrent and caused the issue)

NickyMateev:
> 



> @NickyMateev it included multiple aggregator runs for three systems on dev-main. All three had their specs fetched (this is the place that was concurrent and caused the issue)

Okay, could you also elaborate that when linking the PR fix in the original bug. Thanks!

StanislavStefanov (approved):

PR #3250, opened 2023-08-22
[HOTFIX] Fix concurrent write iteration 2

alextargov:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


installation/scripts/prow/jobs/provision-compass-smoke-test.sh
```diff
@@ -55,6 +55,11 @@ else
     ORD_PR_COMMIT_HASH=$(jq -r '.merge_commit_sha' <<< "${ORD_PR_DATA}")
 fi
 
+# When the immage that is used is from prod registry the value in ORD_PR_NUMBER is actually the short commit hash due to the different naming convention for the dev and prod images
+if [[ -z "$ORD_PR_COMMIT_HASH" || "$ORD_PR_COMMIT_HASH" == "null" ]]; then
+    ORD_PR_COMMIT_HASH="$ORD_PR_NUMBER"
+fi
+
```
PetarTodorovv, line 62: Currently, we're logging an error for `ORD_PR_DATA` in case the image version is not PR but rather a commit hash - 
```
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/pulls/pulls#get-a-pull-request"
}
```
And the `ORD_PR_STATE` will also be "invalid" because the `ORD_PR_DATA` is not what we expect. Could we adapt this part of the code and maybe extract that logic in an `if` statement only when the image version is PR(it will be a PR when the image dir is "dev", we can get it from `global.images.ord_service.dir`)

Also, the names of the variables(logs) are misleading. Could we adapt them as well, e.g. `ORD_PR_NUMBER` to be `ORD_IMAGE_VERSION` or something similar? `ORD_PR_COMMIT_HASH` can be `ORD_COMMIT_HASH` or something similar. And `ORD_COMMIT_HASH` can accept the value from the `ORD_IMAGE_VERSION ` before the `if` statement with the PR logic, and if the image version is PR we'll override the commit hash var with the PR commit in the `if` otherwise it will remain the "prod commit hash" from the post-build job

PR #3254, opened 2023-08-23
Remove unneeded Hydrator retries

StanislavStefanov:
**Description**
Hydrator retries requests regardless of the type of the error that is received. When fetching tenant if the tenant is not found the Hydrator retries the request. The request itself can be slow and retrying it causes timeouts. If tenant is not found with the first request it does not make sense to make a second retry. Retries are useful only for connectivity problems.

- do not retry if the error is not caused by a connectivity issue
- when the certificate from the request is whitelisted do not match the certificate service context provider but only access level context provider
- in Kyma adapter and IAS adapter auth middleware get the tenant from `Client-Id-From-Certificate` header instead of `Tenant` header

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


components/hydrator/internal/director/client.go
```diff
@@ -2,13 +2,16 @@ package director
...
+	"strings"
 	"time"
 
 	authConv "github.com/kyma-incubator/compass/components/director/pkg/auth"
 	"github.com/kyma-incubator/compass/components/director/pkg/model"
 	"github.com/pkg/errors"
 
-	"github.com/kyma-incubator/compass/components/connectivity-adapter/pkg/retry"
+	"github.com/avast/retry-go/v4"
+	//"github.com/kyma-incubator/compass/components/connectivity-adapter/pkg/retry"
```
nyordanoff, line 14: leftover?

tests/external-services-mock/tests/main_test.go
```diff
@@ -16,8 +16,9 @@ import (
 )
 
 type config struct {
-	Auditlog                                  pkgConfig.AuditlogConfig
-	DefaultTestTenant                         string
+	Auditlog          pkgConfig.AuditlogConfig
+	DefaultTestTenant string
+	//DefaultTestSubaccountTenant               string
```
nyordanoff, line 21: leftover?

kaloyanspiridonov (approved):

kaloyanspiridonov (approved):

nyordanoff (approved):

ZdravkoGyurov (approved):

tests/pkg/tenant/tenant_setup.go
```diff
@@ -8,8 +8,9 @@ type TenantStatus string
 type TenantType string
 
 const (
-	testProvider      = "Compass Tests"
-	testDefaultTenant = "Test Default"
+	testProvider                = "Compass Tests"
+	testDefaultTenant           = "Test Default"
+	testDefaultSubaccountTenant = "Test Default Subaccount"
```
PetarTodorovv, line 13: Most probably the new tenant will be inserted by the K8s "tenant-loader" job but I think it doesn't execute every time? As far as I remember only when the director image/component is changed because the job uses the director image. Which in that case will do the job for us. I'm mentioning it to monitor the job and see if the new tenant will be inserted into our environments(dev-main, dev)

PR #3259, opened 2023-08-24
Allow BTP Applications onboarded as Application Templates to consume ORD Service APIs

la4ezar:
**Description**
Currently, we allow BTP SaaS applications onboarded as runtimes to call ORD Service on behalf of their subscribers. With this PR we allow BTP Applications onboarded as Application Templates to consume ORD Service APIs.

- destination fetcher adaptations
- unit tests
- e2e test

**Related issue(s)**

- https://github.com/kyma-incubator/ord-service/pull/97

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/destinationfetchersvc/handler.go
```diff
@@ -47,6 +50,18 @@ func (h *handler) SyncTenantDestinations(writer http.ResponseWriter, request *ht
...
 
+	isTenantSubscribed, err := h.validateTenantSubscription(ctx, tenantID)
+	if err != nil {
+		log.C(ctx).WithError(err).Errorf("Failed to validate tenant %q subscription", tenantID)
+		http.Error(writer, err.Error(), http.StatusBadRequest)
+		return
+	}
+	if !isTenantSubscribed {
+		log.C(ctx).Infof("Tenant %q is not subscribed", tenantID)
+		http.Error(writer, fmt.Sprintf("Tenant %q is not subscribed", tenantID), http.StatusBadRequest)
```
alextargov, line 61: Is "bad request" the most accurate status? Usually, it means a client error, something wrong with the client http request. I'm not saying "don't use error 400 here" but rather questioning if it is suitable.  
P.S. Same applies for the above L56 and for the other handler function

la4ezar: It is not the most accurate status but the whole handler is implemented like that - if there is a problem, return 400

components/director/internal/destinationfetchersvc/handler_test.go
```diff
@@ -168,7 +192,31 @@ func TestHandler_FetchDestinationsSensitiveData(t *testing.T) {
...
+		{
+			Name:      "Error when getting subscribed tenants fails",
+			Request:   reqWithUserContext,
+			DestQuery: namesQueryRaw,
+			DestinationFetcherSvc: func() *automock.DestinationManager {
+				svc := &automock.DestinationManager{}
+				svc.On("GetSubscribedTenantIDs", mock.Anything).Return([]string{}, testErr).Once()
+				return svc
+			},
+			ExpectedStatusCode: http.StatusBadRequest,
```
petar-iv, line 210: Here 500 would be more suitable.

la4ezar: This is a test and that code is expected, because it is returned at handler.go:56. Also, I'm not sure that status code 500 there is appropriate because it's a client error, not a server one.

components/director/internal/destinationfetchersvc/handler.go
alextargov: I'm cool with 400s if they are being widely used in the service.

alextargov (dismissed):

alextargov (approved):

nyordanoff (approved):

petar-iv (approved):

PR #3279, opened 2023-08-30
Propagate scenario group description from BTP System Landscape to TMS

dragobt:
**Description**
Currently when selecting scenario groups for S/4 tokens in BTP Cockpit we send either of the following keys to Extension Service's Token Management Service (TMS):
- UNRESTRICTED
- EVENT_MESH

To eliminate the need for Extension Service to keep a mapping for those keys to the same user-facing names like the ones in the BTP Cockpit we've agreed to propagate the user-facing names to TMS as well. From there Extension Service can reuse the name in S/4.

- Send scenario groups as objects instead of strings
- Update  dependencies

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


dzahariev (approved):

dzahariev (approved):

dzahariev (changes requested):
We need to make it compatible to achieve easier delivery. 
Inside components/director/internal/domain/onetimetoken/service.go on line 433 can catch the error when parsing and to fallback to old approach.

dzahariev (dismissed):

la4ezar (changes requested):

la4ezar (approved):

emmakarova (approved):

PR #3294, opened 2023-09-11
Introduce Constraint priority

StanislavStefanov:
**Description**

In some cases we need to chain more than one opt-in feature modelled as formation constrain on one and the same join point. In some of those cases order is important, therefore we need a way to register those constraints in a way so that we can influence stable execution order.

- Introduce formation constraint priority, createdAt and description fields
- Sort Constraints by priority and createdAt(used as secondary order paraeter) before executing the constraints

**Related issue(s)**

- ...

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


tests/director/tests/formation_constraint_api_test.go
```diff
@@ -16,16 +14,20 @@ import (
 func TestCreateFormationConstraint(t *testing.T) {
 	// GIVEN
 	ctx := context.Background()
+	priority := 7
```
nyordanoff, line 17: Do you think it's a good idea to add the `priority` and `description` fields in the update test (`TestUpdateFormationConstraint`) as well?

components/director/internal/domain/formationconstraint/resolver_test.go
```diff
@@ -607,27 +607,71 @@ func TestResolver_UpdateFormationConstraint(t *testing.T) {
...
+				converter.On("FromInputGraphQL", &formationConstraintInputUpdatedTemplateAndDescription).Return(formationConstraintModelInputUpdatedTemplateAndDescription)
+				converter.On("ToGraphQL", formationConstraintModelUpdatedTemplateAndDescription).Return(gqlFormationConstraintUpdatedTemplateAndDescription)
+
+				return converter
+			},
+			ExpectedOutput: gqlFormationConstraintUpdatedTemplateAndDescription,
+			ExpectedError:  nil,
+		},
+		{
+			Name:  "Success when all updatable fields are provided",
```
nyordanoff, line 654: Probably rename to `Success when only input template and priority are updated`?

PR #3300, opened 2023-09-13
Send certificate trust details in TM notifications

la4ezar:
**Description**
As a TM notification receiver, I want to receive the x509 certificate details of the other party so that I can trust it and automatically enable inbound mTLS integration.

- Add trust details info to ApplicationTemplate / Runtime objects in both Configuration Changed and Application Tenant Mapping template inputs.
- E2e test
- Unit tests

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/domain/webhook/datainputbuilder/cert_subject_builder.go
```diff
@@ -0,0 +1,71 @@
...
+// GetTrustDetailsForObject builds trust details for objects with ID objectID
+func (b *WebhookCertSubjectBuilder) GetTrustDetailsForObject(ctx context.Context, objectID string) (*webhook.TrustDetails, error) {
+	trustDetails, err := b.GetTrustDetailsForObjects(ctx, []string{objectID})
+	if err != nil {
+		return nil, err
+	}
+
+	result, ok := trustDetails[objectID]
+	if !ok {
+		return nil, errors.Errorf("There are no trust details for object with ID %q", objectID)
```
ivantenevvasilev, line 57: Are there cases where there won't be any trust details and if we don't have any, should we throw an error?

ivantenevvasilev (approved):

components/director/internal/domain/webhook/datainputbuilder/cert_subject_builder.go
la4ezar: Currently, I think it is impossible to enter that `if` case because even if there are no trust details for the object the `GetTrustDetailsForObjects` func will return a map where there is an entry with key `objectID` and empty value(`webhook.TrustDetails`). So it's practically impossible to enter that case but I wrote it because it's a bad pattern not to check if there is an entry with a given key in a map.

ivantenevvasilev (approved):

kaloyanspiridonov (approved):

PR #3301, opened 2023-09-13
Kyma 2.4.3

ognyvrac:
**Description**
Update Kyma to 2.4.3

- Update Kyma CLI in benchmark and provision scripts
- Move patch logic to Helm overrides
- Change istio images to external ones as Istio has been updated to 1.14.1; With Istio 1.15 ARM images should be supported out of the box, that is why we have not downloaded the external ones and maintained them locally
- Remove Kyma workplace changes that were needed to exclude outdated resources, kept reusability for local scripts

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


installation/scripts/prow/jobs/provision-compass.sh
```diff
@@ -48,8 +48,10 @@ cp yq "$HOME/bin/yq" && cp yq "/usr/local/bin/yq"
 log::info "Successfully installed yq version: $YQ_VERSION"
 
 # Install Kyma to be later used in run.sh
+KYMA_CLI_VERSION="2.5.0"
 log::info "Installing Kyma CLI version: $KYMA_CLI_VERSION"
-KYMA_CLI_VERSION=$(cat ${COMPASS_SOURCE_DIR}/installation/resources/KYMA_VERSION)
+# TODO: Kyma 2.4.3 release exists, but Kyma CLI 2.4.3 does not
+# KYMA_CLI_VERSION=$(cat ${COMPASS_SOURCE_DIR}/installation/resources/KYMA_VERSION)
```
PetarTodorovv, line 54: Hmm, does that mean we'll use kyma CLI 2.5.0 locally as well? Will this cause any issues? And what are the pros/cons using 2.4.2 CLI?

installation/resources/kyma/kyma-overrides-minimal.yaml
```diff
@@ -144,10 +155,40 @@ monitoring:
...
+          namespace: kyma-system
+          pathPrefix: /
+          port: http-web
+          scheme: https
+          # By default in Kyma, Prometheus injects the Istio certificates from a sidecar
+          tlsConfig:
+            caFile: /etc/prometheus/secrets/istio.default/root-cert.pem
+            certFile: /etc/prometheus/secrets/istio.default/cert-chain.pem
+            keyFile: /etc/prometheus/secrets/istio.default/key.pem
+            insecureSkipVerify: true
```
PetarTodorovv, line 172: You've adapted a part of the mTLS patches to be through helm and the other part to be through bash.
image

Is it possible for all of the patches to be in the helm charts? My concern is that the current patches are all in one place and with the new setup there are in a couple of places(with a couple of different overrides)? it's debatable but it could make the supportability harder?

P.S. Do prom dashboards work fine and show data with the new override setup?

installation/resources/KYMA_VERSION
PetarTodorovv, line 1: The OSS documentation should be adapted as well

PetarTodorovv (changes requested):

installation/scripts/prow/jobs/provision-compass.sh
ognyvrac: This specific script is used only during Prow jobs, the integration jobs for example. Locally (`run.sh`), your own version of the Kyma CLI is used. The Kyma CLI version is different from the actual Kyma release (`--source` during `kyma deploy`).

Personally, I use Kyma CLI version 2.15.1 while installing Kyma 2.3.0 or 2.4.3 and it does not show any issues; I doubt it would be a problem. I have not tested the Kyma CLI 2.4.2 as I thought it was a safer bet to go with 2.5.0.

installation/resources/kyma/kyma-overrides-minimal.yaml
ognyvrac: Unfortunately, not all ServiceMonitors can be moved to the Helm charts and all dashboards look fine.

I went with this approach as I expect the Helm chart syntax to rarely change. When these components are removed from the patch functions you would no longer look at them when doing upgrades. Another possiblity would be to revert them back but place a comment that some of them can be overriden in the Helm chart, but are not.

PetarTodorovv (approved):

Peteva (approved):
+1

PR #3321, opened 2023-09-21
Split director tests in multiple test definitions

ivantenevvasilev:
**Description**
Currently, when one test in the director tests fails, the whole test definition is retested. The director tests are the longest running tests and doubling that makes PR iteration slow. If it is split up into multiple smaller `TestDefiniton` resources, it should be faster in the case of retesting. 
A typical execution of the director tests is currently ~36 minutes, if one test fails it would double. The tests split in the current way take

1. application tests ~5 minutes
2. runtime tests ~4 minutes
3. director tests ~4 minutes
4. notification tests ~15 minutes
5. formation tests ~2 minutes

Worst case scenario, the notification tests would be retested, which would add at most 15 more minutes, as opposed to the 36 minutes.

- move `examplesaver` to its own package, so that it can be used in all director tests
- move `createSubscription` utils to tests pkg package in order to be reusable without import cycles in multiple tests
- split up `formation_api_test.go` into 4 more test files - `application_only_notifications_test.go`, `application_subscription_notifications_test.go`, `runtime_notification_subscription_test.go` and `tenant_hierarchy_notifications_test.go`
- introduce application, runtime, notifications and formation test definitions that are moved away from the main director test definition

**Pull Request status**

- [x] [N/A] Implementation
- [x] [N/A] Unit tests
- [x] Integration tests


dzahariev (approved):

PetarTodorovv (approved):

chart/compass/templates/tests/director/director-test.yaml
```diff
@@ -319,6 +319,1286 @@ spec:
       {{ end }}
       restartPolicy: Never
 ---
+apiVersion: "testing.kyma-project.io/v1alpha1"
+kind: TestDefinition
```
DimitarPetrov, line 323: Splitting that into multiple files would be easier to use.

Not to slow down the merge of that PR we can handle that in the next improvement that will also re-iterate over the environment variables and slim it to the only needed for the particular sub-set of tests.

PR #3337, opened 2023-09-28
Adapt sandbox tests after splitting director test definition

ivantenevvasilev:
**Description**
After splitting the director test definition, the sandbox commands were not adapted. This PR adds specific Makefiles for each new test definition.

- add Makefile in each new director test definition

**Related issue(s)**

- #3321

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/tests/director/Makefile
+++ b/tests/director/Makefile
@@ -52,7 +52,7 @@ run-formation:
 	@kubectl exec -n kyma-system oct-tp-compass-e2e-tests-compass-e2e-director-local-0 -c director-tests -- ./director-formation.test -test.run $(testName) -test.v
 
 run-runtime:
-	@kubectl exec -n kyma-system oct-tp-compass-e2e-tests-compass-e2e-director-local-0 -c director-tests -- ./director-formation.test -test.run $(testName) -test.v
+	@kubectl exec -n kyma-system oct-tp-compass-e2e-tests-compass-e2e-director-local-0 -c director-tests -- ./director-runtime.test -test.run $(testName) -test.v
 
 run-notification:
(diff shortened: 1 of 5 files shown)


tests/director/Makefile
PetarTodorovv, line 1: There is a .md file in the tests directory with a brief explanation of the sandbox tests. Since the setup and commands are changed, could we also adapt the doc as well?

ivantenevvasilev: Nice catch, there needs to be an adaptation to the README.md, I'll open a follow-up PR with the docs, as this could potentially be blocking development.

tests/director/Makefile
```diff
@@ -44,6 +44,19 @@ sandbox-test:
...
 
+
+run-application:
+	@kubectl exec -n kyma-system oct-tp-compass-e2e-tests-compass-e2e-director-local-0 -c director-tests -- ./director-application.test -test.run $(testName) -test.v
+
+run-formation:
+	@kubectl exec -n kyma-system oct-tp-compass-e2e-tests-compass-e2e-director-local-0 -c director-tests -- ./director-formation.test -test.run $(testName) -test.v
+
+run-runtime:
+	@kubectl exec -n kyma-system oct-tp-compass-e2e-tests-compass-e2e-director-local-0 -c director-tests -- ./director-formation.test -test.run $(testName) -test.v
```
la4ezar, line 55: Copy+paste error. Instead of `-- ./director-formation.test` it should be `-- ./director-runtime.test`.

la4ezar (changes requested):

la4ezar (approved):

PR #3338, opened 2023-09-29
Adapt ORD from v1.5.0 to v1.5.3

dragobt:
**Description**
Adopt the ORD version: v1.5.0, v1.5.1, v1.5.2, v1.5.3

- Introduce description field for Product and Tombstone resources
- Remove hierarchy field from Api and Event resources
- Adjust validation rules for ord spec versions from v1.5.0 to v1.5.3
- Add and adapt unit tests
- Introduce db migrations for the added/removed properties

**Related issue(s)**

- https://github.com/kyma-incubator/ord-service/pull/100

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/open_resource_discovery/validation.go
```diff
@@ -701,7 +824,14 @@ func validateORDLinks(value interface{}) error {
 			validation.NilOrNotEmpty,
 			validation.Length(MinDescriptionLength, MaxDescriptionLength),
 		},
-	})
+	}
+	if err := validateJSONArrayOfObjects(value, elementFieldRules); err != nil {
+		return err
+	}
+	if err := validateLinks(value); err != nil {
```
emmakarova, line 831: As far as I see the `validateLinks` function has the same logic as `validateJSONArrayOfObjects` with the difference in the check for link duplicates. Can this function `areThereLinkTitleDuplicates` be used instead `validateLinks`? Or is there a concern to be done like this?

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -565,7 +647,8 @@ func validateVendorInput(vendor *model.VendorInput) error {
 func validateTombstoneInput(tombstone *model.TombstoneInput) error {
 	return validation.ValidateStruct(tombstone,
 		validation.Field(&tombstone.OrdID, validation.Required, validation.Match(regexp.MustCompile(TombstoneOrdIDRegex))),
-		validation.Field(&tombstone.RemovalDate, validation.Required, validation.By(isValidDate)))
+		validation.Field(&tombstone.RemovalDate, validation.Required, validation.By(isValidDate)),
+		validation.Field(&tombstone.Description, validation.NilOrNotEmpty, validation.Length(MinDescriptionLength, MaxDescriptionLength)))
```
emmakarova, line 651: Unit test for validation of `description` field.

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -529,11 +609,13 @@ func validateProductInput(product *model.ProductInput) error {
 
 	return validation.ValidateStruct(product,
 		validation.Field(&product.OrdID, validation.Required, validation.Match(regexp.MustCompile(ProductOrdIDRegex))),
-		validation.Field(&product.Title, validation.Length(1, 255), validation.Required),
+		validation.Field(&product.Title, validation.Length(MinTitleLength, MaxTitleLength), validation.Required),
 		validation.Field(&product.ShortDescription, shortDescriptionRules...),
+		validation.Field(&product.Description, validation.NilOrNotEmpty, validation.Length(MinDescriptionLength, MaxDescriptionLength)),
```
emmakarova, line 614: Unit test for validation of `description` field.

components/director/internal/open_resource_discovery/validation.go
```diff
@@ -452,9 +531,13 @@ func validateEventInput(event *model.EventDefinitionInput, docPolicyLevel *strin
...
-		validation.Field(&event.Description, validation.Required, validation.Length(MinDescriptionLength, MaxDescriptionLength)),
+		validation.Field(&event.Name, validation.Required, validation.NewStringRule(noNewLines, "short description should not contain line breaks"),
+			validation.When(checkResourcePolicyLevel(docPolicyLevel, event.PolicyLevel, PolicyLevelSap), validation.Length(MinTitleLength, MaxTitleLengthSAPCorePolicy), validation.By(validateTitleDoesNotContainsTerms)),
+			validation.Length(MinTitleLength, MaxTitleLength)),
+		validation.Field(&event.ShortDescription, validation.Required, validation.NewStringRule(noNewLines, "short description should not contain line breaks"), validation.RuneLength(1, 256),
+			validation.When(checkResourcePolicyLevel(docPolicyLevel, event.PolicyLevel, PolicyLevelSap), validation.Match(regexp.MustCompile(ShortDescriptionSapCorePolicyRegex)), validation.Length(MinTitleLength, MaxTitleLength), validation.By(validateShortDescriptionDoesNotStartWithResourceName(event.Name)))),
+		validation.Field(&event.Description, validation.Required, validation.Length(MinDescriptionLength, MaxDescriptionLength),
```
emmakarova, line 539: Units tests for validation of `description` and `shortDescription` fields in the case when sap-core policy is set.

components/director/internal/open_resource_discovery/validation.go
dragobt: One more thing which `validateLinks` do is that it checks if the elements of the array are json objects which is not done in `validateJSONArrayOfObjects`. 
Perhaps If i call this function in a different scenario i want it to work as well, so i think that it needs this checks.

emmakarova (approved):

PR #3340, opened 2023-09-29
Fix reporting for FA with READY state in asynchronous case

ivantenevvasilev:
**Description**
When reporting a status for a formation assignment in `READY` state with the reset option, the formation assignments weren't persisted and we relied that they would both pass through `ProcessFormationAssignmentPair`, but that is not always the case. If we have an asynchronous webhook attached to the reverse assignment and we send the configuration to it, we would send the notification and it would change to `INITIAL` and wait for a regular status report. This would leave the assignment update for the assignment not committed. 
Another problem that could happen is when we report a new configuration, but the receiving end does not return any configuration, the assignment would be left in `INITIAL`.

- enable reporting for formation assignments that do not have webhooks, but the reverse assignment has one
- add e2e test that reports status with reset unidirectionally with one webhook
- add e2e test that checks status report with reset option for the asynchronous case

**Related issue(s)**
- #3260

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/formationmapping/handler_test.go
```diff
@@ -1297,14 +1299,16 @@ func TestHandler_ResetFormationAssignmentStatus(t *testing.T) {
 			transactFn: txGen.ThatSucceedsTwice,
 			faServiceFn: func() *automock.FormationAssignmentService {
 				faSvc := &automock.FormationAssignmentService{}
+				faSvc.On("Update", contextThatHasTenant(internalTntID), testFormationAssignmentID, faWithSourceAppAndTargetRuntime(model.ConfigPendingAssignmentState)).Return(nil).Once()
```
nyordanoff, line 1302: Minor: we could cover the error cases of the `Update` as well?

nyordanoff (approved):

PR #3358, opened 2023-10-04
Fix formation assignment race condition during unassign

PetarTodorovv:
**Description**
There is a potential issue when we have at least one async participant(with async webhook) and one(or more) sync participants in formation and try to unassign them. If the async notification is the first one executed, the state of the FA is changed to "DELETING"(in memory) but not committed until the sync notification is completed as well. However, the async participant, meanwhile, can send a status update request to the "status API" where his formation assignment will be retrieved from the DB, and its state will be "READY", it won't be reflected until the sync participant finishes his work. That leads to a problem because instead of deleting the formation assignment we'll update it to a ready state and try to send another formation assignment notifications.

In the `CleanupFormationAssignment` method in the case of async_callback webhook, we update the formation assignment in the "DELETING" state but this is no longer needed and should be removed after we adapt the unassign method to have another separate transaction in which we change the formation assignments states to "DELETING". Since the `CleanupFormationAssignment` is used also in the resync method(`resynchronizeFormationAssignmentNotifications), it should be adapted as well. The same approach is used there as well - a separate transaction that takes care of updating the formation assignments and committing them.

- When executing an unassign operation, change all formation assignments states to DELETING in an isolated DB transaction that will be committed.
- Add unit tests for the new functionality and adapt the old ones
- Remove the extra delay from the destination e2e test + remove changes related to that extra delay
- Adapt the resync method to have a separate transaction that updates formation assignments and commits them.

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/formation/service.go
```diff
@@ -1763,3 +1770,44 @@ func isObjectTypeSupported(formationTemplate *model.FormationTemplate, objectTyp
...
+		if err := s.formationAssignmentService.Update(ctx, fa.ID, fa); err != nil {
+			return errors.Wrapf(err, "while updating formation assignment with ID: '%s' to '%s' state", fa.ID, fa.State)
+		}
+	}
+
+	return nil
+}
+
+// transaction wraps a given function into an isolated DB transaction
+func (s *service) transaction(ctx context.Context, dbCalls func(ctxWithTransact context.Context) error) error {
```
StanislavStefanov, line 1794: Can we rename the method to executeInTransaction as it seems more intuitive

components/director/internal/domain/formation/service.go
```diff
@@ -1763,3 +1770,44 @@ func isObjectTypeSupported(formationTemplate *model.FormationTemplate, objectTyp
 
 	return true
 }
+
+// updateFormationAssignmentsWithStateForObjectID lists all formation assignments for a given objectID that are either source or target of the assignment
+// and update every one of them with the provided state
+func (s *service) updateFormationAssignmentsWithStateForObjectID(ctx context.Context, formationID, objectID, objectType, faState string) error {
```
StanislavStefanov, line 1776: Can we rename the last argument to newState so that it is explicit what the argument is used for

StanislavStefanov (changes requested):

StanislavStefanov (dismissed):

ivantenevvasilev (approved):

dzahariev (approved):

dzahariev (approved):

ivantenevvasilev (dismissed):

components/director/internal/domain/formation/service.go
```diff
@@ -1769,3 +1841,45 @@ func isObjectTypeSupported(formationTemplate *model.FormationTemplate, objectTyp
 
 	return true
 }
+
+// updateFormationAssignmentsWithStateForObjectID lists all formation assignments for a given objectID that are either source or target of the assignment
```
StanislavStefanov, line 1845: Can we mention in the comment that the function also resets the config and error of the FAs

components/director/internal/domain/formationassignment/service.go
```diff
@@ -794,17 +800,7 @@ func (s *service) CleanupFormationAssignment(ctx context.Context, mappingPair *A
 
 	requestWebhookMode := mappingPair.AssignmentReqMapping.Request.Webhook.Mode
 	if requestWebhookMode != nil && *requestWebhookMode == graphql.WebhookModeAsyncCallback {
-		log.C(ctx).Infof("The webhook with ID: %q in the notification is in %q mode. Updating the assignment with ID: %q to state: %q and waiting for the receiver to report the status on the status API...", mappingPair.AssignmentReqMapping.Request.Webhook.ID, graphql.WebhookModeAsyncCallback, assignment.ID, string(model.DeletingAssignmentState))
```
StanislavStefanov, line 797: This is removed as it is already done in the resync method?

StanislavStefanov (changes requested):

dzahariev (dismissed):

components/director/internal/domain/formation/service.go
```diff
@@ -1002,106 +1062,124 @@ func (s *service) ResynchronizeFormationNotifications(ctx context.Context, forma
...
-				ReverseAssignmentReqMapping: reverseReqMapping,
-			},
-		}
-		switch fa.State {
-		case string(model.InitialAssignmentState), string(model.CreateErrorAssignmentState):
-			assignmentPair.Operation = model.AssignFormation
-			if _, err = s.formationAssignmentService.ProcessFormationAssignmentPair(faResyncTransactionCtx, &assignmentPair); err != nil {
-				errs = multierror.Append(errs, err)
+			if notificationForFA != nil && notificationForFA.Webhook.Mode != nil && *notificationForFA.Webhook.Mode == graphql.WebhookModeAsyncCallback && operation == model.UnassignFormation {
+				fa.State = string(model.DeletingAssignmentState)
```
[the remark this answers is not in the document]
PetarTodorovv: IMHO I think it's better to be handled in the `ProcessFormationAssignmentPair` and not in the resync method because that way it's encapsulated in the `ProcessFormationAssignmentPair ` method and if someone is using it won't have to prepare the assignments in INITIAL state, he will directly call the method. E.g. if we do it in the resync method, we have to do it for all other places using the `ProcessFormationAssignmentPair ` method - `formationmapping` package(`processFormationAssignmentNotifications`), assign/ProcessFormationAssignments in the `formationassignment` package, and potentially other places.

We are doing it for Unassign because the management of the "DELETING" state is no longer handled in the `CleanupFormationAssignment` method but rather in the `UnassignFormation` method. And `CleanupFormationAssignment` is used both in `UnassignFormation` and `resync` methods. That's why we have to handle this "async callback deleting" state in the resync method as well.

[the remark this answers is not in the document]
StanislavStefanov: The assignments should not be set to INITIAL in the ProcessFormationAssignmentPair. Setting it there is a workaround as the opertion is deduced from the state. Also if we keep setting the state to initial in ProcessFormationAssignmentPair we will have to introduce another transaction there as this is the reason for the rece conditions in assign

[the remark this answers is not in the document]
PetarTodorovv: Discussed offline and it's a separate issue in the Assign method that will be fix as part of another PR.

components/director/internal/domain/formation/service.go
```diff
@@ -889,64 +881,140 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+	err = nerr
 	if err != nil {
 		if apperrors.IsNotFoundError(err) {
-			commitErr := tx.Commit()
-			if commitErr != nil {
-				return nil, errors.Wrapf(err, "while committing transaction with error")
+			if commitErr := tx.Commit(); commitErr != nil {
+				err = errors.Wrapf(
+					errors.Wrapf(commitErr, "while committing transaction"),
+					errors.Wrapf(nerr, "while generating formation assignment notifications").Error(),
```
StanislavStefanov, line 956: `err` already contains the data from `nerr` so line 956 is not needed it will only duplicate the message inside. Also do we really need the commit error itself. Isnt the message "while committing transaction with error" enough to show what happend?

components/director/internal/domain/formation/service.go
```diff
@@ -889,64 +881,140 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+
 			return formationFromDB, nil
 		}
 		return nil, errors.Wrapf(err, "while generating notifications for %s unassignment", objectType)
 	}
 
-	formationAssignmentsForObject, err := s.formationAssignmentService.ListFormationAssignmentsForObjectID(transactionCtx, formationFromDB.ID, objectID)
-	if err != nil {
-		return nil, errors.Wrapf(err, "while listing formationAssignments for object with type %q and ID %q", objectType, objectID)
+	formationAssignmentClonesForObject := make([]*model.FormationAssignment, 0)
```
StanislavStefanov, line 966: make([]*model.FormationAssignment, 0, len(initialAssignmentsData))

components/director/internal/domain/formation/service.go
```diff
@@ -889,64 +881,140 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
-	if err = s.formationAssignmentService.ProcessFormationAssignments(transactionCtx, formationAssignmentsForObject, rtmContextIDsMapping, applicationIDToApplicationTemplateIDMapping, requests, s.formationAssignmentService.CleanupFormationAssignment, model.UnassignFormation); err != nil {
-		commitErr := tx.Commit()
-		if commitErr != nil {
-			return nil, errors.Wrapf(err, "while committing transaction with error")
+	if nerr = s.formationAssignmentService.ProcessFormationAssignments(transactionCtx, formationAssignmentClonesForObject, rtmContextIDsMapping, applicationIDToApplicationTemplateIDMapping, requests, s.formationAssignmentService.CleanupFormationAssignment, model.UnassignFormation); nerr != nil {
+		err = nerr
+		if commitErr := tx.Commit(); commitErr != nil {
+			err = errors.Wrapf(
+				errors.Wrapf(commitErr, "while committing transaction"),
+				errors.Wrapf(nerr, "while processing formation assignments").Error(),
```
StanislavStefanov, line 988: same as with the previous commit error

components/director/internal/domain/formation/service.go
```diff
@@ -889,64 +881,140 @@ func (s *service) UnassignFormation(ctx context.Context, tnt, objectID string, o
...
+			)
+			return nil, err
 		}
+
+		shouldRevertAssignmentsUpdateFromTransaction = false
```
StanislavStefanov, line 993: Should be after the transaction is commited. Currently if the commit fails the assignments will not be reverted

components/director/internal/domain/formation/service.go
```diff
@@ -1002,106 +1070,126 @@ func (s *service) ResynchronizeFormationNotifications(ctx context.Context, forma
...
-		if err != nil && !apperrors.IsNotFoundError(err) {
-			return nil, err
+		for _, rfa := range resyncableFAs {
+			if rfa.State == string(model.DeleteErrorAssignmentState) {
+				failedDeleteErrorFormationAssignments = append(failedDeleteErrorFormationAssignments, rfa)
+			}
 		}
-		if reverseFA != nil {
-			notificationForReverseFA, err = s.formationAssignmentNotificationService.GenerateFormationAssignmentNotification(faResyncTransactionCtx, reverseFA, operation)
+		resyncableFormationAssignments = resyncableFAs
```
StanislavStefanov, line 1115: It seems the declaration of resyncableFormationAssignments at the beggining of the method is not needed. We can remove it and use resyncableFAs, or use resyncableFormationAssignments on line 1101 instead of resyncableFAs

StanislavStefanov (changes requested):

components/director/internal/domain/formation/service.go
PetarTodorovv: We have two operations here that's why I wanted to log both of them so we could have better log and it will be easier to know the transaction error occurred while another error happened(generate notifications).
Also, the log shouldn't be duplicated? It should be overridden. The initial value of err will be what we have in `nerr` but to 954 line we assign another value.

components/director/internal/domain/formation/service.go
PetarTodorovv: check the comment above

components/director/internal/domain/formation/service.go
PetarTodorovv: we have another flag that is flipped after the transaction is committed, on line 1004. This flag here is needed to cover the case when we have "processFormationAssignments" error and the transaction is successfully committed. because we'll return with the err on line 994 and we don't want to revert the changed FAs, since the inner transaction is successfully committed.

components/director/internal/domain/formation/service.go
PetarTodorovv: The `resyncableFAs` is in the scope of the first transaction but the `resyncableFormationAssignments` is used outside of this scope, it' used in the second transaction. That's why we get the FAs in the first transaction and assigned it to `resyncableFormationAssignments` var so we can use it later.

StanislavStefanov (approved):

PR #3362, opened 2023-10-05
Adapt local env octopus

ivantenevvasilev:
**Description**
Currently when running e2e tests on an arm machine, it uses the amd64 image of the octopus manager and it sometimes schedules pods before they should be, reaching `running > concurrency` and practically running multiple test pods at the same time. This leads to failed tests, as they conflict with the resources, because it is usually the same test pod.

- patch image for arm64 machines to use arm64 image of octopus

**Pull Request status**

- [x] [N/A] Implementation
- [x] [N/A] Unit tests
- [x] [N/A] Integration tests
- [x] [N/A] `chart/compass/values.yaml` is updated 

### DIFF
--- a/installation/scripts/install-compass.sh
+++ b/installation/scripts/install-compass.sh
@@ -96,7 +96,7 @@ echo "Compass installation status ${STATUS}"
 
 if [[ $(uname -m) == "arm64" ]]; then
   echo "Patching image on octopus for arm64..."
- 	kubectl set image -n compass-system statefulset/compass-octopus "manager=europe-west1-docker.pkg.dev/sap-cp-cmp-dev/ucl-dev/octopus:5f353cd5"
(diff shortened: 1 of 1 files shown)


installation/scripts/install-compass.sh
```diff
@@ -94,6 +94,11 @@ echo "Compass installation finished successfully"
 STATUS=$(helm status compass -n compass-system -o json | jq .info.status)
 echo "Compass installation status ${STATUS}"
 
+if [[ $(uname -m) == "arm64" ]]; then
+  echo "Patching image on octopus for arm64..."
+ 	kubectl set image -n compass-system statefulset/compass-octopus "manager=europe-west1-docker.pkg.dev/sap-cp-cmp-dev/ucl-dev/octopus:5f353cd5"
```
ognyvrac, line 99: use `$RELEASE_NS` for the namespace to have it unified everywhere

ognyvrac (approved):

PR #3369, opened 2023-10-06
Fix aggregator log timestamps

alextargov:
**Description**

- ORD Aggregator uses JSON formatting for the logger. This causes the GCP logs to lack information about log's milliseconds. With this, all logs in a certain second have "000" milliseconds leading to messed up logs.
- This is how logs on dev-main look after this change was applied there.
image

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/pkg/log/configure.go
```diff
@@ -48,6 +49,15 @@ var (
 		textFormatterKey: &logrus.TextFormatter{},
 	}
 
+	supportedFormattersWithTimestamp = map[string]logrus.Formatter{
```
dzahariev, line 52: I think is better to have only formatters with timestamps instead.

chart/compass/charts/ord-aggregator/templates/deployment.yaml
```diff
@@ -108,6 +108,8 @@ spec:
               value: "{{ .Values.global.ordAggregator.http.retry.attempts }}"
             - name: APP_HTTP_RETRY_DELAY
               value: "{{ .Values.global.ordAggregator.http.retry.delay }}"
+            - name: APP_LOG_TIMESTAMP_FORMATTED
```
dzahariev, line 111: I think this is not needed. Having timestamps should be default behaviour.

dzahariev (changes requested):
Please simplify this.

dzahariev (approved):

la4ezar (approved):

PR #3371, opened 2023-10-08
Fail when deleting application if it is part of formation

StanislavStefanov:
**Description**
Applications that are part of formation can be deleted without excluding them from the formation. If the application is visible to the tenant in which context the deletion operation is executed but the application labels are not visible the application will be deeleted. The other participants in the formation will not be notified about the application not being part of the formation anymore.

- Use Global List for getting application labels when ensuring the application is not part of a formation
- Upsert tenant access level for application with DO NOTHING on conflict
- Extend e2e test to try to delete an application with tenant for which the application is visible but the scenario label is not

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [x] Integration tests


components/director/internal/domain/application/service.go
```diff
@@ -790,6 +791,29 @@ func (s *service) GetLabel(ctx context.Context, applicationID string, key string
 	return label, nil
 }
 
+// GetScenariosLabelGlobal missing godoc
+func (s *service) GetScenariosLabelGlobal(ctx context.Context, applicationID string) (*model.Label, error) {
```
DimitarPetrov, line 795: If the application is in different formations in different tenants, it will have multiple scenario labels, therefore this probably needs to be List + merge all values in the application code.

tests/ord-service/tests/subscription_flow_test.go
```diff
@@ -645,6 +645,7 @@ func TestConsumerProviderFlow(stdT *testing.T) {
 		require.ElementsMatch(stdT, []string{destination.Name, destinationSecond.Name}, []string{destinationsFromResponse[0].Get("sensitiveData.destinationConfiguration.Name").String(), destinationsFromResponse[1].Get("sensitiveData.destinationConfiguration.Name").String()})
 		stdT.Log("Successfully fetched system with bundles and destinations")
 
+		unassignFromFormation(stdT, ctx, providerApp.ID, "APPLICATION", consumerFormationName, secondaryTenant)
```
nyordanoff, line 648: Why this unassign had to be added here?

nyordanoff (changes requested):

tests/ord-service/tests/subscription_flow_test.go
StanislavStefanov: The application in the test is assigned to formation in secondary tenant so the scenario label that is created has the tenant field set to the secondaryTenants ID and is not visible to subscriptionConsumerSubaccount.
When the unsubscribe is performed in the context of subscriptionConsumerSubaccount we attempt to delete the application. Before the fix the subscriptionConsumerSubaccountID did not see that the application is in formation and deleted it. With the fix it sees that there is a formation and fails to delete it. The unassign is needed before the unsubscribe so that the applcation can be deleted

nyordanoff (approved):

PR #3395, opened 2023-10-13
Improve status codes in status API

ivantenevvasilev:
**Description**

- adapt `InternalServerError` to `BadRequest` in the cases when the entity has been deleted.

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] [N/A] Integration tests


components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -276,6 +278,9 @@ func (a *Authenticator) isFormationAuthorized(ctx context.Context, formationID s
 
 	f, err := a.formationRepo.GetGlobalByID(ctx, formationID)
 	if err != nil {
+		if apperrors.IsNotFoundError(err) {
+			return false, http.StatusBadRequest, errors.Wrapf(err, "while getting formation with ID: %q globally", formationID)
+		}
```
PetarTodorovv, line 283: Isn't the 404 status more appropriate in such cases instead of a bad request? Also, the error message should be adapted as well because currently all of them are `errors.Errorf("An unexpected error occurred while processing the request. X-Request-Id: %s", correlationID)` regardless of the status code
P.S. valid for other places as well

components/director/internal/formationmapping/handler.go
```diff
@@ -137,6 +141,10 @@ func (h *Handler) updateFormationAssignmentStatus(w http.ResponseWriter, r *http
 	formation, err := h.formationService.Get(ctx, formationID)
 	if err != nil {
 		log.C(ctx).WithError(err).Errorf("An error occurred while getting formation from formation assignment with ID: %q", fa.FormationID)
+		if apperrors.IsNotFoundError(err) {
+			respondWithError(ctx, w, http.StatusBadRequest, errResp)
+			return
+		}
```
PetarTodorovv, line 147: the `errResp` is a "generic" err for the internal server errors. But with the new changes if the error is not found IMO I think the 404 is more appropriate(check the other comment as well) and the error message to the client should be adapted as well, e.g. "The formation assignment with ID: %s does not exist" or something similar.

Also, if we improve the status codes in the "status API" we should revisit all of the methods/responses and adapt them accordingly, e.g. 401, 404, 405, 500, 503, etc.

PetarTodorovv (changes requested):

components/director/internal/formationmapping/handler.go
ivantenevvasilev: I agree, we should revisit all responses. The ones I added here are so that we do not get alerts for 5xx requests when the error is on the client side. 
Currently if a user reports a status on the formations status API or formation assignment status API, it returns 500 in cases of not found. This often triggers the high number of 5xx requests alert and a person has to manually check if it is user error or an actual server error. It could also hide the actual 5xx requests.
This PR is more a slight improvement on the current situation, rather than an overhaul of what we return response and status code we return and when.

ivantenevvasilev: I could change the new ones to 404 with a better message, as they should be.

components/director/internal/formationmapping/auth_middleware.go
```diff
@@ -279,7 +279,7 @@ func (a *Authenticator) isFormationAuthorized(ctx context.Context, formationID s
 	f, err := a.formationRepo.GetGlobalByID(ctx, formationID)
 	if err != nil {
 		if apperrors.IsNotFoundError(err) {
-			return false, http.StatusBadRequest, errors.Wrapf(err, "while getting formation with ID: %q globally", formationID)
+			return false, http.StatusNotFound, errors.Wrapf(err, "while getting formation with ID: %q globally", formationID)
```
PetarTodorovv, line 282: we'll return 404 status code and error but in the handler this error is only logged for traceability and another "generic" error is returned - `errors.Errorf("An unexpected error occurred while processing the request. X-Request-Id: %s", correlationID)` which seems a little strange
Same for the FA function/handler

Actually, I saw it's the same comment as the first one

PetarTodorovv (approved):

PR #3410, opened 2023-10-23
Add FormationAssignmentTemplateInput to join point details

StanislavStefanov:
**Description**
Some usecases of the Config Mutator require for it to have access to fields(like baseURL) of the source or target application of the notification.

- add FormationAssignmentTemplateInput to NotificationStatusReturnedOperationDetails

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


components/director/pkg/formationconstraint/validation.go
```diff
@@ -198,6 +200,33 @@ func emptyNotificationStatusReturnedOperationDetails() *NotificationStatusReturn
 			Value: json.RawMessage("\"\""),
 			Error: json.RawMessage("\"\""),
 		},
+		FormationAssignmentTemplateInput: &webhook.ApplicationTenantMappingInput{
+			Operation:   model.AssignFormation,
+			FormationID: "",
+			Formation: &model.Formation{
+				ID:                  "",
+				FormationTemplateID: "",
+				Name:                "",
```
PetarTodorovv, line 209: We could skip these empty strings as the zero/default value is `""` to minimize the size of the structure. Valid for `FormationID` field(+ other fields) as well. e.g. fixApplicationModel -> BaseEntity

components/director/pkg/formationconstraint/validation.go
```diff
@@ -198,6 +200,33 @@ func emptyNotificationStatusReturnedOperationDetails() *NotificationStatusReturn
...
+			TargetApplicationTemplate: &webhook.ApplicationTemplateWithLabels{
+				ApplicationTemplate: fixApplicationTemplateModel(),
+				Labels:              fixLabels(),
+			},
+			TargetApplication: &webhook.ApplicationWithLabels{
+				Application: fixApplicationModel(),
+				Labels:      fixLabels(),
+			},
+			Assignment:        nil,
+			ReverseAssignment: nil,
```
PetarTodorovv, line 228: Shouldn't the `Assignment` and `ReverseAssignment` be similar to the other example where we have the `Value` and `Error` fields "zeroed"?

PetarTodorovv (changes requested):

components/director/pkg/formationconstraint/validation.go
StanislavStefanov: We do not expect the assignments to be used from here as they are also present in the joinpoint details on top level

components/director/pkg/templatehelper/helper.go
```diff
@@ -80,3 +88,20 @@ func updateAndCopy(input json.RawMessage, path string, entries []string) string
 func mkslice(args ...string) []string {
 	return args
 }
+
+func stringify(v interface{}) string {
```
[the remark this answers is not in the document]
StanislavStefanov: As discussed offline the method from the library does not suite our needs and we will proceed with the custom method

components/director/pkg/templatehelper/helper.go
```diff
@@ -23,6 +24,7 @@ func GetFuncMap() template.FuncMap {
 	fm["copy"] = copyFromJSON
 	fm["updateAndCopy"] = updateAndCopy
 	fm["mkslice"] = mkslice
+	fm["stringify"] = stringify
```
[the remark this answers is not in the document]
StanislavStefanov: As discussed offline for now we wont prefix the custom functions

components/director/pkg/formationconstraint/validation.go
StanislavStefanov: Wouldnt it be better to leave them like this so that it would be explicit that we wanted it to be that way

PetarTodorovv (approved):

PR #3417, opened 2023-10-24
Adopt ORD specification 1.6.3

dzahariev:
**Description**
- Make partOfProduct optional for Packages
- Add related entity types to capability
- Make version optional for ORD ID in tombstone
- Introduce EntityTypeMapping concept

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/model/entitytypemapping.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// EntityTypeTarget represents the Entity Type Target entity
+type EntityTypeTarget struct {
+	OrdId         *string `json:"ordId,omitempty"`
+	CorrelationId *string `json:"correlationId,omitempty"`
+}
+
+// Validate missing godoc
+func (entityTypeTarget *EntityTypeTarget) Validate() error {
```
emmakarova, line 111: When both values are missing this validation function will return true.
So I think we should add these validation rules to the fields
for entityTypeTarget.OrdId -> `validation.When(entityTypeTarget.CorrelationId == nil, validation.Required)`
for entityTypeTarget.CorrelationId -> `validation.When(entityTypeTarget.OrdId == nil, validation.Required)`

emmakarova (approved):

nyordanoff (approved):

PR #3433, opened 2023-11-01
Use self-signed keystore certificates for SAML destinations

nyordanoff:
**Description**

This PR adopts the functionality of issuing long living self-signed certificates in the DestinationCreator Operator. If the flow is `SAML Assertion` we want to issue this type of certs by default. This behaviour can be flipped by a parameter passed in the constraint input template.

- adapt the `SAML Assertion` flow in the DestinationCreator Operator to use self-signed certs based on a field passed in the constraint input template (by default it won't be passed so self-signed certs will be requested)
- adapt the DestinationCreator Operator e2e test to use the more detailed `find API` for validating the created destinations/certs: this API doesn't return just the destination details but directly issue creds and provide them for usage
- adaptations in the external-services-mock and the charts

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


tests/director/tests/notifications/fixtures_test.go
```diff
@@ -342,72 +344,96 @@ func resetShouldFailEndpointFromExternalSvcMock(t *testing.T, client *http.Clien
...
+	require.Equal(t, samlAssertionDestinationName, samlAssertionDest.DestinationConfiguration.Name)
+	require.Equal(t, directordestinationcreator.TypeHTTP, samlAssertionDest.DestinationConfiguration.Type)
+	require.Equal(t, samlAssertionDestinationURL, samlAssertionDest.DestinationConfiguration.URL)
+	require.Equal(t, directordestinationcreator.AuthTypeSAMLAssertion, samlAssertionDest.DestinationConfiguration.Authentication)
+	require.Equal(t, directordestinationcreator.ProxyTypeInternet, samlAssertionDest.DestinationConfiguration.ProxyType)
+	require.Equal(t, app1BaseURL, samlAssertionDest.DestinationConfiguration.Audience)
+	require.Equal(t, samlAssertionCertName+directordestinationcreator.JavaKeyStoreFileExtension, samlAssertionDest.DestinationConfiguration.KeyStoreLocation)
+
+	require.Equal(t, len(expectedCertNames), len(samlAssertionDest.CertificateDetails))
+	for i := 0; i < len(expectedCertNames); i++ {
```
StanislavStefanov, line 406: How do we know if the certificarte details will always have that order

tests/director/tests/notifications/fixtures_test.go
```diff
@@ -342,72 +344,96 @@ func resetShouldFailEndpointFromExternalSvcMock(t *testing.T, client *http.Clien
...
+	require.Equal(t, instanceID, clientCertAuthDest.Owner.InstanceID)
+	require.Equal(t, clientCertAuthDestinationName, clientCertAuthDest.DestinationConfiguration.Name)
+	require.Equal(t, directordestinationcreator.TypeHTTP, clientCertAuthDest.DestinationConfiguration.Type)
+	require.Equal(t, clientCertAuthDestinationURL, clientCertAuthDest.DestinationConfiguration.URL)
+	require.Equal(t, directordestinationcreator.AuthTypeClientCertificate, clientCertAuthDest.DestinationConfiguration.Authentication)
+	require.Equal(t, directordestinationcreator.ProxyTypeInternet, clientCertAuthDest.DestinationConfiguration.ProxyType)
+	require.Equal(t, clientCertAuthCertName+directordestinationcreator.JavaKeyStoreFileExtension, clientCertAuthDest.DestinationConfiguration.KeyStoreLocation)
+
+	require.Equal(t, len(expectedCertNames), len(clientCertAuthDest.CertificateDetails))
+	for i := 0; i < len(expectedCertNames); i++ {
```
StanislavStefanov, line 433: Same as above

tests/director/tests/notifications/fixtures_test.go
nyordanoff: Currently, I am not aware of any use-case where 1 Destination can have more than 1 Certificate. So, it's probably safe to assume that the length of that array will be 1.

However, in the future there could be a change which will require support of multiple certs, so I can re-write it with maps to be on the safe side. What do you think?

StanislavStefanov: If there wont be more than one certificate we can leave it as is. If usecase with more than one cert comes we will adapt the test and can adapt the fixture as well.

StanislavStefanov (approved):

PR #3434, opened 2023-11-01
Return null config instead of empty string

ivantenevvasilev:
**Description**
Currently when sending notifications we had workarounds in place to send empty configurations as `""`, but the more appropriate JSON value would be the `null` value. We removed the `<nil>` value that comes from parsing templates after validating the JSON body. This meant that if we were to have `.Assignment.Value` as nil, it would have to be wrapped in quotes, because it would be rendered as `<nil>` in the JSON, which leads to a syntax error.

- remove workarounds for sending empty string as configuration when there is no `Value` in the assignment
- avoid sending `"null"` instead of `""`
- send `null` instead of no configuration
- move the `<nil>` replacement before validating the JSON

**Related issue(s)**

- ...

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x]  [N/A] Integration tests


components/director/pkg/webhook/types.go
```diff
@@ -174,7 +174,9 @@ func parseTemplate(tmpl *string, data interface{}, dest interface{}) error {
 	if err = t.Execute(res, data); err != nil {
 		return err
 	}
-	if err = json.Unmarshal(res.Bytes(), dest); err != nil {
+	resBytes := bytes.ReplaceAll(res.Bytes(), []byte("\"<nil>\""), []byte("\"\""))
+	resBytes = bytes.ReplaceAll(resBytes, []byte("<nil>"), []byte("null"))
```
PetarTodorovv: Could you please add two comments for both replace scenarios? E.g. the first one is when the field in the go code is a pointer(nil/empty) and we have explicit quotes in the templates resulting in the `"<nil>"` but the second case is when we don't have quotes in the template and we have only `<nil>`?

components/director/pkg/webhook/types.go
PetarTodorovv, line 1: And one general comment - we need to verify all of the components/adapters that we have are not affected by this change, i.e. in the new setup I think they should have only a not nil check? Rather than before not nil check + additional non empty string checks - "{}", "null", etc..

PetarTodorovv (changes requested):

PetarTodorovv (dismissed):

PetarTodorovv (approved):

PetarTodorovv (approved):

alextargov (approved):

PetarTodorovv (approved):

alextargov (approved):

PR #3469, opened 2023-11-15
Director pre-install job

ognyvrac:
**Description**
In the director deployment we have certificate rotation cronjobs that create Secrets which are used by other compass services to interact with the auth middleware - this means that these certifcate Secrets are prerequisites. The `pre-install` hook has been used to initially create these certificates, afterwards, the cronjobs can handle the rotation.

The pre-install jobs are only going to be used on our cloud landscapes as locally these certificate jobs interact with the `external-mock-service` which is installed by the same chart - the `pre-install` hook would not work as the `external-mock-service` is not yet deployed.

- move certificate rotation logic to script to avoid code duplication
- introduces all Kubernetes entities needed for the pre-install jobs
- introduce `EXT_NOT_EXTERNAL` env variable to differentiate between the the two certificate jobs; this varible is later used in `chart/compass/charts/director/files/certificate_rotation.sh` to execute the only different logic

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


chart/compass/values.yaml
```diff
@@ -958,6 +958,11 @@ global:
       certValidity: "7"
       clientCertRetryAttempts: "8"
       containerName: "certificate-rotation"
+    preInstallJob:
+      enabled: false
```
nyordanoff, line 962: General question: as there are no overrides for the benchmark installation, do we want to run these pre jobs there or not?

The benchmark is closer to a local setup but on the other hand it is a clean installation on a cluster so that raised my question.

chart/compass/charts/director/templates/ext-svc-certificate-pre-install-job.yaml
```diff
@@ -0,0 +1,131 @@
...
+      serviceAccountName: {{ $.Chart.Name }}-{{ .Values.global.extSvcCertConfiguration.preInstallJob.name }}
+
+      containers:
+        - name: {{ .Values.global.extSvcCertConfiguration.preInstallJob.name }}
+          image: bitnami/kubectl:1.22.4
+          command:
+            - /bin/bash
+            - -c
+            - |
+{{ .Files.Get "files/certificate_rotation.sh" | printf "%s" | indent 14 }}
```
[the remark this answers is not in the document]
ognyvrac: I remember that I made it `chmod +x` to enable it being executed

nyordanoff (approved):

chart/compass/charts/director/files/certificate_rotation.sh
```diff
@@ -0,0 +1,119 @@
...
+
+  if [[ $IS_CERT_ISSUER_LOCALITY_VALID == false ]]; then
+    echo -e "${RED}Couldn't get the client certificate with valid issuer locality after $CLIENT_CERT_RETRY_ATTEMPTS attempts. Exiting... ${NC}"
+    exit 1
+  fi
+}
+
+echo -e "${YELLOW}Issuing token... ${NC}"
+
+if [[ "$EXT_NOT_EXTERNAL" == true ]]; then
```
PetarTodorovv, line 75: What's the meaning behind this `EXT_NOT_EXTERNAL` variable? As far as I can see this var is used to determine if we should encode the certificate or not when saving it to a file? If that's the situation could we use another name for it?
Also, did you try to "unify" both approaches? Either to encode the cert/key for both jobs or both jobs to use non-encoded data? Seems the cert/key files are used only when getting tokens and from the scripts it seems both approaches are working? Is there a specific reason for this encoding?

ognyvrac: I tried unifying and it did not work. I could not understand why one is encoded and the other not, I assumed it is an implementation detail

PetarTodorovv: strange, nevertheless, could we still use a different name for the variable? e.g. `should_encode_cert`? or something similar?

PR #3475, opened 2023-11-16
Use notification report from details

StanislavStefanov:
**Description**
Use dedicated structure NotifiactionStatusReport for storing notification status response in SYNC case and notification status update report from ASYNC CALLBACK case. Validate, process and mutate the state, config and error in the structure instead of directly mutating the formations assignments. That way we have consistent state of the formation assignment in the DB and in memory.

- Store state, config and error from the report/response in NotifiactionStatusReport object
- Mutate the state, config and error only in the NotifiactionStatusReport object
- Update the formation assignement once the response/report processing is finished

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


components/director/internal/domain/formationassignment/service.go
```diff
@@ -1081,21 +1047,108 @@ type AssignmentErrorWrapper struct {
...
+			return string(model.CreateErrorAssignmentState)
+		}
+
+		return string(model.DeleteErrorAssignmentState)
+	}
+
+	if *response.ActualStatusCode == *response.SuccessStatusCode && webhookMode != graphql.WebhookModeAsyncCallback {
+		return string(model.ReadyAssignmentState)
+	}
+	if operation == model.AssignFormation && webhookMode == graphql.WebhookModeAsyncCallback {
```
ivantenevvasilev, line 1080: Maybe add a comment that this state shouldn't reach `Update` or `UpdateWithConstraints`?

components/director/internal/domain/formationassignment/service.go
```diff
@@ -1081,21 +1047,108 @@ type AssignmentErrorWrapper struct {
...
+		return string(model.InitialAssignmentState)
+	}
+
+	if operation == model.AssignFormation && webhookMode == graphql.WebhookModeSync {
+		return string(model.ConfigPendingAssignmentState)
+	}
+	return string(model.DeletingAssignmentState)
```
ivantenevvasilev, line 1087: Maybe add a comment that this state shouldn't reach `Update` or `UpdateWithConstraints`? And that it only should happen when `operation == model.UnassignFormation && webhookMode == graphqlWebhookModeAsyncCallback` with a comment?

components/director/internal/domain/formationassignment/status_service.go
```diff
@@ -64,6 +53,33 @@ func (fau *formationAssignmentStatusService) UpdateWithConstraints(ctx context.C
...
 
+	stateFromReport := notificationStatusReport.State
+	fa.State = stateFromReport
+
+	if isErrorState(model.FormationAssignmentState(stateFromReport)) {
+		assignmentError := json.RawMessage{}
+		if notificationStatusReport.Error != "" {
+			assignmentErrorWrapper := AssignmentErrorWrapper{AssignmentError{
+				Message:   notificationStatusReport.Error,
+				ErrorCode: ClientError,
```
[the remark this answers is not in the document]
StanislavStefanov: The error in the notificationStatusReport is populated by the Client in case they had an error while performing some operation on their end. We are not editing the error field ourselves.

components/director/internal/domain/formationassignment/service.go
StanislavStefanov: A comment was added to explain why we are skipping further notification processing. Also the `calculateState` method cuold be used in many places and for some of them this may not be the case. Generally whether the state shoud or should not reach certain places in the code does not concern the method that caculates the state.

components/director/internal/domain/formationassignment/service.go
StanislavStefanov: The DELETING state actually reaches the UpdateWithConstraints in case of unassign and it should reach it so that we have recorded that the FA is being deleted

components/director/internal/domain/formationassignment/service.go
```diff
@@ -652,64 +650,43 @@ func (s *service) processFormationAssignmentsWithReverseNotification(ctx context
...
 		}
 	}
 
-	assignment.State = *response.State
-
-	// Notification response has been received. The error should be cleared. The old configuration should be discarded and the new one from the response should be used.
-	if assignment.State == string(model.ReadyAssignmentState) {
-		ResetAssignmentConfigAndError(assignment)
+	if err = validateNotificationResponse(response, assignment, mappingPair.Operation, webhookMode); err != nil {
+		updateError := s.SetAssignmentToErrorState(ctx, assignment, err.Error(), ClientError, model.CreateErrorAssignmentState)
```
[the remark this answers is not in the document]
ivantenevvasilev: Discussed offline that this probably should be done separately.

ivantenevvasilev (approved):

ivantenevvasilev (approved):

PR #3498, opened 2023-11-24
Fix notification response with empty error message

ivantenevvasilev:
**Description**

- set error only if there is error, so that we don't fall into the `json.RawMessage{}` case, in which it tries to update the formation assignment error with a non-nil value that is `""`, which is not a valid JSON.
- adapt e2e test to cover this case
- add external services mock response with error state and no message

**Related issue(s)**
- #3484

**Pull Request status**

- [x] Implementation
- [x] [N/A] Unit tests
- [x] Integration tests

### DIFF
--- a/components/external-services-mock/internal/formationnotification/handler.go
+++ b/components/external-services-mock/internal/formationnotification/handler.go
@@ -739,7 +739,7 @@ func (h *Handler) executeFormationAssignmentStatusUpdateRequest(certSecuredHTTPC
 
 	request.Header.Add(correlation.RequestIDHeaderKey, correlationID)
 	request.Header.Add(httphelpers.ContentTypeHeaderKey, httphelpers.ContentTypeApplicationJSON)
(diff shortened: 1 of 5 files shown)


components/external-services-mock/internal/formationnotification/handler.go
```diff
@@ -671,16 +671,39 @@ func (h *Handler) AsyncFail(writer http.ResponseWriter, r *http.Request) {
...
 	config := "test error"
 	h.asyncFAResponse(ctx, writer, r, operation, config, responseFunc)
 }
 
+// AsyncFailNoError handles asynchronous formation assignment notification requests for both Assign and Unassign operations by failing and setting error states.
+func (h *Handler) AsyncFailNoError(writer http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	operation := Assign
+	if r.Method == http.MethodPatch {
+		operation = Assign
```
PetarTodorovv, line 687: We could remove the assign/patch `if` as the assign operation is the default value from the above and leave only one `if` statement for the delete/unassign?

PetarTodorovv (approved):

PR #3500, opened 2023-11-24
Make formation assignment notification order more deterministic

ivantenevvasilev:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/domain/formation/service.go
```diff
@@ -1186,24 +1197,25 @@ func (s *service) resynchronizeFormationAssignmentNotifications(ctx context.Cont
 	}
 
 	alreadyProcessedFAs := make(map[string]bool, 0)
+	assignmentMappingPairs := append(assignmentMappingNoNotificationPairs, append(assignmentMappingSyncPairs, assignmentMappingAsyncPairs...)...)
```
StanislavStefanov, line 1200: We should add a comment with description why we separate the pairs and then append them. Also it should be added in the FA service as well. 

In the new format e2e tests for the config mutator there are two comments that describes two cases based on the notification execution order - the comment should be deleted

StanislavStefanov (approved):

PR #3531, opened 2023-12-07
Introduce instance creator asynchronous flow control operator

ivantenevvasilev:
**Description**
Introduce a new asynchronous flow control operator that upon successful status report resends the notification to the instance creator and modifies the state so that the assignment is not deleted. When attached to `SendNotification` it behaves like the redirect operator.
The new assignment states are `INSTANCE_CREATOR_DELETING` and `INSTANCE_CREATOR_DELETE_ERROR` and can only be updated by the instance creator operator. This required the inclusion of a new consumer type of instance creator.
The new states are not exposed to the graphql API and are handled internally.

- introduce new operator - `AsynchronousFlowControl`
- add e2e test that utilizes the new operator and simulates a status report response from the instance creator
- add new consumer type for instance creator

**Related issue(s)**
- #3331

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/formationassignment/status_service.go
```diff
@@ -116,11 +116,13 @@ func (fau *formationAssignmentStatusService) DeleteWithConstraints(ctx context.C
 		return errors.Wrapf(err, "while enforcing constraints for target operation %q and constraint type %q", model.NotificationStatusReturned, model.PreOperation)
 	}
 
-	if err = fau.repo.Delete(ctx, id, tenantID); err != nil {
-		if apperrors.IsUnauthorizedError(err) {
-			return apperrors.NewNotFoundError(resource.FormationAssignment, id)
+	if notificationStatusReport.State == string(model.ReadyAssignmentState) {
+		if err = fau.repo.Delete(ctx, id, tenantID); err != nil {
+			if apperrors.IsUnauthorizedError(err) {
```
alextargov, line 121: Why for unauthoriazed error we return a NotFound Err? Probably it is intended, but I'm curious about it

components/director/internal/domain/formationconstraint/operators/asynchronous_flow_control_operator.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	if ri.Location.OperationName == model.NotificationStatusReturned && ri.Location.ConstraintType == model.PreOperation {
+		formationAssignment, err := RetrieveFormationAssignmentPointer(ctx, ri.FAMemoryAddress)
+		if err != nil {
+			return false, err
+		}
+		reverseAssignment, err := RetrieveFormationAssignmentPointer(ctx, ri.ReverseFAMemoryAddress)
+		if err != nil {
+			log.C(ctx).Warnf(errors.Wrapf(err, "Reverse assignment not found").Error())
```
alextargov, line 70: Is there a use case for this error that could potentially break some stuff if the flow is not stopped with a return statement?

ivantenevvasilev: The reverse assignment could be `nil` in the one case that it is used, that's why it is not stopped with a return statement. 
The `GenerateFormationAssignmentPair` method handles `nil` reverse assignment, because the reverse could have been deleted at the time of generating notifications for unassign. Looking at the code now this can be moved in the one if that is used.

components/director/internal/domain/formationassignment/status_service.go
ivantenevvasilev: This is due to the way the repo layer `Update` and `Delete` methods work, in cases when there are 0 rows affected in the SQL statement, they return `Unauthorized` rather than `NotFound`. This combined with the transaction isolation level, if some other transaction deletes the assignment before this one, it would return `Unauthorized`, even though this transaction could get the entity a few lines above in the same transaction.

tests/pkg/notifications/operations/execute-status-report.go
```diff
@@ -0,0 +1,129 @@
...
+	listFormationAssignmentsReq := fixtures.FixListFormationAssignmentRequest(formationID, 100)
+	assignmentsPage := fixtures.ListFormationAssignments(t, ctx, gqlClient, o.tenant, listFormationAssignmentsReq)
+
+	formationAssignmentID := getFormationAssignmentIDBySourceAndTarget(t, assignmentsPage, o.assignmentSource, o.assignmentTarget)
+
+	t.Logf("Calling FA status API for formation assignment ID %q", formationAssignmentID)
+	faAsyncStatusAPIURL := strings.Replace(o.externalServicesMockMtlsSecuredURL, fmt.Sprintf("{%s}", formationIDPathParam), formationID, 1)
+	faAsyncStatusAPIURL = strings.Replace(faAsyncStatusAPIURL, fmt.Sprintf("{%s}", formationAssignmentIDPathParam), formationAssignmentID, 1)
+	reqBody := FormationAssignmentRequestBody{
+		State: "READY",
```
nyordanoff, line 94: I may be mistaken here, but since we aim to make this `ExecuteStatusReportOperation` generic, do we want to always respond with hardcoded `READY` state? In case we want to report another state, will we have to create another operation?

components/director/internal/domain/formationconstraint/operators/asynchronous_flow_control_operator.go
```diff
@@ -0,0 +1,124 @@
...
+
+	ri, ok := input.(*formationconstraint.AsynchronousFlowControlOperatorInput)
+	if !ok {
+		return false, errors.Errorf("Incompatible input for operator: %s", AsynchronousFlowControlOperator)
+	}
+
+	log.C(ctx).Infof("Enforcing constraint on resource of type: %q and subtype: %q for location with constraint type: %q and operation name: %q during %q operation", ri.ResourceType, ri.ResourceSubtype, ri.Location.ConstraintType, ri.Location.OperationName, ri.Operation)
+
+	if ri.Operation == model.AssignFormation && ri.Location.OperationName == model.SendNotificationOperation && ri.Location.ConstraintType == model.PreOperation {
+		return e.RedirectNotification(ctx, &ri.RedirectNotificationInput)
```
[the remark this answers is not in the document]
ivantenevvasilev: Yes, that was my idea - it can be in the input template and redirect based on the input template of the `RedirectNotificationInput`

tests/director/tests/notifications/application_only_notifications_new_format_test.go
```diff
@@ -619,4 +637,133 @@ func TestFormationNotificationsWithApplicationOnlyParticipantsNewFormat(t *testi
...
+		expectationsBuilder = mock_data.NewFAExpectationsBuilder().WithParticipant(app2ID)
+		asserter := asserters.NewFormationAssignmentAsserter(expectationsBuilder.GetExpectations(), expectationsBuilder.GetExpectedAssignmentsCount(), certSecuredGraphQLClient, tnt)
+		statusAsserter := asserters.NewFormationStatusAsserter(tnt, certSecuredGraphQLClient)
+		op = operations.NewAssignAppToFormationOperation(app2ID, tnt).WithAsserters(asserter, statusAsserter).Operation()
+		defer op.Cleanup(t, ctx, certSecuredGraphQLClient)
+		op.Execute(t, ctx, certSecuredGraphQLClient)
+
+		t.Logf("Cleanup notifications")
+		op = operations.NewCleanupNotificationsOperation().WithExternalServicesMockMtlsSecuredURL(conf.ExternalServicesMockMtlsSecuredURL).WithHTTPClient(certSecuredHTTPClient).Operation()
+		defer op.Cleanup(t, ctx, certSecuredGraphQLClient)
```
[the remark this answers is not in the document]
ivantenevvasilev: It seems this is how we are doing it in other tests as well. As far as I remember it's to not forget it in cases that it matters and if we eventually add some action, we would not have to add it everywhere.

tests/pkg/notifications/asserters/notifications_count_async.go
```diff
@@ -0,0 +1,53 @@
...
+			externalServicesMockMtlsSecuredURL: externalServicesMockMtlsSecuredURL,
+			client:                             client,
+		},
+		timeout: time.Second * 8,
+		tick:    time.Millisecond * 50,
+	}
+}
+
+func (a *NotificationsCountAsyncAsserter) AssertExpectations(t *testing.T, ctx context.Context) {
+	require.Eventually(t, func() (isOkay bool) {
```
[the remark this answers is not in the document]
ivantenevvasilev: If it is named, the `return` statements below implicitly return false and the last one returns explicitly `true`. If it isn't we would have to add `return false` everywhere. This is how it was done in other util methods with Eventually.

nyordanoff (approved):

components/director/internal/model/formation_assignment.go
```diff
@@ -118,6 +132,69 @@ func (f *FormationAssignment) Clone() *FormationAssignment {
...
+	return state
+}
+
+func (f *FormationAssignment) isInProgressAssignState() bool {
+	return f.State == string(InitialAssignmentState) ||
+		f.State == string(ConfigPendingAssignmentState)
+}
+
+func (f *FormationAssignment) isInProgressUnassignState() bool {
+	return strings.HasSuffix(f.State, string(DeletingAssignmentState))
```
DimitarPetrov, line 195: Will that catch the regular non-opt-in fearure deleting state?

ivantenevvasilev: Yes, I tested it and `HasSuffix` matches `DELETING` state (`DELETING` is a suffix of `DELETING`). Also it is validated in the e2e tests via the FormationStatus asserts

components/director/internal/model/formation_assignment.go
```diff
@@ -118,6 +132,69 @@ func (f *FormationAssignment) Clone() *FormationAssignment {
...
+		operation = UnassignFormation
+	}
+	return operation
+}
+
+// GetNotificationState returns the assignment state that is to be sent for notifications
+// and is exposed to the GraphQL layer
+func (f *FormationAssignment) GetNotificationState() string {
+	state := f.State
+	if strings.HasSuffix(state, string(DeleteErrorFormationState)) {
```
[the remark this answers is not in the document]
ivantenevvasilev: Yes, it is way too similar to the Assignment one, so I missed it. Will update it.

components/director/internal/model/formation_assignment.go
```diff
@@ -118,6 +132,69 @@ func (f *FormationAssignment) Clone() *FormationAssignment {
...
+	}
+	f.State = string(DeletingAssignmentState)
+	return true
+}
+
+// GetOperation returns the formation operation that is determined based on the state of the assignment
+func (f *FormationAssignment) GetOperation() FormationOperation {
+	operation := AssignFormation
+	if f.State == string(DeleteErrorAssignmentState) || f.State == string(DeletingAssignmentState) ||
+		f.State == string(InstanceCreatorDeleteErrorAssignmentState) || f.State == string(InstanceCreatorDeletingAssignmentState) {
```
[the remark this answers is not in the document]
ivantenevvasilev: I forgot to adapt it after moving it. I will adapt it as well.

components/director/internal/model/formation_assignment.go
DimitarPetrov: If that's the case, can we remove the explicit checks for the regular states in the other methods?

ivantenevvasilev: Which method exactly? The `IsInRegularUnassignState` has to explicitly check that it is not a opt-in feature state, for the others I don't see an explicit check. If it is the `GetNotificationState` I adapted it with the previous commit.

DimitarPetrov: You are right, maybe I overlooked something.

nyordanoff (dismissed):

nyordanoff (approved):

alextargov (approved):

PR #3534, opened 2023-12-11
Fix update of resources in the ord aggregator

emmakarova:
**Description**

In the ORD aggregation when a resource is already present in the db, update operation is executed. But if the package ID of the resource has changed this will not result in actually changing it in our db, because the packageID that is passed in the update input always comes from the one that was stored in db.
This problem exists for API, Event, Capability, Integration dependency and Entity Type.
The PR fixes this behaviour by passing the packageID as an additional parameter to the Update functions.

- pass `packageID` as a parameter to the Update functions of api, event, capability, integration dependency and entity type
- use this `packageID` in the update logic instead of the packageID that was passed from the input object
- adapt unit tests

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/domain/capability/service.go
```diff
@@ -109,14 +109,14 @@ func (s *service) Create(ctx context.Context, resourceType resource.Type, resour
...
-func (s *service) Update(ctx context.Context, resourceType resource.Type, id string, in model.CapabilityInput, capabilityHash uint64) error {
+func (s *service) Update(ctx context.Context, resourceType resource.Type, id string, packageID *string, in model.CapabilityInput, capabilityHash uint64) error {
 	capability, err := s.getCapability(ctx, id, resourceType)
 	if err != nil {
 		return errors.Wrapf(err, "while getting Capability with ID %s for %s", id, resourceType)
 	}
 
 	resourceID := getParentResourceID(capability)
-	capability = in.ToCapability(id, resourceType, resourceID, capability.PackageID, capabilityHash)
+	capability = in.ToCapability(id, resourceType, resourceID, packageID, capabilityHash)
```
dragobt, line 119: Same here, packageID can be nil.

components/director/internal/domain/integrationdependency/service.go
```diff
@@ -105,13 +105,13 @@ func (s *service) Create(ctx context.Context, resourceType resource.Type, resour
...
 // Update updates an existing Integration Dependency.
-func (s *service) Update(ctx context.Context, resourceType resource.Type, resourceID string, id string, in model.IntegrationDependencyInput, integrationDependencyHash uint64) error {
+func (s *service) Update(ctx context.Context, resourceType resource.Type, resourceID string, id string, packageID *string, in model.IntegrationDependencyInput, integrationDependencyHash uint64) error {
 	integrationDependency, err := s.getIntegrationDependency(ctx, id, resourceType)
 	if err != nil {
 		return errors.Wrapf(err, "while getting Integration Dependency with ID %s for %s", id, resourceType)
 	}
 
-	integrationDependency = in.ToIntegrationDependency(id, resourceType, resourceID, integrationDependency.PackageID, integrationDependencyHash)
+	integrationDependency = in.ToIntegrationDependency(id, resourceType, resourceID, packageID, integrationDependencyHash)
```
dragobt, line 114: Same here, packageID can be nil.

dragobt (changes requested):

dragobt (approved):

PR #3543, opened 2023-12-14
Fix ORD Aggregator does not persist certain ORD validation errors

dragobt:
**Description**
Right now we store only the error message which is caused by the processing logic of the ORD. If there is a validation error we only log its message.

With this change we store all error messages which are generated into an ORD execution.
- If there is only a validation error, we mark the operation as completed and save the error message.
- If there is only processing error, we mark the operation as failed, because of the processing failure, and store the message.
- If there are both validation and processing error, we mark the operation as failed, because of the processing failure, and store the error message.
- unit tests
- regenerated mocks

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/operation/service.go
```diff
@@ -88,17 +88,27 @@ func (s *service) DeleteMultiple(ctx context.Context, ids []string) error {
...
 
 // MarkAsCompleted marks an operation as completed
-func (s *service) MarkAsCompleted(ctx context.Context, id string) error {
+func (s *service) MarkAsCompleted(ctx context.Context, id, errorMsg string) error {
 	op, err := s.opRepo.Get(ctx, id)
 	if err != nil {
 		return errors.Wrapf(err, "while getting opreration with id %q", id)
 	}
 
+	if errorMsg != "" {
```
kaloyanspiridonov, line 97: remove else:
op.Error = json.RawMessage("{}")
if errorMsg != "" {
....
....
}

kaloyanspiridonov (changes requested):

kaloyanspiridonov (approved):

PR #3547, opened 2023-12-15
Reduce logs in asynchronous logs in notification tests

ivantenevvasilev:
**Description**

- add LogOnce logger that only prints the first instance of given log to reduce unnecessary logs in e2e tests

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/tests/pkg/testing/testing.go
+++ b/tests/pkg/testing/testing.go
@@ -18,6 +18,48 @@ func NewT(t *testing.T) *T {
 	}
 }
 
+type Logger interface {
+	Log(args ...any)
+	Logf(format string, args ...any)
+	Errorf(format string, args ...interface{})
+	FailNow()
+}
+
+type LoggerWithOnlyOnce struct {
+	logger        Logger
+	alreadyLogged map[string]bool
+}
+
+func NewLoggerWithOnlyOnce(logger Logger) LoggerWithOnlyOnce {
+	return LoggerWithOnlyOnce{
+		logger:        logger,
+		alreadyLogged: map[string]bool{},
+	}
+}
+
+func (l *LoggerWithOnlyOnce) Log(log string, args ...any) {
+	if l.alreadyLogged[log] {
+		return
+	}
+	l.alreadyLogged[log] = true
+	l.logger.Log(append([]any{log}, args...)...)
+}
+
+func (l *LoggerWithOnlyOnce) Logf(format string, args ...any) {
+	if l.alreadyLogged[format] {
+		return
+	}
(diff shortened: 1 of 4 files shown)


tests/pkg/testing/testing.go
```diff
@@ -18,6 +18,48 @@ func NewT(t *testing.T) *T {
 	}
 }
 
+type Logger interface {
+	Log(args ...any)
+	Logf(format string, args ...any)
+	Errorf(format string, args ...interface{})
+	FailNow()
```
PetarTodorovv, line 25: Do we need these two(Errorf and FailNow)? because we don't have additional logic behind them, just calling the underlying methods?

tests/pkg/testing/testing.go
```diff
@@ -18,6 +18,48 @@ func NewT(t *testing.T) *T {
...
+}
+
+type LoggerWithOnlyOnce struct {
```
PetarTodorovv, line 28: How does it sound "OnceLogger" or "LoggerOnce" to you? Or something similar? 
And we could also add a short godoc about the idea of this logger and recommendations to use it together with the retry/async asserts to reduce the logs

PetarTodorovv (changes requested):

tests/pkg/testing/testing.go
ivantenevvasilev: It's required by the methods that accept `require.TestingT`, so I'm just passing it through, so that we don't have to pass in the old `t` for those and for the logs the new one.

PetarTodorovv: I found the same approach is used already in the "test-compass/tests/pkg/notifications/asserters/formation_assignments_async.go" file in the `json.AssertJSONStringEquality(...)` fixture where we pass the testing `t` object instead of "onceLogger"? Couldn't we use the same thing here?
Also, if I understand the idea correctly is to use this only for logs that should be printed only once? And not as a substitute for the testing object(in the fixtures and other places)? And not avoiding it purposely and passing one object on both places?

ivantenevvasilev: If we use the regular object and there are logs within the helper methods, they will be logged multiple times, so it's better to use tOnce in those cases. In the `json.AssertJSONStringEquality` I've just forgot to adapt it to use `tOnce`. In the case of `AssertJSONStringEquality` it just does not print anything, so it doesn't matter which one is used.

PetarTodorovv (dismissed):

PetarTodorovv (approved):

PR #3548, opened 2023-12-15
Fix nil pointer dereference in hydrator component

PetarTodorovv:
**Description**
In some cases during certificate subject mapping loading in the hydrator component, we could get into a situation where there is no error after a graphql request was executed but the response(certificate subject mapping structure/object) is nil. And trying to get something from that response will result in a nil pointer dereference error.

- Add additional check for nil during certificate subject mapping loading

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/hydrator/internal/certsubjectmapping/loader.go
```diff
@@ -160,8 +161,18 @@ func (cl *certSubjectMappingLoader) loadCertSubjectMappings(ctx context.Context,
...
+		// The graphql response could be nil, and there could be NO graphql error in case something else failed.
+		// E.g., an error from the oathkeeper or other components
+		if csmGQLPage == nil {
+			return mappings, errors.Errorf("the certificate subject mappings page cannot be nil")
+		}
 		csmTotalCount = csmGQLPage.TotalCount
 		mappings = append(mappings, convertGQLCertSubjectMappings(csmGQLPage.Data)...)
+
+		if csmGQLPage.PageInfo == nil {
+			return mappings, errors.Errorf("the certificate subject mappings page info be nil")
```
StanislavStefanov, line 174: the certificate subject mappings page info cannot be nil

StanislavStefanov (changes requested):

components/hydrator/internal/certsubjectmapping/loader.go
PetarTodorovv: It could in case of non-graphql error(e.g. ory). We even encounter that issue causing the hydrator component to panic

StanislavStefanov: I meant that the message should be `the certificate subject mappings page info cannot be nil`. The word `cannot` is missed in the current version

PetarTodorovv: Oh, my bad. I thought the comment was about the `if` statement. I fix the error message.

StanislavStefanov (approved):

PR #3566, opened 2023-12-28
Add discovery consumers allowlist to formation templates

DimitarPetrov:
**Description**

- Add discovery consumers allowlist as a column of the formation template entity
- Add JSONB index to scenario labels for faster querying

**Related issue(s)**
- https://github.com/kyma-incubator/ord-service/pull/118

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/pkg/graphql/formation_template.go
+++ b/components/director/pkg/graphql/formation_template.go
@@ -11,6 +11,7 @@ type FormationTemplate struct {
 	Webhooks               []*Webhook    `json:"webhooks"`
 	LeadingProductIDs      []string      `json:"leadingProductIDs"`
 	SupportsReset          bool          `json:"supportsReset"`
+	DiscoveryConsumers     []string      `json:"discoveryConsumers"`
 }
 
 // FormationTemplateExt  is an extended types used by external API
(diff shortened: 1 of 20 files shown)


components/director/pkg/graphql/formation_template.go
```diff
@@ -11,6 +11,7 @@ type FormationTemplate struct {
 	Webhooks               []*Webhook    `json:"webhooks"`
 	LeadingProductIDs      []string      `json:"leadingProductIDs"`
 	SupportsReset          bool          `json:"supportsReset"`
+	DiscoveryConsumers     []string      `json:"discoveryConsumers"`
```
nyordanoff, line 14: In the corresponding validation file: `formation_template_validation.go` we can check if the values of the discovery consumers are among the application/runtime types:

Example check:
`validation.Field(&i.DiscoveryConsumers, validation.In(i.RuntimeTypes, i.ApplicationTypes)),`

If you find it necessary, you can extend the unit tests in the `formation_template_validation_test.go`

nyordanoff (approved):

PR #3581, opened 2024-01-15
Allow assign only for applications that are ready

ivantenevvasilev:
**Description**
Currently we allow applications to be assigned even though they are marked as deleted.

- disallow assign of applications that are not ready or are being deleted

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] [N/A] Integration tests


components/director/internal/domain/formation/service.go
```diff
@@ -1758,6 +1761,21 @@ func (s *service) isValidRuntimeType(ctx context.Context, tnt string, runtimeID
...
+	if err != nil {
+		return errors.Wrapf(err, "while getting application with ID %q", applicationID)
+	}
+
+	if !application.Ready {
+		return apperrors.NewInvalidOperationError(fmt.Sprintf("application with ID %q is not ready", applicationID))
+	}
+	if application.DeletedAt != nil {
+		return apperrors.NewInvalidOperationError(fmt.Sprintf("application with ID %q is currently being deleted", applicationID))
+	}
```
DimitarPetrov, line 1775: Maybe we can switch the checks and check first for the deleted at and then for the ready as those properties are correlated anyways. The Delete At timestamps gives more information about why the system is not available and if the deletedAt is set the Ready will always be false.

components/director/internal/domain/formation/assign_formation_test.go
```diff
@@ -945,6 +971,81 @@ func TestServiceAssignFormation(t *testing.T) {
 			InputFormation:     inputFormation,
 			ExpectedErrMessage: "cannot convert label value to slice of strings",
 		},
+		{
```
PetarTodorovv, line 974: `error when the application is not in ready state`?

components/director/internal/domain/formation/service.go
```diff
@@ -770,6 +770,9 @@ func (s *service) checkFormationTemplateTypes(ctx context.Context, tnt, objectID
 		if err := s.isValidApplicationType(ctx, tnt, objectID, formationTemplate); err != nil {
 			return errors.Wrapf(err, "while validating application type for application %q", objectID)
 		}
+		if err := s.isValidApplication(ctx, tnt, objectID); err != nil {
+			return errors.Wrapf(err, "while application %q", objectID)
```
PetarTodorovv, line 774:
```suggestion
			return errors.Wrapf(err, "while validating application with ID: %q", objectID)
```

PetarTodorovv (changes requested):

PetarTodorovv (approved):

PR #3284, opened 2023-09-01
Support sending notifications for self-referenced formation assignments

ivantenevvasilev:
**Description**
We currently do not support sending of notifications for self-referenced formation assignments, with this PR the functionality is added.

Changes proposed in this pull request:
- remove workarounds for self-referenced formation assignments
- revert deletion of formation assignments at the start of unassign formation #3105
- introduce new global constraint `DoNotGenerateFormationAssignmentNotificationForLoopsGlobalApplication` that disables notifications for self-referenced formation assignments
- refactor unassign formation to work like the unassign flow in the status API. The formation assignment operations come first, then the scenario label manipulation. They are both in separate transactions.
- split up `ResynchonizeFormationNotifications` into multiple transactions
- refactor unassign formation tests to reduce boilerplate
- add e2e tests for the following scenarios 
    1.  assigning and unassigning an application with synchronous notification for self-referenced formation assignments
    2. assigning and unassigning an application with asynchronous notification for self-referenced formation assignments
    3. resynchronizing notifications for synchronous self-referenced formation assignment
    4.  resynchronizing notifications for asynchronous self-referenced formation assignment
    5.  general test with assign, unassign and resynchronization with multiple participants that have notifications for self-referenced formation assignments
- self-referenced formation assignments do not automatically become in `READY` state when in formation lifecycle flows, they are treated the same way as regular assignments and stay in `INITIAL` until the formation becomes `READY`

**Related issue(s)**
- #3105
- #3134
- #2798 (Skip sending notifications for formation assignments with matching Source and Target)

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [x] Integration tests
- [x] `chart/compass/values.yaml` is updated 
- [x] [N/A] Mocks are regenerated, using the automated script

PR #3583, opened 2024-01-16
Fix status report for self-referenced formation assignment

ivantenevvasilev:
**Description**
In the case of self-referenced assignments the second unassign from scenario label will always fail, because the first one will have deleted it.

- only unassign `Target` if it is different from `Source` 

**Related issue(s)**
- #3284

**Pull Request status**

- [x] Implementation
- [x] [N/A] Unit tests
- [x] [N/A] Integration tests


components/director/internal/formationmapping/handler.go
```diff
@@ -393,8 +393,10 @@ func (h *Handler) processFormationUnassign(ctx context.Context, formation *model
 		return errors.Wrapf(err, "while unassigning object with type: %q and ID: %q", fa.SourceType, fa.Source)
 	}
 
-	if err = h.unassignObjectFromFormationWhenThereAreNoFormationAssignments(unassignCtx, fa, formation, fa.Target, fa.TargetType); err != nil {
-		return errors.Wrapf(err, "while unassigning object with type: %q and ID: %q", fa.TargetType, fa.Target)
+	if fa.Source != fa.Target {
```
PetarTodorovv, line 396: Do you think a comment with info will be helpful here?
e.g.
```
// Skip second unassign from formation in case the formation assignment is self-referenced one, 
// and if happened to be the last one. It will be handled above.
```

ivantenevvasilev: Yes, it probably will be useful, I'll add it.

PetarTodorovv (dismissed):

PetarTodorovv (approved):

PR #3591, opened 2024-01-18
Introduce unassignFormationApplicationGlobal mutation

ivantenevvasilev:
**Description**

- add new mutation - `unassignFormationApplicationGlobal`
- add new scope - `formation:global_write`
- unit tests

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/pkg/graphql/schema.graphql
```diff
@@ -2011,6 +2011,7 @@ type Mutation {
 	- [unassign tenant from formation](examples/unassign-formation/unassign-tenant-from-formation.graphql)
 	"""
 	unassignFormation(objectID: ID!, objectType: FormationObjectType!, formation: FormationInput!): Formation! @hasScopes(path: "graphql.mutation.unassignFormation")
+	unassignFormationApplicationGlobal(applicationID: ID!, formationID: ID!): Formation! @hasScopes(path: "graphql.mutation.unassignFormationApplicationGlobal")
```
DimitarPetrov, line 2014: Why don't we follow the pattern above for unassignFormation? We can just call the new mutation unassignFormationGlobal and accept objectId and objectType. The formation argument is better to work with ID instead of a whole input as we know from experience that the old api was not designed in the best possible way in that aspect.

If we don't want to follow the pattern, we can still improve the naming, for example - unassignApplicationFromFormationGlobal

ivantenevvasilev: I made it as such, because it was needed only for applications, but I will generalize it to be as similar to `unassignFormation` to accept objectType and objectID. 
The problem the name is that we have 2 unique constraints - the ID primary key and <name, tenant_id> . If we want it to be global (without a tenant) it has to be done through the ID.

chart/compass/charts/director/config.yaml
```diff
@@ -85,6 +85,7 @@ graphql:
     deleteFormation: ["formation:write"]
     assignFormation: [ "formation:write" ]
     unassignFormation: [ "formation:write" ]
+    unassignFormationGlobal: [ "formation:global_write" ]
```
nyordanoff, line 88: Copy this in `components/director/hack/config-local.yaml` as well so that the mutation can be used in local run of director.

nyordanoff (changes requested):

nyordanoff (approved):

ekaterina-mitova (approved):
Looks good to me.

PR #3592, opened 2024-01-18
Postgres15 preparations

ognyvrac:
**Description**
Change the Postgres images that we use to version 

- update localdb image for local installation
- update schema migrator images

**Pull Request status**

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests


chart/localdb/charts/postgresql/values.yaml
```diff
@@ -15,7 +15,7 @@ global:
 image:
   registry: docker.io
   repository: bitnami/postgresql
-  tag: 11.15.0-debian-10-r93
```
nyordanoff, line 18: Shouldn't we change the value in the `values-production.yaml` as well? [Link](https://github.com/kyma-incubator/compass/blob/main/chart/localdb/charts/postgresql/values-production.yaml#L17)

nyordanoff:
Do we want to adapt the db version during local director execution - https://github.com/kyma-incubator/compass/blob/main/components/director/run.sh#L91?

Maybe here as well - https://github.com/kyma-incubator/compass/blob/main/components/director/hack/jwt_generator.sh#L18

nyordanoff (approved):

PR #3593, opened 2024-01-19
Add timestamps to formations and formation assignments

PetarTodorovv:
**Description**

- ...

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/schema-migrator/migrations/director/20240119172919_add-formation-assignments-timestamps.up.sql
```diff
@@ -0,0 +1,21 @@
+BEGIN;
+
+ALTER table formations
+    ADD COLUMN last_state_change_timestamp TIMESTAMP,
+    ADD COLUMN last_notification_sent_timestamp TIMESTAMP;
+
+UPDATE formations SET last_state_change_timestamp = now()
```
DimitarPetrov, line 7: You can use DEFAULT value during column creation instead of updating each row of the table.

danielgospodinow:
Ah, splendid effort, old chap! Your GitHub pull request is an absolute triumph, a testament to your coding prowess. The meticulous attention to detail and the sheer brilliance of your implementation are nothing short of extraordinary. One cannot help but be thoroughly impressed by the elegant finesse with which you've tackled the task at hand. Jolly good show, indeed! Tip my hat to you, sir, for a job exceedingly well done. Bravo! 🎩

PetarTodorovv:
> Ah, splendid effort, old chap! Your GitHub pull request is an absolute triumph, a testament to your coding prowess. The meticulous attention to detail and the sheer brilliance of your implementation are nothing short of extraordinary. One cannot help but be thoroughly impressed by the elegant finesse with which you've tackled the task at hand. Jolly good show, indeed! Tip my hat to you, sir, for a job exceedingly well done. Bravo! 🎩

Haha 😄 I was pleasantly surprised to hear from you, pal! Your comment made me laugh, as always, eloquent, honey-mouthed and facetious. Thank you for your kind words, mate! Cheers

components/director/internal/domain/formation/repository.go
```diff
@@ -147,8 +147,21 @@ func (r *repository) Update(ctx context.Context, model *model.Formation) error {
...
-	log.C(ctx).Debugf("Updating formation with ID: %q and name: %q...", model.ID, model.Name)
-	return r.updater.UpdateSingleGlobal(ctx, r.conv.ToEntity(model))
+	newEntity := r.conv.ToEntity(model)
+
+	var retrievedEntity Entity
+	if err := r.globalGetter.GetGlobal(ctx, repo.Conditions{repo.NewEqualCondition("id", model.ID)}, repo.NoOrderBy, &retrievedEntity); err != nil {
+		return err
+	}
+
+	if retrievedEntity.State != newEntity.State {
```
StanislavStefanov, line 157: What if we are in an error state and after executing a notification the response contains error that is different than the previous one? Shouldn't we update the timestamp in this case also? The `last_notification_sent_timestamp` will be updated with the execution of the notification, but the  `last_state_change_timestamp` will not despite we received a response.

components/director/internal/domain/formationassignment/repository.go
```diff
@@ -329,8 +330,20 @@ func (r *repository) Update(ctx context.Context, model *model.FormationAssignmen
...
 	}
+	newEntity := r.conv.ToEntity(model)
 
-	return r.updaterGlobal.UpdateSingleGlobal(ctx, r.conv.ToEntity(model))
+	var retrievedEntity Entity
+	if err := r.globalGetter.GetGlobal(ctx, repo.Conditions{repo.NewEqualCondition("id", model.ID)}, repo.NoOrderBy, &retrievedEntity); err != nil {
+		return err
+	}
+
+	if retrievedEntity.State != newEntity.State {
```
StanislavStefanov, line 340: Here we have the same issue as with the formation entity. The FA can be in `CONFIG_PENDING` and after another wave of notifications it can still be in a `CONFIG_PENDING` but with a completely different `value`. We should consider whether the `value` and `error` fields are changed along with the `state` field.

StanislavStefanov (changes requested):

components/director/internal/domain/formation/repository.go
PetarTodorovv: As far as I understand we only want to update this timestamp(`last_state_change_timestamp`) in case the state is changed? So in the scenario that you described the state will be the same.

components/director/internal/domain/formationassignment/repository.go
PetarTodorovv: Same as the comment above, I think the timestamp only represents the change in the state not in the response(that is composite of state, error, value)?

components/director/internal/domain/formation/repository.go
PetarTodorovv: discussed offline and the code is adjusted.

components/director/internal/domain/formationassignment/repository.go
PetarTodorovv: discussed offline and the code is adjusted.

StanislavStefanov (approved):

dzahariev (approved):

components/director/internal/model/formation_assignment.go
```diff
@@ -23,16 +24,18 @@ const (
...
+	FormationID                   string                  `json:"formation_id"`
+	TenantID                      string                  `json:"tenant_id"`
+	Source                        string                  `json:"source"`
+	SourceType                    FormationAssignmentType `json:"source_type"`
+	Target                        string                  `json:"target"`
+	TargetType                    FormationAssignmentType `json:"target_type"`
+	State                         string                  `json:"state"`
+	Value                         json.RawMessage         `json:"value"`
+	Error                         json.RawMessage         `json:"error"`
+	LastStateChangeTimestamp      *time.Time
```
DimitarPetrov, line 37: Why no json tags? Is that intentional? Where do we use model structs for json (un)marshalling?

PetarTodorovv: I think these tags are not used directly because we use the converters but I added them for consistency in [this PR](https://github.com/kyma-incubator/compass/pull/3615).

PR #3596, opened 2024-01-23
List all formations for participant globally

StanislavStefanov:
**Description**
As part of the deprovisioning flows the Landscape Resource operator deletes Applications from the backend when the Landscape resource is being deleted. The LRO needs to know whether the Application corresponding to the LR is part of any formations and unassign it from them before proceeding with the deletion.

- add query for listing all formation that a object is part of

**Related issue(s)**

- ...

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


components/director/pkg/graphql/schema.graphql
```diff
@@ -1759,6 +1760,7 @@ type Query {
 	- [query formations](examples/query-formations/query-formations.graphql)
 	"""
 	formations(first: Int = 200, after: PageCursor): FormationPage! @hasScopes(path: "graphql.query.formations")
+	formationsForParticipant(participantId: String!): [Formation!]! @hasScopes(path: "graphql.query.formationsForParticipant")
```
DimitarPetrov, line 1763: We usually refer generically to the entities by "object", for example

formationsForObject(objectId: String....

If that object is not part of any formation it is actually not a "participant". Thinking in that direction, what would be the result here in that case?

components/director/internal/domain/formationassignment/repository.go
```diff
@@ -288,6 +290,21 @@ func (r *repository) ListAllForObject(ctx context.Context, tenant, formationID,
 	return r.multipleFromEntities(entities), nil
 }
 
+// ListAllForObjectGlobal retrieves all FormationAssignment objects that have objectID as `target` or `source` from the database across all tenants
+func (r *repository) ListAllForObjectGlobal(ctx context.Context, objectID string) ([]*model.FormationAssignment, error) {
+	var entities EntityCollection
+	conditions := repo.Or(repo.ConditionTreesFromConditions([]repo.Condition{
```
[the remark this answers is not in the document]
StanislavStefanov: No, there is no need to introduce new indices. The composite index is used when executing this query, like with the `ListAllForObject` method.

components/director/pkg/graphql/schema.graphql
nyordanoff: Reminder not to forget to rename it in the places where the scopes are set.

tests/pkg/fixtures/fomation_requests.go
```diff
@@ -8,6 +8,19 @@ import (
 	gcli "github.com/machinebox/graphql"
 )
 
+func FixGetFormationsForParticipantRequest(participantID string) *gcli.Request {
```
nyordanoff, line 11: I think it will be a good idea to generate an example for the new query under `examples` dir so that we have it by hand when we need it.

tests/pkg/notifications/context-keys/keys.go
```diff
@@ -3,6 +3,10 @@ package context_keys
 const (
 	FormationIDKey           = "formationIDKey"
 	FormationNameKey         = "formationNameKey"
+	SecondFormationIDKey           = "secondFormationIDKey"
+	SecondFormationNameKey         = "secondFformationNameKey"
```
nyordanoff, line 7: 1. the file can be formatted
2. typo in the value of the `SecondFormationNameKey` var -> should be `secondFormationNameKey`
3. `ApplicationID` seems to be unused so it can be deleted

tests/director/tests/formation/formation_api_test.go
```diff
@@ -122,6 +127,167 @@ func TestListFormations(t *testing.T) {
 	})
 }
 
+func TestListFormationsForParticipantGlobal(t *testing.T) {
+	ctx := context.Background()
+	tnt := tenant.TestTenants.GetDefaultTenantID()
+	otherTnt := tenant.TestTenants.GetIDByName(t, tenant.TenantSeparationTenantName)
+
+	formationTemplateName := "list-formations-for-participant-global"
```
nyordanoff, line 135: minor: can you prefix it with `e2e-test` for easier cleanup if necessary?

components/director/internal/domain/formationassignment/repository_test.go
```diff
@@ -419,6 +419,36 @@ func TestRepository_ListAllForObject(t *testing.T) {
...
 }
 
+func TestRepository_ListAllForObjectGlobal(t *testing.T) {
+	suite := testdb.RepoListTestSuite{
+		Name:       "List All Formations Assignments for object globally",
+		MethodName: "ListAllForObjectGlobal",
+		SQLQueryDetails: []testdb.SQLQueryDetails{
+			{
+				Query:    regexp.QuoteMeta(`SELECT id, formation_id, tenant_id, source, source_type, target, target_type, state, value, error FROM public.formation_assignments WHERE (source = $1 OR target = $2)`),
+				Args:     []driver.Value{TestSource, TestSource},
```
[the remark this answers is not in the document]
StanislavStefanov: The first and the second arguments should be the same as the method lists all formation assignments where the object is either source or target

ekaterina-mitova (approved):
Looks good to me.

PR #3617, opened 2024-01-28
Constraint inconsistencies fix

StanislavStefanov:
**Description**

Fix fomration constraints inconsistencies

- Do not allow a constraint to be attached to formation multiple times
- Retrun global constraints for formationConstraintsByFormationType
- Allow constraint description of size 2048

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/schema-migrator/migrations/director/20240127192653_constraint-validation.up.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+
+ALTER TABLE formation_template_constraint_references ADD CONSTRAINT unique_template_id_constraint_id UNIQUE (formation_template_id, formation_constraint_id);
```
nyordanoff, line 3: If there are multiple records for the same `formation_template_id <-> formation_constraint_id` the creation of this constraint will fail duplication error.

Have you checked if such "duplicated" records exist on all environments and overall what will be our strategy to ensure that the constraint creation won't fail?

nyordanoff (changes requested):

components/schema-migrator/migrations/director/20240127192653_constraint-validation.up.sql
StanislavStefanov: before the release such records should be removed manually. As the constraints are maintained by us there wont be any records created during the release

nyordanoff (approved):

PR #3642, opened 2024-02-07
Handle design time destinations as raw json

StanislavStefanov:
**Description**

The only authenticationType for design time destinations is NoAuthorization. In case a new format needs to be supported a code change is needed. 

For the design time destinations there is no processing performaed on the provided payload and it is only forwarded. There is no technical reason to unmarshal the provided payload into a go structure and the forwardong it. Better approach is to handle the destination details as raw json which enables us to handle new types of design time destinations without any code changes.

- handle design time destination data as raw json

**Pull Request status**

- [X] Implementation
- [X] Unit tests
- [X] Integration tests


components/director/internal/domain/formationconstraint/operators/utils.go
```diff
@@ -68,3 +71,12 @@ func RetrieveNotificationStatusReportPointer(ctx context.Context, notificationSt
 
 	return notificationStatusReport, nil
 }
+
+func contains(s []string, e string) bool {
```
ivantenevvasilev, line 75: If this is only used in `destination_creator_operator.go`, maybe `slices.Contains` can be used? It basically does the same.

components/director/internal/destinationcreator/service.go
```diff
@@ -107,42 +107,36 @@ func NewService(
...
-	destReqBody := &NoAuthDestinationRequestBody{
-		BaseDestinationRequestBody: BaseDestinationRequestBody{
-			Name:                 destinationName,
-			URL:                  destinationDetails.URL,
-			Type:                 destinationcreatorpkg.Type(destinationDetails.Type),
-			ProxyType:            destinationcreatorpkg.ProxyType(destinationDetails.ProxyType),
-			AuthenticationType:   destinationcreatorpkg.AuthType(destinationDetails.Authentication),
-			AdditionalProperties: destinationDetails.AdditionalProperties,
-		},
+	stripInternalFields, err := destinationDetails.StripInternalFields()
```
ivantenevvasilev, line 129: Maybe `destinationWithStrippedInternalFields` or something that implies that this is still a destination? `destWithStrippedInternalFields`?

PR #3643, opened 2024-02-08
Improve error handling during unassign/delete operations

PetarTodorovv:
**Description**
With a recent change that adds timestamps to formation and formation assignment entities, we've missed handling the not found error in case of unassign/delete operations. Leading to errors when the entity is already deleted.

- Ignore not found error when updating formation/formation assignment entity and not found error is received

**Related issue(s)**

- https://github.com/kyma-incubator/compass/pull/3593

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/formation/service.go
```diff
@@ -1529,6 +1529,9 @@ func (s *service) SetFormationToErrorState(ctx context.Context, formation *model
 	formation.Error = marshaledErr
 
 	if err := s.formationRepository.Update(ctx, formation); err != nil {
+		if state == model.DeleteErrorFormationState && apperrors.IsNotFoundError(err) {
```
ivantenevvasilev, line 1532: In the cases where you are checking for `DeleteErrorFormationState`, maybe you should also check the cases where the formation is in `DeletingErrorFormationState` also have to be covered. In theory someone could be resynchronizing the formation and resending the notification (it is still in DELETING), but the participant actually responds and deletes it during this operation. Same for each check that is only for `DELETE_ERROR` only.

components/director/internal/domain/formation/service_test.go
```diff
@@ -1984,6 +2018,39 @@ func TestServiceDeleteFormation(t *testing.T) {
 			ExpectedFormation:  nil,
 			ExpectedErrMessage: testErr.Error(),
 		},
+		{
+			Name: "Error when processing formation notifications fails and formation assignment is missing",
```
ivantenevvasilev, line 2022: `Error when processing formation notifications fails and formation is missing`

ivantenevvasilev (changes requested):

components/director/internal/domain/formation/service.go
PetarTodorovv: I didn't understand that, here the error state could be either `CreateErrorFormationState` or `DeleteErrorFormationState` based on the operation - `CreateFormation` or `DeleteFormation`? The `DeletingFormationState` is not propagated at all here? Same is valid for the other check in the file

ivantenevvasilev (approved):

PR #3653, opened 2024-02-13
Introduce short living cache in hydrator

dzahariev:
**Description**

- Introduce short living cache in hydrator

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/hydrator/cmd/main.go
+++ b/components/hydrator/cmd/main.go
@@ -185,7 +185,7 @@ func registerHydratorHandlers(ctx context.Context, router *mux.Router, authentic
 	runtimeMappingHandlerFunc := getRuntimeMappingHandlerFunc(ctx, internalDirectorClientProvider, cfg.JWKSSyncPeriod)
 
 	logger.Infof("Registering Authentication Mapping endpoint on %s...", cfg.Handler.AuthenticationMappingEndpoint)
-	authnMappingHandlerFunc := authnmappinghandler.NewHandler(oathkeeper.NewReqDataParser(), httpClient, authnmappinghandler.DefaultTokenVerifierProvider, authenticators)
+	authnMappingHandlerFunc := authnmappinghandler.NewHandler(oathkeeper.NewReqDataParser(), httpClient, authnmappinghandler.DefaultTokenVerifierProvider, authenticators, 30*time.Second)
 
 	logger.Infof("Registering Tenant Mapping endpoint on %s...", cfg.Handler.TenantMappingEndpoint)
 	tenantMappingHandlerFunc, err := getTenantMappingHandlerFunc(authenticators, internalDirectorClientProvider, internalGatewayClientProvider, cfg.StaticGroupsSrc, cfgProvider, cfg.ConsumerClaimsKeys, metricsCollector, cfg.TenantSubstitutionLabelKey)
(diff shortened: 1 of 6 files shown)


components/hydrator/internal/authnmappinghandler/token_data_cache.go
```diff
@@ -0,0 +1,71 @@
...
+type Entry struct {
+	TokenData TokenData
+	Taken     time.Time
+}
+
+// TokenDataCache is a local short-lived cache
+type TokenDataCache struct {
+	mutex sync.Mutex
+	ttl   time.Duration
+	data  map[string]map[string]Entry
```
PetarTodorovv, line 18: Could we simplify this and make it `map[string]Entry`, removing the double map? Holding the "token" -> Entry data. Because in the token we'll have embedded the issuer URL in the claims? That would simplify also the `PutTokenData`, `GetTokenData`, and `Cleanup` methods.

components/hydrator/cmd/main.go
```diff
@@ -185,7 +185,7 @@ func registerHydratorHandlers(ctx context.Context, router *mux.Router, authentic
 	runtimeMappingHandlerFunc := getRuntimeMappingHandlerFunc(ctx, internalDirectorClientProvider, cfg.JWKSSyncPeriod)
 
 	logger.Infof("Registering Authentication Mapping endpoint on %s...", cfg.Handler.AuthenticationMappingEndpoint)
-	authnMappingHandlerFunc := authnmappinghandler.NewHandler(oathkeeper.NewReqDataParser(), httpClient, authnmappinghandler.DefaultTokenVerifierProvider, authenticators)
+	authnMappingHandlerFunc := authnmappinghandler.NewHandler(oathkeeper.NewReqDataParser(), httpClient, authnmappinghandler.DefaultTokenVerifierProvider, authenticators, 30*time.Second)
```
PetarTodorovv, line 188: I'm wondering if we should extract this as configuration? In case we have to change/adapt in the future

PetarTodorovv (changes requested):

DimitarPetrov:
As far as I understand, the problem at hand is that we may overload the authz server (xsuaa) because our implementation is doing an online JWT token validation as opposed to offline token validation (or if it is not us that is causing the problem, we are immediately affected because the online token validation assumes that the authorization server is available).

Is there a way to implement an offline token validation instead of the cache? I believe that the library we use supports offline validation (downloading the jwks keys and validating the signature locally), however I don't know if it will work out of the box with xsuaa and their subdomain based tenancy approach (different issuer for each tenant). However, we can check if the official internal libraries have such functionality and how it is implemented, usually offline validation is a must for scalability reasons and it is very likely that there are guides and examples how to achieve it with xsuaa.

components/hydrator/internal/authnmappinghandler/token_data_cache.go
PetarTodorovv, line 1: One general comment about this "token cache" approach - it seems the slow operation(s) is initially, the first time when we load token issuers and JWKS(that we cannot escape it) + there is cache optimization for the JWKS as part of the library that we use. The PR will definitely improve the situation but the interesting question is how much? Because if it's negligible we need to consider other suggestions as well

dzahariev:
@DimitarPetrov and @PetarTodorovv, 
from the response in the ticket, I get that this cache is not needed. 
I am closing this PR.

PR #3655, opened 2024-02-13
Introduce tenant_discovery_operator consumer type in backend

petar-iv:
**Description**

- Introducing a new consumer type - "tenant_discovery_operator". A new operator will be introduced which takes an application from compass (hence, adding the "application:read" scope to the new consumer type) and creates a representation of this application in the declarative world.

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [N/A] Unit tests
- [N/A] Integration tests
- [N/A] Mocks are regenerated, using the automated script


components/director/pkg/inputvalidation/validators_cert_subject_mappings.go
```diff
@@ -25,6 +25,8 @@ const (
 	ManagedApplicationConsumerType = "Managed Application Consumer"
 	// LandscapeResourceOperatorType represents a landscape resource operator consumer type
 	LandscapeResourceOperatorType = "Landscape Resource Operator"
+	// TenantDiscoveryOperatorType represents a tenant discovery operator consumer type
```
PetarTodorovv, line 28: One more place needed to be adapted - `components/director/pkg/inputvalidation/validators_cert_subject_mappings.go` It's a consumer type validation when certificate subject mappings are created. I guess you will create CSM for this operator?

P.S. I'm not sure why all the consumers are duplicated in the mentioned file, I'll check that and make a PR to fix it and use the ones from the consumer package.

PetarTodorovv (changes requested):

PR #3667, opened 2024-02-19
Handle empty config in instance creator asynchronous flow control operator

ivantenevvasilev:
**Description**
Currently when there is no configuration in the flow control operator, it throws an error while unmarshallling.

- only Unmarshal configuration when it is not empty

**Related issue(s)**
- #3531

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] [N/A] Integration tests

### DIFF
--- a/components/director/internal/domain/formationconstraint/operators/asynchronous_flow_control_operator_test.go
+++ b/components/director/internal/domain/formationconstraint/operators/asynchronous_flow_control_operator_test.go
@@ -87,7 +87,7 @@ func TestConstraintOperators_AsynchronousFlowControlOperator(t *testing.T) {
 		},
 		// Assign during PreStatusReturned
 		{
-			Name:                             "Success when transitioning to READY state with inbound credentials",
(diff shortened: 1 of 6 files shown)


components/director/internal/domain/formationconstraint/operators/asynchronous_flow_control_operator_test.go
```diff
@@ -87,7 +87,7 @@ func TestConstraintOperators_AsynchronousFlowControlOperator(t *testing.T) {
 		},
 		// Assign during PreStatusReturned
 		{
-			Name:                             "Success when transitioning to READY state with inbound credentials",
+			Name:                             "Error when formation assignment config is invalid",
```
PetarTodorovv, line 90: Could we change the value of `invalidFAConfig` to `invalid-config` or something similar? Just to remove the destination specifics

PetarTodorovv (approved):

PR #3672, opened 2024-02-19
Extend error checks

PetarTodorovv:
**Description**
In the linked PR below we suppress not found error disguised as unauthorized error in case of update. However, during update operations we also execute get operation(s) which return not found error that we miss to handle.

- Ignore not found error when updating formation/formation assignment entity and not found error is received

**Related issue(s)**

- https://github.com/kyma-incubator/compass/pull/3643

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/chart/compass/values.yaml
+++ b/chart/compass/values.yaml
@@ -174,7 +174,7 @@ global:
       name: compass-pairing-adapter
     director:
       dir: dev/incubator/
-      version: "PR-3667"
+      version: "PR-3672"
       name: compass-director
     hydrator:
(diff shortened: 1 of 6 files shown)


ivantenevvasilev (commented):
The changes are okay, but I think the changes in #3643 were also for formation update in `formation/service.go`, I took a look at `Update` in the formation repository and it also behaves the same way - get the entity (not found error) and then update it (unauthorized).

ivantenevvasilev (approved):

PR #3673, opened 2024-02-20
Support Oauth2mTLS destinations

StanislavStefanov:
**Description**

- support oauth2mTLS destinations in Destination Creator Formation Operator

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/domain/destination/service.go
```diff
@@ -374,6 +376,65 @@ func (s *Service) createOAuth2ClientCredentialsDestinations(ctx context.Context,
...
+		if err := s.createOAuth2MTLSDestinations(ctx, destinationDetails, oauth2mTLSCredentials, formationAssignment, correlationIDs, skipSubaccountValidation); err != nil {
+			return errors.Wrapf(err, "while creating oauth2 mTLS destination with name: %q", destinationDetails.Name)
+		}
+	}
+	return nil
+}
+
+// createOAuth2MTLSDestinations is responsible to create an oauth2 mTLS destination resource in the remote destination service as well as in our DB
+func (s *Service) createOAuth2MTLSDestinations(ctx context.Context, destinationDetails operators.Destination, oauth2mTLSCredentials *operators.OAuth2mTLSAuthentication, formationAssignment *model.FormationAssignment, correlationIDs []string, skipSubaccountValidation bool) error {
+	subaccountID, err := s.destinationCreatorSvc.DetermineDestinationSubaccount(ctx, destinationDetails.SubaccountID, formationAssignment, skipSubaccountValidation)
```
PetarTodorovv, line 391: Instead of `s.destinationCreatorSvc.DetermineDestinationSubaccount` here we need to use the `EnsureDestinationSubaccountIDsCorrectness` method in the outside method(`CreateOAuth2mTLSDestinations`, the exported one) similar to the SAMLAssertion and ClientCertAuth destinations.
What is the reason for that: for these destinations, we have certificates and during certificate creation, we use this `EnsureDestinationSubaccountIDsCorrectness` method to ensure the subaccount is only one and we create the certificate/key store in that subaccount. After that during destination creation we need to ensure the subaccount is again only one. Because it will try to find the certificate/key store in that subaccount. If for some reason the subaccount is one but not the correct one, then it will fail, which is fine.

components/external-services-mock/pkg/destinationcreator/types.go
```diff
@@ -71,6 +72,19 @@ func (c *OAuth2ClientCredentialsDestination) GetType() string {
 	return OAuth2ClientCredentialsType
 }
 
+// OAuth2ClientCredentialsDestination is a structure representing a oauth2 client credentials destination entity and its data from the remote destination service
```
PetarTodorovv, line 75: Wrong type/comment

components/external-services-mock/pkg/destinationcreator/types.go
```diff
@@ -129,3 +143,10 @@ type DestinationSvcOAuth2ClientCredsDestResponse struct {
 	DestinationConfiguration OAuth2ClientCredentialsDestination `json:"destinationConfiguration"`
 	AuthTokens               []AuthTokensDetails                `json:"authTokens"`
 }
+
+// DestinationSvcOAuth2MTLSDestResponse contains the response data from destination service 'find API' request for destination of type 'OAuth2ClientCredentials'
```
PetarTodorovv, line 147: Wrong type `OAuth2ClientCredentials` in the comment

PetarTodorovv (changes requested):

PetarTodorovv (approved):

PR #3687, opened 2024-02-27
Trigger systems sync only for customers and accounts

kaloyanspiridonov:
**Description**

- Limit the requests to the system fetcher only for customer and account tenants

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/components/director/internal/domain/tenant/resolver.go
+++ b/components/director/internal/domain/tenant/resolver.go
@@ -2,6 +2,7 @@ package tenant
 
 import (
 	"context"
+	tenantpkg "github.com/kyma-incubator/compass/components/director/pkg/tenant"
 
 	"github.com/kyma-incubator/compass/components/director/pkg/inputvalidation"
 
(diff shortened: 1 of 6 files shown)


components/director/internal/domain/tenant/resolver.go
```diff
@@ -278,7 +279,13 @@ func (r *Resolver) Write(ctx context.Context, inputTenants []*graphql.BusinessTe
 		return nil, err
 	}
 
-	r.syncSystemsForTenants(ctx, tenantIDs)
+	tenantIDs := make([]string, 0, len(tenantsMap))
+	for tenantID, tenantType := range tenantsMap {
+		tenantIDs = append(tenantIDs, tenantID)
+		if tenantType == tenantpkg.Account || tenantType == tenantpkg.Customer {
```
alextargov, line 285: Can you create a function that determines that? Something like: 
```
func isSyncableTenant(tenantType tenantpkg.Type) {
   return tenantType == tenantpkg.Account || tenantType == tenantpkg.Customer
}
```

then use it on L286 and L314

alextargov (approved):

PR #3702, opened 2024-03-06
Introduce List Applications by Local Tenant Id and Filter API

petar-iv:
**Description**

- New GraphQL API for listing applications by local tenant id and optionally - a filter

**Related issue(s)**

N/A

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/components/director/internal/domain/application/repository_test.go
+++ b/components/director/internal/domain/application/repository_test.go
@@ -991,6 +991,149 @@ func TestPgRepository_GetBySystemNumber(t *testing.T) {
 	suite.Run(t)
 }
 
+func TestPgRepository_ListByLocalTenantID_NoFilter(t *testing.T) {
+	localTenantID := "local-tenant-id"
+	app1ID := "aec0e9c5-06da-4625-9f8a-bda17ab8c3b9"
+	app2ID := "ccdbef8f-b97a-490c-86e2-2bab2862a6e4"
+	appEntity1 := fixDetailedEntityApplication(t, app1ID, givenTenant(), "App 1", "App desc 1")
+	appEntity2 := fixDetailedEntityApplication(t, app2ID, givenTenant(), "App 2", "App desc 2")
+
+	appModel1 := fixDetailedModelApplication(t, app1ID, givenTenant(), "App 1", "App desc 1")
+	appModel2 := fixDetailedModelApplication(t, app2ID, givenTenant(), "App 2", "App desc 2")
+
+	suite := testdb.RepoListPageableTestSuite{
+		Name: "List Applications by Local Tenant ID no filter",
+		SQLQueryDetails: []testdb.SQLQueryDetails{
+			{
+				Query: regexp.QuoteMeta(`
+					SELECT id, app_template_id, system_number, local_tenant_id,
+						name, description, status_condition, status_timestamp, system_status,
+						healthcheck_url, integration_system_id, provider_name, base_url, application_namespace,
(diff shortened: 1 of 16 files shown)


components/director/internal/domain/application/repository_test.go
```diff
@@ -991,6 +991,149 @@ func TestPgRepository_GetBySystemNumber(t *testing.T) {
 	suite.Run(t)
 }
 
+func TestPgRepository_ListByLocalTenantID_NoFilter(t *testing.T) {
+	localTenantID := "local-tenant-id"
```
dragobt, line 995: I am not sure why the tests are passing, because this `localTenantID` value, which is passed to the query here - `...WHERE (local_tenant_id = $1...` is different from the application's `localTenantID` value which is returned. 
Passed `localTenantID` value - `local-tenant-id`
Applicaiton's `localTenantID` value - `local-tenant-id-123`

dragobt: I think the same is valid and for the other repo layer unit tests and probably for the service tests as well.

components/director/internal/domain/root_resolver.go
```diff
@@ -570,6 +570,11 @@ func (r *queryResolver) ApplicationBySystemNumber(ctx context.Context, systemNum
 	return r.app.ApplicationBySystemNumber(ctx, systemNumber)
 }
 
+// ApplicationsByLocalTenantID returns applications retrieved by local tenant id and optionally - a filter
+func (r *queryResolver) ApplicationsByLocalTenantID(ctx context.Context, localTenantID string, filter []*graphql.LabelFilter, first *int, after *graphql.PageCursor) (*graphql.ApplicationPage, error) {
```
dragobt, line 574: Consider adding an E2E test for this query, so we make sure it works.

dragobt (changes requested):

dragobt:
Do not forget to add a description to the PR.

tests/director/tests/application/application_api_test.go
```diff
@@ -2025,6 +2025,127 @@ func TestGetApplicationsAPIEventDefinitions(t *testing.T) {
...
+	tenantId := tenant.TestTenants.GetDefaultSubaccountTenantID()
+
+	localTenantID := "local-tenant-id-1234"
+
+	nameField := "name"
+	displayNameField := "displayName"
+	localTenantIDField := "localTenantId"
+
+	createAppTemplate := func(name string) graphql.ApplicationTemplate {
+		input := fixtures.FixAppTemplateInputWithDefaultDistinguishLabel(name, conf.SubscriptionConfig.SelfRegDistinguishLabelKey, name+" "+conf.SubscriptionConfig.SelfRegDistinguishLabelValue)
```
dragobt, line 2040: I think there is no need to use `FixAppTemplateInputWithDefaultDistinguishLabel`, because this "distinguish label" is responsible for representing SaaS Applications and is used for the subscription flow e2e tests.

dragobt (changes requested):

tests/director/tests/application/application_api_test.go
petar-iv: I also prefer to have only what's needed for the test to run. I first tried it with another function (I think it was this one https://github.com/kyma-incubator/compass/blob/94decb80bf142d5affc60604979194644cbfd873/tests/pkg/fixtures/application_template_requests.go#L14), but got an error (either distinguish label had to be provided or something else) and then I switched to this as the other tests in this file also use it.

dragobt (approved):

PR #3754, opened 2024-03-14
Fix cost-object id trimming

kaloyanspiridonov:
**Description**

- When creating an Account tenant with a parent of type 'cost-object,' the cost-object ID must not be trimmed.

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/tenant/converter_test.go
```diff
@@ -153,6 +152,10 @@ func TestConverter_ToGraphQLInput(t *testing.T) {
 }
 
 func TestConverter_InputFromGraphQL(t *testing.T) {
+	retrieveTenantTypeFn := func(ctx context.Context, t string) (string, error) {
```
dragobt, line 155: The arguments of the func are not used, maybe we can use `_`

components/director/internal/domain/tenant/converter_test.go
```diff
@@ -281,18 +322,59 @@ func TestConverter_MultipleInputFromGraphQL(t *testing.T) {
...
+				Provider:       testProvider,
+			},
 		}
 
 		// THEN
 		require.Equal(t, len(expected), len(res))
 		require.Equal(t, expected, res)
 	})
+
+	retrieveTenantTypeFnReturnsErr := func(ctx context.Context, t string) (string, error) {
```
dragobt, line 356: same here

dragobt (changes requested):

dragobt (approved):

PR #3777, opened 2024-03-27
Add operations resolver

kaloyanspiridonov:
**Description**

- Extend the existing application and applications GraphQL query to also be able to request the latest ORD operation as part of the request.

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/root_resolver.go
```diff
@@ -1181,6 +1187,11 @@ func (r *applicationResolver) EventingConfiguration(ctx context.Context, obj *gr
 	return r.app.EventingConfiguration(ctx, obj)
 }
 
+// Operations missing godoc
```
dragobt, line 1190: add meaningful go doc

components/director/pkg/graphql/schema.graphql
```diff
@@ -197,6 +204,11 @@ enum RuntimeStatusCondition {
 	FAILED
 }
 
+enum ScheduledOperationType {
```
dragobt, line 207: maybe just `OperationType`, or even `Type` instead of `ScheduledOperationType` is enough?

dragobt (changes requested):

dragobt (approved):

PR #3801, opened 2024-04-10
Do not delete constraint if it is attached

StanislavStefanov:
**Description**
Currently a Formation Constraint can be deleted even if it is attached to a Formation Template. Deleting attached constraint can cause isses with the existing formations or newly created formations from the attached Formation Type.
- Fail if the formation Constraint is attached to any Formation Template

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/formationconstraint/service.go
```diff
@@ -150,6 +151,20 @@ func (s *service) ListByFormationTemplateIDs(ctx context.Context, formationTempl
 
 // Delete deletes formation constraint by id
 func (s *service) Delete(ctx context.Context, id string) error {
+	formationTemplateConstraintReferences, err := s.formationTemplateConstraintReferenceRepo.ListByConstraintID(ctx, id)
+	if err != nil {
+		return errors.Wrapf(err, "while listing Formation Template Constraint References for Constraint with ID %s", id)
+	}
+
+	formationTemplateIDs := make([]string, 0, len(formationTemplateConstraintReferences))
```
DimitarPetrov, line 159: Why we need this, can we just check the other slice's len?

components/director/internal/domain/formationconstraint/service.go
```diff
@@ -150,6 +151,20 @@ func (s *service) ListByFormationTemplateIDs(ctx context.Context, formationTempl
 
 // Delete deletes formation constraint by id
 func (s *service) Delete(ctx context.Context, id string) error {
+	formationTemplateConstraintReferences, err := s.formationTemplateConstraintReferenceRepo.ListByConstraintID(ctx, id)
```
DimitarPetrov, line 154: Can we just use exists?

DimitarPetrov (commented):
General comment - can we just remove the DB cascade and make sure that the DB is guarding us as a list mile?

components/director/internal/domain/formationconstraint/service.go
StanislavStefanov: We could use Exist - but then we wouldnt know which are the Formation Templates from which the constraint has to be potentialy detached

components/director/internal/domain/formationtemplateconstraintreferences/repository.go
```diff
@@ -59,6 +59,16 @@ func (r *repository) ListByFormationTemplateIDs(ctx context.Context, formationTe
 	return r.multipleFromEntities(entityCollection)
 }
 
+// ListByConstraintID lists formationTemplateConstraintReferences for the provided constraint ID
+func (r *repository) ListByConstraintID(ctx context.Context, constraintID string) ([]*model.FormationTemplateConstraintReference, error) {
+	var entityCollection EntityCollection
+
+	if err := r.lister.ListGlobal(ctx, &entityCollection, repo.NewEqualCondition(formationConstraintColumn, constraintID)); err != nil {
+		return nil, errors.Wrap(err, "while listing formationTemplate-constraint references by constraint ID")
```
la4ezar, line 67: Add the actual constraint ID in the log.

la4ezar (changes requested):

la4ezar (approved):

PR #3812, opened 2024-04-12
Introduce Global Applications API

kaloyanspiridonov:
**Description**

- Introduce new graphql Global Applications query. This query provides the opportunity to retrieve applications filtered by the provided label filter and ignoring the tenancy 

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/domain/application/converter.go
```diff
@@ -299,3 +299,31 @@ func (c *converter) statusGraphQLToModel(in *graphql.ApplicationStatus) *model.A
 		Timestamp: time.Time(in.Timestamp),
 	}
 }
+
+type appWithTenantsConverter struct {
+	appConv    ApplicationConverter
+	tenantConv TenantConverter
+}
+
+// NewAppWithTenantsConverter creates new application with tenant converter
```
dzahariev, line 308: tenant->tenants

dzahariev (commented):
- I think the scope need to be added also to hydrator's administrator group mps-superadmin here: https://github.com/kyma-incubator/compass/blob/main/chart/compass/charts/hydrator/static-groups.yaml
- Add this new role also to local config here: https://github.com/kyma-incubator/compass/blob/main/components/director/hack/config-local.yaml
- E2e test is missing.

dzahariev (dismissed):

dzahariev (approved):

PR #3818, opened 2024-04-17
Fix ASA tenancy

la4ezar:
**Description**
When we delete a runtime that has ASA for it we should delete the associated scenarios labels and ASA for it. The problem is that the scenarios label is in the parent tenant and we are using the child tenant which comes from the context. But there are no such records in the database(for the child) and we don't process the scenarios labels, ASAs, and don't unassign the runtime from the formation.

- Fix ASA tenancy

**Related issue(s)**

- ...

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


components/director/internal/domain/runtime/resolver.go
```diff
@@ -862,35 +863,46 @@ func (r *Resolver) EventingConfiguration(ctx context.Context, obj *graphql.Runti
...
 	if err != nil {
 		return err
 	}
 
-	for _, scenario := range scenarios {
-		scenarioAssignment, err := r.scenarioAssignmentService.GetForScenarioName(ctx, scenario)
+	for _, parentTenantID := range tntMapping.Parents {
+		ctxWithParentTenant := tenant.SaveToContext(ctx, parentTenantID, "")
+
+		scenariosLbl, err := r.runtimeService.GetLabel(ctxWithParentTenant, runtimeID, model.ScenariosKey)
```
StanislavStefanov, line 879: The `scenarioLabel` will soon be removed. Instead of relying on it we should list the ASAs where the `Tenant` property of the ASA is amoung the `tntMapping.Parents`. Then we can remove the assigned scenarios using the ASAs. This will reduce the complexity and will remove unneeded calls to the DB.

components/director/internal/domain/formation/service.go
```diff
@@ -1509,7 +1513,7 @@ func (s *service) DeleteAutomaticScenarioAssignment(ctx context.Context, in *mod
 // RemoveAssignedScenarios removes all the scenarios that are coming from any of the provided ASAs
 func (s *service) RemoveAssignedScenarios(ctx context.Context, in []*model.AutomaticScenarioAssignment) error {
```
StanislavStefanov, line 1514: Can we rename the method to something like `UnassignFormationsCommingFromASA`. The current name is a bit missleading.

StanislavStefanov: Same applies to the asaEngine method as well

components/director/internal/domain/scenarioassignment/repository.go
```diff
@@ -111,6 +113,31 @@ func (r *repository) GetForScenarioName(ctx context.Context, tenantID, scenarioN
...
+	var collection EntityCollection
+
+	scenarioConditions := make(repo.Conditions, 0, len(scenarioNames))
+	for _, name := range scenarioNames {
+		scenarioConditions = append(scenarioConditions, repo.NewEqualCondition(scenarioColumn, name))
+	}
+
+	conditions := repo.Or(repo.ConditionTreesFromConditions(scenarioConditions)...)
+
+	if err := r.conditionLister.ListConditionTree(ctx, resource.AutomaticScenarioAssigment, tenantID, &collection, conditions); err != nil {
```
StanislavStefanov, line 127: You can use the standard lister with 	`repo.NewInConditionForStringValues(scenarioColumn, scenarioNames)` instead of this one

StanislavStefanov (changes requested):

components/director/internal/domain/runtime/resolver.go
```diff
@@ -862,35 +863,46 @@ func (r *Resolver) EventingConfiguration(ctx context.Context, obj *graphql.Runti
 // deleteAssociatedScenarioAssignments ensures that scenario assignments which are responsible for creation of certain runtime labels are deleted,
 // if runtime doesn't have the scenarios label or is part of a scenario for which no scenario assignment exists => noop
 func (r *Resolver) deleteAssociatedScenarioAssignments(ctx context.Context, runtimeID string) error {
```
StanislavStefanov, line 865: Rename the method  as we no longer delete the ASAs

components/director/internal/domain/runtime/resolver.go
```diff
@@ -876,19 +882,9 @@ func (r *Resolver) deleteAssociatedScenarioAssignments(ctx context.Context, runt
...
-		if err != nil && !notFound {
-			return err
-		}
-
-		if notFound {
-			return nil
-		}
-
-		scenarios, err := labelPkg.ValueToStringsSlice(scenariosLbl.Value)
+		scenarios, err := r.asaEngine.GetScenariosFromMatchingASAs(ctx, runtimeID, graphql.FormationObjectTypeRuntime)
```
StanislavStefanov, line 885: use `ctxWithParentTenant`

StanislavStefanov (changes requested):

components/director/internal/domain/scenarioassignment/repository.go
```diff
@@ -37,6 +38,7 @@ type repository struct {
 	creator         repo.CreatorGlobal
 	singleGetter    repo.SingleGetter
 	lister          repo.Lister
+	conditionLister repo.ConditionTreeLister
```
StanislavStefanov, line 41: this lister is unused

StanislavStefanov (approved):

ognyvrac (approved):

emmakarova (approved):

PR #3822, opened 2024-04-18
Remove runtime mapping handler

la4ezar:
**Description**
Removing the runtime mapping handler, all graphQL queries / mutations that were related to it as well as the corresponding configurations.

- Remove runtime mapping handler

**Pull Request status**

- [ ] Implementation
- [ ] Unit tests
- [ ] Integration tests


PetarTodorovv (changes requested):
In the description, you mentioned that you removed GraphQL query/mutation but I don't see such changes, is the description outdated? Should we remove any queries/mutations that the runtime mapping handler has used or they may be used by someone else? For example `runtimeByTokenIssuer`

PR #3865, opened 2024-05-10
Add dedicated method for formation/formation assignments timestamps update

PetarTodorovv:
**Description**
Currently, we update the entire entity which in some cases can lead to a problem. With this PR we change that logic to use a dedicated method for `last_notification_sent_timestamp` updates for both formation and formation assignment entities.

- Introduce a new repo method that will take care to update only the `last_notification_sent_timestamp` of formations and formation assignments
- Adapt unit tests

**Related issue(s)**

- https://github.com/kyma-incubator/compass/pull/3593

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/formation/notifications.go
```diff
@@ -268,12 +266,11 @@ func (ns *notificationsService) updateLastNotificationSentTimestamp(ctx context.
 
 	if fa != nil {
 		log.C(ctx).Infof("Updating the last notification sent timestamp for formation assignment with ID: %s", fa.ID)
-		fa.SetLastNotificationSentTimestamp(time.Now())
-		if err := ns.formationAssignmentRepo.Update(ctx, fa); err != nil {
+		if err := ns.formationAssignmentRepo.UpdateLastNotificationSentTimestamps(ctx, fa.ID); err != nil {
 			// That covers the case when we send two unassign notifications to one participant
 			// and the response of the first notification is returned and processed, which deletes the formation assignment,
 			// while the second notification still hasn't been sent.
-			if webhookNotificationReq.GetOperation() == model.UnassignFormation && (apperrors.IsNotFoundError(err) || apperrors.IsUnauthorizedError(err)) { // the not found error is disguised behind the unauthorized error in case of update
```
ivantenevvasilev, line 276: Is the unauthorized not applicable anymore? It's still an update.

ivantenevvasilev: I think even it might not return NotFound either, since it's not using the regular `updater`. But we should handle errors correctly, in order not to reintroduce race conditions again.

components/director/internal/domain/formation/notifications.go
```diff
@@ -257,9 +256,8 @@ func (ns *notificationsService) updateLastNotificationSentTimestamp(ctx context.
 	fa := webhookNotificationReq.GetFormationAssignment()
 	if fa == nil && f != nil {
 		log.C(ctx).Infof("Updating the last notification sent timestamp for formation with ID: %s", f.ID)
-		f.SetLastNotificationSentTimestamp(time.Now())
-		if err := ns.formationRepo.Update(ctx, f); err != nil {
-			if webhookNotificationReq.GetOperation() == model.DeleteFormation && (apperrors.IsNotFoundError(err) || apperrors.IsUnauthorizedError(err)) { // the not found error is disguised behind the unauthorized error in case of update
+		if err := ns.formationRepo.UpdateLastNotificationSentTimestamps(ctx, f.ID); err != nil {
```
ivantenevvasilev, line 259: Same as the other comment.

ivantenevvasilev (changes requested):

components/director/internal/domain/formation/notifications.go
PetarTodorovv: About the first part of the sentence - we use SQL directly but if the error is `sql.ErrNoRows` in the `MapSQLError` func we can still receive a not found error. That's why I left the not found check.
Regarding the second part of the sentence - I didn't understand what we should do?

ivantenevvasilev: My bad, I didn't realize that the MapSQLError returns a NotFound error in such a case. In that case it seems ok as is.

ivantenevvasilev (approved):

PR #3868, opened 2024-05-10
Make the hydrator component bootstrap more resilient

PetarTodorovv:
**Description**
In some cases we can have an invalid certificate subject mapping(CSM) in our DB and during a restart of the hydrator component(or pod eviction) the new pod(s) will start failing because we validate the CSMs. This can cause disruptions and pods to enter in "CrashLoopBackOff" state

- During hydrator bootstrap do not fail/return an error if there are invalid CSMs but rather log the error and proceed.
- Extend the supported consumer types for CSMs
- Fix incorrect error messages in different components

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


kaloyanspiridonov (dismissed):

kaloyanspiridonov (approved):

kaloyanspiridonov (approved):

components/hydrator/internal/subject/cert_subject.go
```diff
@@ -40,7 +40,8 @@ func NewProcessor(ctx context.Context, certSubjectMappingCache certsubjectmappin
 
 	for _, m := range mappings {
 		if err := m.Validate(); err != nil {
-			return nil, err
+			// in case the certificate subject mapping is invalid, only log an error and continue so the hydrator component can still start
```
nyordanoff, line 43: minor: the go-doc of this func can be adapted as well:

`If the subject-to-consumer mapping is invalid, an error is returned.` -> `If the subject-to-consumer mapping is invalid, an error is logged.`

components/hydrator/internal/subject/cert_subject.go
```diff
@@ -40,7 +40,8 @@ func NewProcessor(ctx context.Context, certSubjectMappingCache certsubjectmappin
...
+			log.C(ctx).Errorf("Certificate subject mapping with subject: '%s' is not valid: %s", m.Subject, err.Error())
```
nyordanoff, line 44: Since we no longer return an error in this func, do you think we can change it to return only `*processor` instead of `(*processor, error)`?

components/director/pkg/inputvalidation/validators_cert_subject_mappings.go
```diff
@@ -30,6 +30,8 @@ var SupportedConsumerTypes = map[consumer.Type]bool{
 	consumer.InstanceCreator:                    true,
 	consumer.FormationViewer:                    true,
 	consumer.TechnicalClient:                    true,
+	consumer.ExternalCertificate:                true,
```
nyordanoff, line 33: Shouldn't we update the director version in the hydrator's `go.mod`, because during the validation in the hydrator we import the `/director/pkg/inputvalidation` and use this map

nyordanoff (approved):

kaloyanspiridonov (approved):

PR #3871, opened 2024-05-13
Ignore tenants with slashes

dzahariev:
**Description**

- Ignore tenants with slashes

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/tenant/fetcher_on_demand.go
```diff
@@ -50,6 +52,10 @@ func NewFetchOnDemandService(client Client, config FetchOnDemandAPIConfig) Fetch
 
 // FetchOnDemand calls an API which fetches details for the given tenant from an external tenancy service, stores the tenant in the Compass DB and returns 200 OK if the tenant was successfully created.
 func (s *fetchOnDemandService) FetchOnDemand(ctx context.Context, tenant, parentTenant string) error {
+	if strings.Contains(tenant, "/") {
```
emmakarova, line 55: Edit the comment so that the case for tenants with slashes is included (right now from the comment is expected that the code will make the request and return 200 in that case.

emmakarova (approved):

PR #3897, opened 2024-05-23
Return all operation entities for each application in operations subresolver

dragobt:
**Description**

Right now if the application has an application template it will search for both values in the operation data which is not correct. The application can contain an ORD webhook and it must be searched for operation with data with only the specific application id. If there is no such operation then we try to see if there is an operation with both the application id and its application template id.

- add implementation
- adjust unit tests

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


components/director/internal/domain/application/resolver.go
```diff
@@ -1344,6 +1340,26 @@ func (r *Resolver) getApplication(ctx context.Context, get func(context.Context)
...
+			if err == nil || apperrors.IsNotFoundError(err) {
+				return op, nil
+			}
+		} else {
+			return op, nil
+		}
+	}
+
+	return nil, errors.Wrap(err, "while getting operation")
+}
```
alextargov, line 1361: Here's a simpler function but I may be missing a case:

```
func (r *Resolver) getOperation(ctx context.Context, obj *graphql.Application) (*model.Operation, error) {
	appTemplateID := ""
	if obj.ApplicationTemplateID != nil {
		appTemplateID = *obj.ApplicationTemplateID
	}

	op, err := r.opSvc.GetByDataAndType(ctx, data.NewOrdOperationData(obj.ID, appTemplateID), model.OperationTypeOrdAggregation)
	if err != nil && !apperrors.IsNotFoundError(err) {
		return nil, errors.Wrap(err, "while getting operation")
	}

	return op, nil
}
```

dragobt: This was exactly the code before the change. Here if the application has an app template we will search for operation with appID and appTemplate id - this means that the processing will be for the application in the context of app template's ord webhook, but there is a case where an application that has an app template, but also has an ord webhook and the processing must be in the context of the application's ord webhook.

alextargov (approved):

PR #3900, opened 2024-05-30
Disallow empty labels creation

dzahariev:
**Description**

- Disallow empty labels creation

**Pull Request status**
- [x] Implementation
- [x] Unit tests
- [ ] Integration tests


dragobt:
It looks good, but when executing **TestRegisterApplicationFromTemplateWithPlaceholderPayload** e2e test it looks like there is still an empty label created which I am not sure where it came from. Can we please double-check that we are not missing something?
image

dragobt (approved):

PR #3915, opened 2024-06-11
Ignorelist specific ord rules

alextargov:
**Description**

- Define a mechanism to ignore specific ORD rules for apps/app templates with a given namespace.
- Ignoring means not deleting the invalid ORD resources. We mark the operation as failed, we persist the errors but we do not remove the invalid ord data.

**Related issue(s)**

- ...

**Pull Request status**

- [x] Implementation
- [x] Unit tests
- [x] Integration tests


components/director/internal/open_resource_discovery/global_registry.go
```diff
@@ -71,7 +71,7 @@ func (s *globalRegistryService) SyncGlobalResources(ctx context.Context) (map[st
 		return nil, errors.Wrapf(err, "while fetching global registry documents from %s", s.config.URL)
 	}
 
-	validationErrors, err := s.documentValidator.Validate(ctx, documents, s.config.URL, map[string]bool{}, docsString, "sap:base:v1")
+	validationErrors, err := s.documentValidator.Validate(ctx, documents, s.config.URL, map[string]bool{}, docsString, "sap:base:v1", "")
```
dzahariev, line 74: Probably is good to introduce "Global" or similar namespace, instead using empty string. Otherwise it will be difficult to provide global ignore list.

components/director/internal/open_resource_discovery/validator.go
```diff
@@ -52,7 +58,9 @@ func (v *DocumentValidator) Validate(ctx context.Context, documents []*Document,
 
 		if len(currentDocumentErrors) > 0 {
 			log.C(ctx).Infof("There are %d validation errors from API Metadata Validator", len(currentDocumentErrors))
-			deleteInvalidResourcesFromDocument(documents[i], currentDocumentErrors)
+
```
dzahariev, line 61: Probably should merge Global ignore list and name-spaced ignore list here before proceed with clearing the results.

dzahariev (commented):
Handling Global ignore list needs a little attention. Otherwise the change looks Ok.

dzahariev (dismissed):

dzahariev (approved):

